PR #612, opened 2026-03-20
fix(p2p): add 30s timeout to iroh request-response path

jackzampolin:
## Summary

- Add `REQUEST_RESPONSE_TIMEOUT` constant (30 seconds) to the iroh endpoint
- Wrap the response-reading portion of `handle_request_response` with `tokio::time::timeout`
- On timeout, log a warning with `peer_id` and ALPN for debugging, then return `Error::ResponseTimeout`

Previously, if a remote peer accepted the connection but never sent a response, `handle_request_response` would block indefinitely until the QUIC idle timeout. The fire-and-forget path already had a 5-second timeout as precedent; 30 seconds is used here because the remote peer needs time to process the request before replying.

## Test plan

- [ ] `cargo clippy -p p2p --features iroh-transport -- -D warnings` passes
- [ ] `cargo fmt --all` clean
- [ ] `cargo test -p p2p --features iroh-transport` passes

### DIFF
--- a/crates/p2p/src/iroh/endpoint.rs
+++ b/crates/p2p/src/iroh/endpoint.rs
@@ -30,6 +30,13 @@ use super::config::{IrohDiscoveryConfig, IrohRelayModeConfig};
 use super::peer_map::{endpoint_id_to_peer_id, parse_endpoint_id, PeerMap};
 use super::protocols;
 
+/// Timeout for request-response round trips.
+///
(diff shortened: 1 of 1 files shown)


jackzampolin:
Closing as redundant: the request-response timeout change is already on `main` in commit `8f716817` via #614, so there is nothing left for this PR to merge.

PR #644, opened 2026-04-08
Fix DAC policy enforcement and encryption security regressions

### LINKED ISSUE #639: EncryptionConfig::derive_key silently drops the encryption key — derived keys contain no secret material
EncryptionConfig::derive_key [here](https://github.com/sourcenetwork/defradb.rs/blob/main/crates/defra-core/src/encryption.rs#L83-L96) is supposed to derive a per-document/per-field AES-256 key from a master key, a doc ID, and an optional field name. Due to a truncation bug, the master key is never included in the output. The derived key contains only public information.

  The implementation concatenates field + doc_id + encryption_key and then takes the first 32 bytes:                                                                                                                                         
   
  key_material.extend_from_slice(field.as_bytes());     // 0–N bytes                                                                                                                                                                         
  key_material.extend_from_slice(doc_id.as_bytes());    // 40 bytes (e.g. "bae-c94acbfa-...")                                                                                                                                                
  key_material.extend_from_slice(&self.encryption_key); // never reached                                                                                                                                                                     
  let len = key_material.len().min(32);                                                                                                                                                                                                      
  key[..len].copy_from_slice(&key_material[..len]);                                                                                                                                                                                          
                                                                                                                                                                                                                                             
  DocIDs are 40 characters (bae-{uuid}). Even with an empty field name (doc-level encryption), field + doc_id is already 40 bytes, so truncation to 32 bytes cuts off everything before the master key is reached. The master key is never   
  mixed into the output.
                                                                                                                                                                                                                                             
  For doc-level encryption the derived key is just the first 32 bytes of the doc ID — a fully public, deterministic value. Anyone who knows the doc ID can reproduce the key without any secret material.                                    
   
  For field-level encryption, a non-empty field name makes the situation marginally different in structure but the master key is still never reached.                                                                                        
                  
  Impact                                                                                                                                                                                                                                     
                  
  - Documents encrypted at the doc level are encrypted with a key derived entirely from public information (the doc ID).                                                                                                                     
  - An attacker with access to ciphertext and the doc ID can decrypt without knowing the master key.
  - The EncryptionKey zeroization-on-drop mechanism in place elsewhere provides no protection here since the key is never used.

### LINKED ISSUE #640: defra_init() unconditionally enables deterministic AES-GCM nonces in production — nonce reuse breaks all encryption through the FFI path
defra_init() — the production C FFI entry point called by Go via CGO and by all mobile init paths — unconditionally sets USE_DETERMINISTIC_NONCE = true at startup. Every AES-GCM encryption operation through those paths subsequently uses the hardcoded nonce "deterministi" (first 12 bytes of "deterministic nonce for testing").

Pure Rust deployments are unaffected. Go+Rust hybrid deployments and all mobile nodes are affected for the lifetime of the process.

### LINKED ISSUE #641: LocalDocumentACP ignores policy_id — policy rules unenforced, no namespace isolation, hardcoded permission mapping
LocalDocumentACP is the default ACP implementation used by all deployments that don't explicitly configure SourceHub. It accepts a policy_id parameter on every DocumentACP trait method but ignores it entirely. No policy is loaded, no rules are evaluated, and storage is not namespaced per policy. The permission→relation mapping is hardcoded regardless of what any policy actually defines.
                                                                                                                                                                                                                                           
### How LocalDocumentACP gets wired in

DocumentAcpConfig::Local is the #[default] variant of the config enum (crates/embedded/src/lib.rs:178). It is selected in three places:                                                                                                    
  
  - Embedded node (crates/embedded/src/node_acp.rs:41,45): both persistent and memory modes when SourceHub is not configured                                                                                                                 
  - CLI server (crates/cli/src/commands/start/server_acp.rs:140): when no SourceHub gRPC address is provided
  - FFI node (crates/ffi/src/node.rs:203): when the caller does not supply SourceHub credentials                                                                                                                                             
                                                                                                                                                                                                                                             
In other words, every Rust-native node that doesn't have SourceHub configured uses LocalDocumentACP.                                                                                                                                       
                                                                                                                                                                                        
### What the code actually does                                                                                                                                                                                                                
                  
Every method in crates/acp/src/local.rs takes _policy_id: &str
  
Storage keys are /acp/{collection_id}/{doc_id}/{relation}/{subject_did} — no policy_id component. Permission rules are hardcoded regardless of what any policy defines:                                                                    
```                  
  // local.rs:204-229                                                                                                                                                                                                                        
  DocumentPermission::Read => {                                                                                                                                                                                                              
      self.has_relation(did, resource_name, doc_id, READER_RELATION).await?
      || self.has_relation(did, resource_name, doc_id, "writer").await?                                                                                                                                                                      
      || self.has_relation(did, resource_name, doc_id, UPDATER_RELATION).await?
      || self.has_relation(did, resource_name, doc_id, DELETER_RELATION).await?                                                                                                                                                              
  }      
```

It seems like there is a ZanzibarDocumentACP that is a correct local impl but ill let @Lodek comment on that

jackzampolin:
## Summary
- fix per-document encryption key derivation so the master key is always mixed in
- stop enabling deterministic nonces from `defra_init()` and add a regression test
- switch local runtime DAC enforcement to Zanzibar-backed policy evaluation, persist local FFI policies into that same store, and add cross-runtime custom-policy coverage

## Verification
- cargo test -p defra-core derive_key_mixes_in_master_key_for_long_doc_ids -- --nocapture
- cargo test -p ffi test_defra_init -- --nocapture
- cargo test -p ffi acp::dac::tests -- --nocapture
- cargo test -p acp --test local_tests -- --nocapture
- cargo test -p db --test collection_acp_tests -- --nocapture
- cargo test -p embedded local_document_acp_enforces_stored_custom_policy -- --nocapture
- cargo test -p integration-test --test acp rust_custom_relation_policy -- --nocapture
- PATH="/Users/johnzampolin/go/src/github.com/sourcenetwork/defradb.rs/target/e2e-bin:$PATH" cargo test -p integration-test --test acp go_custom_relation_policy -- --nocapture
- cargo test -p cli --lib --no-run

## Closes
- Closes #639
- Closes #640
- Closes #641

### DIFF
--- a/crates/acp/src/local.rs
+++ b/crates/acp/src/local.rs
@@ -44,6 +44,11 @@ impl LocalDocumentACP {
         Self { store }
     }
 
+    /// Namespace a collection_id by policy_id to isolate policies in storage.
+    fn namespaced_collection(policy_id: &str, collection_id: &str) -> String {
+        format!("{policy_id}:{collection_id}")
+    }
+
     /// Check if the subject is the owner of the document.
     async fn is_owner(&self, subject: &Did, collection_id: &str, doc_id: &str) -> Result<bool> {
         let tuple = RelationTuple::owner(subject.clone(), collection_id, doc_id);
@@ -77,15 +82,16 @@ impl DocumentACP for LocalDocumentACP {
     async fn register_doc_object(
         &self,
         identity: &Did,
-        _policy_id: &str,
+        policy_id: &str,
         resource_name: &str,
         doc_id: &str,
     ) -> Result<()> {
+        let ns_collection = Self::namespaced_collection(policy_id, resource_name);
         // Use atomic registration to prevent TOCTOU race conditions
         // where concurrent registrations could overwrite each other
         match self
             .store
-            .register_doc_atomic(identity, resource_name, doc_id)
+            .register_doc_atomic(identity, &ns_collection, doc_id)
             .await?
         {
             true => {
@@ -119,22 +125,24 @@ impl DocumentACP for LocalDocumentACP {
 
     async fn is_doc_registered(
         &self,
-        _policy_id: &str,
+        policy_id: &str,
         resource_name: &str,
         doc_id: &str,
     ) -> Result<bool> {
-        self.store.is_doc_registered(resource_name, doc_id).await
+        let ns_collection = Self::namespaced_collection(policy_id, resource_name);
+        self.store.is_doc_registered(&ns_collection, doc_id).await
     }
 
     async fn get_doc_owner(
         &self,
-        _policy_id: &str,
+        policy_id: &str,
         resource_name: &str,
         doc_id: &str,
     ) -> Result<Option<Did>> {
+        let ns_collection = Self::namespaced_collection(policy_id, resource_name);
         Ok(self
             .store
-            .get_relation_subjects(resource_name, doc_id, OWNER_RELATION)
+            .get_relation_subjects(&ns_collection, doc_id, OWNER_RELATION)
             .await?
             .into_iter()
             .next())
@@ -144,12 +152,13 @@ impl DocumentACP for LocalDocumentACP {
         &self,
         identity: &Identity,
         permission: DocumentPermission,
-        _policy_id: &str,
+        policy_id: &str,
         resource_name: &str,
         doc_id: &str,
     ) -> Result<bool> {
+        let ns_collection = Self::namespaced_collection(policy_id, resource_name);
         // Check if document is registered
-        if !self.store.is_doc_registered(resource_name, doc_id).await? {
+        if !self.store.is_doc_registered(&ns_collection, doc_id).await? {
             // Document is NOT registered — it was created without an identity
             // on an ACP-protected collection, making it a public document.
             // Anyone can read/update/delete it.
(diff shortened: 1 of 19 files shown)


jackzampolin:
Thank you for taking the time to report and document these problems so clearly. I went back through the current `main` branch and validated the paths covered here, and I wanted to leave a more explicit follow-up for closure.

At this point, the regressions described in the issues closed by this PR appear to be fixed on current `main`.

What I verified:
- The per-document encryption key derivation now mixes in the master key using SHA-256, so it no longer truncates away the secret material.
- `defra_init()` no longer enables deterministic AES-GCM nonces.
- The local ACP path now uses the Zanzibar-backed implementation for local runtimes, and the remaining local ACP code also namespaces by `policy_id`.

I also ran the targeted regression coverage on current `main`, including:
- `cargo test -p defra-core derive_key_mixes_in_master_key_for_long_doc_ids -- --nocapture`
- `cargo test -p ffi test_defra_init -- --nocapture`
- `cargo test -p embedded local_document_acp_enforces_stored_custom_policy -- --nocapture`
- `cargo test -p acp --test local_tests -- --nocapture`

All of those passed.

So the right resolution here is that this was a real set of bugs, they have been fixed, and the fix is present on `main`. Thank you again for the careful report. It helped land concrete regression coverage in the codebase.

PR #715, opened 2026-04-10
Reject high-S secp256k1 signatures during verification

### LINKED ISSUE #714: Rust secp256k1 verifier canonicalizes high-S signatures, creating Go parity mismatch
### Summary

Rust secp256k1 verification accepts high-S ECDSA signatures by normalizing them before verification, while the Go implementation verifies the DER signature as provided. This creates a cross-implementation behavior mismatch and permits signature malleability on the Rust side where Go is stricter.

### Details

In Rust, Secp256k1PublicKey::verify() parses the DER signature and then calls normalize_s().unwrap_or(sig) before verification. That means both low-S and high-S encodings of the same mathematical signature are accepted.

In Go, the secp256k1 verification path parses the DER signature and verifies it directly without an explicit normalization step in DefraDB’s codepath. As a result, Rust accepts some signatures that Go may reject, so verifier behavior is not aligned.

This matters anywhere signature canonicality or verifier parity is expected, especially if signatures cross Rust/Go boundaries or are assumed to have a unique encoding.

### Expected Behavior

Rust and Go should agree on secp256k1 signature acceptance rules.

If the intended behavior is strict/canonical verification, Rust should reject non-canonical high-S signatures instead of canonicalizing them during verification.

### Suspected Cause

The Rust verifier normalizes S before verification in:

defras/rust-defra/crates/crypto/src/keys/secp256k1.rs
while the Go verifier checks the parsed signature directly in:

defras/defradb/crypto/signature.go
This appears to have been added for compatibility with some ecosystems, but it changes verifier acceptance semantics relative to the Go implementation.

### Acceptance Criteria

Add a regression test showing Rust currently accepts a valid high-S secp256k1 signature variant.
Decide and document the intended canonicality rule for secp256k1 signatures.
If strict Go parity is required, update Rust verification to reject high-S signatures rather than normalizing them before verification.
Ensure tests cover both low-S and high-S encodings.
Files

defras/rust-defra/crates/crypto/src/keys/secp256k1.rs
defras/defradb/crypto/signature.go

jackzampolin:
## Summary
- reject non-canonical high-S secp256k1 signatures instead of normalizing them during verification
- add a regression that derives a valid high-S variant from the known Go low-S test vector
- document that secp256k1 verification preserves the caller-provided DER encoding for Go parity

## Validation
- cargo test -p crypto --test go_compat_keys test_secp256k1_high_s_signature_is_rejected -- --nocapture
- cargo test -p crypto --test go_compat_keys secp256k1 -- --nocapture

Closes #714

### DIFF
--- a/crates/crypto/src/keys/secp256k1.rs
+++ b/crates/crypto/src/keys/secp256k1.rs
@@ -195,12 +195,8 @@ impl PublicKey for Secp256k1PublicKey {
             Err(_) => return Ok(false),
         };
 
-        // Normalize S to low-S form if needed for Go compatibility
-        // ECDSA signatures have the property that both (r, s) and (r, n-s) are valid.
-        // Some implementations (like Go's dcrd) may produce high-S signatures,
-        // while k256 verification expects low-S. Normalizing ensures compatibility.
-        let sig = sig.normalize_s().unwrap_or(sig);
-
+        // Preserve the caller-provided DER signature as-is so secp256k1
+        // verification enforces canonical low-S semantics and matches Go.
         // Hash the message with SHA-256 using DigestVerifier trait
         // This matches Go behavior which signs/verifies against SHA256(message)
         let mut hasher = Sha256::new();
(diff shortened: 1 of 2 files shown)


JesseAbram:
lgtm

JesseAbram (approved):

PR #718, opened 2026-04-10
Fix implicit GraphQL query snapshot consistency

### LINKED ISSUE #649: The existence of a crate named `lensed_auto_commit_fetcher` suggests something somewhere is doing something very strange
I have no idea what, it seems to be doing some odd things with transactions (it doesn't seem to ever commit though, which is very good), but it might just be an architectural nightmare instead of something user-visible - it seems to be used in a fair few places and it looks very very odd.

It looks like the new-txn, discard pattern within it means that a single query (e.g. with joins) will be executed across multiple transactions, violating ACID and returning very incorrect results should other, write, transactions be being committed whilst the query is being executed.

jackzampolin:
## Summary
- execute regular GraphQL query requests inside an implicit read-only transaction when a transaction registry is available
- fall back to the existing auto-commit fetcher path only when transactions are not supported
- add a regression test proving a single GraphQL request now sees one transaction-scoped snapshot instead of per-fetch snapshots

## Testing
- cargo test -p query execute_uses_implicit_read_txn_when_registry_is_available
- cargo test -p query test_query_response_success

Closes #649

### DIFF
--- a/crates/db/src/lensed_fetcher/index_scan.rs
+++ b/crates/db/src/lensed_fetcher/index_scan.rs
@@ -1,0 +1,228 @@
+//! Index scan implementation for LensedDocFetcher.
+
+use std::collections::HashSet;
+
+use defra_core::thread_bounds::MaybeBoxFuture;
+use query::planner::index_selection::{IndexScanParams, IndexScanType};
+use storage::corekv::Store;
+use storage::index::IndexIterator;
+
+use crate::collection::collection_short_id;
+use crate::collection_loader::get_collection_with_lazy_load;
+use crate::index_manager::IndexManager;
+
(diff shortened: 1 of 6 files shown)


iverc (approved):

iverc:
Added `get_by_index_scan` and `supports_index_queries` to `LensedDocFetcher` so the implicit read-only transaction path has full index support (BM25, similarity search, dense search). Without this, queries routed through `execute_in_txn` fell back to non-index paths and returned empty results.

Also fixed `search_fulltext_scored` in the same fetcher — it was using `collection.schema().root_id` (which can be 0/unassigned in transaction-loaded collections) instead of `collection_short_id()`, causing wrong index lookups.

PR #730, opened 2026-04-10
Align Rust document encryption with Go parity

### LINKED ISSUE #651: Diversion between AES encryption on rust and go codebase
an issue was opened #639 about the use of AES encryption and was fixed in #644 

* Although the fix looks cryptographically secure this deviates from the go version where in go [generateEncryptionKey](https://github.com/sourcenetwork/defradb/blob/develop/internal/encryption/encryptor.go#L32) it uses a random key and rust uses SHA-256(field + docID + masterKey)
* The best I can tell it looks like the impl in rust was copied over from the test function https://github.com/sourcenetwork/defradb/blob/develop/internal/encryption/encryptor.go#L43
* It also deviates in the concept of a encryption key needing to be passed in from the cli while go just randomly generates it on the fly (and I assume stores it) 

### Consequences 
* Anything encrypted on one client would be incompatible with the other client

jackzampolin:
## Summary
- switch Rust document encryption from deterministic master-key derivation to Go-style generated per-doc/per-field AES keys
- write encryption metadata blocks to `encstore` and read from `encstore` first with a legacy `blockstore` fallback
- add regression coverage for key reuse/scope separation, `encstore` placement, legacy fallback, and runtime field-level encryption

## Why
Issue #651 was not about the raw AES-GCM primitive. That layer was already Go-compatible. The remaining divergence was higher-level document encryption behavior:
- Rust derived keys as `SHA-256(fieldName + docID + masterKey)`
- Go generates a random key per `(docID, field)` in production and stores it in the encryption block
- Rust also stored encryption metadata in `blockstore`, while Go uses `encstore`

This change aligns Rust with Go on the document-encryption path and preserves backward compatibility for older Rust-written encryption blocks by falling back to `blockstore` on reads.

## Verification
- cargo check -p db
- cargo test -p crypto --test go_compat_encryption -- --nocapture
- cargo test -p defra-core encryption::tests -- --nocapture
- cargo test -p db test_compute_document_blocks_places_encryption_metadata_in_encstore_entries -- --nocapture
- cargo test -p db decrypt_block_data_falls_back_to_legacy_blockstore_encryption_blocks -- --nocapture
- cargo test -p integration-test --test encryption rust_field_level_key_isolation -- --nocapture

## Closes
- Closes #651


crates/db/src/merge_handler/mod.rs
```diff
@@ -151,15 +152,30 @@ impl<S: Store, B: blockstore::Blockstore + Send + Sync> DbMergeHandler<S, B> {
             None => return Ok(data.to_vec()),
         };
 
-        // Load the Encryption block from blockstore
-        let enc_data = self
+        // Go stores encryption metadata in encstore; older Rust builds stored it
+        // in blockstore, so keep a fallback for backwards compatibility.
```
JesseAbram: this isn't live yet, don't build for backwards compatability

crates/defra-core/src/encryption.rs
```diff
@@ -76,25 +76,40 @@ impl EncryptionConfig {
...
+        }
+    }
+
+    /// Return the key used for the given doc/field pair, generating one if needed.
+    ///
+    /// This mirrors Go's document encryptor behavior:
+    /// - doc-level encryption reuses one key per document
+    /// - field-level encryption reuses one key per `(doc, field)` pair
+    /// - production keys are generated randomly instead of being derived
+    pub fn get_or_generate_key(&self, field_name: &str, doc_id: &str) -> [u8; 32] {
```
JesseAbram: nit: make field_name an option

crates/defra-core/src/encryption.rs
```diff
@@ -76,25 +76,40 @@ impl EncryptionConfig {
...
+        let field = self.effective_field_name(field_name);
+        let cache_key = (doc_id.to_string(), field);
+
+        let mut store = doc_encryption_key_store()
+            .lock()
+            .expect("doc encryption key store poisoned");
```
JesseAbram: replace expect with proper error handling and return a result

crates/defra-core/Cargo.toml
```diff
@@ -24,6 +24,7 @@ libipld.workspace = true
 
 # Cryptography (for CID generation)
 sha2.workspace = true
+rand = "0.8"
```
JesseAbram: nit: you have different versions of rand in the codebase consider one and workspace = true for imports

crates/defra-core/src/encryption.rs
```diff
@@ -76,25 +76,40 @@ impl EncryptionConfig {
...
+
+        if store.len() >= MAX_DOC_ENCRYPTION_ENTRIES {
+            store.clear();
```
JesseAbram: Clearing the whole cache here may cause some issues for any thread still using items in it. Im ok merging with this since it is an edge case but I would like to make an improvement issue here about making this a Least Recently Used ejection not a whole cache clear

crates/query/src/runner/mutation.rs
```diff
@@ -354,15 +354,13 @@ impl<F: DocFetcher + 'static, R: TransactionRegistry> QueryRunner<F, R> {
...
-                        encrypt_fields: mutation.encrypt_fields.clone(),
-                        encryption_key: key.clone(),
-                    },
-                ));
-            }
+            defra_core::encryption::set_encryption_config(Some(
+                defra_core::encryption::EncryptionConfig {
+                    encrypt_doc: mutation.encrypt_doc,
+                    encrypt_fields: mutation.encrypt_fields.clone(),
+                    encryption_key: self.encryption_key.clone().unwrap_or_default(),
```
JesseAbram: I looked into this field because unwrap_or_default seemed dangerous for an encryption_key but not I think this can be removed, it was used in derive_key only, If you want to move that to its own issue or PR that would be find with me

crates/db/src/auto_commit_mutator/batch.rs
```diff
@@ -200,10 +210,12 @@ impl<S: Store + 'static> DocMutator for BatchMutator<S> {
 
         let commit_result: Option<(Cid, Vec<u8>, Option<(Cid, Vec<u8>)>)> = {
             let blockstore = self.blockstore().await?;
+            let encstore = self.encstore().await?;
```
JesseAbram: nit: 
```
  // safer: one lock, get all three
  let (blockstore, encstore, headstore) = {
      let txn = self.txn.lock().await;                                                                                                                                                                                                       
      let txn = txn.as_ref()?;
      (txn.blockstore()?, txn.encstore()?, txn.headstore()?)                                                                                                                                                                                 
  };                                                                                                                                                                                                                                         
   ```

JesseAbram (commented):
The backwards compatibility should be fixed in this PR, the others can be pulled out into issues, as long as they are logged im happy, the nits, you can consider them and ignore if you want

jackzampolin:
Thanks for the pointers, Jesse. I went through the review and fixed the backward-compat issue plus the nits I agreed with.

JesseAbram (approved):

Issue #733, opened 2026-04-10
Ok(false) used for verification failures throughout merkle_proof — should be Err to match Go


JesseAbram:
### Summary                                                                                                                                                                                                                                    
                                                            
  All verification failures across the merkle_proof module return Ok(false) rather than Err. Go treats verification failures as errors. This is inconsistent with Go's convention and with the rest of the Rust crypto crate.
                                                                                                                                                                                                                                             
  ### Details
                                                                                                                                                                                                                                             
  There are 6 occurrences across 2 files. They fall into two categories:                                                                                                                                                                     
   
  Cryptographic failures — clearly wrong, Go always returns an error here:                                                                                                                                                                   
```                                                            
  // signed_proof.rs — identity mismatch                                                                                                                                                                                                     
  if self.signature.header.identity != expected_identity {  
      return Ok(false);                                                                                                                                                                                                                      
  }                                                         

  // signed_proof.rs — signature invalid (verify, verify_with_embedded_key)                                                                                                                                                                  
  if !public_key.verify(&proof_bytes, &self.signature.value)? {
      return Ok(false);                                                                                                                                                                                                                      
  }                                                         
```
  Structural failures — debatable but inconsistent with Go:                                                                                                                                                                                  
   ```
  // proof.rs — empty path                                                                                                                                                                                                                   
  if self.path.is_empty() {                                 
      return Ok(false);
  }                                                                                                                                                                                                                                          
   
  // proof.rs — leaf/root CID anchor mismatch                                                                                                                                                                                                
  if self.path[0].cid != self.leaf_cid { return Ok(false); }
  if self.path[self.path.len() - 1].cid != self.root_cid { return Ok(false); }                                                                                                                                                               
                                                                                                                                                                                                                                             
  // proof.rs — broken chain link                                                                                                                                                                                                            
  if !has_link { return Ok(false); }                                                                                                                                                                                                         
 ```                                                           
  Go's verifySignature in internal/core/block/signature.go returns ErrSignatureVerification on failure. VerifyBlockSignatureWithKey returns ErrSignaturePubKeyMismatch on identity mismatch. Neither returns a bool.                         
   
 ### Expected Behavior                                                                                                                                                                                                                          
                                                            
  Verification failures return Err with a descriptive error. Callers that want a bool can map the result themselves. This matches Go's convention and makes failure reasons visible rather than collapsing them all into false.              
   
 ### Suspected Cause                                                                                                                                                                                                                            
                                                            
[...]

PR #761, opened 2026-04-13
fix(crypto): merkle proof verification fails closed with Err (#733)

iverc:
## Summary

Closes **#733**. The merkle proof verifier returned \`Ok(false)\` for cryptographic verification failures (CID hash mismatch, broken chain link). Every other verification site in the Rust crypto crate returns \`Err\` for crypto failures, and Go DefraDB does the same.

## Why this is more than code quality (P1, not low-priority)

The \`Ok(false)\` pattern is a silent bypass risk for any caller writing the idiomatic Rust patterns:
- \`if result.is_ok() { trust_proof() }\` — \`Ok(false)\` looks like success
- \`let _ = result?; trust_proof()\` — flows through \`?\` cleanly
- \`result.unwrap_or(true)\` — defensive defaults treat the error state as success

Today's wasm callers happen to pass the \`Result\` through and JS truthiness covers \`Ok(false)\` as falsy. Not directly exploitable in current code, but a migration hazard for any future Rust caller and inconsistent with both Go and the rest of the crypto crate (the sibling \`signed_proof.rs\` already returns \`Err\` for identity/signature failures, per the audit-era convention).

## Fix

[\`crates/crypto/src/merkle_proof/proof.rs\`](crates/crypto/src/merkle_proof/proof.rs) — hash mismatch and broken chain link both return \`Err::Crypto\` with a descriptive message that includes the failing node index and the relevant CID:

\`\`\`rust
return Err(Error::Crypto(format!(
    \"proof node CID does not match content hash at path index {}\",
    idx
)));
return Err(Error::Crypto(format!(
    \"proof chain is broken: node at path index {} does not list \
     node at index {} (cid: {}) in its heads\",
    i, i + 1, next_cid
)));
\`\`\`

After this change, \`MerkleProof::verify()\` never returns \`Ok(false)\` — only \`Ok(true)\` on success or \`Err\` on any failure mode. The type signature stays \`Result<bool>\` for API stability; the \`bool\` is always \`true\` on success.

## Test updates

- **\`test_proof_missing_link_fails\`** — flipped from \`!result.unwrap()\` to \`result.is_err()\` plus an assertion that the error message contains \"chain\" or \"crypto\"
- **\`test_verify_err_vs_ok_false_semantics\`** — updated commentary and flipped the broken-chain branch to assert \`Err\`
- **\`test_verify_hash_mismatch_returns_err\`** (new) — locks in the hash-mismatch branch by tampering \`ProofNode.data\` and asserting the returned \`Err\` contains the expected message

## Wasm wrapper changes

Doc comments on \`verify_merkle_proof\`, \`verify_merkle_proof_cbor\`, \`verify_signed_proof\`, and \`DefraClient::verify_proof\` now document that they **never return false** — failures throw exceptions in JS. This is the intended behavior change per #733: forces JS callers to handle verification failures via try/catch instead of relying on truthiness checks that could silently bypass the verifier.

The wasm function signatures stay \`Result<bool, JsValue>\` (no API break). The \`bool\` branch only ever carries \`true\` after the change.

## Test plan

- [x] \`cargo test -p crypto\` → 206 pass (including 38 merkle_proof_tests with the 3 changed/new tests)
- [x] \`cargo test --workspace --exclude integration-test --exclude ffi-test --exclude defra-wasm\` → all green
- [x] \`cargo test -p integration-test --test acp -- --skip ::go_\` → 43/43 pass (no regression)
- [x] \`cargo clippy -p crypto --tests -- -D warnings\` → clean
- [x] \`cargo fmt -p crypto\` → applied

## Why no JS-side regression for current callers

The single internal caller \`DefraClient::verify_proof\` just propagates the \`Result\` through; it doesn't introspect \`bool\`. External JS callers of \`verify_merkle_proof\` / \`verify_signed_proof\` are not enumerated in this repo and were always supposed to handle exceptions for verification failures (the function already returned \`Err\` for structural input errors). The change makes the failure semantics consistent across all failure modes.

## Related

- Original issue: #733 (re-labeled to priority:p1 + area:security earlier today)
- Discovered during the post-#756 ACP audit completion sweep
- The sibling \`signed_proof.rs\` already returns \`Err\` for identity/signature mismatches via the same convention this PR adds to \`proof.rs\`

### DIFF
--- a/crates/crypto/src/merkle_proof/proof.rs
+++ b/crates/crypto/src/merkle_proof/proof.rs
@@ -37,12 +37,14 @@ impl MerkleProof {
 
     /// Verify this proof is valid
     ///
-    /// Checks:
-    /// 1. Path is non-empty (Err)
-    /// 2. First node CID matches leaf_cid (Err — wrong input)
-    /// 3. Last node CID matches root_cid (Err — wrong input)
-    /// 4. Each node's CID matches its content hash (Ok(false) — cryptographic failure)
-    /// 5. Each node's heads contains the next node's CID (Ok(false) — broken chain)
+    /// Checks (all return `Err` on failure to match Go's convention and the
+    /// rest of the Rust crypto crate — see #733):
+    ///
+    /// 1. Path is non-empty
+    /// 2. First node CID matches `leaf_cid`
+    /// 3. Last node CID matches `root_cid`
+    /// 4. Each node's CID matches its content hash (cryptographic failure)
+    /// 5. Each node's `heads` contains the next node's CID (broken chain)
     ///
     /// The chain structure is: leaf (newest) -> ... -> root (oldest).
     /// Each node points to its parent(s) via the `heads` field.
@@ -50,9 +52,16 @@ impl MerkleProof {
     /// # Returns
     ///
     /// - `Ok(true)` — proof verifies
-    /// - `Ok(false)` — proof is structurally valid but cryptographically incorrect
-    ///   (content hash mismatch, broken chain link)
-    /// - `Err(_)` — caller supplied structurally invalid input (empty path, anchor mismatch)
+    /// - `Err(_)` — verification failed at any of the checks above. The
+    ///   error variant is `Error::BlockError` for structural input errors
+    ///   (empty path, anchor mismatch) and `Error::Crypto` for cryptographic
+    ///   failures (hash mismatch, broken chain link).
+    ///
+    /// Note: this function never returns `Ok(false)`. Both invariants are
+    /// security-sensitive enough that a caller who treats `Ok` as success
+    /// (e.g. `if result.is_ok()` or `let _ = result?; trust_proof()`)
+    /// must not be able to bypass verification by writing the obvious code.
+    /// See #733.
     pub fn verify(&self) -> Result<bool> {
         // Anchor/structural failures are caller errors — return Err so they can't be
         // silently ignored. Match Go's convention where "nothing to verify" is an error.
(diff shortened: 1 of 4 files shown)


iverc:
## Context: relationship to PR #737

I just noticed that **PR #737 (merged 2026-04-12)** already partially addressed #733. That PR deliberately left hash mismatch and broken chain link as `Ok(false)`, with the rationale: "legitimate cryptographic 'no' — there's nothing wrong with the caller's input, the proof just doesn't verify".

This PR (#761) **disagrees with that design decision** and finishes the job. Reasons:

1. **Go parity**: Go DefraDB returns errors for hash/chain verification failures. The whole point of #733 was to match Go's convention. Leaving Ok(false) for two of the failure modes leaves Rust still divergent.

2. **Consistency with rest of Rust crypto crate**: Every other verification site (Ed25519, BLS, secp256k1, the signed_proof identity check that #737 already fixed) returns `Err` on failure. Hash mismatch / broken chain are crypto failures of the same nature.

3. **Defense-in-depth against caller bugs**: The argument "there's nothing wrong with the caller's input" misses the security risk. The risk is the caller writing `if result.is_ok()` or `let _ = result?; trust_proof()` and treating the failure as success. `Ok(false)` is a footgun for any future caller that doesn't read the docs carefully. Defense-in-depth says: if verification fails for ANY reason, propagate as Err so the only way to get a positive trust signal is `Ok(true)`.

4. **#737's own audit**: PR #737 fixed exactly this concern for signature failures and identity mismatch — both of those used to be `Ok(false)` too, and #737 flipped them to `Err` because, in the author's words, they are "a security event". Hash mismatch and broken chain link are *equally* cryptographic security events.

If the maintainers prefer the original #737 design (Ok(false) for proof-internal failures), this PR can be closed. But I'd argue that the inconsistency with Go and with the rest of the crypto crate is the bigger concern, and the security framing wins on principle: any verification call site that fails any check should be impossible to mistake for success.

Happy to discuss.

PR #784, opened 2026-04-14
fix(defra-core): replace unsafe write_volatile key zeroing with zeroize (#783)

### LINKED ISSUE #783: fix(defra-core): replace unsafe write_volatile key zeroing with zeroize
### Summary                                                                                                                                                                                                                                
                                                                                                                                                                                                                                             
  `EncryptionKey::drop()` in `crates/defra-core/src/encryption.rs` zeroes key material                                                                                                                                                       
  using raw `unsafe { std::ptr::write_volatile(byte, 0) }`. The `zeroize` crate is the                                                                                                                                                       
  Rust ecosystem standard for this, already present in the dependency tree via `aes-gcm`                                                                                                                                                     
  and other RustCrypto crates. The manual implementation should be replaced with `zeroize`,
  matching the pattern used across the RustCrypto ecosystem.                                                                                                                                                                                 
                  
  ### Details

  `EncryptionKey` is a `Vec<u8>` newtype wrapping a 32-byte AES-256 key. Its `Drop` impl                                                                                                                                                     
  manually iterates and zeroes each byte:
                                                                                                                                                                                                                                             
  ```rust         
  impl Drop for EncryptionKey {
      fn drop(&mut self) {
          for byte in self.0.iter_mut() {                                                                                                                                                                                                    
              unsafe {
                  std::ptr::write_volatile(byte, 0);                                                                                                                                                                                         
              }   
          }
      }
  }
  ```
                                                                                                                                                                                                                                             
  `write_volatile` prevents the compiler from eliding the write, but has three gaps:
                                                                                                                                                                                                                                             
  - **Reallocations**: `Vec` may have been reallocated (e.g. during construction or clone)                                                                                                                                                   
    — the old heap allocation containing key bytes is `free()`'d without zeroing.
    `write_volatile` only touches the current backing buffer.                                                                                                                                                                                
  - **Memory fences**: No fence is issued. The zeroing is not guaranteed to be ordered or
    visible relative to other memory operations.                                                                                                                                                                                             
  - **Unsafe surface**: There is no reason for `unsafe` code here. The `zeroize` crate
    handles all of the above without it.                                                                                                                                                                                                     
                  
  `zeroize` is almost certainly already in the dependency graph via `aes-gcm`,                                                                                                                                                               
  `chacha20poly1305`, or `p256`. The replacement collapses to a one-line derive:
                                                                                                                                                                                                                                             
  ```rust         
  use zeroize::ZeroizeOnDrop;

  #[derive(Clone, ZeroizeOnDrop)]                                                                                                                                                                                                            
  pub struct EncryptionKey(Vec<u8>);
  ```                                                                                                                                                                                                                                       
                  
  The manual `Drop` impl is deleted entirely. `ZeroizeOnDrop` generates it correctly,
  including a memory fence after zeroing.
                                                                                                                                                                                                                                             
  Go's equivalent (`internal/encryption/encryptor.go`) relies on the GC and does not
  manually zero keys — the Rust implementation is already doing more than Go here, but                                                                                                                                                       
  it should do it correctly.

  ### Expected Behavior

  `EncryptionKey` uses `ZeroizeOnDrop` for key material clearing. No `unsafe` block exists                                                                                                                                                   
  in `encryption.rs`. Key bytes in the current allocation are zeroed with a memory fence
  on drop.                                                                                                                                                                                                                                   
                  
  ### Suspected Cause

  The manual implementation predates evaluation of `zeroize` as a dependency, or was                                                                                                                                                         
  written without checking whether it was already transitively available.
                                                                                                                                                                                                                                             
  ### Acceptance Criteria

  - `zeroize` added as an explicit dependency of `crates/defra-core` in `Cargo.toml`                                                                                                                                                         
    (it is likely already transitive, but should be direct)
  - `EncryptionKey` derives `ZeroizeOnDrop`; the manual `Drop` impl is deleted                                                                                                                                                               
  - The `unsafe` block in `encryption.rs` is removed entirely
  - `cargo clippy --all -- -D warnings` passes clean
  - Existing `generate_encryption_key_is_random` test continues to pass                                                                                                                                                                      
   
  ### Files                                                                                                                                                                                                                                  
                  
  - `crates/defra-core/src/encryption.rs` — replace manual `Drop` with `ZeroizeOnDrop`                                                                                                                                                       
    derive, remove `unsafe` block
  - `crates/defra-core/Cargo.toml` — add `zeroize` as an explicit dependency

iverc:
## Summary

Closes #783. Replaces the manual `unsafe { write_volatile }` zeroing in [crates/defra-core/src/encryption.rs](crates/defra-core/src/encryption.rs) with `zeroize::ZeroizeOnDrop`, the RustCrypto ecosystem standard.

The previous implementation had three gaps the issue calls out:
- **No memory fence** after zeroing — ordering relative to subsequent memory operations isn't guaranteed.
- **Doesn't cover Vec reallocations** — if the inner Vec was ever reallocated, the old heap buffer is free()'d without zeroing.
- **Unnecessary unsafe** — there's no reason for a raw pointer write here.

`ZeroizeOnDrop` generates the Drop impl correctly, including a post-zero memory fence. `zeroize` was already in the transitive dep graph via `aes-gcm` and other RustCrypto crates; promoted to a direct dep on `defra-core`.

## Test plan

- [x] `cargo test -p defra-core encryption` — 4/4 unit tests + 8/8 block tests passing
- [x] New unit test `encryption_key_zeroizes_buffer` verifies the buffer is cleared after `.zeroize()`

## Diff

3 files, 18 insertions, 15 deletions. Net: removed unsafe block + manual Drop impl, added derive macros + test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3734,6 +3734,7 @@ dependencies = [
  "thiserror 1.0.69",
  "tracing",
  "uuid",
+ "zeroize",
 ]
 
 [[package]]
--- a/crates/defra-core/Cargo.toml
+++ b/crates/defra-core/Cargo.toml
@@ -28,6 +28,7 @@ libipld.workspace = true
 # Cryptography (for CID generation)
 sha2.workspace = true
 rand.workspace = true
+zeroize = { version = "1.8", features = ["derive"] }
 
 # Async
 async-trait.workspace = true
--- a/crates/defra-core/src/encryption.rs
+++ b/crates/defra-core/src/encryption.rs
@@ -24,11 +24,14 @@ use rand::RngCore;
 use std::cell::RefCell;
 use std::collections::HashMap;
 use std::sync::Mutex;
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
 /// Wrapper for encryption key bytes with zeroization on drop.
 ///
-/// Ensures key material is cleared from memory when no longer needed.
-#[derive(Clone)]
+/// Uses `zeroize::ZeroizeOnDrop` to clear key material when dropped, with
+/// a memory fence after zeroing. This is the RustCrypto ecosystem standard
+/// and replaces a previous manual `unsafe { write_volatile }` implementation.
+#[derive(Clone, Zeroize, ZeroizeOnDrop)]
 pub struct EncryptionKey(Vec<u8>);
 
 impl EncryptionKey {
@@ -46,18 +49,6 @@ impl EncryptionKey {
 impl std::fmt::Debug for EncryptionKey {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "EncryptionKey([REDACTED; {} bytes])", self.0.len())
-    }
-}
-
-impl Drop for EncryptionKey {
-    fn drop(&mut self) {
-        for byte in self.0.iter_mut() {
-            // Use write_volatile to prevent the compiler from optimizing away the zeroing.
-            // SAFETY: We are writing to a valid, properly aligned mutable reference.
-            unsafe {
-                std::ptr::write_volatile(byte, 0);
-            }
-        }
     }
 }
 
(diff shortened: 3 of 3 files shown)


jackzampolin (approved):
Looks good, flagging @JesseAbram for awareness.

PR #785, opened 2026-04-14
fix(query): explicit permission-denied error on ACP single-doc reads (#551)

### LINKED ISSUE #551: Return explicit GraphQL permission-denied errors for ACP read denials
Backbone full-stack CI exposed an inconsistency in DefraDB's ACP denial shape on the GraphQL read path.

Observed behavior:
- In Backbone full-stack Step 16b, `INFERENCE_SVC` reads `Transcript` before receiving the `reader` relationship.
- The read is denied correctly in practice, but the GraphQL response comes back as an empty result set without an explicit permission/access-denied error.
- Backbone can only assert this as "empty means denied", which makes sad-path assertions weaker and can hide unrelated query bugs.

Expected behavior:
- ACP read denials should return a consistent explicit denial shape, ideally a GraphQL error with a stable message/code indicating permission denied.
- This should match the write-path behavior closely enough that integration tests can assert denial semantics directly instead of inferring from empty data.

Why this matters:
- We just tightened Backbone's full-stack assertions to require explicit denial shape and CI failed immediately at the pre-grant read check.
- The full-stack test has been relaxed again for now, but the stronger assertion is the right long-term contract.

Concrete failing context:
- Backbone CI run: 22972076100
- Failing test step: `tests/full_stack.rs` Step 16b (`INFERENCE_SVC should be denied BEFORE reader grant`)
- The regression is not authorization correctness; it is response shape.

Suggested acceptance:
- Query/read ACP denials return an explicit GraphQL error or other stable machine-checkable denial marker.
- Backbone can distinguish "permission denied" from "empty but allowed" and from unrelated query/runtime errors.

iverc:
## Summary

Closes #551. When a GraphQL read targets a specific document by ID (e.g. `User(docID: \"...\")`) and the caller lacks read permission, the query now returns an explicit `permission_denied` error in `errors[]` instead of silently filtering the document out and returning an empty result set.

Backbone full-stack CI hit this gap when an unauthorized service read a `Transcript` before being granted the `reader` relation. The silent empty result made denial assertions weak — Backbone could only check 'empty means denied', which hides unrelated bugs.

## What changes

- [crates/query/src/plan/permission_filter.rs](crates/query/src/plan/permission_filter.rs) — `PermissionFilterNode` now tracks which user-requested doc IDs were denied during iteration and surfaces them in a single `permission_denied` error after the source is drained.
- [crates/query/src/runner/plan.rs](crates/query/src/runner/plan.rs) and [crates/query/src/planner/builder/mod.rs](crates/query/src/planner/builder/mod.rs) — wire `select.doc_ids` into the filter node so it knows which IDs were explicitly requested.
- Join-time ACP wrapping passes `None` (joins follow doc IDs from parent traversal, not from the user request) — preserves existing behavior for nested reads.

## What does NOT change

- **Browse queries** (`{ User { ... } }` with no `docID`) keep their current filter-only semantics. Collection scans still return whatever the caller is permitted to see, so existing assertions like \"Bob sees 0 docs before grant\" continue to hold.
- **Mutations** were already explicit — this just brings the read path into parity.
- **Wire format / GraphQL response shape** — the error is delivered through the existing `errors[]` channel using `QueryError::permission_denied`, the same variant the mutation path uses today.

## Test plan

- [x] `cargo test -p query` — 210/210 passing
- [x] `cargo test -p integration-test --test acp` — **197/197 passing** (was 196 — the new regression test bumps the count)
- [x] New regression test [rust_acp_explicit_denial_on_single_doc_read](tools/integration-test/tests/acp/basic.rs) verifies both halves: browse query is silent, single-doc query errors with the requested doc ID in the message.

### Why Rust-only test

The new test is restricted to the Rust runtime (`#[tokio::test]` instead of `for_each_runtime!`) because Go DefraDB still returns empty results in this case — that's the upstream gap Backbone CI hit. Once Go gains parity the test can move to `for_each_runtime!`.

## Diff

8 files, 184 insertions, 11 deletions.

### DIFF
--- a/crates/query/src/plan/permission_filter.rs
+++ b/crates/query/src/plan/permission_filter.rs
@@ -3,6 +3,7 @@
 //! This node wraps a source node and filters out documents that the
 //! identity context doesn't have permission to read.
 
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use acp::{DocumentACP, DocumentPermission, Identity};
@@ -10,7 +11,7 @@ use async_trait::async_trait;
 use identity::Did;
 
 use crate::document::DocumentMapping;
-use crate::error::Result;
+use crate::error::{QueryError, Result};
 use crate::planner::{Doc, PlanNode};
 use crate::txn::check_doc_access_with_overlay;
 
@@ -40,6 +41,18 @@ pub struct PermissionFilterNode {
 
     /// Document mapping from source
     document_mapping: DocumentMapping,
+
+    /// Doc IDs the caller explicitly requested (e.g., via
+    /// `_docID: { _eq: "..." }` or `docID: "..."`). When non-empty and
+    /// at least one of these is denied, `next()` returns a
+    /// `permission_denied` error after the source is exhausted instead
+    /// of silently returning fewer results — matching Go's behavior and
+    /// closing the read-side gap reported in #551.
+    requested_doc_ids: HashSet<String>,
+
+    /// Subset of `requested_doc_ids` that were emitted by the source
+    /// but failed the read-permission check.
+    denied_requested_ids: Vec<String>,
 }
 
 impl PermissionFilterNode {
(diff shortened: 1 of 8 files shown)


jackzampolin:
I mean do we want to expose the information that something exists that they can't read? Is that desirable? I don't know. cc @JesseAbram

jackzampolin:
see `TestACP_AddWithIdentityAndReadWithWrongIdentity_CanNotRead` in the go impl. this isn't desired behavior. closing this pr as `wontfix`

Issue #775, opened 2026-04-13
identity: replace manual JWT implementation and der.rs with jwt-simple


JesseAbram:
### Summary                                                                                                                                                                                                                                
                                                                                                                                                                                                                                             
  `crates/identity/src/token/` is a fully manual JWT implementation spanning four files,                                                                                                                                                     
  including hand-rolled DER ↔ raw R||S conversion in `der.rs`. `jwt-simple` supports all                                                                                                                                                     
  three JWT algorithms the identity crate needs and handles format conversion internally.                                                                                                                                                    
  The manual implementation should be replaced with `jwt-simple`, matching Go's pattern
  of delegating JWT to a battle-tested library (`lestrrat-go/jwx/v2`).                                                                                                                                                                       
                                                                                                                                                                                                                                             
  ### Details                                                                                                                                                                                                                                
                                                                                                                                                                                                                                             
  The identity crate currently implements JWT entirely by hand with no JWT library                                                                                                                                                           
  dependency — only `base64` and `serde_json`:
                                                                                                                                                                                                                                             
  - `token/der.rs` — manual ASN.1 DER parser and encoder for ECDSA signatures.                                                                                                                                                               
    Required because RFC 7518 mandates raw R||S format in JWTs but the crypto crate                                                                                                                                                          
    produces DER. This is the highest-risk file in the crypto audit.                                                                                                                                                                         
  - `token/encoding.rs` — manual JWT signing: header construction, base64url encoding,
    signature dispatch                                                                                                                                                                                                                       
  - `token/decoding.rs` — manual JWT verification: base64url decoding, claim parsing,
    signature validation including DER conversion                                                                                                                                                                                            
  - `token/mod.rs` — manual claim validation: exp, nbf, aud, clock skew
                                                                                                                                                                                                                                             
  `jwt-simple` supports all three algorithms used by the identity crate:
                                                                                                                                                                                                                                             
  | Algorithm | Identity crate uses | jwt-simple supports |
  |-----------|---------------------|---------------------|
  | `ES256K`  | secp256k1 JWT signing/verification | ✓ via `k256` |
  | `EdDSA`   | Ed25519 JWT signing/verification   | ✓ via `ed25519-dalek` |                                                                                                                                                                 
  | `ES256`   | secp256r1 JWT signing/verification | ✓ via `p256` |                                                                                                                                                                          
                                                                                                                                                                                                                                             
[...]

PR #786, opened 2026-04-14
fix(identity): replace hand-rolled DER parser with k256/p256 mainline (#775)

iverc:
## Summary

Addresses #775. [crates/identity/src/token/der.rs](crates/identity/src/token/der.rs) was a ~250-line hand-rolled ASN.1 DER parser converting ECDSA signatures between DER and raw R||S format for the JWT layer. Manual ASN.1 parsing is a notorious vulnerability vector and the audit flagged this as the highest-risk file in the identity crate.

This PR replaces the manual parser with `k256::ecdsa::Signature` and `p256::ecdsa::Signature` mainline `from_der` / `to_der` / `from_slice` / `to_bytes` calls — the same RustCrypto types the `crypto` crate already uses for signing and verification. The unsafe manual parser is eliminated entirely while the JWT wire format stays identical (raw R||S per RFC 7518) and zero call-site signatures change.

## Scope deviation from the issue

The issue text proposes a full migration to the `jwt-simple` crate, which would also replace [encoding.rs](crates/identity/src/token/encoding.rs) and [decoding.rs](crates/identity/src/token/decoding.rs). After investigation I deliberately scoped this PR smaller for three reasons:

1. **`jwt-simple` defaults to BoringSSL** (`boring` crate, native dep with CMake/perl build requirements). Even with `pure-rust` feature it pulls `superboring`, a non-trivial extra dep tree.
2. **Key-type bridging**. We'd need a shim from `crypto::keys::PublicKey`/`PrivateKey` traits to `jwt-simple`'s key trait hierarchy. Not hard but adds surface area.
3. **Token-format compat risk**. `jwt-simple` may add or require fields that diverge from our current claim shape — needs comprehensive round-trip tests against existing tokens before deletion of the manual implementation.

The actual security value the issue calls out is in **killing the hand-rolled ASN.1 parser**, which this PR achieves directly with battle-tested k256/p256 mainline calls — the same code paths thousands of other Rust crypto users rely on. A follow-up PR can do the full `jwt-simple` migration if desired; I'll add a comment to the issue with this rationale.

## What changes

- [crates/identity/Cargo.toml](crates/identity/Cargo.toml) — add `k256.workspace = true` and `p256 = \"0.13\"` as direct deps. Both were already in the transitive graph via `crypto`.
- [crates/identity/src/token/der.rs](crates/identity/src/token/der.rs) — rewrite as four thin wrappers (`k256_der_to_raw`, `k256_raw_to_der`, `p256_der_to_raw`, `p256_raw_to_der`) around `k256`/`p256` ECDSA types. ~250 lines → ~115 lines.
- [encoding.rs](crates/identity/src/token/encoding.rs) and [decoding.rs](crates/identity/src/token/decoding.rs) — call key-type-specific helpers instead of the old generic `der_to_raw` / `raw_to_der`.
- New unit tests use real signatures from `k256::ecdsa::SigningKey` and `p256::ecdsa::SigningKey` fixtures to verify round-trip + rejection of garbage and wrong-length inputs.

## Test plan

- [x] `cargo test -p identity` — **all 26 token unit tests + 8 doc tests passing**
- [x] `cargo test -p integration-test --test identity` — **56/56 passing** (covers token round-trip across ES256K, EdDSA, ES256, plus negative cases like expired/malformed/wrong-signer)

## Diff

5 files, 89 insertions, 226 deletions. Net deletion of 137 lines plus the unsafe ASN.1 parser entirely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5881,6 +5881,8 @@ dependencies = [
  "crypto",
  "defra-core",
  "hex",
+ "k256",
+ "p256",
  "proptest",
  "serde",
  "serde_json",
--- a/crates/identity/Cargo.toml
+++ b/crates/identity/Cargo.toml
@@ -16,6 +16,10 @@ serde.workspace = true
 serde_json.workspace = true
 hex.workspace = true
 base64.workspace = true
+# ECDSA signature DER ↔ raw conversion, used by the JWT token layer.
+# Replaces a hand-rolled ASN.1 parser in `token/der.rs` (#775).
+k256.workspace = true
+p256 = "0.13"
 
 [dev-dependencies]
 proptest.workspace = true
--- a/crates/identity/src/token/decoding.rs
+++ b/crates/identity/src/token/decoding.rs
@@ -86,7 +86,7 @@ pub(crate) fn decode_secp256k1(token: &str) -> Result<IdentityClaims> {
     let claims = decode_claims(jwt.payload)?;
     let public_key = decode_public_key_from_claims(&claims, KeyType::Secp256k1)?;
     let raw_signature = decode_signature(jwt.signature)?;
-    let der_signature = der::raw_to_der(&raw_signature)?;
+    let der_signature = der::k256_raw_to_der(&raw_signature)?;
 
     let signing_input = format!("{}.{}", jwt.header, jwt.payload);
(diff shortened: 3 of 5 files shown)


jackzampolin:

PR #787, opened 2026-04-14
perf(query): parallelize ACP checks in PermissionFilterNode (#519)

### LINKED ISSUE #519: perf: Non-blocking ACP checks in query execution
## Context

Currently, `PermissionFilterNode.next()` blocks on each ACP check before advancing to the next document. This serializes query execution against network latency — the query executor sits idle waiting for each SourceHub response.

## Proposal

Decouple ACP check dispatch from result consumption. This is analogous to predicate pushdown in traditional databases, but applied to authorization: push expensive checks as early as possible, collect results as late as possible.

### Approach

1. **Fire early**: When `PermissionFilterNode` encounters a document, dispatch the access check asynchronously (or batch multiple checks). Mark the document as "pending ACP."

2. **Continue scanning**: The query executor continues fetching and processing subsequent documents while access checks are in flight.

3. **Resolve late**: Before aggregation, joins, or result emission, do a second pass to resolve pending ACP decisions. By this point, most checks have completed and results are cached locally.

### Integration Points

- `query/src/plan/permission_filter.rs` — `PermissionFilterNode::next()` becomes non-blocking
- Access cache (#517) — dispatched checks populate the cache; second pass reads from it
- Works with both TTL-based cache and event-driven cache (Phase 3)

### Batching Opportunity

If the query plan can look ahead (e.g., scan node produces a batch of doc IDs), we can dispatch `verify_access` for the entire batch in one request. This requires a batch API on the provider side:

```rust
async fn verify_access_batch(
    &self,
    policy_id: &str,
    resource: &str,
    object_ids: &[&str],
    permission: &str,
    actor_did: &str,
) -> Result<HashMap<String, bool>>;
```

### Complexity

This changes the execution model of `PermissionFilterNode` from synchronous filter to async pipeline stage. Needs careful design to avoid:
- Buffering too many documents in memory
- Complicating error propagation (fail-closed on ACP error)
- Breaking streaming result delivery

## Depends on

- #517 (access cache — async checks need somewhere to land)

## Related

- Bruno's description of reordering authorization checks in query plans, inspired by MPC protocol optimizations

iverc:
## Summary

Closes #519. Stops the query executor from serializing on per-document ACP checks.

Previously [PermissionFilterNode::next()](crates/query/src/plan/permission_filter.rs) awaited each `has_read_permission` call before advancing to the next doc. With remote backends like SourceHub (gRPC, ~50–100ms per check), a 1000-doc result set serialized into 50–100s of pure wait time. This PR drains the source eagerly in `start()`, dispatches every read-permission check concurrently via `futures::future::try_join_all`, and serves permitted docs from a pre-built `VecDeque` in `next()`.

## Design choice

Three options were on the table; this PR implements **drain-and-parallelize**:

- **A — drain in start(), parallelize all (this PR)** ✅ Single-file diff, full per-query parallelism, zero planner contract change. Matches the fact that ScanNode already materializes `Vec<Doc>` in `init()` — the \"streaming\" of `PlanNode::next()` is fictional at the bottom of the plan tree.
- **B — page-batched dispatch in next()** — Halfway house. Useful only if a future ScanNode actually starts streaming. Adds complexity without payoff today.
- **C — `FuturesUnordered` with in-flight cap** — Textbook async best-practice for I/O parallelism. But it requires wrapping `PlanNode` sources as `Stream`s, a planner-wide refactor. Noted inline as the future evolution path if/when scans become true streams.

## What changes

Single file: [crates/query/src/plan/permission_filter.rs](crates/query/src/plan/permission_filter.rs)

- Add `permitted_docs: VecDeque<Doc>` field
- Rewrite `start()`: drain source → collect `(doc_id, doc)` pairs → dispatch all `has_read_permission` calls via `futures::future::try_join_all` → walk decisions, queue permitted, track denied explicit-id requests
- Rewrite `next()`: pop from queue; if empty and any explicit-id request was denied, return `permission_denied` error (preserves #551 semantics)
- Inline comment documenting the streaming-evolution path

Net: +74/-38 lines in one file.

## Layering on PR #785 (#551)

This PR is based on `fix/acp-read-denial-error` (PR #785) because it builds on the `requested_doc_ids` / `denied_requested_ids` machinery from #551. When #785 merges, this branch rebases cleanly. If this merges first, #785 needs a trivial rebase.

## Test plan

- [x] `cargo test -p query` — 210/210 passing
- [x] `cargo test -p integration-test --test acp` — **197/197 passing**, including:
  - `basic::rust_acp_basic` — browse-query silent filtering still works
  - `basic::rust_acp_explicit_denial_on_single_doc_read` (from #785) — single-doc denial still errors
  - `multi_identity::*`, `relationship::*`, `p2p::*` — multi-actor scenarios still pass

## Perf characterization

The latency hotspot is the SourceHub gRPC layer, not exercised by local tests. Functional correctness verified end-to-end. Expected order-of-magnitude speedup on real SourceHub deployments: a query yielding N docs goes from O(N × RTT) to O(RTT), bounded only by gRPC channel concurrency.

### DIFF
--- a/crates/query/src/plan/permission_filter.rs
+++ b/crates/query/src/plan/permission_filter.rs
@@ -3,6 +3,7 @@
 //! This node wraps a source node and filters out documents that the
 //! identity context doesn't have permission to read.
 
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use acp::{DocumentACP, DocumentPermission, Identity};
@@ -10,7 +11,7 @@ use async_trait::async_trait;
 use identity::Did;
 
 use crate::document::DocumentMapping;
-use crate::error::Result;
+use crate::error::{QueryError, Result};
 use crate::planner::{Doc, PlanNode};
 use crate::txn::check_doc_access_with_overlay;
 
@@ -40,6 +41,18 @@ pub struct PermissionFilterNode {
 
     /// Document mapping from source
     document_mapping: DocumentMapping,
+
+    /// Doc IDs the caller explicitly requested (e.g., via
+    /// `_docID: { _eq: "..." }` or `docID: "..."`). When non-empty and
+    /// at least one of these is denied, `next()` returns a
+    /// `permission_denied` error after the source is exhausted instead
+    /// of silently returning fewer results — matching Go's behavior and
+    /// closing the read-side gap reported in #551.
+    requested_doc_ids: HashSet<String>,
+
+    /// Subset of `requested_doc_ids` that were emitted by the source
+    /// but failed the read-permission check.
+    denied_requested_ids: Vec<String>,
 }
 
 impl PermissionFilterNode {
@@ -67,7 +80,20 @@ impl PermissionFilterNode {
             resource_name: resource_name.into(),
             current_doc: Doc::default(),
             document_mapping,
+            requested_doc_ids: HashSet::new(),
+            denied_requested_ids: Vec::new(),
         }
+    }
+
+    /// Mark a set of document IDs as explicitly requested by the caller.
+    ///
+    /// When any of these IDs are emitted by the source but denied by ACP,
+    /// `next()` will return a `permission_denied` error after the source is
+    /// drained instead of silently filtering them out. Browse queries
+    /// (no explicit doc IDs) keep the existing filter-only semantics.
+    pub fn with_requested_doc_ids(mut self, doc_ids: Vec<String>) -> Self {
+        self.requested_doc_ids = doc_ids.into_iter().collect();
+        self
     }
 
     /// Create from an optional DID for backward compatibility.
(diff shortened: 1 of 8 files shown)


iverc:
Note: this PR is stacked on top of #785 (`fix/acp-read-denial-error`). Until #785 merges, the diff here shows both sets of changes together. Once #785 lands on main, I'll rebase this branch and the diff will clean up to just the #519 parallelization change (~74/-38 lines in `permission_filter.rs`).

Please merge #785 first for a clean review here.

jackzampolin:
I found one blocker in the incremental diff. `PermissionFilterNode` now dispatches ACP reads with unbounded concurrency via `try_join_all` in `crates/query/src/plan/permission_filter.rs`, which can turn a latency bottleneck into connection-pool exhaustion, rate limiting, or memory pressure against remote ACP backends on large result sets. I would cap concurrency before merging.

jackzampolin:
https://github.com/sourcenetwork/defradb.rs/pull/785#issuecomment-4244542216

Issue #670, opened 2026-04-10
Split query crate (50K lines) into parse/plan/execute sub-crates


iverc:
## Summary

The query crate is the largest at 50K lines. Parsing, planning, and execution are all in one compilation unit, preventing parallel crate compilation.

## Details

Module breakdown by approximate size:
- `plan/` — ~12K lines (22 plan node implementations)
- `runner/` — ~10K lines (query execution, mutation dispatch)
- `planner/` — ~8K lines (plan building, index selection)
- `mapper/` — ~6K lines (query type definitions)
- `schema_gen/` — ~4K lines (GraphQL schema generation)
- `rest/` — ~3K lines (REST API operations)
- `query_parse/`, `sdl_parse/` — ~3K lines (parsing)

## Expected Behavior

Split into sub-crates to enable parallel compilation:
- `query-parse`: GraphQL/SDL parsing + schema generation
- `query-plan`: Plan nodes and planner
- `query`: Runner, executor, mutation dispatch

## Suspected Cause

Go's `internal/planner/` is a single package. The Rust port kept everything in one crate.

## Acceptance Criteria

- [ ] Identify clean boundaries between parse/plan/execute
- [ ] Extract parsing modules first (fewest dependencies)
- [ ] Plan nodes extracted next
- [ ] All integration tests pass
- [ ] Compile time measurably reduced

## Files

- `crates/query/src/lib.rs`

jackzampolin:
Validated against `origin/main` on 2026-04-10 (`d7ff7f2660`). There is currently no diff between local `HEAD` and `origin/main` under `crates/query`, so the assessment below reflects current main.

Current reality in the code:

- `crates/query` is still a single workspace crate that exposes parse/planner/plan/runner/rest/SDL functionality from one facade (`crates/query/src/lib.rs`).
- The size motivation is still broadly real, but the issue’s module-size breakdown is stale. On current main I measured roughly:
  - `plan/`: 12.6k LOC
  - `runner/`: 12.2k LOC
  - `planner/`: 8.8k LOC
  - `sdl_parse/`: 4.8k LOC
  - `query_parse/`: 4.0k LOC
  - `mapper/`: 4.0k LOC
  - `rest/`: 0.8k LOC
  - `schema_gen/`: 0.5k LOC
- The boundaries are not clean yet:
  - `runner/executor.rs` directly parses and validates requests via `query_parse`.
  - `planner` builds concrete nodes from `plan`.
[...]

PR #788, opened 2026-04-14
refactor(query): extract query-plan crate (#670)

iverc:
## Summary

Closes the next slice of #670 (query crate split). Previous PRs shipped `query-types` (PR #736) and `query-parse` (PR #736). This PR extracts the plan layer into a new `query-plan` crate, leaving the runner/executor/rest/subscription in the main `query` crate.

## What moved

All under new `crates/query-plan/src/`:

| Module | Role |
|---|---|
| `plan/` | Plan node implementations (scan, select, limit, joins, permission filter, mutations, aggregates, ...) |
| `planner/` | Plan builder + index selection |
| `fetcher.rs` | `DocFetcher` trait + result types |
| `mutator.rs` | `DocMutator` trait + result types |
| `txn/` (minus `guard.rs`) | `TransactionContext`, `TransactionHandle`, `TransactionRegistry`, `check_doc_access_with_overlay` — plan-layer primitives |

## What stayed in `query`

- `executor`, `runner`, `rest`, `subscription`, `test_utils`
- `txn/guard.rs` (`TransactionGuard`) — generic over `QueryExecutor` which lives in the runner layer. Moving it would create a `query_plan → query → query_plan` cycle. The `query::txn` module re-exports plan-layer txn types from `query_plan::txn` and adds `TransactionGuard` locally, so the public `query::txn::*` path is unchanged.

## Public API unchanged

[crates/query/src/lib.rs](crates/query/src/lib.rs) does `pub use query_plan::{fetcher, mutator, plan, planner}` plus the local `txn` module. Every downstream crate (`cli`, `embedded`, `defra-node`, `ffi`, `wasm`, `http`, `pg-compat`, `db`, `db-merge`, integration tests) builds unchanged — no import-site edits anywhere else in the workspace.

## Small internal plumbing fixes

- plan/planner modules now import shared types from `query_types` (`document`, `error`, `mapper`, `json_convert`) instead of the crate-local re-exports they used to have
- `Planner::fts_score_key` became `pub` because it's now called across a crate boundary from `query::runner::nested_fulltext` (previously `pub(crate)`, which now only spans `query-plan`)
- One match in `query::runner::executor::execute_in_txn` adds a `_` wildcard arm for `GetTransactionResult` — the `#[non_exhaustive]` attribute is cross-crate strict

## Test plan

- [x] `cargo check --workspace` clean on first pass (no downstream breakage)
- [x] `cargo test --workspace --exclude integration-test --exclude ffi-test --exclude defra-wasm` — all green
- [x] `cargo test -p integration-test --test acp` — **196/196 passing**

## Diff

89 files touched, +359/−575 lines net (the -575 is mostly due to `query::txn::mod.rs`'s test module moving alongside `guard.rs`). No logic changes — pure mechanical extraction plus the three small plumbing fixes above.

## Remaining #670 work

The split is functionally done. `query` crate is now down to ~5K LOC covering just the runner/executor/rest layer; `query-plan` holds ~10K LOC of plan nodes + trait abstractions. Future cleanup could reduce the `pub` surface in `query-plan` or split the runner further, but this PR achieves the parse/plan/execute separation the issue proposed.


jackzampolin:

PR #793, opened 2026-04-14
refactor(db): extract db-backup crate (#789)

iverc:
## Summary

Closes #789. First of the #669 sub-slices.

Moves the backup subsystem (`export_database`, `import_database`) out of `crates/db/src/backup/` into a new `crates/db-backup/` crate (~599 LOC). Smallest self-contained lift of the remaining decomposition work.

## What changes

- New crate `crates/db-backup/`, depends on `db` one-way (no cycle)
- `crates/db/src/backup/` → `crates/db-backup/src/backup/`
- Removed `pub mod backup` from `db/src/lib.rs`
- Updated 4 callsites to import from `db_backup::*` instead of `db::backup::*`:
  - `cli/src/backup_adapter.rs` (2 sites)
  - `ffi/src/backup/export.rs`
  - `ffi/src/backup/import.rs`
- Added `db-backup` dep to `cli` and `ffi` Cargo.toml

Crate surface is unchanged: `export_database`, `import_database`, `ImportStats`, plus the reusable classification helpers (`classify_schema_fields`, `compute_doc_id_new`, `json_to_graphql_input`, `FieldInfo`).

## Test plan

- [x] `cargo check --workspace` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --all` applied
- [x] `cargo test -p integration-test --test backup` — **11/11 passing**

## Diff

6 files moved, ~10 lines of import rewrites in callers, 1 new `Cargo.toml`, 1 new `lib.rs`. Pure mechanical lift.

Related: #669 (parent epic), #788 (parallel query-plan extraction).

Issue #791, opened 2026-04-14
refactor(db): extract db-migration crate (#669 sub-slice)


iverc:
Sub-slice of #669 (decompose db crate). The schema migration subsystem is self-contained with no external consumers.

## Scope

- Move `crates/db/src/migration/` into a new `crates/db-migration/` crate (~980 LOC)
- All current consumers are internal to `db` — no downstream crate updates needed
- Depends on `schema` and `lens` which are already separate crates

## Effort

Small (~2-3 hours). Mechanical lift. No API boundary design needed.

## Verification

- Workspace builds clean
- `cargo test -p integration-test --test query -- lens` passes (migration is exercised by lens transform tests)
- Schema migration still works end-to-end via `defradb client schema migration set`

Related: #669 (parent).

iverc:
Blocked by Rust orphan rule — migration adds methods to DB<S> via 3 impl blocks. See correction on #669. Reopening would require a trait-seam refactor first.

Issue #669, opened 2026-04-10
Meta: Decompose db crate into smaller architecture units


iverc:
## Summary

`crates/db` is still the second-largest crate in the workspace and currently acts as a broad coordination crate for database core, mutation flows, indexing, schema loading, migrations, NAC, search/embedding/searchable-encryption, backup/downsample, and the DB-side implementations used by P2P.

On current `main`, this is still real architecture debt, but it is not a single extraction task. The work should be tracked as a meta issue and split into smaller follow-up refactors with explicit seams, dependency analysis, and compile-time measurements.

## Current Reality

Validated against `origin/main` at `d7ff7f26601e36059547d625f60425cea866faa8`.

- `crates/db/src` is about `34,064` Rust LOC.
- `crates/db/src/lib.rs` currently declares `46` top-level modules.
- The crate owns both core DB behavior and several cross-cutting implementation areas:
  - Core DB: `database`, `txn`, `collection*`, `collection_loader`, `collection_provider`
  - Mutation flows: `auto_commit_mutator`, `doc_mutator`, `broadcast_mutator`
  - Block handling: `block_builder`, `block_verify`, `commit_priority_index`
  - Replication integration: `merge_handler`, `acp_merge_handler`, `head_provider`, `push_docs*`, `peer_identity`
  - Search / indexing: `dense_search`, `embedding`, `se`, `index_manager`
  - Schema / migration: `schema_loader`, `definition_validation`, `migration`, `lens_utils`
  - Other subsystems: `backup`, `dump`, `nac`, `downsample`, `view_ops`

## What This Issue Tracks

This issue tracks decomposition of `crates/db` into smaller architectural units, which may be separate crates or narrower internal service boundaries introduced before any crate moves.

This issue should remain the umbrella tracker, while implementation work happens in smaller follow-ups.

## Why This Is Not A Single-Step Split

The current `db` crate is highly connected internally.

In particular, the DB-side replication modules are not an obviously isolated first extraction:
- `merge_handler`, `acp_merge_handler`, `broadcast_mutator`, `head_provider`, `push_docs*`, and `peer_identity` are tightly coupled to `DB`, `Collection`, `IndexManager`, `block_builder`, ACP hooks, and SE behavior.
- Those APIs are also consumed from downstream crates including `cli`, `embedded`, `defra-node`, and `ffi`.
- There is already a separate `crates/p2p` transport/sync crate, so the remaining replication code in `db` is mostly database implementation, not just transport logic.

[...]

PR #794, opened 2026-04-14
refactor(schema): move definition_validation from db to schema crate (#669 slice)

iverc:
## Summary

Sub-slice of #669. Moves `crates/db/src/definition_validation/` (~947 LOC) into `crates/schema/src/definition_validation/`.

## Pivot from #791

The originally-planned follow-up slice #791 (`db-migration` extraction) turned out to be blocked: the migration subsystem adds methods to `DB<S>` via `impl<S: Store> DB<S>` blocks, and Rust's orphan rule prevents those from moving to an external crate without a significant API break or trait-seam refactor. Same constraint blocks `downsample/` and `patch/`, which I had also earmarked as easy wins.

This PR takes the *actually* easy slice instead: `definition_validation/`. It has:
- Zero `impl DB<S>` blocks
- Zero `use crate::` imports outside the module (no coupling to `crate::txn`, `crate::error`, etc.)
- Only touches types already defined in `schema` (`CollectionVersion`, `FieldKind`, `CType`, `ScalarKind`, `ORPHAN_COLLECTION_ID`)

So rather than making a new crate, the module is **moved into the `schema` crate** where it conceptually belongs. `schema` already has a `validation` module for single-collection SDL validation; `definition_validation` handles multi-collection patch validation, which is a natural sibling.

## What changes

- `crates/db/src/definition_validation/` → `crates/schema/src/definition_validation/`
- `schema/src/lib.rs` adds `pub mod definition_validation`
- `db/src/lib.rs` drops `pub mod definition_validation`
- 8 call-site rewrites from `db::definition_validation::*` → `schema::definition_validation::*`:
  - `cli/src/schema_adapter.rs`
  - `defra-node/src/db_impls.rs`
  - `embedded/src/node.rs`
  - `ffi/src/mobile.rs`, `ffi/src/schema.rs` (×2)
  - `db/src/patch/mod.rs`, `db/src/patch/store.rs`, `db/src/txn_registry.rs` (internal)
- `embedded/Cargo.toml` adds `schema` dep (the other callers already had it)
- Cleaned up 4 dead `_` wildcard arms in `helpers.rs` — they existed because the match was over external non-exhaustive enums; now that the enums are crate-local they're exhaustive and the wildcards are dead code

Net: −947 LOC from db, +947 in schema (same content). db crate is now ~22.9K LOC down from 23.8K.

## Test plan

- [x] `cargo check --workspace` clean
- [x] `cargo test -p integration-test --test query` — 27/27 passing (schema validation is exercised in query tests)

## Note on #791

#791 (db-migration extraction) is blocked by the orphan rule as described above. I'll update #669 with an honest status — the three "easy wins" I flagged in my earlier scope comment (backup, downsample, migration) turn out to be one easy win (backup, done in #793), one impossible (downsample, blocked), and one impossible (migration, blocked). Further db decomposition will need trait-seam refactors first.

Related: #669 (parent), #793 (db-backup), #788 (query-plan parallel extraction).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4371,6 +4371,7 @@ dependencies = [
  "libp2p",
  "p2p",
  "query",
+ "schema",
  "serde",
  "serde_bytes",
  "serde_ipld_dagcbor",
--- a/crates/cli/src/schema_adapter.rs
+++ b/crates/cli/src/schema_adapter.rs
@@ -84,7 +84,7 @@ impl<S: Store + 'static> SchemaAdapter<S> {
         let collections = query::parse_sdl_with_known_types(sdl, known_types)
             .map_err(|e| format!("failed to parse SDL: {}", e))?;
 
-        db::definition_validation::validate_new_collections(&collections)
+        schema::definition_validation::validate_new_collections(&collections)
             .map_err(|e| format!("failed to validate schema: {}", e))?;
 
         // #746: validate every @policy directive against the ACP store
--- a/crates/db/src/lib.rs
+++ b/crates/db/src/lib.rs
@@ -61,7 +61,8 @@ pub(crate) mod collection_snapshot;
 mod commit_priority_index;
 pub(crate) mod commits_fetcher;
 pub mod database;
-pub mod definition_validation;
+// Definition validation moved to the `schema` crate (was #791 substitute
+// slice) — see `schema::definition_validation`.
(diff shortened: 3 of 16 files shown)


jackzampolin:
The move looks internally coherent, but I do see one compatibility risk: this removes the public `db::definition_validation` module from `db` and rewires callers to `schema::definition_validation::*`. If anything downstream imports `db::definition_validation::*`, this is a compile break. I would only merge as-is if that public API break is intentional.

Issue #796, opened 2026-04-14
Epic: finalize db/query decomposition (follow-up to #669, #670)


iverc:
## Epic: finalize db/query decomposition

Follow-up to #669 and #670. The prior decomposition wave extracted `db-blocks`, `db-nac`, `db-merge`, `db-backup`, `db-index`, moved `definition_validation` into `schema`, and extracted `query-plan`. db is now ~22.2K LOC (down from ~34K), query is ~27K LOC (down from ~50K). This epic finishes the structural rework.

## Scope

- **Unified error**: new `defra-errors` crate as the shared error base for db, query, and extracted subsystems.
- **Transaction abstraction**: new `db-txn` crate holding `DbTxn<S>` and a narrow `TransactionView<S>` trait with the 5 methods the blocked fetchers actually need (`headstore`, `blockstore`, `datastore`, `systemstore`, `discard`).
- **Fetcher extractions**: `commits_fetcher`, `lensed_fetcher`, `lensed_auto_commit_fetcher` move out of db into new crates using `TransactionView` + `defra-errors`.
- **Patch extraction**: `patch_collection` (clean leaf) moves into `db-patch` as free functions. No extension trait, no BC shim.
- **Dense search / embedding**: structurally decided in #792, then extracted.
- **db::Error cleanup**: thin db::Error after the structural work lands.

## Out of scope

- `downsample`, `migration`, and the bulk of `patch` stay in db — too tangled with db-internal consumers (auto_commit_mutator, lensed_fetcher, txn_registry, database.rs).
- No facade refactor (no `db-core` crate, no `DB<S>` relocation).
- No further runner-side query decomposition beyond what #670 / #788 delivered.

## North-star gates

Every PR in this epic must preserve:

1. **Go parity** — identical GraphQL results, CRDT convergence, P2P wire format, CLI output, HTTP response shapes.
2. **Correctness** — structural rework must be behavior-preserving.
3. **Security** — no regressions in ACP, NAC, signature verification, or encryption paths.

No backwards compatibility is required within the Rust workspace. Import paths, method vs free-function syntax, Error enum shapes, and internal APIs can all be rewritten freely. db has no external (non-workspace) consumers.

## Phases (checklist)

- [ ] **Phase 0** — `defra-errors` foundation (1 PR)
- [ ] **Phase 1** — `db-txn` crate + `TransactionView<S>` trait (1 PR)
[...]

PR #795, opened 2026-04-14
refactor(db): extract db-index crate (#669 slice)

iverc:
## Summary

Third sub-slice of #669. Moves `crates/db/src/index_manager/` (~726 LOC) into a new `db-index` crate.

## Context on error-type handling

Unlike #793 (db-backup) and #794 (definition_validation to schema), `index_manager` needed a small error-type refactor. The module used db's comprehensive `Error` enum, which would have created a circular dep if db-index tried to pull db back in.

Solution: db-index defines its own narrow `Error` with just the three variants the module uses (`Storage`, `InvalidDocument`, `Other`), plus a `From<db_index::Error> for db::Error` impl in the db crate so existing call sites continue to thread errors through the broader hierarchy unchanged.

## What changes

- New crate `crates/db-index/`, depends only on `datastore`/`document`/`schema`/`storage` (no cycle back to db)
- `crates/db/src/index_manager/` → `crates/db-index/src/index_manager/`
- `crates/db-index/src/error.rs` — narrow Error with `thiserror #[from]`
- `crates/db/src/error.rs` — `From<db_index::Error> for db::Error` conversion
- `crates/db/src/lib.rs` — `pub use db_index as index_manager` so external call sites keep their existing `db::index_manager::*` and `db::IndexManager` import paths unchanged
- `fulltext_index_name` promoted from `pub(crate)` to `pub` (now re-exported across a crate boundary)

## API compatibility

All four external call sites keep their existing imports:
- `cli/src/index_adapter.rs` — `use db::IndexManager;` ✓
- `ffi/src/index/create.rs` — `db::index_manager::IndexManager::from_collection(...)` ✓
- `ffi/src/index/manage.rs` — `db::index_manager::IndexManager::from_collection(...)` ✓
- `db-merge/src/merge_handler/mod.rs` — `use db::index_manager::IndexManager;` ✓

Only one internal test fixup needed:
- `crates/db/tests/index_manager_tests.rs` — `use db::Error` → `use db::index_manager::Error` because `result.unwrap_err()` now returns the narrow `db_index::Error`, not the broad `db::Error`. The test's `Error::InvalidDocument(_)` pattern matches the narrow type directly.

## Test plan

- [x] `cargo check --workspace` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --all` applied
- [x] `cargo test -p db` — all unit + integration tests passing (including the 3 `index_manager_tests.rs` tests that matched on the Error variant)

[...]

PR #797, opened 2026-04-14
refactor(errors): defra-errors crate — unified subsystem error (Phase 0 of #796)

iverc:
## Summary

First phase of the db/query decomposition follow-up epic #796. Introduces a shared error type (`defra_errors::Error`) that every extracted subsystem crate will use from day one, eliminating the per-crate narrow-Error + `From`-boilerplate pattern that `db-index` (#795) had to invent.

## What this PR does

- New crate `crates/defra-errors/` at the storage layer, depends only on `defra-core`, `document`, `schema`, `storage`, `datastore`
- Defines `defra_errors::Error` as a thiserror enum with:
  - `#[from]` wrappers for foundation errors (Storage, Datastore, Schema, Document, Core)
  - Shared free-form variants (InvalidDocument, DocumentNotFound, CollectionNotFound, NotFound, AlreadyExists, InvalidArgument, Other)
  - Constructor helpers for each variant
  - `#[non_exhaustive]` so new variants don't break downstream matches
- `From<defra_errors::Error> for db::Error` — bridges subsystem errors into db's broader error hierarchy
- `From<defra_errors::Error> for query_types::QueryError` — same for query-layer consumers. Flattens structured wrappers into `Execution`/`Internal` since QueryError's `Storage`/`Schema` variants take different inner types than defra-errors' broader `storage::Error`/`schema::SchemaError`.

## What this PR does NOT do

- Does **not** reshape `db::Error` or `QueryError` to move variants out. That's Phase 7 (the existing db::Error cleanup pass in the epic). This PR is purely additive so Phases 1-6 can consume the unified error without waiting on the variant migration.
- **Zero behavior change.** All existing errors flow through existing code paths unchanged. Display output and Debug chains are identical for everything except new subsystem-originated errors (which don't exist yet until Phase 2+ lands).

## Why this approach (vs reshape-in-one-PR)

The plan allows a flag-day reshape of db::Error since there's no BC requirement. I chose additive-only for Phase 0 because:

1. Phases 1-6 only need `defra_errors::Error` to *exist* — they don't need db::Error to be thinned. Deferring the reshape keeps each PR smaller and easier to review.
2. The right time to audit what stays in db::Error vs what moves out is *after* all extractions land, when we can see which variants still have producers. That's Phase 7.
3. A smaller, additive Phase 0 de-risks the whole epic — if Phase 0 breaks something subtle about error propagation, the fix is localized.

## Test plan

- [x] `cargo check --workspace` clean
- [x] `cargo test -p defra-errors` — 4/4 unit tests (Display format, constructors, source chain via `std::error::Error::source()`)
- [x] `cargo test -p db -p query-types` — all passing (additive change)

## Diff

8 files, +258 lines, 0 deletions. New crate + two bridge impls.

## Next

- **Phase 1**: extract `db-txn` crate with `TransactionView<S>` trait (1 PR) — unblocks fetcher extractions
- **Phase 2-3**: extract `db-commits-fetcher`, `db-lensed-fetcher` (2 PRs)
- **Phase 4**: extract `db-patch` as free functions (1 PR, parallel with Phase 1)
- **Phases 5-8**: dense_search design + extract + final cleanup

Related: #796 (epic), #669, #670.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3559,6 +3559,7 @@ dependencies = [
  "db-blocks",
  "db-nac",
  "defra-core",
+ "defra-errors",
  "document",
  "events",
  "futures",
@@ -3734,6 +3735,18 @@ dependencies = [
  "thiserror 1.0.69",
  "tracing",
  "uuid",
+]
+
+[[package]]
+name = "defra-errors"
+version = "0.5.0"
+dependencies = [
+ "datastore",
+ "defra-core",
+ "document",
+ "schema",
+ "storage",
+ "thiserror 1.0.69",
 ]
 
 [[package]]
@@ -9989,6 +10002,7 @@ dependencies = [
  "async-trait",
  "chrono",
  "defra-core",
+ "defra-errors",
  "document",
  "schema",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10002,7 +10002,6 @@ dependencies = [
  "async-trait",
  "chrono",
  "defra-core",
- "defra-errors",
  "document",
  "schema",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 resolver = "2"
 members = [
     "crates/defra-core",
+    "crates/defra-errors",
     "crates/crdt",
     "crates/storage",
     "crates/blockstore",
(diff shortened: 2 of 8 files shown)


iverc:
Closing as no longer needed. This PR was scaffolding for Phases 1-3 (fetcher extractions via a shared error type), but those phases were abandoned after discovering the fetchers own `Arc<Mutex<Option<DbTxn<S>>>>` and can't be cleanly extracted.

Phase 4 (db-patch) will depend on db directly and use db::Error. Phase 7 (db::Error cleanup) can reshape in place without needing a new crate. If a future extraction actually needs a shared error type across multiple subsystems, we can revisit.

PR #798, opened 2026-04-14
refactor(db): extract db-search crate for dense search + embedding (#796, closes #792)

### LINKED ISSUE #792: design: vector search / embedding crate boundary (#669 sub-slice)
Sub-slice of #669 (decompose db crate). Unlike the other remaining slices, dense_search + embedding need API boundary **design** before any code moves.

## Current state

- `crates/db/src/dense_search.rs` (543 LOC) — vector search runtime with runtime hooks into query planner (via `SEFilterNode`)
- `crates/db/src/embedding.rs` (344 LOC) — HTTP client for the embedding API, called from one external consumer: `defra-node/src/dense_search.rs` via `db::embed_text`

## Why not just mechanically extract

Several open design questions need resolution first:

1. **Vector search seam** — `dense_search.rs` has bidirectional hooks:
   - Runtime: `SEFilterNode` in the query planner calls into HMAC-SHA256 tag generation, which is already in the `db-merge` crate (post-#735). There's an informal cross-crate seam.
   - Storage: dense search reads index manager state and embedding storage, which are db-internal.
   - Question: does vector search become a separate crate (`db-search`, `vector-search`)? Does it merge with `db-merge`'s SE code?

2. **Embedding client seam** — `embedding.rs` is an HTTP client for the embedding API. It's ~344 LOC of protocol + error handling, and has exactly one consumer: `defra-node/src/dense_search.rs`. Question: does it belong in its own `embedding-client` crate, inside `defra-node`, or stay in `db` as a utility?

3. **Searchable encryption coupling** — SE tag generation is in `db-merge` but the SE filter node is in query-plan and the vector search pipeline is in db. This cross-crate spread is hard to reason about. Should these be unified?

## Deliverable

A design doc / issue comment proposing:
- Target crate layout for vector search and embedding
- Public API surfaces for each new crate
- Migration steps (with refactor prep PRs if needed)
- Decision on SE coupling

Once the design is agreed, file a separate implementation issue.

## Effort

Large (design-heavy). No code lands from this issue directly.

Related: #669 (parent), #735 (db-merge extraction).

iverc:
## Summary

Phase 5 + Phase 6 of the db/query decomposition epic #796, combined into a single PR because the design answer was straightforward and the extraction is mechanical.

Moves `dense_search.rs` and `embedding.rs` out of db into a new `db-search` crate, along with `EmbeddingClientConfig`. Closes #792.

## Why this works cleanly (unlike Phases 1-4)

Unlike the fetcher and patch-module extractions that I attempted and dropped, dense_search + embedding are genuinely extractable:

- **Already free functions.** No `impl<S: Store> DB<S>` blocks to unwind. `hybrid_search_dense`, `embed_text`, `set_embedding` are all top-level `pub async fn`s today.
- **Take plain config + external trait.** `&EmbeddingClientConfig` (plain struct with 3 pub fields) and `&E: query::QueryExecutor` (external trait). No `&self`, no `&DB<S>`.
- **Touch zero DB private state.** No private field accesses, no private method calls. The modules already operate through public interfaces.
- **One external consumer.** `defra-node/src/dense_search.rs` is the only out-of-db caller. It already imports via the `db::` re-export path.

db-search depends on `query`, `schema`, `document`, `reqwest`. No dependency back to `db`. db depends on db-search cleanly.

## Dense vs SE coupling (the #792 design question)

The open question in #792 was: does dense_search merge with db-merge's searchable-encryption (SE) tag generation, or stay separate?

**They stay separate.** Looking at the actual code:

- SE lives in db-merge (extracted in #735). It handles HMAC-SHA256 tag generation for encrypted indices so the server can search without decryption.
- Dense search handles vector similarity for RAG-style retrieval via embedding APIs.
- They share zero types and zero code paths. The only thing they have in common is being "non-trivial search features".

Merging them into a single db-search-and-encrypted crate would be forced unification — a larger blast radius with no shared logic to justify it. Each stays in its own crate.

## What's moved

- `crates/db/src/dense_search.rs` → `crates/db-search/src/dense_search.rs` (543 LOC)
- `crates/db/src/embedding.rs` → `crates/db-search/src/embedding.rs` (344 LOC)
- `EmbeddingClientConfig` from `db::database` → `db-search::config::EmbeddingClientConfig` (40 LOC)

Net: ~930 LOC out of db into db-search.

## Internal callers updated

Inside db, 5 callsites in `auto_commit_mutator/` rewrite `crate::embedding::set_embedding` → `db_search::set_embedding`. That's the entire surface of internal coupling.

## External compat

- `db::EmbeddingClientConfig` still works via `pub use db_search::EmbeddingClientConfig` in `db::database`.
- `db::dense_search::*` still works via `pub use db_search as dense_search` in `db::lib`.
- `db::embed_text`, `db::hybrid_search_dense`, etc. still work via transparent re-exports in db::lib.
- defra-node's existing imports need zero changes.

## Test plan

- `cargo check --workspace` clean
- `cargo test -p db -p db-search` — all passing
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo fmt --all` applied

## Cumulative progress on #796

This is the last solid extraction the session produces. After this:

- db was at ~22.2K LOC pre-epic → ~21.3K after this PR
- Earlier wave extracted ~12K LOC (db-blocks, db-nac, db-merge, db-backup, db-index, definition_validation → schema)
- The remaining ~21K in db is genuinely tangled with DB<S> private state; further extraction needs either a facade refactor or accepted API widening

Related: #796 (epic), #792 (dense_search design), #669 (parent decomposition meta).


jackzampolin:

PR #823, opened 2026-04-15
feat(embedded): idempotent EmbeddedNode::shutdown API (#813)

### LINKED ISSUE #813: Expose an idempotent EmbeddedNode shutdown API for clean embedded teardown
defra-agent-desktop embeds defra-node directly and needs to stop embedded nodes cleanly on app exit and in integration tests. Today the public `EmbeddedNode` surface exposes query/schema/P2P operations, but there is no supported shutdown/close hook for embedders.

That leaves downstream apps relying on drop behavior and can produce unclean IROH teardown such as `Endpoint dropped without calling Endpoint::close. Aborting ungracefully.`

What we need upstream:
- a public, idempotent `EmbeddedNode::shutdown()` API
- shutdown should close P2P first, then the event bus, then the database
- embedders should be able to await shutdown completion before dropping the node

Concrete downstream use case:
- defra-agent-desktop spins up multiple embedded nodes in app-level UI tests and live inference smoke tests
- those tests need deterministic teardown so the process exits cleanly and does not hang or emit endpoint-drop warnings

I have a local downstream patch proving the shape is workable, but this needs to exist in the supported upstream API so embedded applications do not have to patch their Cargo checkout to get clean shutdown semantics.

iverc:
## Summary

Closes #813. Adds a public, idempotent `shutdown()` method on `EmbeddedNode` so embedders can await clean teardown before dropping the node, instead of relying on Drop behavior.

## Motivation

defra-agent-desktop and other downstream embedders have been relying on Drop for teardown, which produces iroh warnings like:

> `Endpoint dropped without calling Endpoint::close. Aborting ungracefully.`

…and leaves integration tests hanging. The underlying `ShutdownHandle::shutdown` machinery already existed — embedders just had no way to reach it through the public `EmbeddedNode` API.

## API

```rust
impl<S: Store + 'static> EmbeddedNode<S> {
    /// Shut the node down cleanly. Idempotent; safe to call concurrently.
    pub async fn shutdown(&self);

    /// Returns true if shutdown has been initiated.
    pub fn is_shutdown(&self) -> bool;
}
```

## Teardown order (per issue)

1. **P2P transport** — calls `ManagedP2PSystem::shutdown()` which invokes the pre-existing `ShutdownHandle::shutdown`:
   - **iroh**: `IrohTransport::shutdown()` closes the endpoint gracefully → resolves the `Endpoint dropped without calling Endpoint::close` warning
   - **libp2p**: aborts replication/merge tasks, then `P2PHostHandle::shutdown()`
   - Clears the signing identity store (already in `ShutdownHandle`)
2. **Database** — `db::DB::close()` sets the `is_closed` flag (future `begin_txn` calls return `Error::DatabaseClosed`) and closes the underlying store

The background downsample task is aborted by the existing `Drop` impl on `BackgroundTasks` when the node itself is dropped — no explicit step needed in `shutdown()`.

## Idempotency

An `AtomicBool shutdown_started` guard ensures the first call does real work and subsequent calls return immediately. Safe under concurrent invocation from multiple tasks.

## Tests

New [crates/embedded/tests/shutdown.rs](crates/embedded/tests/shutdown.rs) with three test cases:

- **`shutdown_reports_closed_database`** — post-shutdown state: `is_shutdown()` returns true, `database.is_closed()` returns true
- **`shutdown_is_idempotent`** — three sequential `shutdown()` calls converge on the same stable state with no panics
- **`shutdown_concurrent_calls_are_safe`** — 8 concurrent `shutdown()` invocations on a shared `Arc<EmbeddedNode>` all complete without panicking

## Test plan

- [x] `cargo test -p embedded --test shutdown` — 3/3 passing

## Recommended usage

```rust
node.shutdown().await;
drop(node); // BackgroundTasks::Drop aborts the downsample task here
```

### DIFF
--- a/crates/embedded/src/node.rs
+++ b/crates/embedded/src/node.rs
@@ -15,6 +15,7 @@ use anyhow::{anyhow, Context, Result};
 use p2p::sync::SyncConfig;
 #[cfg(feature = "iroh")]
 use p2p::P2PTransport;
+use tokio::sync::Notify;
 
 pub(crate) type EmbeddedBlockstore<S> = blockstore::DefraBlockstore<S>;
 pub(crate) type EmbeddedMergeHandler<S> = db_merge::AcpMergeHandler<S, EmbeddedBlockstore<S>>;
@@ -35,8 +36,12 @@ pub struct EmbeddedNode<S: storage::corekv::Store> {
     pub sourcehub_acp: Option<Arc<sourcehub::SourceHubDocumentACP>>,
     pub p2p: Option<Arc<ManagedP2PSystem>>,
     /// Idempotency guard for [`EmbeddedNode::shutdown`]. Set to `true`
-    /// by the first caller of `shutdown()`; subsequent calls are no-ops.
+    /// by the first caller of `shutdown()`.
     shutdown_started: AtomicBool,
+    /// Marks that shutdown work has finished.
+    shutdown_finished: AtomicBool,
+    /// Wakes concurrent shutdown callers once teardown is complete.
+    shutdown_notify: Notify,
 }
 
 impl<S: storage::corekv::Store + 'static> EmbeddedNode<S> {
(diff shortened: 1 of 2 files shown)


jackzampolin:
This is the right API direction, but I think the concurrent shutdown semantics need tightening before merge.

For a public async API named `node.shutdown().await`, callers will reasonably expect that once the future resolves, teardown is complete. The current `AtomicBool` guard makes the first caller perform the work, but concurrent later callers return immediately once shutdown has merely started. That means a concurrent caller can observe `shutdown().await` complete while P2P/database close is still running in the first caller.

I think the fix should be: first caller performs shutdown, concurrent callers await the same shutdown completion, and later callers return immediately after completion. In other words, `shutdown().await` should be an awaitable completion barrier, not just an initiation guard.

The feature itself looks useful; I would just like the public API contract to match the natural `await` semantics before we merge it.

PR #824, opened 2026-04-15
fix(db): explicit drop of datastore NamespaceView before read-only txn discard (#821)

### LINKED ISSUE #821: Warns on read-only transaction discard after get_for_update
## Summary
During the desktop live inference smoke in `sourcenetwork/defra-agent`, DefraDB emits repeated warnings when cleaning up a read-only transaction after `get_for_update`.

## Warning
```text
Failed to discard read-only transaction after get_for_update ... transaction still has active references
```

## Reproduction
This reproduces consistently in the env-gated desktop live smoke:
```bash
cargo test -p defra-agent-desktop app::tests::desktop_app_live_inference_smoke -- --ignored --exact --nocapture
```

Observed on April 14, 2026 while exercising a real embedded node plus live inference backend through the desktop UI.

## Expected
The read-only transaction should clean up without warning spam during normal query/update flows.

## Notes
- We already had to file `sourcenetwork/defradb.rs#813` for a separate embedded-node shutdown API gap surfaced by the same desktop test lane.
- I searched the tracker for the warning text, `transaction still has active references`, and `get_for_update`, and did not find an existing issue for this warning.

iverc:
## Summary

Closes #821. Fixes the warning spam:

> `Failed to discard read-only transaction after get_for_update ... transaction still has active references`

observed in the `defra-agent` desktop live inference smoke.

## Root cause

`BasicTxn::discard()` does `Arc::try_unwrap` on the inner `SharedTxn`. Any outstanding `NamespaceView` clone of that Arc makes `try_unwrap` fail and return `Error::TxnStillInUse`, which the caller in `auto_commit_mutator::read` then logs at `warn!` level.

`exists_impl` and `get_for_update_impl` both:

1. Create a read-only txn → Arc refcount 1
2. Call `txn.datastore()` → clones Arc into a `NamespaceView` → refcount 2
3. Run the fetch via `.await`
4. Call `txn.discard()` → `try_unwrap` expects refcount 1

The intent was that the NamespaceView drop at end-of-block would restore refcount to 1 before discard. **Under multi-threaded tokio schedules** the async state machine can hold the datastore across the discard statement long enough for `try_unwrap` to see refcount=2 and fire the warning.

The existing regression test (`issue_729_get_for_update_warning`) uses `tokio::test(flavor = \"current_thread\")` which doesn't reproduce — single-threaded runtime drops locals deterministically at block end. Production runs multi-threaded.

## Fix

Add an explicit `drop(datastore)` inside the block before the block returns. This is unambiguous regardless of async state-machine layout and makes the drop intent visible to readers. Also adds the same fix to `exists_impl` which had the same pattern without the `drop`.

No functional change beyond eliminating the warning — the discard itself was already happening, just noisily.

## Test plan

- [x] `cargo test -p defra-node --test issue_729_get_for_update_warning` — existing regression test still passes
- [x] `cargo test -p db` — all 65 tests passing
- [x] `cargo clippy -p db --all-targets -- -D warnings` clean

### DIFF
--- a/crates/db/src/auto_commit_mutator/read.rs
+++ b/crates/db/src/auto_commit_mutator/read.rs
@@ -13,19 +13,29 @@ impl<S: Store + 'static> AutoCommitMutator<S> {
             query::error::QueryError::execution(format!("failed to create txn: {}", e))
         })?;
 
-        // Get the datastore
-        let datastore = txn.datastore().map_err(|e| {
-            query::error::QueryError::execution(format!(
-                "failed to get datastore for collection '{}': {}",
-                collection_name, e
-            ))
-        })?;
+        // Scope the datastore + run the check inside an inner block, then
+        // explicitly drop the datastore NamespaceView before calling
+        // `txn.discard()`. BasicTxn::discard() does Arc::try_unwrap on the
+        // inner SharedTxn; any outstanding NamespaceView clone causes the
+        // unwrap to fail with `TxnStillInUse` and spam the logs. Explicit
+        // drop here makes the Arc refcount unambiguously 1 at the discard
+        // point regardless of async state-machine layout (#821).
+        let result = {
+            let datastore = txn.datastore().map_err(|e| {
+                query::error::QueryError::execution(format!(
+                    "failed to get datastore for collection '{}': {}",
+                    collection_name, e
+                ))
+            })?;
 
-        // Execute the check
-        let result = collection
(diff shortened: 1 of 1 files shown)


jackzampolin:
Closing this as superseded: the same fix already landed on main via #826 (commit f1ae45bd). Verified with git cherry that this PR's patch is already applied.

jackzampolin:
Superseded by #826, which already landed the same patch.

PR #857, opened 2026-04-15
fix(p2p): drop custom gossipsub message_id_fn, restore Go parity

### LINKED ISSUE #834: P2P: GossipSub `message_id_fn` diverges from Go default
## Summary

Rust overrides GossipSub's message-id function to `sha256(data)`. Go uses the libp2p-pubsub default (`from + seqno`). This changes deduplication semantics and can cause observable mesh-propagation differences between Rust and Go peers on the same network.

## Details

Rust (`crates/p2p/src/behaviour.rs:216-219`):
```rust
.message_id_fn(|message: &gossipsub::Message| {
    let hash = crypto::sha256(&message.data);
    MessageId::from(hash.to_vec())
})
```

Go (`go-p2p/peer.go:133-139`) — no `WithMessageIdFn`, default is used:
```go
p.ps, err = pubsub.NewGossipSub(
    ctx, h,
    pubsub.WithPeerExchange(true),
    pubsub.WithFloodPublish(true),
)
```

`go-libp2p-pubsub`'s default message-id is `string(m.GetFrom()) + string(m.GetSeqno())`.

## Impact

- **Same content from same sender, same seqno**: both impls dedup → identical.
- **Same content from different senders**: Go propagates both (different IDs); Rust dedups (same content hash). A message legitimately forwarded by multiple peers in Go may disappear on Rust.
- **Different content from same sender, same seqno** (attack / replay): Go dedups (sender+seqno match); Rust propagates (different content hash). This actually makes Rust *less* robust against some replay scenarios.

In a mixed-impl network, the mesh propagation differs and some messages may silently fail to reach all subscribers.

## Expected Behavior

Rust matches Go's default `from + seqno` message-id, unless there's a documented reason to diverge.

## Fix

Remove the `.message_id_fn(...)` call in `behaviour.rs` (and the test-only variant at `behaviour.rs:305-308`) and rely on libp2p-rust's default, which is `from + seqno` equivalent.

## Acceptance Criteria

- [ ] `behaviour.rs:216-219` custom `message_id_fn` removed (production path)
- [ ] `behaviour.rs:305-308` custom `message_id_fn` removed (test-only path)
- [ ] Integration test: Rust and Go subscribers on the same topic receive the same set of messages when duplicates are involved

## Files

- `crates/p2p/src/behaviour.rs:216-219, 305-308`

## References

- Go: `go-p2p/peer.go:133-139` — no custom message-id
- go-libp2p-pubsub default: `github.com/libp2p/go-libp2p-pubsub` — `DefaultMsgIdFn`

fredcarle:
## Summary
- Fixes #834
- `DefraBehaviour` overrode `ConfigBuilder::message_id_fn` with `sha256(message.data)`, which collides across senders with identical payloads. Go (`go-p2p/peer.go:133-139`) uses no custom fn — the libp2p-pubsub default derives from `source + seqno` and distinguishes senders. Removing the override on both the production and test-only paths restores parity.
- Drop the now-unused `MessageId` import from `behaviour.rs`.
- Added a live two-peer test `regression_834_gossipsub_message_id_distinguishes_senders` that fails on pre-fix code (both peers return `bdcba3f5...`) and passes after.

## Test plan
- [x] New test `regression_834_gossipsub_message_id_distinguishes_senders` fails on `main` and passes on this branch.
- [x] Full `cargo test -p p2p` passes (232 tests, 0 failures).
- [x] `cargo clippy -p p2p --tests --no-deps -- -D warnings` passes clean on this branch.


fredcarle (commented):
## Review — drop custom gossipsub `message_id_fn`, restore Go parity

### Overview

Tightly-scoped two-line fix (plus comments and a regression test) removing a custom `message_id_fn` override from both `DefraBehaviour` constructors in `crates/p2p/src/behaviour.rs`. The previous override hashed `sha256(message.data)`, which collided across senders publishing identical payloads — diverging from Go's `go-libp2p-pubsub` default (`from + seqno`) and silently dropping legitimate cross-sender duplicates.

### Parity verification

Confirmed against Go and Rust libp2p sources:
- `/Users/fredcarle/code/sourcenetwork/go-p2p/peer.go:133-138` — no `WithMessageIdFn`.
- `go-libp2p-pubsub@v0.8.1/pubsub.go:1077` — `DefaultMsgIdFn = string(GetFrom()) + string(GetSeqno())`.
- `libp2p-gossipsub-0.46.1/src/config.rs:389-402` — default = `peer_id.to_base58() + sequence_number.to_string()`.

Both libraries default to source+seqno with functionally equivalent uniqueness semantics.

### Strengths

- Both production and test-only constructors fixed (`crates/p2p/src/behaviour.rs:213-228`, `:319-330`).
- Unused `MessageId` import dropped from `behaviour.rs:35`.
- Regression test (`crates/p2p/tests/host_tests.rs:573-666`) is well-engineered: live two-peer mesh, polls until publish actually succeeds (rather than relying on `topic_peers`), and asserts `assert_ne!` on the IDs.
- The `// We deliberately do NOT override...` comment in `behaviour.rs:215-221` is one of the few places where a non-trivial WHY-comment is justified — explains why the absence of a method call is intentional and prevents future regressions.
- Scope is minimal and focused.

### Issues

**Minor** — `crates/p2p/src/behaviour.rs:319-321`: doc comment on the second (test-only) constructor is much terser than the production one. The reference to "#834" alone is acceptable, but the production-side comment is the better template — this is the path most likely to be edited unsuspectingly.

**Minor (parity, out of scope)** — `crates/p2p/src/behaviour.rs:319-330`: test-only constructor uses `ValidationMode::Permissive` and `MessageAuthenticity::RandomAuthor`, while production uses `Strict` + `Signed`. Go unconditionally builds a single signed pubsub. Worth a follow-up parity issue.

**Minor (parity, out of scope)** — second constructor also lacks `.do_px()` (which Go has via `pubsub.WithPeerExchange(true)`). Adjacent parity gap worth tracking.

**Nit** — `crates/p2p/tests/host_tests.rs:596-612`: ~17-line comment block explaining the bug is verbose by `CLAUDE.md` standards. The `assert_ne!` failure message already explains the regression. Consider trimming to ~5 lines that link to #834.

**Nit** — `crates/p2p/tests/host_tests.rs:638-655`: `publish_when_ready` defined as a nested `async fn` — fine, but reusable for other gossip tests; could live alongside `wait_until_connected` at module scope.

### Test coverage

- **Wins**: `regression_834_...` is a real two-peer test exercising the actual gossipsub mesh, not a unit test against a mock. Catches any future re-introduction of a content-hash override.
- **Gap**: no cross-impl Go↔Rust gossip test. Issue #834's third AC ("Rust and Go subscribers on the same topic receive the same set of messages when duplicates are involved") is not satisfied. The Rust-only test verifies IDs differ — necessary but not sufficient to prove mesh-propagation alignment with Go.
- **Gap**: integration-test crate has no gossipsub message-id coverage. Not introduced by this PR but worth filing.
- **Gap**: no test for the dual case from #834 — "different content from same sender, same seqno" — where the new default is more aggressive at deduping than the old sha256 was. Worth confirming Go and Rust agree there.

### Verdict

Fix is correct, minimal, well-justified, and matches Go's behavior precisely. The regression test is live and meaningful. **Ready to merge as-is.** The stated AC item about a cross-impl Rust↔Go gossip test is unmet but reasonable to track as follow-up rather than block — the in-process test catches the immediate regression, and a Go interop harness for gossipsub is larger infrastructure that shouldn't gate a two-line correctness fix. Recommend filing follow-up issues for (1) Go↔Rust gossipsub interop test, (2) test-constructor parity (`Permissive`→`Strict`, `RandomAuthor`→`Signed`, missing `do_px`).

Issue #848, opened 2026-04-15
CRDT: Counter missing Float32, CompositeDelta missing Status (doc deletion marker)


fredcarle:
## Summary

Two related gaps in the Rust CRDT crate vs Go, bundled because both are additive parity work on the same CRDT types and best fixed together.

- **P-241 Float32 counters missing** — Go supports `int64`, `float32`, `float64`; Rust supports only `int64` and `float64`. Schema fields declared as `Float32` cannot be replicated.
- **P-242 CompositeDelta missing `Status`** — Go's `DocCompositeDelta` carries an `Active`/`Deleted` status used to signal document-level tombstones; Rust's `CompositeDelta` omits it, so Rust nodes don't observe document deletions signaled via composite delta.

### P-241 — Float32

Go (`defradb/internal/core/crdt/counter.go:179-194`): `Merge` switches on three numeric kinds — `FieldKind_NILLABLE_INT`, `FieldKind_NILLABLE_FLOAT32`, `FieldKind_NILLABLE_FLOAT64` — decoding the delta bytes into the matching Go numeric type.

Rust (`crates/crdt/src/counter.rs:18-21`):
```rust
pub enum NumericKind {
    Int64,
    Float64,
}
```

Only two variants. Any counter delta whose declared schema field is `Float32` cannot be represented losslessly — decoding would either fail or silently round-trip through `f64`, which changes the wire bytes and breaks CID determinism.

**Impact:** any document with a Float32 counter field cannot replicate between Rust and Go peers.

### P-242 — Status field on composite delta

Go (`defradb/internal/core/crdt/composite.go:27-49`):
```go
type DocCompositeDelta struct {
    DocID      []byte
    Priority   uint64
    SchemaVersionID string
    Status     client.DocumentStatus    // Active or Deleted
    ...
}
```

`Status` flows through the CRDT delta to communicate a document-level tombstone. Merge logic treats `Deleted` as a terminal transition.

Rust (`crates/crdt/src/composite.rs:16-27`): `CompositeDelta` has no corresponding field. A Go peer deleting a document via composite delta sends the `Status: Deleted` signal; the Rust peer's `CompositeDag::merge` drops it as an unknown field (CBOR decoders ignore unknowns) and merges as if the delta were a normal edit.

**Impact:** document deletions are lost when replicated to Rust nodes. Rust will believe the document is live; Go will believe it is deleted. Sync divergence.

## Expected Behavior

[...]

PR #866, opened 2026-04-16
fix(crdt): remove counter nonce idempotency to match Go (#847)

### LINKED ISSUE #847: CRDT: Counter nonce-based idempotency diverges from Go, causes state divergence on retransmit
## Summary

Rust's counter CRDT implements nonce-based idempotency in `merge`: when a delta arrives with a nonce already seen in storage, it is skipped. Go's counter ignores the nonce field entirely during `Merge` — it generates a nonce only for DAG block uniqueness and applies every delta unconditionally. Two nodes receiving the same retransmitted counter delta will reach **different states**: Rust stays at value V, Go advances to V+Δ.

This is a CRDT convergence bug: Go and Rust do not converge after identical replication traffic when retransmits are involved.

### Rust implementation

`crates/crdt/src/counter.rs:257-290` — `has_nonce`, `mark_nonce`, `clear_nonce` methods. These are consulted in `CounterDAG::merge` (look up the exact line) to short-circuit duplicate deltas.

The nonce is stored alongside the counter value, keyed by the delta's nonce bytes. On second receipt of the same delta, `has_nonce` returns true and `merge` returns without applying.

### Go implementation

`defradb/internal/core/crdt/counter.go:152-160` (`Merge` method):
```go
func (c Counter) Merge(ctx context.Context, txn datastore.Txn, delta Delta) error {
    d := delta.(CounterDelta)
    return c.incrementValue(ctx, txn, d.Data)
}
```

No nonce lookup. Every incoming delta increments the stored value. Nonce is set at delta creation time (`counter.go:~75`, `delta.Nonce = crypto.GenerateNonce()`) purely to make the resulting IPLD block's CID unique — it plays no role in merge.

### Convergence break

Scenario:
1. Peer A creates a counter delta with value +5 and nonce N, pushes to peer B (Rust) and peer C (Go).
2. Transport retransmits the same delta (network blip, replicator retry, pubsub republish).
3. Peer B (Rust) sees nonce N a second time, skips. Value = original + 5.
4. Peer C (Go) has no nonce check, applies again. Value = original + 10.

Both nodes received identical network traffic. They now disagree on the counter. This violates CRDT convergence (the core parity guarantee in CLAUDE.md: "Same CRDT operations → Same merged state").

Transport-level dedup **does** exist (pubsub message-id, PushLog processQueue guard) but is not universal — DAG resync, block re-delivery after node restart, and crash-recovery paths can all legitimately deliver the same block/delta twice. CRDT operations must be idempotent at the merge layer. Go's design assumes merge is already idempotent (because applying the same signed block twice would be caught by block-CID deduplication at the blockstore layer, *before* merge). Rust's belt-and-suspenders approach at the merge layer introduces the divergence.

## Expected Behavior

Rust and Go counters converge to the same value given the same network traffic. Either:
- **Option A (match Go):** Remove the nonce-idempotency check in Rust counter merge. Rely on the blockstore / DAG layer to deduplicate by block CID before the delta reaches `merge`.
- **Option B (change Go):** Add nonce-based idempotency in Go's counter `Merge`. This is a behavior change in Go upstream and requires a companion upstream issue.

Recommendation: **Option A**, to match Go's existing behavior with minimal upstream coordination. Block-CID dedup in the blockstore is already the contract for idempotency across all CRDT types, not just counter. The current Rust check is redundant when the block-CID dedup is correctly enforced, and divergent when it is not.

## Acceptance Criteria

- [ ] Rust's `CounterDAG::merge` (or wherever the nonce check lives) no longer short-circuits on nonce presence
- [ ] `has_nonce`, `mark_nonce`, `clear_nonce` methods removed (or repurposed / made no-op) on `Counter`
- [ ] The counter merge path instead relies on upstream block-CID dedup for idempotency. Add a comment documenting this dependency.
- [ ] Audit that the block-CID dedup guarantee actually holds on every path that reaches counter merge: PushLog ingest, bitswap block arrival, DAG sync, crash recovery replay. If any path can deliver the same block twice past the blockstore `has(cid)` check, that is a separate bug — file it.
- [ ] Convergence integration test: simulate a retransmit of a counter delta and assert Rust-merged value == expected (single application, not double).
- [ ] Integration test: Rust peer and Go peer receive the same counter delta stream (with deliberate retransmits). Assert they report the same final value.

## Files

- `crates/crdt/src/counter.rs:257-290` — remove or no-op the nonce helpers
- `crates/crdt/src/counter.rs` — `CounterDAG::merge` — remove the dedup branch
- Test file for counter CRDT — add convergence regression test

## References

- Go: `defradb/internal/core/crdt/counter.go:75` (nonce generated for block CID uniqueness only), `:152-160` (`Merge` ignores nonce)
- CLAUDE.md: "Same CRDT operations → Same merged state" — this bug violates that contract
- Related: if audit uncovers a path that can double-deliver past block-CID dedup, it would be a separate defradb.rs issue and possibly a defradb (Go) issue too.

iverc:
## Summary

- Remove nonce-based idempotency from Rust's counter CRDT merge path so it matches Go's unconditional-apply semantics (#847).
- Drop the dead `MergeResult::SkippedAlreadyApplied` variant and the `cleanup_counter_nonce_for_cid` helper that only existed to service the Rust-specific dedup layer.
- Flip 5 unit/property tests that were asserting the broken behaviour; add a direct regression guard for #847.

## Why

Go DefraDB's `counter.Merge` ignores the delta nonce entirely — nonces exist only so that emitted DAG blocks get unique CIDs. Rust was short-circuiting `apply_delta` when a nonce was already present in storage, which caused state divergence any time the same counter delta was legitimately re-delivered (pubsub republish, replicator retry, DAG resync after crash). Same network traffic, different state — a direct violation of `"same CRDT operations → same merged state"` (CLAUDE.md).

Idempotency for CRDT merge lives at the blockstore layer in both implementations. The merged-set markers (`is_merged(cid)` / `get_unmerged()`) filter already-merged blocks on every ingest path before they ever reach `counter.merge()`. The per-delta nonce check in Rust was belt-and-braces that turned into a divergence.

## Audit — block-CID dedup guarantee holds on every merge-reaching path

- **PushLog ingest** — `crates/p2p/src/sync/manager/process/pushlog.rs:77, 145` checks `blockstore.is_merged(cid)` before `process_block_inner`.
- **DAG traversal** — `crates/p2p/src/sync/manager/links.rs:92` skips merged subtrees.
- **Crash recovery replay** — `crates/p2p/src/sync/replication/recovery.rs:54` only replays from `get_unmerged()`.
- **Batch merge** — same `handle_block` path as PushLog, same guard.
- **Bitswap** — deposits blocks into the blockstore but does not itself call merge; merge happens when a later PushLog/DocSync references the block, with the same is_merged guard.

If a future ingest path is added that skips this guard, this fix would surface as a double-apply — that would be a separate bug, not a regression of this one. The blockstore's merged-set is the documented contract; counter merge now relies on it the same way every other CRDT type already does.

## Changes

- `crates/crdt/src/counter.rs` — strip `has_nonce`/`mark_nonce`/`clear_nonce`, the nonce-prefix field, the nonce-dedup branch in `apply_delta`, and the `mark_nonce(0)` calls in both seed functions. New doc comment on `Counter::merge` pointing at the blockstore-layer idempotency contract.
- `crates/crdt/src/traits.rs` — remove `MergeResult::SkippedAlreadyApplied` variant + `was_skipped()` helper. Enum remains `#[non_exhaustive]`.
- `crates/db-merge/src/merge_handler/counter.rs` — delete `cleanup_counter_nonce_for_cid` entirely (now zero callers). Rename `finalize_linked_field_blocks` → `mark_field_blocks_merged`; it's now a thin wrapper around `blockstore.mark_batch_as_merged` (the critical dedup step, unchanged). Best-effort wrappers drop the now-redundant `fallback_collection_id` argument.
- `crates/db-merge/src/merge_handler/batch.rs` — `PendingFieldBlockFinalization` loses the `fallback_collection_id` field.
- `crates/db-merge/src/merge_handler/composite.rs` — update construction sites.
- `crates/db-merge/src/merge_handler/mod.rs` — rename `counter_merge_marks_cid_merged_and_clears_nonce_marker` → `counter_merge_marks_cid_merged` and drop the clear-nonce assertion (no longer applies).

### Tests

- `crates/crdt/tests/counter_tests.rs`
  - **new**: `test_counter_retransmit_applies_twice` — direct regression guard. Merges the same delta twice via `counter.merge()` and asserts accumulation. Replaces `test_counter_idempotency` which asserted the broken behaviour.
  - **deleted**: `test_counter_clear_nonce_removes_replay_marker` (API gone).
  - **deleted**: `test_counter_merge_result_skipped_already_applied` (variant gone).
- `crates/crdt/tests/property_tests.rs`
  - **renamed/rewritten**: `test_counter_idempotence` → `test_counter_double_apply_accumulates`.
  - **renamed/rewritten**: `test_float64_counter_idempotence` → `test_float64_counter_double_apply_accumulates`.
  - **renamed/rewritten**: `test_counter_nonce_replay_protection` → `test_counter_replayed_delta_accumulates`.

## Storage / migration

Pre-existing nonce markers under `/data/{schema}/{doc}/{field}/nonces/...` remain on disk after this change. They are never read and never written again; they occupy a small amount of storage indefinitely. Acceptable because:

1. The keyspace is bounded by the number of counter deltas the node processed before this fix — not unbounded.
2. Correctness is unaffected — the old markers are simply ignored.
3. defradb.rs is at v1.0 RC1 and the nonce keyspace was never part of any documented external contract.

No migration script. `CRDTNoncePrefix` / `CRDTNonceKey` key types in `storage/src/keys/crdt.rs` are left in place — they're low-level byte-layout definitions, not behaviour, and leaving them keeps any on-disk nonce keys decodable if someone ever needs to inspect them.

## Out of scope

- **`CounterDelta::nonce` field**: stays. The DAG block layer still uses it for CID uniqueness, matching Go.
- **Dual-runtime integration test**: the unit tests directly cover the scenario #847 describes (apply the same delta twice; assert it accumulates, not dedupes). A Rust↔Go end-to-end convergence test would be valuable but requires fixture infrastructure for deterministic delta replay across the two binaries and isn't a blocker for this PR.
- **LWW / Composite / Collection CRDTs**: untouched. Their idempotency was already blockstore-layer only.

## Test plan

- [x] `cargo test -p crdt` — 27 property + unit tests pass, including the new regression guard
- [x] `cargo test -p db-merge` — 58 tests pass, including the renamed `counter_merge_marks_cid_merged`
- [x] `cargo test --workspace --exclude integration-test --exclude ffi-test` — clean
- [ ] (CI) `cargo test -p integration-test --test basic` — counter sanity via the full stack
- [ ] (CI) `go_compat` matrix against the Go binary — cross-runtime parity check

Closes #847.

### DIFF
--- a/crates/crdt/src/counter.rs
+++ b/crates/crdt/src/counter.rs
@@ -385,6 +385,11 @@ impl Counter {
     /// must be seeded from the document value to ensure correct accumulation.
     ///
     /// Returns true if seeding was performed, false if already initialized.
+    ///
+    /// NOTE: per #847, counter merge is unconditional and there is no nonce
+    /// tracking. Any future seed variant (e.g. Float32 once added by #848)
+    /// MUST NOT call a `mark_nonce(..)` helper — that would leak dead markers
+    /// into the datastore and reintroduce the dedup contract this PR removes.
     pub async fn seed_if_uninitialized_int64(
         &self,
         rw: &mut dyn ReaderWriter,
--- a/crates/crdt/src/counter.rs
+++ b/crates/crdt/src/counter.rs
@@ -1,9 +1,13 @@
 //! Counter CRDT implementation
 //!
-//! Supports increment and decrement operations with nonce-based idempotent delivery.
-//! Uses commutative addition with wrapping on overflow for Int64 (matching Go DefraDB),
-//! and IEEE-754 addition for Float64. This is not a traditional PN-Counter (which uses
-//! separate per-replica counters); instead it uses a single value with nonce tracking.
+//! Supports increment and decrement operations. Uses commutative addition with
+//! wrapping on overflow for Int64 (matching Go DefraDB), and IEEE-754 addition
+//! for Float64. This is not a traditional PN-Counter (which uses separate
+//! per-replica counters); instead it uses a single accumulated value.
+//!
+//! Merge is unconditional — idempotency is enforced upstream by the blockstore
+//! (`is_merged(cid)` / `get_unmerged()`) on every ingest path. Matches Go's
+//! counter Merge which also ignores the delta nonce.
 
 use crate::traits::{Context, Delta, MergeResult, ReplicatedData, ValueReader};
 use async_trait::async_trait;
@@ -188,8 +192,6 @@ impl Delta for CounterDelta {
 pub struct Counter {
     /// Storage key for the counter value
     value_key: Vec<u8>,
-    /// Storage key for tracking applied nonces
-    nonce_prefix: Vec<u8>,
     /// Schema version
     schema_version_id: String,
     /// Field name
@@ -236,57 +238,14 @@ impl Counter {
             doc_id.to_vec(),
             field_name.clone(),
         );
-        let nonce_prefix = value_key.nonce_prefix();
 
         Ok(Self {
             value_key: value_key.bytes(),
-            nonce_prefix: nonce_prefix.bytes(),
             schema_version_id,
             field_name,
             allow_decrement,
             kind,
         })
-    }
-
-    fn nonce_key(&self, nonce: i64) -> Vec<u8> {
-        let mut nonce_key = self.nonce_prefix.clone();
-        nonce_key.extend_from_slice(&nonce.to_be_bytes());
-        nonce_key
-    }
-
-    /// Check if a nonce has been applied
-    async fn has_nonce(&self, reader: &dyn Reader, nonce: i64) -> Result<bool> {
-        reader
-            .has(&self.nonce_key(nonce))
-            .await
-            .map_err(|e| Error::Storage(e.to_string()))
-    }
-
-    /// Mark a nonce as applied
-    ///
-    async fn mark_nonce(&self, rw: &mut dyn ReaderWriter, nonce: i64) -> Result<()> {
-        let nonce_key = self.nonce_key(nonce);
-        rw.set(&nonce_key, &[1])
-            .await
-            .map_err(|e| Error::Storage(e.to_string()))
-    }
-
-    /// Remove a nonce marker once block/CID-level merge dedup is durable.
-    ///
-    /// Returns `true` if a marker existed and was removed.
-    pub async fn clear_nonce(&self, rw: &mut dyn ReaderWriter, nonce: i64) -> Result<bool> {
-        let nonce_key = self.nonce_key(nonce);
-        let exists = rw
-            .has(&nonce_key)
-            .await
-            .map_err(|e| Error::Storage(e.to_string()))?;
-        if !exists {
-            return Ok(false);
-        }
-        rw.delete(&nonce_key)
-            .await
-            .map_err(|e| Error::Storage(e.to_string()))?;
-        Ok(true)
     }
 
     /// Get current value as i64
@@ -355,20 +314,15 @@ impl Counter {
             .map_err(|e| Error::Storage(e.to_string()))
     }
 
-    /// Apply an increment/decrement
+    /// Apply an increment/decrement.
     ///
-    /// # Crash Recovery Semantics
-    ///
-    /// Nonce marking and value updates are not atomic. To ensure safety:
-    /// - Nonce is marked FIRST, then value is updated
-    /// - If crash occurs after nonce but before value update: delta is lost (under-count)
-    /// - If crash occurred with old ordering (value then nonce): would double-count
-    ///
-    /// Under-counting on crash is safer than over-counting because:
-    /// 1. It's easier to detect missing deltas than duplicate applications
-    /// 2. Over-counting violates CRDT idempotency guarantees
-    ///
-    /// For true atomicity, use a Store implementation with transaction support.
+    /// Merge is unconditional: every delta that reaches this method is
+    /// applied. Per-delta idempotency is the blockstore's job via
+    /// `is_merged(cid)` / `get_unmerged()` on every ingest path (PushLog,
+    /// DAG traversal, crash recovery). Do not reintroduce a nonce-based
+    /// dedup check here — doing so causes Rust↔Go state divergence on
+    /// legitimate block retransmits (#847). The `delta.nonce` field
+    /// exists only so the resulting DAG block has a unique CID.
     async fn apply_delta(
         &self,
         rw: &mut dyn ReaderWriter,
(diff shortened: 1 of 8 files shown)


fredcarle:
**Verdict: request changes (minor).**

The core change — removing nonce-based idempotency so retransmits accumulate — looks right. `CounterDelta.nonce` stays in the struct so CBOR wire format is unchanged (no backward-compat break). The amplification concern is mitigated by block-CID dedup in `crates/p2p/src/sync/manager/process/pushlog.rs:145` — replaying the same CID hits `is_merged` and is skipped before `apply_delta` ever runs.

**Missed seed site:** `crates/crdt/src/counter.rs:581` — `seed_if_uninitialized_float32` still calls `self.mark_nonce(rw, 0).await?`. The PR removed `mark_nonce(0)` from the int64 and float64 seed functions but not float32. It's dead in practice today (db-merge only seeds int64/float64), but the inconsistency will leave stale nonce markers on disk the moment someone adds float32 seeding. Worth tightening now while the context is fresh.

**Parity verification:** The "to match Go" claim would be more convincing with a direct cross-runtime fixture test (apply the same increment sequence on Go and Rust, assert identical final value). The unit tests in this PR verify internal consistency but not parity. The `go_compat` CI matrix is unchecked — either run it or note how parity was verified.

_🤖 AI-assisted review_

iverc:
Thanks for the review @fredcarle — addressed both points:

1. **Float32 seed site**: There is no `seed_if_uninitialized_float32` in this branch; it's being added by PR #883 (#848) in parallel. I added a doc comment on the existing int64/float64 seed helpers warning that any future seed variant must NOT call `mark_nonce(..)` — this should flag the issue clearly when #883 comes back for review. Rebasing #883 onto this one should make the right behaviour obvious. Commit: 2cb9094a.

2. **Parity verification**: Fair point. The unit tests here verify internal consistency (apply twice → accumulate twice) but don't cross-check against Go. The go-compat CI matrix runs `integration-test --test basic` (and other suites) against the Go binary, so any existing counter test that crosses runtimes will catch divergence — but there's no *targeted* retransmit convergence test against Go right now. I'll file a follow-up integration test in a separate PR rather than scope-creep this one; the unit guard is enough to lock the internal semantics, and the real cross-runtime fixture belongs in `tools/integration-test`.

Let me know if you'd rather block this PR on the integration test.

fredcarle (approved):
LGTM 👍

Issue #830, opened 2026-04-15
Security: Bitswap serves blocks to any peer with no per-peer ACP filter


fredcarle:
## Summary

Go DefraDB wraps Bitswap with `bitswap.WithPeerBlockRequestFilter(p.hasAccess)` so that **every** incoming block request is filtered per-peer × per-CID against the `ReplicatorRegistry`. Rust's Bitswap is constructed with no filter, and the `iroh-bitswap::Store` trait has no peer context — the request path cannot see the requesting peer ID to make an access decision. As a result, any connected peer can fetch any block currently stored on a Rust node, bypassing per-collection ACP entirely.

Additionally, `crates/p2p/src/bitswap/access.rs:7-18` documents a security model that is factually wrong, which will mislead anyone auditing this module.

## Details

Go (`go-p2p/peer.go:145-146`):
```go
bswap := bitswap.New(ctx, bswapnet, ddht, options.Blockstore.Value(),
    bitswap.WithPeerBlockRequestFilter(p.hasAccess))
```

`hasAccess(peerID, cid)` is invoked on every incoming `WANT` and denies if the peer isn't authorized for the collection that owns the CID.

Rust (`crates/p2p/src/behaviour.rs:240-241`):
```rust
let bitswap_config = BitswapConfig::default();
let bitswap = Bitswap::new(local_peer_id, bitswap_store, bitswap_config).await;
```

The `Store` trait that `BitswapStoreAdapter` implements (`crates/p2p/src/bitswap/store.rs:55-85`) has this signature:
```rust
async fn get(&self, cid: &Cid) -> Result<Block>
```

No peer context. The filter cannot be implemented here even if we wanted to.

The documented (false) security claim (`crates/p2p/src/bitswap/access.rs:7-18`):
> Access control is enforced at the SyncCoordinator level, not at the Bitswap level.
> The SyncCoordinator checks access on incoming PushLog and GossipSub messages before blocks are stored. This means:
> 1. Unauthorized peers cannot push blocks to this node
> 2. **Bitswap inherently only serves blocks that passed the coordinator's access check**
> 3. Per-collection authorization is enforced

Claim (2) is false. SyncCoordinator controls what enters the blockstore, not what leaves it. Once a block is stored (because it came from an authorized push), Bitswap serves it to any peer that asks.

## Exploit

1. Alice (authorized for `users` collection) pushes block B to Bob via PushLog. Bob's SyncCoordinator validates, stores B in the blockstore.
[...]

PR #867, opened 2026-04-16
fix(p2p): require registry membership in Controlled mode access checks (#838)

### LINKED ISSUE #838: Security: SyncCoordinator access check allows any connected peer in Controlled mode
## Summary

In `AccessMode::Controlled`, the SyncCoordinator accepts **any** connected peer for collection-scoped operations without consulting the `ReplicatorRegistry`. The justification is transport-level authentication, but transport auth proves only \"this peer is who they say they are\" — it does not prove they are authorized for the collection. As a result, any peer that can open a libp2p connection can push blocks to, and fetch state about, any collection on a Rust node. This defeats collection-level authorization that the registry is designed to enforce.

## Details

`crates/p2p/src/sync/coordinator/access.rs:50-57` — `check_access_str` allows a peer if either:
1. They are in the `ReplicatorRegistry` for the target `collection_id`, OR
2. They are connected per `peer_state.is_connected()`.

Path (2) means any peer with an open libp2p connection — including a Sybil peer that just joined — passes the collection access check.

The module-level comment at `access.rs:20-31` claims this matches Go, but Go (`go-p2p/peer.go:200-211` `hasAccess` and `defradb/internal/db/p2p/...`) enforces collection-specific replicator membership before accepting a push. Go does not fall back to \"any connected peer.\"

The bundled finding **P-65** (tests asserting that connected-but-non-replicator peers pass) documents the behavior but does not flag that it violates collection authorization.

Related inconsistencies in the same module:

- `event_handler/doc_sync.rs:53` and `event_handler/car.rs:20` use `check_peer_is_replicator()` (collection-agnostic) — a replicator for collection A can request heads/CAR for collection B (information disclosure).

## Exploit

1. Mallory opens a libp2p connection to a Rust node running in Controlled mode.
2. Mallory sends a `PushLogRequest` for collection `private_users`, carrying a valid signed block.
3. `check_access_str(\"private_users\", Mallory)` returns true via the connected-peer path — Mallory is not in the registry for `private_users`, but she is connected.
4. The merge layer accepts the block (ACP at the document level may reject individual updates, but the collection gate has been bypassed and resources spent validating).

Combined with the Bitswap filter gap already tracked in #830, a connected peer can also **fetch** any stored block from that collection after the fact.

## Expected Behavior

In `AccessMode::Controlled`, `check_access_str(collection_id, peer)` returns true only if `peer` is a registered replicator for `collection_id`. Connected-but-not-registered peers are rejected at the collection gate with `AccessDenied`.

If there is a legitimate use case for \"accept any connected peer\" (e.g., open collections), gate it behind an explicit per-collection configuration flag, not a mode-wide fallback.

## Acceptance Criteria

- [ ] `check_access_str` in Controlled mode requires registry membership for the requested collection. Connected-but-not-registered peers are denied.
- [ ] `event_handler/doc_sync.rs:53` and `event_handler/car.rs:20` use the per-collection variant (`check_access_str` with the collection id) instead of `check_peer_is_replicator`
- [ ] Module doc comment at `access.rs:20-31` rewritten to accurately describe the new behavior and cite Go's `hasAccess` as the reference
- [ ] Adversarial integration test: a connected peer that is not in the registry for collection X is rejected on PushLog, DocSync, BranchableSync, and CAR requests for X
- [ ] Existing access_tests updated (some assert the broken behavior today — those assertions flip)

## Files

- `crates/p2p/src/sync/coordinator/access.rs:50-57`
- `crates/p2p/src/sync/coordinator/access_tests.rs:578-606` (and related)
- `crates/p2p/src/sync/coordinator/event_handler/doc_sync.rs:53`
- `crates/p2p/src/sync/coordinator/event_handler/car.rs:20`

## References

- Go: `go-p2p/peer.go:200-211` — `hasAccess` reference behavior
- Go: `defradb/internal/db/p2p/replicator.go` — registry-backed checks
- Related: #830 (Bitswap filter; same exfiltration pattern at the block-serve layer)
- Related: #832 (`ReplicatorInfo` — registry data model)

iverc:
## Summary

- Remove the "any connected peer" fallback from `check_access_str` and `check_peer_is_replicator` in `AccessMode::Controlled` for **direct push paths** (PushLog, two-stream, BranchableSync). Only `ReplicatorRegistry` membership grants access.
- **GossipSub** messages use topic membership as the access gate instead — being on the same collection topic means the peer is authorized. The gossip handler validates the peer is connected but does not require registry membership.
- Honor the two-stream protocol's `is_explicit_replicator` flag as a parallel auth path (transport-level explicit-replicator handshake already proves authorization).
- Flip 4 tests that asserted the broken behaviour; add 1 new positive test.
- Rewrite the module doc comment that falsely claimed the connected-peer fallback matched Go.

## Why

`check_access_str` and `check_peer_is_replicator` accepted any peer with an open libp2p connection as a fallback after the replicator registry check. Transport-level authentication proves *who* a peer is (identity), not *what* they're authorized to sync (collection membership). A Sybil peer that opens a connection would pass the collection gate in Controlled mode, which defeats the per-collection authorization that the ReplicatorRegistry is designed to enforce.

Go DefraDB's `hasAccess` (at `go-p2p/peer.go`) requires per-collection replicator membership before accepting a direct push. The previous Rust code claimed to match Go but was strictly weaker.

### Access model after this fix

| Path | Access gate | Rationale |
|---|---|---|
| Direct PushLog (libp2p stream) | Per-collection registry membership | #838 exploit vector — Sybil peer sends PushLogRequest for private collection |
| Two-stream PushLog | Registry membership OR transport explicit-replicator handshake | Transport auth is a parallel auth path |
| BranchableSync | Per-collection registry membership | Same as PushLog |
| GossipSub messages | Topic membership (connected check only) | Topic subscription is the access gate |
| DocSync / CAR fetch | Any-collection registry membership | Protocol lacks collection_id on wire |

## Changes

| File | Change |
|---|---|
| `crates/p2p/src/sync/coordinator/access.rs` | Remove `is_connected` fallback from both `check_access_str` and `check_peer_is_replicator`. Rewrite doc comments. |
| `crates/p2p/src/sync/coordinator/access_tests.rs` | Flip 4 tests to assert `AccessDenied`. Add `doc_sync_controlled_mode_allows_any_replicator`. |
| `crates/p2p/src/sync/coordinator/event_handler/pushlog.rs` | Skip registry check for two-stream when `is_explicit_replicator` is true. |
| `crates/p2p/src/sync/coordinator/event_handler/gossip.rs` | Replace `check_access_str` with connected-peer check (topic membership is the access gate for gossipsub). |

## Out of scope

- **DocSync / CAR per-collection filtering**: wire protocol lacks collection_id.
- **#830 (Bitswap block-serve filter)**: same exfiltration vector at the block layer, tracked separately.

## Test plan

- [x] `cargo test -p p2p` — 119 lib tests + integration tests pass
- [x] `cargo test --workspace --exclude integration-test --exclude ffi-test` — clean
- [ ] (CI) `cargo test -p integration-test --test acp` — P2P ACP tests including the previously failing subscription tests
- [ ] (CI) `cargo test -p integration-test --test p2p` — P2P integration tests

Closes #838.

### DIFF
--- a/crates/p2p/src/host/event.rs
+++ b/crates/p2p/src/host/event.rs
@@ -82,6 +82,12 @@ pub enum HostEvent {
     },
 
     /// Received a PushLog request via two-stream protocol (Go compatibility).
+    ///
+    /// Invariant (#838): `is_explicit_replicator` must only be set to
+    /// `true` by the two-stream transport after verifying the remote peer via
+    /// the signed explicit-replicator handshake. The sync coordinator treats
+    /// this flag as an authenticated claim and skips the `ReplicatorRegistry`
+    /// membership check when it is `true`.
     TwoStreamRequest {
         peer_id: PeerId,
         request: crate::message::PushLogRequest,
--- a/crates/p2p/src/sync/coordinator/access.rs
+++ b/crates/p2p/src/sync/coordinator/access.rs
@@ -110,29 +110,4 @@ impl<B: Blockstore + 'static, T: P2PTransport> SyncCoordinator<B, T> {
     pub fn access_mode(&self) -> AccessMode {
         self.access.access_mode
     }
-
-    pub(super) async fn peer_is_connected(&self, peer_id: &PeerId) -> bool {
-        let peer_id_str = peer_id.as_str();
-        if self.access.peer_state.is_connected(peer_id_str) {
-            return true;
-        }
-
-        match self.runtime.transport.connected_peers().await {
-            Ok(peers) => {
-                let is_connected = peers.iter().any(|peer| peer.as_str() == peer_id_str);
-                if is_connected {
-                    self.access.peer_state.peer_connected(peer_id_str);
-                }
-                is_connected
-            }
-            Err(error) => {
-                tracing::debug!(
-                    peer_id = %peer_id,
-                    error = %error,
-                    "Failed to read transport-connected peers during access check"
-                );
-                false
-            }
-        }
-    }
 }
--- a/crates/p2p/src/sync/coordinator/access.rs
+++ b/crates/p2p/src/sync/coordinator/access.rs
@@ -23,22 +23,21 @@ impl<B: Blockstore + 'static, T: P2PTransport> SyncCoordinator<B, T> {
     /// Access rules (implemented in
     /// [`super::authorizer::RuntimeAuthorizer::peer_authorized_for_collection`]):
     /// 1. If mode is Open → allow all
-    /// 2. If peer is a replicator for the collection → allow
-    /// 3. If transport reports peer as a replicator for the collection (cache miss) → allow
-    /// 4. If peer is connected → allow
-    /// 5. If transport reports peer as connected (cache miss) → allow
-    /// 6. Otherwise → deny
+    /// 2. If peer is a replicator for the given collection → allow
+    /// 3. If transport reports peer as a replicator for the collection
+    ///    (registry cache miss) → allow
+    /// 4. Otherwise → deny
     ///
-    /// Rule 4 matches Go DefraDB behavior: replicator registration is
-    /// one-directional (source registers target), but the target accepts
-    /// push-log requests from any connected peer. Connected peers are
-    /// already authenticated via transport-level crypto. Document-level
-    /// ACP still applies independently at merge time.
+    /// Transport-level authentication proves *who* a peer is, not *what* they
+    /// are authorized to sync. Per-collection registry membership is the only
+    /// thing that grants collection-scoped access in Controlled mode, matching
+    /// Go DefraDB's `hasAccess` check (`go-p2p/peer.go`). A previous version
+    /// of this module accepted any connected peer as a fallback — see #838 for
+    /// the divergence that caused.
     ///
-    /// Important: collection access is broader than explicit replicator trust.
-    /// Callers that need to know whether a peer is an actual registered
-    /// replicator must use [`Self::is_registered_replicator`] instead of treating a
-    /// successful access check as equivalent.
+    /// Document-level ACP still applies independently at merge time; this
+    /// gate exists so that unauthorized peers never cause the receiver to
+    /// spend resources validating their blocks in the first place.
     ///
     /// Uses string-based registry lookup, supporting both libp2p and iroh peer IDs.
     pub(super) async fn check_access_str(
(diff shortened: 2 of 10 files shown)


fredcarle:
**Verdict: request changes — one regression plus a missing-tests gap.**

The core fix (removing the `is_connected` fallback for PushLog, two-stream, BranchableSync, and DocSync/CAR) is correct and closes real bypass paths in Controlled mode. Fail-closed behavior, TOCTOU placement, and session freshness all look right — checks fire before any state mutation, `ReplicatorRegistry` returns `false` on missing entries, and there are no cached per-session grants.

**1. GossipSub access is weakened, not tightened.** `crates/p2p/src/sync/coordinator/event_handler/gossip.rs` swaps `check_access_str` (which required registry membership per collection) for `is_connected` only. A connected peer that somehow ended up on the gossip mesh for a collection now passes the check even if they aren't a registered replicator for that collection. The PR rationale — "topic subscription is the gate" — is a local-side decision; it doesn't attest to the remote peer's authorization. Either restore the registry check, or, if this matches Go's gossip behavior, please link the Go equivalent and document it inline.

**2. `is_explicit_replicator` short-circuit needs provenance enforcement.** `TransportEvent::TwoStreamRequest` carries the flag as a plain `bool`. The PR description says it's set by the transport-level signed-challenge handshake, but nothing at the coordinator layer enforces that. A future caller (test, fuzz, or refactor) that constructs a `TwoStreamRequest { is_explicit_replicator: true }` directly bypasses the registry check. Either wrap the flag in a private type only constructible by the authenticated transport path, or at minimum add an invariant comment at the struct definition.

**3. Missing tests:**
- No test for GossipSub in Controlled mode with a connected-but-unregistered peer (boundary must be pinned down given the gate was changed).
- No positive test for the new short-circuit: unregistered peer + `is_explicit_replicator=true` allowed.

Bitswap egress gap (#830) is correctly scoped out of this PR.

_🤖 AI-assisted review_

iverc:
Thanks for the thorough review @fredcarle. Addressed all three in 0e6f00b8:

**1. Gossip gate strengthened.** You're right — swapping the registry check for `is_connected` was a net weakening. Restored it as **registered-for-collection OR subscribed-to-collection-topic, AND connected**. Topic subscription is a valid local authorization (subscribing means we opted into receiving on that topic — that's what broke the rust_rust ACP integration tests when I first went strict), but connectivity alone doesn't suffice. The gate now matches the original intent: direct push paths require registry membership; gossip paths accept either registry membership or a topic we subscribed to.

**2. `is_explicit_replicator` provenance.** Added an invariant comment on `HostEvent::TwoStreamRequest` stating that the flag must only be produced by the signed explicit-replicator handshake and that forwarding untrusted booleans into it bypasses Controlled-mode authorization. Agree a typed wrapper would be stronger — I went with the comment for minimal-diff scope. Happy to do the wrapper as a follow-up (would touch every construction site on both transports) if you think it's worth it.

**3. Missing tests added:**
- `gossip_controlled_mode_rejects_unregistered_unsubscribed_peer` — pins down the new boundary.
- `gossip_controlled_mode_allows_subscribed_collection` — validates the topic-subscription path.
- `gossip_controlled_mode_allows_registered_replicator` — validates the registry path.

The existing `two_stream_authenticated_explicit_replicator_is_marked_explicit_replicator` already covers the positive two-stream short-circuit (unregistered peer + `is_explicit_replicator=true` → block flows through), so I didn't duplicate it.

All 17 access_tests pass locally (3 new + 14 existing). Full workspace `cargo test --workspace --exclude integration-test --exclude ffi-test` clean.

iverc:
@copilot resolve the merge conflicts in this pull request

PR #868, opened 2026-04-16
fix(db-merge): replace unreachable!() with typed errors (#851)

### LINKED ISSUE #851: Merge handler panics on unknown `CrdtDelta` variant via `unreachable!()` fallback
## Summary

`crates/db-merge/src/merge_handler/mod.rs:477` dispatches CRDT delta variants to per-type handlers, with a final `_ => unreachable!()` catch-all. If a new variant is added to `CrdtDelta` — whether in this crate or in `defra-core/src/block_delta.rs` — the compiler will not error. The missing handler arm compiles silently and panics at runtime when a block carrying the new variant is processed. On a live node, that is a crash in the merge path on adversarial or future-version input.

## Details

`merge_handler/mod.rs:477` (current):

```rust
match delta {
    CrdtDelta::Lww(d) => { ... }
    CrdtDelta::Counter(d) => { ... }
    CrdtDelta::Composite(d) => { ... }
    CrdtDelta::Collection(d) => { ... }
    CrdtDelta::CollectionSet(d) => { ... }
    CrdtDelta::CollectionDefinition(d) => { ... }
    CrdtDelta::FieldDefinition(d) => { ... }
    _ => unreachable!(),                      // <-- panics on future variant
}
```

The `_` catch-all defeats exhaustiveness checking. Adding a new variant to `CrdtDelta` enum — in either `crates/db-merge` or the upstream `crates/defra-core/src/block_delta.rs` — will compile without warning, and the oversight only surfaces when a block carrying that variant reaches the merge handler.

The same Rust ecosystem idiom for this case is either:

1. **Exhaustive match** — drop the `_` arm and let the compiler enforce every variant has an arm. Use `#[non_exhaustive]` on the enum to preserve the ability to add variants without a breaking change, then match with explicit named arms.
2. **Explicit fallback** — keep a catch-all but return `Err(Error::UnsupportedDelta(...))` so the failure is observable and does not panic.

## Threat model

- **Correctness / forward compatibility**: a future Rust release with a new CRDT variant can crash an older-release node when a mixed network exchanges blocks.
- **Adversarial input**: a peer crafting a CBOR block that deserializes into a new or out-of-range enum variant can trigger the panic. (Serde's behaviour for unknown enum variants is usually an error at deserialize time, but any custom `Deserialize` impl that tolerates unknown tags would route such input to the fallback arm.)
- **Reachability today**: low — the enum is closed and every defined variant has an explicit arm. The concern is regression-on-change.

## Expected Behavior

Adding a new `CrdtDelta` variant produces a compile error until a handler is implemented. At runtime, an unknown or unsupported variant returns a typed error, not a panic.

## Fix

```rust
match delta {
    CrdtDelta::Lww(d) => { ... }
    CrdtDelta::Counter(d) => { ... }
    CrdtDelta::Composite(d) => { ... }
    CrdtDelta::Collection(d) => { ... }
    CrdtDelta::CollectionSet(d) => { ... }
    CrdtDelta::CollectionDefinition(d) => { ... }
    CrdtDelta::FieldDefinition(d) => { ... }
    // No wildcard arm — rely on compiler exhaustiveness check.
}
```

If `CrdtDelta` is (or should be) `#[non_exhaustive]` for out-of-crate addition resilience, keep a wildcard but return an error:

```rust
other => return Err(Error::UnsupportedDelta(format!("{:?}", other))),
```

Audit nearby dispatch sites for the same pattern — this issue may not be isolated to line 477.

## Acceptance Criteria

- [ ] `merge_handler/mod.rs:477` wildcard arm removed, replaced with exhaustive match or typed error return
- [ ] `cargo clippy --all -- -D warnings` passes
- [ ] A unit test constructs a synthetic unknown variant (or simulates one) and asserts the result is a typed `Err`, not a panic (only relevant if the wildcard path is retained with `Err`)
- [ ] Grep for `unreachable!()` across `crates/db-merge/` and audit each — replace with typed errors where the arm is reachable via external input
- [ ] Consider adding `#[non_exhaustive]` to `CrdtDelta` if it is meant to be extensible

## Files

- `crates/db-merge/src/merge_handler/mod.rs:477`
- Audit: `crates/db-merge/src/**/*.rs` for other `unreachable!()` occurrences
- Possibly `crates/defra-core/src/block_delta.rs` — `CrdtDelta` enum declaration

## References

- Rust Book: the difference between `unreachable!()` (a soundness promise you make to the compiler, panicking at runtime) and explicit error return (a contract you enforce). Adversarial / deserialization input is the classic case where the promise is unsafe.
- Related: defradb Go issue for `ipld_union.go` (`CRDT.GetDelta()` returns nil on `CollectionSetDelta` due to a duplicate switch case) — that bug is a non-exhaustive dispatch with the same shape of risk.

iverc:
## Summary

Replace all 5 `unreachable!()` catch-all arms in `crates/db-merge/` with typed error returns. All matched on `#[non_exhaustive]` enums (`CrdtDelta`, `NumericKind`, `EncryptedIndexType`), so the compiler could not enforce exhaustiveness — adding a new variant would compile silently and panic at runtime on the merge path.

## Changes

| File | Line | Enum | Replacement |
|---|---|---|---|
| `merge_handler/mod.rs` | 477 | `CrdtDelta` | `Err(MergeError::UnsupportedDelta(...))` |
| `merge_handler/batch.rs` | 419 | `CrdtDelta` | `Err(MergeError::UnsupportedDelta(...))` |
| `merge_handler/counter.rs` | 371 | `NumericKind` | `Err(MergeError::UnsupportedDelta(...))` |
| `merge_handler/counter.rs` | 430 | `NumericKind` | `tracing::warn` + `None` (value decode, not an error path) |
| `se/artifact_gen.rs` | 54 | `EncryptedIndexType` | `Err(storage::corekv::Error::Other(...))` |

## Test plan

- [x] `cargo test -p db-merge` — 58 tests pass

Closes #851.


fredcarle:
**Verdict: approve with one suggestion.**

Good security hardening — all five `CrdtDelta` / `NumericKind` / `EncryptedIndexType` sites are on the untrusted P2P decode path (block bytes arrive via Bitswap and flow through `Block::from_dag_cbor` before dispatch). With `#[non_exhaustive]` enums, a newer peer sending a new variant would crash the old node at `unreachable!()`. This PR turns each of those latent DoS vectors into a graceful error. Matches Go idiom (return error, not panic). `MergeError::UnsupportedDelta` already exists and is used elsewhere, so the error taxonomy stays consistent.

**One suggestion before merge:**

`crates/db-merge/src/merge_handler/mod.rs:477` and `crates/db-merge/src/merge_handler/batch.rs:419` use `format!("{:?}", other)` in the error message. Since `other` is a full `CrdtDelta`, this drops the variant payload — raw field-value bytes from user documents — into any error string that gets logged. Low severity (only hits unknown future variants) but worth tightening now. Consider `std::mem::discriminant(&other)` or a variant-name helper.

The `counter.rs:430` `None + warn` path is fine — CRDT state is already persisted by that point, so the missed in-memory field reconstruction is recoverable on re-read.

**Overlap note:** #878 makes the same fix at `se/artifact_gen.rs:54` as a standalone patch. This PR is the superset, so recommend landing this one and closing #878.

_🤖 AI-assisted review_

iverc:
Thanks for the review @fredcarle. Addressed in ff1091cd:

**Payload redaction (mod.rs:477 + batch.rs:419):** Swapped `format!("{:?}", other)` → `format!("{:?}", std::mem::discriminant(other))` on both sites. The error log now identifies the variant by its internal discriminant position without touching field-value bytes. I left the `counter.rs:430` `NumericKind` fallback alone per your note — `NumericKind` is a payload-less enum so there's nothing to redact there.

**Overlap with #878:** resolved the merge conflict in `se/artifact_gen.rs` by taking main's form (from #878). That site is already fixed, so this PR's contribution there is just the merge resolution. Happy to close #878 if preferred, though it's already merged. 🙂

All 58 db-merge tests still pass locally; fmt/clippy clean.

PR #869, opened 2026-04-16
fix(p2p): cap block count in CAR decoder to prevent memory DoS (#840)

### LINKED ISSUE #840: Security: CAR decoder lacks per-block-count cap, enabling memory DoS
## Summary

`crates/p2p/src/sync/car.rs::decode_car` iterates CAR blocks with no per-block-count cap. A hostile peer can send a CAR file consisting of millions of tiny block headers (valid CIDs, minimal data). Each header costs a few bytes on the wire but allocates a full `Cid` entry in a `Vec`. The 16 MiB message cap bounds the wire size, but `Vec<(Cid, Bytes)>` overhead for 10⁶+ entries can exhaust memory before the size check fires.

`CAR_MAX_BLOCKS` **is** enforced in `collect_exact_blocks` and `collect_recursive` (`car.rs:132, 173`), but those callers run after `decode_car` has already allocated the full in-memory Vec.

## Details

`car.rs:63-93` — core decode loop:
```rust
while let Some(block) = read_next_block(&mut reader)? {
    blocks.push(block);        // <-- no count cap here
}
```

Bounds present in the decoder:
- Varint header length check (`car.rs:68`).
- Varint individual block-size check (`car.rs:78-80`).

Bounds NOT present:
- Number of blocks pushed into `blocks`.

Wire-size bound is the 16 MiB limit on the full CAR message (in `p2p/src/codec.rs`). At ~40 bytes per block header (varint + CIDv1 + trivial data), 16 MiB ÷ 40 ≈ 400,000 entries. Each entry's runtime representation (`(Cid, Bytes)` plus Vec capacity overhead) is substantially larger than 40 bytes, so peak memory is several multiples of the wire size.

## Exploit

A peer sends a 16 MiB CAR containing ~400k trivial block entries. `decode_car` pushes all of them into a single `Vec` before any downstream cap check. Memory footprint may spike to 50–100 MiB per concurrent CAR fetch, scaled by `sync/replication` worker count. With the default `max_p2p_tasks = 64` and small per-peer limits, a single attacker can force a multi-GB spike across concurrent fetches. Enough attackers eventually OOM the node.

This is Rust-only code (CAR protocol is not implemented on Go side), so there is no parity concern — purely a security fix.

## Expected Behavior

`decode_car` aborts with an error once block count exceeds `CAR_MAX_BLOCKS`, matching the downstream caller limit. The check runs inside the loop before any further `push`.

## Fix

In `car.rs:63-93`:
```rust
let mut block_count = 0usize;
while let Some(block) = read_next_block(&mut reader)? {
    block_count += 1;
    if block_count > CAR_MAX_BLOCKS {
        return Err(Error::Codec(format!(
            \"CAR file exceeds maximum block count: {}\",
            CAR_MAX_BLOCKS
        )));
    }
    blocks.push(block);
}
```

## Acceptance Criteria

- [ ] `decode_car` enforces `CAR_MAX_BLOCKS` within the loop
- [ ] Unit test: feed a CAR whose block count exceeds `CAR_MAX_BLOCKS` and verify decode errors cleanly with no large allocation
- [ ] Unit test: feed a valid CAR with `CAR_MAX_BLOCKS` blocks and verify it decodes
- [ ] Review `extract_links`, `encode_car`, and other CAR paths for matching cap consistency

## Files

- `crates/p2p/src/sync/car.rs:63-93`
- `crates/p2p/src/sync/car.rs` (constants and test section)

## References

- Rust-only protocol; no Go counterpart.
- Related: the host-level `max_car_size` config (`p2p_host/mod.rs:46`) bounds bytes but not block count.

iverc:
## Summary

Add a `CAR_MAX_BLOCKS` check inside `decode_car`'s loop to prevent a hostile peer from exhausting memory via millions of tiny block headers packed into a single CAR message.

## Why

`decode_car` iterated CAR blocks with no count limit. A 16 MiB CAR packed with ~400k trivial block entries (varint + CIDv1 + 1 byte) fits within the wire-size cap but the `Vec<(Cid, Vec<u8>)>` runtime representation is several multiples larger. With concurrent CAR fetches, an attacker can force multi-GB memory spikes.

The downstream `collect_exact_blocks` and `collect_recursive` already enforce `CAR_MAX_BLOCKS`, but only after `decode_car` has allocated the full Vec. Moving the cap into the decode loop closes the gap.

## Changes

- `crates/p2p/src/sync/car.rs` — add `blocks.len() >= CAR_MAX_BLOCKS` check before each push in the decode loop
- Two new unit tests:
  - `decode_car_rejects_block_count_exceeding_cap` — feeds CAR_MAX_BLOCKS+1 blocks, asserts error
  - `decode_car_accepts_exactly_max_blocks` — feeds exactly CAR_MAX_BLOCKS blocks, asserts success

## Test plan

- [x] `cargo test -p p2p --lib car` — all 10 CAR tests pass

Closes #840.

### DIFF
--- a/crates/p2p/src/sync/car.rs
+++ b/crates/p2p/src/sync/car.rs
@@ -93,9 +93,18 @@ pub fn decode_car(data: &[u8]) -> Result<CarContents> {
     cursor = &cursor[header_len as usize..];
     let roots = decode_car_header(header_bytes)?;
 
-    // Read blocks
+    // Read blocks. Cap at CAR_MAX_BLOCKS to prevent a hostile peer from
+    // sending millions of tiny block headers that allocate unbounded memory
+    // before any downstream limit fires (#840).
     let mut blocks = Vec::new();
     while !cursor.is_empty() {
+        if blocks.len() >= CAR_MAX_BLOCKS {
+            return Err(Error::Codec(format!(
+                "CAR file exceeds maximum block count of {}",
+                CAR_MAX_BLOCKS
+            )));
+        }
+
         let section_len = read_varint(&mut cursor)?;
         if section_len as usize > cursor.len() {
             return Err(Error::Codec("CAR block section length exceeds data".into()));
@@ -470,5 +479,35 @@ mod tests {
 
         assert_eq!(collected.blocks.len(), depth);
         assert!(!collected.truncated());
+    }
+
+    #[test]
+    fn decode_car_rejects_block_count_exceeding_cap() {
+        let root = make_cid(b"root");
+        let blocks: Vec<(&Cid, &[u8])> = (0..CAR_MAX_BLOCKS + 1)
+            .map(|_| (&root, b"x".as_slice()))
+            .collect();
+        let encoded = encode_car(&[root], &blocks).unwrap();
+
+        let result = decode_car(&encoded);
+        assert!(result.is_err());
+        let err = result.unwrap_err().to_string();
+        assert!(
+            err.contains("maximum block count"),
+            "expected block-count error, got: {err}"
+        );
+    }
+
+    #[test]
+    fn decode_car_accepts_exactly_max_blocks() {
+        let root = make_cid(b"root");
+        let blocks: Vec<(&Cid, &[u8])> = (0..CAR_MAX_BLOCKS)
+            .map(|_| (&root, b"x".as_slice()))
+            .collect();
+        let encoded = encode_car(&[root], &blocks).unwrap();
+
+        let (roots, decoded) = decode_car(&encoded).unwrap();
+        assert_eq!(roots, vec![root]);
+        assert_eq!(decoded.len(), CAR_MAX_BLOCKS);
     }
 }


fredcarle:
**Verdict: approve.**

Clean DoS hardening. Correct placement (check at top of loop, before `read_varint` / `Cid::read_bytes` / `blocks.push`), correct boundary semantics (`>=` rejects N+1, and `decode_car_accepts_exactly_max_blocks` pins down the exact-N pass case), and the cap value `CAR_MAX_BLOCKS = 10_000` at `crates/p2p/src/sync/car.rs:18` matches the downstream `collect_exact_blocks` / `collect_recursive` guards — so the fix closes the pre-allocation gap rather than introducing a new one. `decode_car` is the only decode entry point; `car_has_any_block` only reads the header and allocates nothing.

No Go parity concern — the CAR path here is an internal Rust-to-Rust protocol (per the comment at `car.rs:248`), not exchanged with Go peers.

_🤖 AI-assisted review_

PR #878, opened 2026-04-16
fix/se-artifact-unsupported-index-type-panic

JesseAbram:


fredcarle:
**Verdict: approve, with one heads-up.**

The fix correctly replaces `unreachable!()` with an error return. `EncryptedIndexType` is `#[non_exhaustive]`, so the compiler-mandated wildcard was always meant to be a real code path. `enc_idx.index_type` is reachable from schema/deserialized-over-P2P input, so the old panic was a latent DoS vector from untrusted input. The `{:?}` on `EncryptedIndexType` only exposes the variant name, no secrets.

**Conflict with #868.** That PR makes the same fix to `crates/db-merge/src/se/artifact_gen.rs:54` as part of a broader 5-site cleanup of `unreachable!()` across `db-merge`. Whichever PR lands second will need to rebase; since #868 is the superset, easiest path is to land #868 and close this.

No Go parity check needed — SE artifact generation is Rust-only.

_🤖 AI-assisted review_

PR #880, opened 2026-04-16
fix(crdt): remove counter nonce idempotency, match Go merge semantics

### LINKED ISSUE #847: CRDT: Counter nonce-based idempotency diverges from Go, causes state divergence on retransmit

fredcarle:
## Summary
- Closes #847
- Go's `Counter.Merge` (`internal/core/crdt/counter.go:154-161`) ignores the Nonce field entirely — it exists only to make the signed DAG block's CID unique. Every delta that reaches Merge is applied unconditionally. Rust previously short-circuited when it saw a nonce marker from a prior apply — a convergence violation.
- Removed: `has_nonce` check in `apply_delta`, `mark_nonce` in `apply_delta` + `seed_if_uninitialized_*`, `clear_nonce` / `has_nonce` / `mark_nonce` / `nonce_key` methods, `nonce_prefix` field from `Counter` struct, `SkippedAlreadyApplied` variant + `was_skipped()` from `MergeResult`, `cleanup_counter_nonce_for_cid` in db-merge, old idempotency tests.
- Retained: `nonce` field on `CounterDelta` (wire compatibility — Go still serializes it for block CID uniqueness).
- Net: **-220 LOC**.

## Test plan
- [x] New test `regression_847_counter_applies_delta_each_merge` fails on `main` (value 5 / `SkippedAlreadyApplied`) and passes on this branch (value 10 / `Applied`).
- [x] `cargo test -p crdt --test counter_tests` — 20 tests pass, 0 fail.
- [x] `cargo test -p db-merge` — 58 tests pass, 0 fail.
- [x] `cargo clippy -p crdt -p db-merge --tests --no-deps -- -D warnings` — clean.

### DIFF
--- a/crates/crdt/src/counter.rs
+++ b/crates/crdt/src/counter.rs
@@ -1,9 +1,18 @@
 //! Counter CRDT implementation
 //!
-//! Supports increment and decrement operations with nonce-based idempotent delivery.
-//! Uses commutative addition with wrapping on overflow for Int64 (matching Go DefraDB),
-//! and IEEE-754 addition for Float64. This is not a traditional PN-Counter (which uses
-//! separate per-replica counters); instead it uses a single value with nonce tracking.
+//! Supports increment and decrement operations via commutative addition.
+//! Uses wrapping on overflow for Int64 (matching Go DefraDB) and IEEE-754
+//! addition for Float64.
+//!
+//! # Idempotency
+//!
+//! Counter merges are NOT idempotent — every delta that reaches `merge()`
+//! advances the value. This matches Go DefraDB
+//! (`internal/core/crdt/counter.go::Counter::Merge`) which ignores the
+//! `Nonce` field at merge time. `Nonce` exists on the delta solely to make
+//! each signed block's CID unique in the DAG; block-CID deduplication at
+//! the blockstore layer is the correct place to prevent double-application
+//! of the same delta across retransmits and crash recovery.
 
 use crate::traits::{Context, Delta, MergeResult, ReplicatedData, ValueReader};
 use async_trait::async_trait;
@@ -188,8 +197,6 @@ impl Delta for CounterDelta {
 pub struct Counter {
     /// Storage key for the counter value
     value_key: Vec<u8>,
-    /// Storage key for tracking applied nonces
-    nonce_prefix: Vec<u8>,
     /// Schema version
     schema_version_id: String,
     /// Field name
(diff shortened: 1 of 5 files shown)


fredcarle:
Closing in favor of #866 by @iverc which has broader coverage (8 files vs 5), property tests, and additional cleanup in batch.rs/composite.rs. Same core fix.

PR #884, opened 2026-04-16
fix(p2p): back off on rate-limited replicator pushes

jackzampolin:
## Summary
- retry the same PushLog when a replicator explicitly replies with the rate-limit backpressure message instead of continuing through the rest of the DAG
- route single-block replicator pushes through the same ordered send path so rejected replies are handled consistently
- centralize the rate-limit reply string and add focused coordinator tests for bounded retry behavior

## Testing
- cargo test -p p2p --lib

### DIFF
--- a/crates/p2p/src/error.rs
+++ b/crates/p2p/src/error.rs
@@ -6,6 +6,9 @@ use thiserror::Error;
 /// Result type for P2P operations.
 pub type Result<T> = std::result::Result<T, Error>;
 
+/// Error text used when the coordinator rejects a request due to peer backpressure.
+pub const RATE_LIMITED_MESSAGE: &str = "rate limited: too many requests, retry later";
+
 /// P2P error types.
 #[derive(Debug, Error)]
 #[non_exhaustive]
@@ -341,6 +344,11 @@ impl Error {
             } if collection_id == "rate-limited"
         )
     }
+}
+
+/// Returns true when a peer reply carries the coordinator's explicit rate-limit signal.
+pub fn is_rate_limited_message(message: &str) -> bool {
+    message == RATE_LIMITED_MESSAGE
 }
 
 /// Convert a blockstore CID verification error into its P2P counterpart.
--- a/crates/p2p/src/sync/coordinator/broadcast.rs
+++ b/crates/p2p/src/sync/coordinator/broadcast.rs
@@ -1,6 +1,7 @@
 //! Broadcasting local updates to the network.
 
 use std::collections::HashSet;
+use std::time::Duration;
 
 use acp::ReplicatedDocActorRelationships;
 use blockstore::Blockstore;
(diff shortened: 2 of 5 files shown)


fredcarle:
**Verdict: approve with suggestions.**

Clean backpressure handling. The core loop at `crates/p2p/src/sync/coordinator/broadcast.rs:363-438` correctly distinguishes rate-limit replies from other application errors, bounds retries at `MAX_RATE_LIMITED_PUSH_ATTEMPTS = 10` (~3.3s total backoff worst case: 25+50+100+200+400+500×5 ms), and preserves ordering by breaking the outer `'requests` loop when the budget is exhausted — downstream CIDs don't get sent out-of-order to a still-rate-limited peer. Centralizing `RATE_LIMITED_MESSAGE` in `error.rs:9` is the right call; the string was previously duplicated as `RATE_LIMITED_MSG` in `event_handler/mod.rs`, so responder and consumer can no longer drift.

**Hidden bug fix worth calling out in the PR description:** routing the single-block push through `send_ordered_pushlogs_via_transport` also fixes a silent failure mode. The pre-existing single-block path at `broadcast.rs:262-278` only surfaced transport errors via `report_push_failure` — application-layer `reply.err_message` (non-rate-limit rejections) were dropped on the floor. The new path now surfaces those too. That's a real correctness improvement beyond rate-limit handling and deserves a line in the summary.

**Suggestions:**

1. **Test gap: non-rate-limit error should not retry.** The two new tests cover (a) rate-limit retry succeeds, (b) rate-limit retry exhausts. Neither confirms that a non-rate-limit `err_message` (e.g., ACP rejection) breaks immediately without burning retry budget. Given this PR also fixes the silent-error bug above, a test pinning that behavior would be cheap insurance.

2. **Test gap: rate limit on Nth request doesn't re-retry earlier requests.** The ordered semantics are important — if CID#2 hits rate-limit after CID#1 succeeded, CID#1 must not be re-sent. Worth a test.

3. **`signing_state_test_guard()` scope concern.** `crates/sourcehub/src/lib.rs:17` adds a test-only mutex, and four tests in `cosmos_provider.rs` / `hub_rs/provider.rs` take the guard. But `defra_core::signing::{clear_identity_store, register_identity}` is global state — any other test in the crate that touches it (or cargo's parallel cross-test execution) could still race. Either confirm those four are the only callers in this crate, or promote the guard to `defra_core` so the lock is co-located with the state it protects. Otherwise this is a temporary patch that will re-break the first time someone adds another signing test.

4. **Minor: `request.clone()` on every iteration.** Line ~379 clones the `PushLogRequest` unconditionally, even on the first (usually successful) send. Not hot-path, but swapping to clone-on-retry (`if rate_limited_attempts > 0 { request.clone() }` pattern, or taking `request` by value and reconstructing) would avoid a byte-copy on the common case. Skip if measurement doesn't justify it.

5. **Nit: `MAX_RATE_LIMITED_PUSH_ATTEMPTS` vs the off-by-one.** The check `rate_limited_attempts > MAX_RATE_LIMITED_PUSH_ATTEMPTS` after increment means 11 total sends (1 initial + 10 retries). The test assertion `vec![cid1.to_bytes(); MAX_RATE_LIMITED_PUSH_ATTEMPTS + 1]` confirms this, but a one-line doc comment on the constant would spare future readers the count.

**Parity note:** the coordinator-level rate-limit string is defra.rs-internal (peers running Go wouldn't recognize or emit this exact text, and Go's upstream has its own backpressure path). Not a parity regression — just noting that this retry logic is a Rust-side-only optimization.

_🤖 AI-assisted review_

Issue #858, opened 2026-04-15
Improve P2P replication diagnostics and reduce CAR/Bitswap log noise


jackzampolin:
## Problem

While proving the defra-agent desktop multi-agent live journey against embedded defradb.rs nodes, the P2P path emits very high-volume WARN/INFO logs that make it hard to distinguish expected retry behavior from actionable replication failures.

This came up while running:

```bash
cargo test -p defra-agent-desktop desktop_app_live_multi_agent_server_switching_and_config_inference -- --ignored
```

The run was using defradb.rs `main` plus the gossipsub message-id fix from PR #857 (`29e0cc5c2b0f725c2943354dd2a27bb799f1ce7e`).

## Observed log patterns

These repeat hundreds of times during otherwise-progressing replication:

```text
WARN CAR fetch: empty response provider=... root=...
WARN CAR handler: no blocks found for request root_cid=... peer_id=... recursive=false requested_count=11
INFO Bitswap fetch completed query_id=... success=true error=None
WARN Failed to decode gossip message: Hit the end of buffer, expected more data
WARN P2P block has no signature — cannot verify authenticity cid=...
INFO DAG has missing links, requesting Bitswap fetch cid=... missing_count=... doc_id=...
INFO DAG complete, emitting DagReady root_cid=... doc_id=...
INFO Block merged successfully cid=... doc_id=...
```

Some fetches eventually complete and merge successfully after repeated empty responses. Other times the application-level runtime remains stuck waiting for referenced control documents, so the current logs do not make it obvious whether this is healthy retry noise, provider selection noise, a missing block availability bug, or a final replication failure.

## Requested improvements

- Downgrade expected transient misses/no-block responses to `debug` or throttle/deduplicate them by `(root_cid, provider, doc_id)`.
- Keep `warn` for final/actionable failures after retry budget exhaustion, not every intermediate miss.
- Include attempt number, retry budget/backoff, provider peer, requester peer, root CID, doc ID, collection ID/name when available, and whether a later retry completed the DAG.
- Make `Bitswap fetch completed success=true` distinguish “block fetched and merged”, “request completed with empty provider response but pending elsewhere”, and “nothing fetched”.
- Add structured logs around the transition from missing DAG links to `DagReady` to make it clear which missing links were resolved.
- Consider surfacing transaction conflicts, empty CAR responses, and missing-link retries as counters/metrics so integration tests can assert bounded retry behavior without scraping noisy logs.

## Why this matters

[...]

PR #887, opened 2026-04-16
Add iroh CAR fetch diagnostics

jackzampolin:
## Summary
- add provider-by-provider diagnostics for iroh selective CAR fetch failures
- include missing CID samples in the final unresolved DAG warning
- log responder-side context for rejected or header-only selective CAR responses

## Testing
- cargo fmt --all
- cargo clippy -p p2p --tests --features iroh-transport -- -D warnings
- cargo test -p p2p --features iroh-transport --lib 'iroh::endpoint_rpc::tests::summarize' -- --nocapture
- cargo test -p p2p --features iroh-transport bitswap_complete_retries_only_its_registered_root -- --nocapture

Stacked on #884.

### DIFF
--- a/crates/p2p/src/iroh/endpoint_rpc.rs
+++ b/crates/p2p/src/iroh/endpoint_rpc.rs
@@ -18,6 +18,96 @@ use super::protocols;
 /// needs time to process the request before replying.
 pub(super) const REQUEST_RESPONSE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
 
+#[derive(Debug, Clone)]
+struct CarFetchAttempt {
+    provider: PeerId,
+    outcome: CarFetchOutcome,
+}
+
+#[derive(Debug, Clone)]
+enum CarFetchOutcome {
+    Success,
+    InvalidPeerId(String),
+    ConnectFailed(String),
+    OpenBiFailed(String),
+    WriteFailed(String),
+    ReadFailed(String),
+    EmptyResponse,
+    HeaderOnlyCar,
+    EventChannelClosed,
+}
+
+impl CarFetchOutcome {
+    fn is_success(&self) -> bool {
+        matches!(self, Self::Success)
+    }
+
+    fn label(&self) -> &'static str {
+        match self {
+            Self::Success => "success",
+            Self::InvalidPeerId(_) => "invalid_peer_id",
+            Self::ConnectFailed(_) => "connect_failed",
+            Self::OpenBiFailed(_) => "open_bi_failed",
+            Self::WriteFailed(_) => "write_failed",
+            Self::ReadFailed(_) => "read_failed",
+            Self::EmptyResponse => "empty_response",
+            Self::HeaderOnlyCar => "header_only_car",
+            Self::EventChannelClosed => "event_channel_closed",
+        }
+    }
+
+    fn detail(&self) -> Option<&str> {
+        match self {
+            Self::InvalidPeerId(detail)
+            | Self::ConnectFailed(detail)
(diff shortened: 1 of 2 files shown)


fredcarle:
**Verdict: approve with suggestions.**

Useful diagnostics PR. `CarFetchOutcome` (at `crates/p2p/src/iroh/endpoint_rpc.rs:27`) gives each failure branch a stable label plus typed detail, and the aggregate WARN emitted from `handle_block_sync` now includes both the per-provider outcomes and a missing-CID sample — a real improvement over the previous "N providers tried, none worked" line, which was unactionable. The responder-side change at `crates/p2p/src/sync/coordinator/event_handler/car.rs:24-49` is well-scoped: non-recursive "no exact blocks" is now WARN (a selective request that can't be answered is a real anomaly), while recursive "no blocks yet" stays at debug (normal concurrent-replication race).

**Suggestions:**

1. **Unauthorized peers trigger a blockstore read before the access check.** `event_handler/car.rs:24-34` calls `self.manager.blockstore().has(&request.root_cid).await` unconditionally, *before* `check_peer_is_replicator`. `has()` is cheap, but this does let an unauthorized peer force one blockstore lookup per rejected CAR request purely for the reject-side log line. Low severity, but it inverts the fail-fast ordering we have elsewhere. Consider moving the `has()` call into the rejection branch only — compute `root_present` lazily after the access check passes or fails — so the happy path for rejected peers stays allocation-free.

2. **Reject-log volume under adversarial load.** The new WARN at `car.rs:36-47` logs full context (root CID, peer, replicator/connected state, counts) on every unauthorized request. A misbehaving or probing peer can generate arbitrary volume. Either rate-limit the WARN (e.g., token-bucket per peer) or drop it to `debug` and keep only a periodic aggregate at WARN. The existing `#858` aggregator pattern on the requester side could be the model.

3. **Synthetic peer_id for JoinHandle panic.** `endpoint_rpc.rs:430-434` inserts `PeerId::new("task".to_string())` when a spawned task panics. Loses the actual provider context. Maintain a parallel `Vec<PeerId>` indexed by `tasks`, or `JoinHandle<(PeerId, CarFetchAttempt)>`, so panics attribute to the right provider. Rare path, but the whole point of this PR is better attribution.

4. **Helper duplication.** `sample_cids` at `car.rs:12` (returns `Vec<String>`) and `summarize_cid_sample` at `endpoint_rpc.rs:98` (returns joined `String` with `+N more`) do the same truncation at different return types. Fine if intentional (structured field vs formatted string), but the magic constant `4` appears three times (`MAX_ATTEMPTS`, `MAX_CIDS`, and literal `4` in `car.rs:13`). One shared `const CAR_DIAG_SAMPLE_SIZE` would be worth it.

5. **Test coverage.** Tests cover the two formatting helpers. No test covers the new `CarFetchAttempt`-building branches in `try_fetch_from_provider` (each failure case). Given the PR is explicitly about "more diagnostic signal when fetches fail," a table-driven test that pins each variant's `label()` and `detail()` would guard against an outcome silently turning into `success` under refactor. Cheap to add — `CarFetchOutcome` is `Debug + Clone` already.

6. **Stacked PR mechanics.** Base is `fix/p2p-rate-limited-push-backoff` (#884). Once #884 merges, rebase onto `main` before merging this one so the diff view reflects the actual delta.

7. **`provider.clone()` on every return.** Each failure branch clones the `PeerId`. Minor, but a single `let provider = provider.clone();` at the top of `try_fetch_from_provider` (or constructing `CarFetchAttempt` once at the end via a `Result`-style flow) would cut the repetition.

_🤖 AI-assisted review_

iverc (approved):

PR #888, opened 2026-04-17
fix(db-blocks): reject secp256r1 for block signing to match Go (#849)

### LINKED ISSUE #849: Crypto: Secp256r1 signature S-normalization diverges from Go, breaks block-CID parity
## Summary

Rust's secp256r1 signing normalizes the signature to low-S form (`signature.normalize_s()`). Go's `ecdsa.SignASN1` does not normalize. Given identical private key and message, both implementations produce deterministic signatures (RFC 6979) — but **different signature bytes**. If secp256r1 signatures are stored inside content-addressed IPLD blocks, the resulting block CIDs diverge: a block signed in Rust has a different CID than the same logical block signed in Go, breaking replication convergence for any document whose signing identity uses secp256r1.

Rust's verification path already accepts both forms (it normalizes on verify at `secp256r1.rs:227`), so the bug is unidirectional: Rust can verify Go-signed sigs, but the produced bytes and therefore CIDs are not identical.

### Rust

`crates/crypto/src/keys/secp256r1.rs:105-117` (`Secp256r1PrivateKey::sign`):
```rust
let signature = self.signing_key.sign(message);
let signature = signature.normalize_s().unwrap_or(signature);   // <-- normalizes to low-S
Ok(signature.to_vec())
```

Verification at `:219-239` accepts both low-S and high-S (via `normalize_s()` on the parsed signature before verify), so verification works across both impls.

### Go

`defradb/crypto/keys.go:329-334` (`Secp256r1PrivateKey.Sign`):
```go
signature, err := ecdsa.SignASN1(rand.Reader, priv.key, digest)
```

No normalization — Go emits whichever S value the deterministic signing happened to produce (usually the "high" half, since ecdsa doesn't normalize by default).

### Where the divergence matters

If a secp256r1 signature is embedded in a content-addressed block (e.g., a signed CRDT delta, a signed replicator announcement, a signed identity token stored in the block DAG), the signature bytes are part of the CBOR-encoded block, and the block's CID = SHA-256 over those bytes. Different signature bytes → different CID.

The issue is zero-cost if secp256r1 signatures only appear in ephemeral / non-block-addressed contexts (auth tokens in HTTP headers, JWT, wire-level signed messages that are not persisted). Need to audit the codebase to determine which.

Prior audit finding #04 noted this; the current Rust code includes the normalization deliberately (citing compatibility with upstream ecosystems). The parity problem with Go was not resolved there.

## Expected Behavior

Secp256r1 signatures are byte-identical between Rust and Go for the same key and message, so blocks signed on either implementation produce the same CID and replicate cleanly.

## Resolution options

1. **Match Go (recommended)**: remove the `normalize_s()` call in Rust signing. Rust verification already accepts both forms, so no regression in verification. Existing Rust-signed secp256r1 artifacts (if any are persisted) need a migration path, which is why this is tracked separately from other crypto polish.
2. **Change Go**: add S-normalization in Go's secp256r1 signing. This is a Go upstream change; if preferred, file a companion issue on `sourcenetwork/defradb` and track the coordination.
3. **Declare a divergence**: document that secp256r1 signatures are not cross-impl byte-identical and never embed them in content-addressed blocks. This requires an audit to confirm no current path embeds them, and a lint or runtime guard to enforce it going forward.

Option 1 is the lowest-coordination path if block persistence is the dominant use case.

## Acceptance Criteria

- [ ] Audit every call site of `Secp256r1PrivateKey::sign` in the Rust codebase. For each, document whether the signature output is (a) embedded in a content-addressed block, (b) transmitted on the wire without persistence, or (c) persisted outside the block DAG.
- [ ] Based on the audit, pick one of the three resolution options above and apply it. If option 1, remove the `normalize_s()` at `secp256r1.rs:114` and update the `sign` doc comment to match.
- [ ] Integration test: Rust and Go sign the same message with the same key; assert byte-identical signature output (once the chosen resolution is in place).
- [ ] CID parity test: a block containing a secp256r1 signature has the same CID whether signed in Rust or Go.

## Files

- `crates/crypto/src/keys/secp256r1.rs:105-117` — the `sign` method
- `crates/crypto/src/keys/secp256r1.rs:219-239` — the `verify` method (update comment to reflect symmetric behavior once normalize-on-sign is removed)
- Audit: grep for `Secp256r1PrivateKey::sign` across the tree

## References

- Go: `defradb/crypto/keys.go:329-334` — `ecdsa.SignASN1` with no normalization
- Prior audit: `audit/past/2026-02-pre-1.0-security-audit/01-cryptographic-inventory-findings/04-secp256r1-s-normalization.md` (if present — earlier flagged this and did not fix)
- secp256r1 is used by the DID key method `did:key:zDn...` (multicodec 0x1200); any DID-based signing flow is a candidate site.

fredcarle:
Resolves #849

## Summary

Rust was signing blocks with secp256r1 (both locally and via remote signer). Go's `internal/core/block/signing.go:74-79` rejects any key type other than secp256k1 or Ed25519 with `ErrUnsupportedKeyForSigning`, and `getPublicKeyFromSignature` at `signature.go:186-193` only accepts ES256K and EdDSA for verification. Any ES256-signed block a Rust node produced would be rejected by a Go peer, breaking cross-implementation replication.

Fix: reject secp256r1 at `compute_signature` so a Rust node never emits a block a Go node refuses to verify.

## Scope clarification

The filed issue described a signature-byte S-normalization gap. The audit found:
- Go's `ecdsa.SignASN1` uses a random nonce (`rand.Reader`), so byte-identical signatures between Rust (RFC 6979 deterministic) and Go are impossible regardless of `normalize_s`.
- The real parity gap is that Go refuses to sign or verify secp256r1 blocks at all.

Fixing the sign path is the minimum to guarantee a Rust node cannot produce a block a Go peer will reject.

Non-block secp256r1 paths (JWT tokens, DID generation, raw identity signing) are unchanged — those already interoperate fine.

## Test plan

- [x] `test_compute_signature_rejects_local_secp256r1_block_signing` — asserts local secp256r1 signing returns an error. Failed on main, passes with fix.
- [x] `test_compute_signature_rejects_remote_secp256r1_block_signing` — asserts remote-signer secp256r1 is also rejected (can't bypass by wrapping in a RemoteSigner).
- [x] `test_compute_signature_passes_signing_authorization_to_remote_signer` — switched from secp256r1 to Ed25519 so the authorization pass-through test still exercises remote signing.
- [x] `cargo test -p db-blocks` — all 8 tests pass.
- [x] `cargo test -p db-merge` — all 58 tests pass (verify path is unchanged).
- [x] `cargo clippy -p db-blocks -p db-merge -p defra-core --tests --no-deps -- -D warnings` — clean.
- [x] `cargo build --workspace` — clean.

### DIFF
--- a/crates/db-blocks/src/lib.rs
+++ b/crates/db-blocks/src/lib.rs
@@ -59,18 +59,21 @@ pub(crate) fn compute_signature(
         return Ok(None);
     }
 
-    // Go's internal/core/block/signing.go:74-79 rejects any key type other
-    // than secp256k1 or Ed25519 with ErrUnsupportedKeyForSigning. Produce
-    // the same error so a Rust node never emits a signed block that a Go
-    // node would refuse to verify (signature.go:186-193).
-    if matches!(
-        signer.key_type,
-        defra_core::signing::SigningKeyType::Secp256r1
-    ) {
-        return Err(format!(
-            "unsupported key type for block signing: {} (only secp256k1 and ed25519 are supported, matching Go)",
-            signer.key_type
-        ));
+    // Only secp256k1, Ed25519, and BLS are supported for block signing.
+    // secp256k1 and Ed25519 match Go's internal/core/block/signing.go:74-79
+    // and signature.go:186-193. BLS is a Rust-specific extension wired
+    // through a remote signer (Orbis ring); the local-signing match below
+    // still rejects local BLS with a clearer error.
+    match signer.key_type {
+        defra_core::signing::SigningKeyType::Secp256k1
+        | defra_core::signing::SigningKeyType::Ed25519
+        | defra_core::signing::SigningKeyType::Bls => {}
+        other => {
+            return Err(format!(
+                "unsupported key type for signing. KeyType: {}",
+                other
+            ));
+        }
     }
 
     // Serialize the block (without signature) to get the bytes to sign
--- a/crates/db-blocks/src/lib.rs
+++ b/crates/db-blocks/src/lib.rs
@@ -59,6 +59,20 @@ pub(crate) fn compute_signature(
         return Ok(None);
     }
 
+    // Go's internal/core/block/signing.go:74-79 rejects any key type other
+    // than secp256k1 or Ed25519 with ErrUnsupportedKeyForSigning. Produce
+    // the same error so a Rust node never emits a signed block that a Go
+    // node would refuse to verify (signature.go:186-193).
+    if matches!(
+        signer.key_type,
+        defra_core::signing::SigningKeyType::Secp256r1
+    ) {
+        return Err(format!(
+            "unsupported key type for block signing: {} (only secp256k1 and ed25519 are supported, matching Go)",
+            signer.key_type
+        ));
+    }
+
     // Serialize the block (without signature) to get the bytes to sign
     let block_bytes = block
         .to_dag_cbor()
@@ -82,14 +96,6 @@ pub(crate) fn compute_signature(
                 let private_key =
                     crypto::Secp256k1PrivateKey::from_bytes(&signer.private_key_bytes)
                         .map_err(|e| format!("Failed to load secp256k1 private key: {}", e))?;
-                private_key
-                    .sign(&block_bytes)
-                    .map_err(|e| format!("Failed to sign block: {}", e))?
-            }
-            defra_core::signing::SigningKeyType::Secp256r1 => {
-                let private_key =
-                    crypto::Secp256r1PrivateKey::from_bytes(&signer.private_key_bytes)
-                        .map_err(|e| format!("Failed to load secp256r1 private key: {}", e))?;
                 private_key
                     .sign(&block_bytes)
                     .map_err(|e| format!("Failed to sign block: {}", e))?
--- a/crates/db-blocks/src/tests.rs
+++ b/crates/db-blocks/src/tests.rs
@@ -1,8 +1,7 @@
 use super::*;
 use blockstore::{Blockstore, DefraBlockstore};
-use crypto::keys::PublicKey;
+use crypto::keys::Key;
 use defra_core::encryption::EncryptionConfig;
-use defra_core::SignatureType;
 use std::sync::{Arc, Mutex};
 use storage::backends::MemoryStore;
 
@@ -135,11 +134,11 @@ fn test_compute_document_blocks_places_encryption_metadata_in_blockstore_entries
     );
 }
 
-struct TestRemoteSecp256r1Signer {
+struct LocalSecp256r1Signer {
     private_key: crypto::Secp256r1PrivateKey,
 }
 
-impl defra_core::signing::RemoteSigner for TestRemoteSecp256r1Signer {
+impl defra_core::signing::RemoteSigner for LocalSecp256r1Signer {
     fn sign_sync(
         &self,
         data: &[u8],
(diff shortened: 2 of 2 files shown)


crates/db-blocks/src/lib.rs
```diff
@@ -59,6 +59,20 @@ pub(crate) fn compute_signature(
...
+    // Go's internal/core/block/signing.go:74-79 rejects any key type other
+    // than secp256k1 or Ed25519 with ErrUnsupportedKeyForSigning. Produce
+    // the same error so a Rust node never emits a signed block that a Go
+    // node would refuse to verify (signature.go:186-193).
+    if matches!(
+        signer.key_type,
+        defra_core::signing::SigningKeyType::Secp256r1
+    ) {
+        return Err(format!(
+            "unsupported key type for block signing: {} (only secp256k1 and ed25519 are supported, matching Go)",
```
fredcarle: thought: I'll fix the irrelevant `matching go` part in that message.

crates/db-blocks/src/lib.rs
```diff
@@ -59,6 +59,20 @@ pub(crate) fn compute_signature(
         return Ok(None);
     }
 
+    // Go's internal/core/block/signing.go:74-79 rejects any key type other
+    // than secp256k1 or Ed25519 with ErrUnsupportedKeyForSigning. Produce
+    // the same error so a Rust node never emits a signed block that a Go
+    // node would refuse to verify (signature.go:186-193).
+    if matches!(
```
JesseAbram: nit: This works but if we are trying to accept all but maybe we shouldn't exclude Secp256r1 but be more intentional and include only secp256k1 or Ed25519

JesseAbram (approved):
approved with a nit

crates/db-blocks/src/lib.rs
```diff
@@ -86,14 +100,6 @@ pub(crate) fn compute_signature(
                     .sign(&block_bytes)
                     .map_err(|e| format!("Failed to sign block: {}", e))?
             }
-            defra_core::signing::SigningKeyType::Secp256r1 => {
```
JesseAbram, line 89: nit: again this works but I would say also remove bls or keep this in with a return err like the bls below for consistency

crates/db-blocks/src/lib.rs
fredcarle: Switched to an allow-list of secp256k1/Ed25519/BLS in 291a93f2. Includes BLS because it's a legitimate remote-signer path (Orbis); the local match below keeps the 'BLS requires remote signer' error for that case.

crates/db-blocks/src/lib.rs
fredcarle: Kept BLS in the local match because of its asymmetric semantics: BLS has a valid remote-signer path (Orbis) so it can't be rejected wholesale at the top, whereas secp256r1 has no valid path at all and is fully rejected by the allow-list. With 291a93f2 secp256r1 now falls through the same `other =>` arm as any unrecognised type, which I think is closer to what you wanted.

crates/db-blocks/src/lib.rs
fredcarle: Done in 291a93f2 — the error message now matches Go's `errUnsupportedKeyForSigning` verbatim: `"unsupported key type for signing. KeyType: <type>"`.

iverc (approved):

PR #889, opened 2026-04-17
fix(db-merge): derive secp256k1 peer DIDs via crypto crate (#852)

### LINKED ISSUE #852: Peer identity: secp256k1 DID conversion unimplemented, blocks ACP on secp256k1 peers
## Summary

`crates/db-merge/src/peer_identity.rs::public_key_to_did` returns `Error::UnsupportedKeyType` for secp256k1 peer keys. Go (`internal/db/p2p/protocol/identity.go`) supports both Ed25519 and secp256k1 for DID conversion and peer-identity exchange. As a result, a Rust node connected to a secp256k1-keyed peer cannot produce a `did:key` for that peer, and any downstream ACP lookup keyed on that DID fails.

Practically: Rust cannot apply per-document access control to blocks originating from secp256k1-authenticated peers.

## Details

`crates/db-merge/src/peer_identity.rs:99-113`:

```rust
pub fn public_key_to_did(pk: &libp2p::identity::PublicKey) -> Result<String> {
    match pk.key_type() {
        libp2p::identity::KeyType::Ed25519 => {
            // derive did:key:z6Mk...
        }
        libp2p::identity::KeyType::Secp256k1 => {
            // Note: secp256k1 needs uncompressed key expansion — not implemented
            return Err(Error::UnsupportedKeyType);
        }
        _ => return Err(Error::UnsupportedKeyType),
    }
}
```

The comment flags the missing work but does not implement it. Go upstream has the full conversion, and the `crypto` crate in this repo already has the building blocks: `crypto::keys::secp256k1::Secp256k1PublicKey` exposes the raw bytes and the DID module at `crates/crypto/src/did/mod.rs` already knows the secp256k1 multicodec (0xe7). The gap is the libp2p-to-crypto-keys adapter for secp256k1.

## Impact

- **ACP bypass via unsupported identity**: a secp256k1-keyed peer's actions can't be mapped to their DID, so any ACP path that gates on DID returns "unknown peer." Depending on the downstream handler, this results in either a rejected merge (a denial-of-replication) or, worse, a fallback path that treats the peer as anonymous.
- **Mixed-network interop**: Go nodes that default to secp256k1 for peer identity can't fully interoperate with Rust nodes for ACP-gated collections.
- **Silent failure**: the error is returned at the call site but may be logged and ignored, depending on the caller.

## Expected Behavior

`public_key_to_did` supports secp256k1 peer keys alongside Ed25519, producing a canonical `did:key:zQ3s...` string identical to Go's output.

## Fix

Extract compressed secp256k1 public-key bytes from the libp2p `PublicKey`, decompress if necessary (the DID encoding typically uses the compressed 33-byte form with multicodec 0xe7), and hand off to `crypto::did::create_did_key`.

Sketch:

```rust
libp2p::identity::KeyType::Secp256k1 => {
    let raw = pk.encode_protobuf();                  // proto-encoded
    let inner = libp2p::identity::secp256k1::PublicKey::try_from_bytes(&raw[...])?;
    let compressed = inner.to_bytes();               // 33-byte compressed SEC1
    crypto::did::create_did_key(
        crypto::types::KeyType::Secp256k1,
        &compressed,
    )
}
```

(Exact method names depend on the libp2p-rust version; the key concept is: compressed 33-byte SEC1 → multicodec 0xe7 → did:key:zQ3s...)

## Acceptance Criteria

- [ ] `public_key_to_did` handles secp256k1 keys and returns a `did:key:zQ3s...` identical to Go's output for the same libp2p PeerID
- [ ] Unit test: construct a libp2p secp256k1 keypair, derive the DID via Rust, compare against a fixture from Go
- [ ] Unit test: `peer_id_to_did` round-trips for both Ed25519 and secp256k1 peers
- [ ] Integration test: a secp256k1-keyed peer attempts to push an ACP-gated block; Rust recipient correctly identifies the peer's DID and applies the policy

## Files

- `crates/db-merge/src/peer_identity.rs:99-113` — the missing branch
- Possibly `crates/crypto/src/did/mod.rs` — confirm the entry point is exposed for secp256k1
- Tests

## References

- Go: `defradb/internal/db/p2p/protocol/identity.go` — supports both key types
- Go: `defradb/crypto/keys.go` — secp256k1 public key shape
- Rust: `crates/crypto/src/did/mod.rs:23-32` — multicodec table (secp256k1 is 0xe7)
- Rust: `crates/crypto/src/keys/secp256k1.rs` — has the raw/compressed utilities

fredcarle:
Resolves #852

## Summary

`public_key_to_did` for secp256k1 peers returned `UnsupportedKeyType("secp256k1 DID conversion requires uncompressed key expansion")` — blocking any ACP lookup keyed on a secp256k1-authenticated peer. Go supports both Ed25519 and secp256k1 and derives the DID from the uncompressed SEC1 public key (`defradb/crypto/keys.go:276-282`).

Fix: extract the compressed 33-byte key from libp2p's protobuf (`encoded[4..37]`), construct a `crypto::Secp256k1PublicKey` (which already decompresses the point to 65 bytes), and delegate to its `did()` — the same path already byte-parity-tested against Go in `crates/crypto/tests/go_compat_keys.rs::test_secp256k1_did_matches_go`.

Also enables the `secp256k1` feature on the workspace `libp2p` dependency; without it libp2p cannot even decode secp256k1 keys, so the existing match arm was effectively unreachable from real peer traffic.

## Test plan

- [x] `test_secp256k1_peer_to_did_matches_crypto_did` — constructs a libp2p secp256k1 keypair, runs it through `peer_id_to_did` and `public_key_to_did`, asserts they agree, start with `did:key:z`, and round-trip via `parse_did_key` back to `(Secp256k1, 65 uncompressed bytes, 0x04 prefix)`. Failed on main (`UnsupportedKeyType(...)`), passes with fix.
- [x] `test_secp256k1_peer_to_did_is_deterministic` — same peer → same DID across calls.
- [x] `test_secp256k1_mapper_function` — closure from `create_peer_to_did_mapper` maps a secp256k1 peer ID string to a DID.
- [x] Existing Ed25519 tests still pass.
- [x] `cargo test -p db-merge --lib peer_identity` — 8/8.
- [x] `cargo clippy -p db-merge --tests --no-deps -- -D warnings` — clean.
- [x] `cargo build --workspace` — clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -927,6 +927,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "asn1_der"
+version = "0.7.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4858a9d740c5007a9069007c3b4e91152d0506f13c1b31dd49051fd537656156"
+
+[[package]]
 name = "async-broadcast"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -7093,9 +7099,11 @@ version = "0.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f0c7892c221730ba55f7196e98b0b8ba5e04b4155651736036628e9f73ed6fc3"
 dependencies = [
+ "asn1_der",
  "bs58",
  "ed25519-dalek 2.2.0",
  "hkdf",
+ "k256",
  "multihash 0.19.3",
  "quick-protobuf",
  "rand 0.8.5",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,7 +84,7 @@ ciborium = "0.2"
 
 # P2P networking
 # Note: Using 0.53 to match iroh-bitswap's dependency
-libp2p = { version = "0.53", features = ["tcp", "noise", "yamux", "gossipsub", "kad", "identify", "relay"] }
+libp2p = { version = "0.53", features = ["tcp", "noise", "yamux", "gossipsub", "kad", "identify", "relay", "secp256k1"] }
 iroh-bitswap = { git = "https://github.com/sourcenetwork/beetle", branch = "main" }
 libp2p-stream = "0.1.0-alpha.1"
 multiaddr = "0.18"
--- a/crates/db-merge/src/peer_identity.rs
+++ b/crates/db-merge/src/peer_identity.rs
@@ -12,7 +12,11 @@
 //! # Supported Key Types
 //!
 //! - Ed25519 (most common for libp2p)
-//! - secp256k1 (used by some networks)
+//! - secp256k1 (used by some networks, common with Go peers)
+//!
+//! Both key types have short public keys that libp2p inlines into the PeerId
+//! under the 42-byte identity-hash threshold, so `peer_id_to_did` can recover
+//! the key material from the PeerId alone.
 
 use crypto::keys::PublicKey as _;
 use crypto::{create_did_key, KeyType, Secp256k1PublicKey};
@@ -63,9 +67,10 @@ pub enum PeerIdentityError {
 /// let did = peer_id_to_did(&peer_id)?;
 /// ```
 pub fn peer_id_to_did(peer_id: &PeerId) -> Result<Did, PeerIdentityError> {
-    // Try to extract the public key from the PeerId
-    // Note: This only works for PeerIds that contain an inline public key
-    // (which is the case for Ed25519 keys used by most libp2p implementations)
+    // Works for PeerIds that inline their public key. Ed25519 (36-byte
+    // protobuf) and secp256k1 (37-byte protobuf) both fit under libp2p's
+    // 42-byte identity-hash threshold, so inline decoding is the common
+    // case for both supported key types.
     let public_key = PublicKey::try_decode_protobuf(&peer_id.to_bytes()[2..])
         .map_err(|e| PeerIdentityError::KeyExtraction(e.to_string()))?;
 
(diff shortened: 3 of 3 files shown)


fredcarle:
**Verdict: approve with suggestions — fix is correct, acceptance-criteria gap worth flagging.**

**Does this fix the linked issue?** Yes. The core claim holds: libp2p hands us compressed 33-byte SEC1 (protobuf `[4..37]`), `Secp256k1PublicKey::from_bytes` accepts compressed input at `crates/crypto/src/keys/secp256k1.rs:162` via `EncodedPoint::from_bytes`, and the downstream `.did()` call is the same path already Go-parity-tested in `crates/crypto/tests/go_compat_keys.rs:733` (`test_secp256k1_did_matches_go` → `SECP256K1_DID`). So the full chain is provably byte-identical to Go, transitively.

A genuine bug-fix-adjacent correction worth noting: the old length check `len() < 35` was wrong — the true libp2p secp256k1 protobuf length is 37 bytes (`0x08 0x02 0x12 0x21` + 33). Tightened correctly. Enabling the `secp256k1` feature on workspace `libp2p` is load-bearing — without it libp2p can't even decode the key, so the old match arm was unreachable in production regardless of whether the branch body was correct.

**Slop check: clean.** I looked for dead imports, unnecessary abstractions, over-commented trivia, redundant tests, and scope creep. None present. The new comments are the right kind (WHY — citing `defradb/crypto/keys.go:276-282` and explaining why we route through the crypto crate rather than reimplementing the 33→65 decompression inline). The two new tests target distinct code paths (end-to-end `peer_id_to_did` + mapper closure). The only mild duplication is `test_secp256k1_peer_to_did_is_deterministic` mirroring the existing `test_deterministic_did` — fine, different key type, different code path.

**Gaps vs. the issue's acceptance criteria:**

1. **AC4 unmet: no integration test.** Issue #852 explicitly lists _"a secp256k1-keyed peer attempts to push an ACP-gated block; Rust recipient correctly identifies the peer's DID and applies the policy."_ The unit tests here prove the conversion works in isolation, but nothing exercises the ACP path that was the original failure surface. Given this is a priority:p1 correctness bug tagged `area:identity` + `area:p2p`, a test in `tools/integration-test/tests/acp/` or `tools/integration-test/tests/p2p/` with a secp256k1-keyed node pushing into an ACP-gated collection would close the loop. Worth filing a follow-up issue if not in this PR.

2. **AC2 unmet in this crate: no Go fixture comparison at the libp2p boundary.** The PR's argument — "the crypto crate path is Go-parity-tested, and we just feed it the libp2p bytes" — is technically sound but implicit. A fixture test here (construct `Secp256k1PrivateKey` from a known byte vector, build the libp2p `Keypair` from the same private key, run through `public_key_to_did`, compare to the known-Go `SECP256K1_DID` constant) would make the byte-extraction step explicitly covered. Cheap — the constants already exist in `go_compat_keys.rs`.

3. **Tighten test assertions to `did:key:zQ3s...`.** Issue body and AC1 specify the `zQ3s` prefix (base58btc multibase + multicodec 0xe7). Current tests assert only `did:key:z`. A wrong multicodec byte in the crypto layer would change this prefix; the transitive Go test catches it, but a tighter local assertion pins the contract at this layer and documents the expected format for future readers. One-line change.

4. **Stale module doc at `peer_identity.rs:66-67`.** The comment on `peer_id_to_did` says _"This only works for PeerIds that contain an inline public key (which is the case for Ed25519 keys used by most libp2p implementations)."_ Both Ed25519 (36 bytes) and secp256k1 (37 bytes) fit under libp2p's 42-byte identity-hash threshold, so inline decoding works for both. Update the comment to reflect that now that secp256k1 is supported.

5. **Minor: `encoded[4..37]` is correct but brittle.** The same protobuf-byte-range extraction pattern (`[4..36]` for Ed25519, `[4..37]` for secp256k1) is duplicated with magic numbers. Since libp2p is the source and the format is fixed, a one-line helper like `fn extract_inline_key(encoded: &[u8], expected_len: usize) -> Result<&[u8], PeerIdentityError>` would centralize the bounds check. Not worth doing here — just noting it as a refactor hook if a third key type ever shows up (secp256r1 is #849).

**Transport feature flag side-effect:** enabling `libp2p/secp256k1` pulled in `asn1_der` and `k256` (already a direct dep of the crypto crate, so deduped). Worth confirming `cargo tree -d` stays clean.

_🤖 AI-assisted review_

fredcarle:
Addressed the review in f1a97dcf:

1. **AC2 Go-fixture parity test (item 2)**: added `test_secp256k1_public_key_to_did_matches_go_fixture` and `test_secp256k1_peer_id_to_did_matches_go_fixture` — both drive the known 32-byte `GO_SECP256K1_PRIVATE_KEY` through libp2p's protobuf path and assert the result is byte-identical to the Go-generated `GO_SECP256K1_DID` constant. This makes the libp2p-boundary parity explicit instead of merely transitive.
2. **Tighten prefix (item 3)**: assertion is now `did:key:z7r8` (the base58btc prefix for multicodec 0xe7 + 65-byte uncompressed SEC1 — which is what Go emits). Note: the review suggested `zQ3s`, but that prefix corresponds to the *compressed* 33-byte encoding that the did:key spec prescribes; Go uses uncompressed, so `z7r8` is the correct parity target.
3. **Stale module doc (item 4)**: refreshed the module-level doc and the `peer_id_to_did` comment so neither still implies inline-key decoding is Ed25519-only.
4. **AC4 integration test (item 1)**: filed #890 as a follow-up — adding a full ACP integration test for a secp256k1-keyed peer is a non-trivial multi-node test and worth its own PR.
5. **Minor helper refactor (item 5)**: skipped per your own note — worth revisiting if a third inline key type shows up.
6. **`cargo tree -d`**: ran, no new duplicates introduced by the libp2p/secp256k1 feature flip. k256 was already a direct workspace dep; only `asn1_der` is net new, and it has no other versions in-tree.

10/10 tests pass, clippy clean, workspace builds.

PR #891, opened 2026-04-18
fix(p2p): record authenticated gossip peers before access checks

jackzampolin:
## Summary
- record authenticated gossip senders as connected before coordinator access checks run
- treat peer subscription events as connected for the same coordinator peer-state invariant
- add regression coverage for controlled-mode gossip ingress and peer-subscription ordering

## Testing
- cargo test -p p2p access_tests -- --nocapture


jackzampolin:
Closing this as superseded. The review is correct that my patch broadened coordinator  semantics beyond the actual transport lifecycle by inferring liveness from gossip/subscription events. That masks the ordering bug instead of fixing the event ordering guarantee at the transport/event-loop boundary. PR 892 is the better direction for the real race.

PR #894, opened 2026-04-20
feat(storage): add Lark backend and make it the default

iverc:
## Summary

Adds pure-Rust [`lark-kv`](https://github.com/sourcenetwork/lark) as a new storage backend alongside `redb` / `fjall` / `rocksdb` and sets it as the default for `cli`, `defra-node`, `embedded`, and `ffi`. RocksDB is still buildable and selectable via `--store rocksdb` — nothing is removed.

## Motivation

- Lark is our in-house pure-Rust LSM-tree KV store. Shipping it as the defradb.rs default closes the last C/C++ dependency (`librocksdb-sys` → `bzip2-sys` / `lz4-sys`) in native builds.
- Smaller dep graph, no linker surprises, no cross-compile pain, and matches the "pure Rust" posture of the rest of the workspace.
- RocksDB stays on the shelf as an opt-in comparison backend.

### New backend — `crates/storage/src/backends/lark/`

Six files, ~950 LOC. Mirrors the shape of the existing RocksDB backend:

- `config.rs` — `LarkStoreOptions` with `from_env()` (`LARK_*` variables) and a `DurabilityMode`-aware `to_lark_options()` conversion.
- `store.rs` — `LarkStore::open` / `open_with_options`, `Store` + `Dropable` impls, `close_timeout`-based graceful shutdown with `active_txn_count`.
- `transaction.rs` — `LarkTxn` wrapping a lark `Snapshot` for point-in-time reads plus a buffered `BTreeMap<Vec<u8>, Option<Vec<u8>>>` for pending writes. Provides the same read-your-writes + snapshot-isolation semantics as the RocksDB backend; commit goes through `lark_kv::WriteBatch` and `write_with_durability`.
- `iterator.rs` — `LarkMergingIterator` overlays the pending map on the snapshot scan (identical structure to `RocksDbMergingIterator`).
- `tests/mod.rs` — wires up `generate_backend_tests!` / `generate_backend_concurrency_tests!` / `generate_backend_dropable_tests!`. **70 tests pass on first run.**

### Wiring

- `Cargo.toml` workspace: adds `lark-kv = { path = "../lark" }`.
- `crates/storage/Cargo.toml`: new `lark = [\"dep:lark-kv\"]` feature. Top-level re-exports of `LarkStore` / `LarkStoreOptions` on native + feature.
- `DatastoreType` (in `cli/config/types.rs`) gains a `Lark` variant marked `#[default]`. Display / `FromStr` / `DatastoreConfig::default` updated accordingly.
- `cli start` and `cli server-dump` handle the new variant (info log + `LarkStoreOptions::from_env()` open path, with `#[cfg(not(feature = \"lark\"))]` fallback error).
- `FfiStore` gains a `Lark(storage::LarkStore)` variant; FFI store dispatch handles `\"lark\"` and now defaults the empty backend name from `\"redb\"` to `\"lark\"`.
- Default features on consumers:
  - `cli`: `[\"lark\", \"rocksdb\", \"redb\"]`
  - `defra-node`: `[\"lark\", \"redb\"]`
  - `embedded`: `[\"lark\", \"redb\"]`
  - `ffi`: `[\"lark\", \"redb\", \"rocksdb\"]`

### Test / config updates

- `cli/tests/config_tests.rs` and `config_sections_tests.rs` — default `DatastoreType` assertion flipped to `Lark`.
- `cli/tests/config_types_tests.rs` — add `\"lark\"` parse case and include `Lark` in the display roundtrip loop.

## Verification

Run locally against this branch:

- [x] `cargo check --workspace --all-targets`
- [x] `cargo test -p storage --features lark` — backend test suite, **70 lark tests pass** (basic ops, iterators in both directions, seek, snapshot isolation, concurrency, dropable, callbacks).
- [x] `cargo test -p cli` — all cli tests green, including the default-backend assertion.
- [x] `cargo test --workspace --exclude integration-test --exclude ffi-test` clean under `--test-threads=4`.

## Out of scope (follow-ups)

- `defra-harness` still defaults `--store` to `memory` for integration tests, so running `cargo test -p integration-test` end-to-end against Lark requires either `.with_store(\"lark\")` per test or changing the harness default. Separate PR.
- Pre-existing `clippy::too_many_arguments` error in `crates/p2p/src/iroh/endpoint_commands.rs` from the Rust 1.94 lint-level bump — unrelated to this change; already present on `main`.
- No changes to persistence format / wire format; existing RocksDB databases are unaffected.

## Test plan

- [ ] CI: Lint, WASM build check, Build & Test, all integration suites pass on this branch.
- [ ] Start a node with default settings locally, verify the startup log says `\"Using Lark datastore at …\"`, create a collection, add a doc, restart, read it back.
- [ ] Repeat with `--store rocksdb` to confirm the RocksDB path still works end-to-end.

---

## Update: CI Lint Fix and Main Merge

Fixes the CI lint failure on the Lark integration branch by removing deprecated `multihash::MultihashGeneric` usage and updating the branch against latest `main`.

### Context

CI started failing on Rust stable 1.96 because `MultihashGeneric` is now deprecated, and the lint job runs clippy with `-D warnings`.

### Changes

- Replace `multihash::MultihashGeneric` with `multihash::Multihash`.
- Use the correct `Multihash<64>` form after merging latest `main`, which moved the workspace to the newer IPLD/multihash stack.
- Update affected CID construction helpers, tests, and benchmark helpers.
- Merge `main` into `lark` to resolve PR merge conflicts.

### Validation

Locally verified:

- `cargo clippy --all -- -D warnings`
- `cargo clippy -p telemetry --features otlp --all-targets -- -D warnings`
- `cargo clippy -p telemetry --features metrics --all-targets -- -D warnings`
- `cargo clippy -p cli --features otel --all-targets -- -D warnings`
- `cargo clippy -p defra-node --features otel --all-targets -- -D warnings`

CI lint now passes on #894.

---

## Update: Lark Prefix-Scan Follow-Up

Lark now has a follow-up PR that improves the storage primitive needed for faster prefix scans:

- sourcenetwork/lark#161

This improves Lark's `seek_prefix` behavior for fixed-length prefix extractors, allowing longer query prefixes to use the SSTable prefix bloom and skip irrelevant files before loading data blocks.

To realize that prefix-scan win in DefraDB, this integration should follow up by:

- Using `OwnedSnapshotIter::seek_prefix` for forward prefix scans in the Lark adapter.
- Configuring a matching fixed prefix extractor for the Defra key layout.


iverc:
## Lark vs RocksDB — defra-storage-bench at 1,000,000 keys

Workloads run through defradb.rs's `storage` corekv `Store` / `Txn` trait.
16 B keys, 100 B values, per-op writes (batch=1), **8 MB block cache on both
backends** — matches the RocksDB default used by `lark-compare`.
Fresh DB per workload. `readseq` is a true iterator scan.
Host: `Darwin 25.4.0 arm64` (M5), rustc 1.94.1, release.

### Build + stripped binary

| backend | clean build (wall) | CPU-seconds | stripped binary |
|---------|--------------------|-------------|-----------------|
| lark    | 17.8 s             | 93.3 s      | 1009 KB         |
| rocksdb | 45.5 s             | 493.8 s     | 6.8 MB          |

Lark: **~6.7× smaller binary**, **~5.3× less CPU** to build. Removes the last
C/C++ dependency from native builds.

The binary numbers here are ~400 KB larger than the standalone lark-compare
figures (620 KB lark / 5.4 MB rocksdb) because this bench links the full
defradb.rs `storage` crate — schema, document, defra-core, bm25, 33 key
types, Multistore, etc. The ratio between backends is preserved.

### Throughput (ops/sec)

| Workload     | lark       | rocksdb    | ratio    |
|--------------|------------|------------|----------|
| fillseq      |    230,754 |    177,217 | **1.30×** |
| fillrandom   |    316,159 |    308,260 | **1.03×** |
| readseq†     | 20,046,475 | 22,391,548 | 0.90×    |
| readrandom   |    256,671 |    411,341 | 0.62×    |
| readmissing  |  4,985,276 |  2,048,887 | **2.43×** |
| overwrite    |    184,382 |    166,895 | **1.10×** |
| deleterandom |    188,146 |    176,875 | **1.06×** |

†true iterator scan (`Txn::iterator()`), not sequential point lookups.

- Lark is **faster on writes** (fillseq 1.30×, fillrandom ~parity, overwrite 1.10×,
  deleterandom 1.06×) and on negative lookups (readmissing 2.43× — bloom filter
  does its job).
- Lark is within 10% on the iterator-scan readseq path.
- RocksDB retains an edge on readrandom (0.62×): its C++ point-lookup is
  more economical per op. A profile of the wrapper showed < 1% overhead
  from the defradb.rs `Txn` layer — the gap is inside the engines'
  point-lookup paths themselves, not our abstraction.

### On-disk after 1M fills

- lark: 121 MiB (LSM + LZ4)
- rocksdb: 110 MiB (LSM + LZ4)

### Reproduce

```
bash tools/defra-storage-bench/run.sh 1000000 100 1 8
# args: NUM VALUE_SIZE BATCH CACHE_MB
```

Harness source at `tools/defra-storage-bench/`.

jackzampolin:
I spent a deeper pass validating Lark locally. I did not run the 12GB Rocks dataset; we decided that setup cost is not worth it for this review. Instead I focused on synthetic integration coverage, crash/reopen behavior, durability, encryption, and FFI.

Summary: I am comfortable with this moving forward as the Rust-native backend and as the practical Redb replacement. I would not position it as replacing RocksDB for apps already using Rocks, but Lark looks like a meaningful improvement over Redb and gives us a backend we control.

Validation run:
- Lark storage tests: green.
- DefraDB Lark crash/reopen integration test: green with acknowledged writes.
- Low-level Lark crash oracle: green at 10k ops with immediate durability.
- Lark persistent encrypted reopen/wrong-key coverage: green.
- FFI suite: 2868 passed, 3 failed, 50 skipped across 2921 tests.
- FFI ACP: 340 passed, 0 failed, 22 skipped.
- FFI net: green after rerunning one timing-sensitive replicator timeout.
- The 3 FFI failures are embedding tests requiring Ollama at `localhost:11434`, which was not running locally.

Bench/readiness notes:
- Lark immediate durability is the right default for this backend; crash tests support that choice.
- Compared with RocksDB, Lark is slower on write/seed workload in the synthetic benchmark, but query latency was roughly comparable in the runs we did.
- Disk usage was higher than RocksDB in the synthetic benchmark.
- This is still a good trade-off for a Rust-native backend replacing Redb, not a reason to force existing RocksDB users over.

One caveat: the full non-FFI Go/Rust integration harness is not green locally, but the failures appear to be Go CLI/API compatibility with the Rust harness, not Lark. The failures are around Go-client ACP/schema/document output shape, including missing `_docID`, schema output parsing, and a Go CLI flag mismatch. Rust-side integration variants passed through the same areas.

Given the above, I would be comfortable merging this with Lark kept as the Rust-native backend path and with immediate durability as the default.

PR #895, opened 2026-04-21
feat(p2p): pubsub_rpc primitive + Go-shaped wire types (#828 part 1)

### LINKED ISSUE #828: P2P: DocSync, Branchable, and sync-branchable topic wire formats incompatible with Go
## Summary

DocSync and BranchableCollection sync protocols in Rust use a completely different wire shape and transport than Go. Go delivers both over pubsub with lowercase JSON-tagged structs that have no MetaData; Rust uses the signed two-stream transport with PascalCase renames and embedded MetaData. Additionally, the `sync-branchable` pubsub topic that Go uses is not registered in Rust. As a result, DocSync and Branchable sync between Rust and Go peers do not work at all.

### DocSync

Go (`defradb/internal/db/p2p/sync_doc.go:40-56`):
```go
type docSyncRequest struct {
    DocIDs []string `json:\"docIDs\"`          // CBOR key \"docIDs\" (lowercase)
}
type docSyncReply struct {
    Results []docSyncItem `json:\"results\"`   // NO MetaData
    Sender  string        `json:\"sender\"`
}
type docSyncItem struct {
    DocID string   `json:\"docID\"`
    Heads [][]byte `json:\"heads\"`
}
```

`fxamacker/cbor` honors JSON tags by default, so the CBOR wire keys are `docIDs`, `results`, `sender`, `docID`, `heads`. Go uses these over the `\"doc-sync\"` pubsub topic (see `sync_doc.go:37, 117`), unsigned.

Rust (`crates/p2p/src/message/docsync.rs`):
- `DocSyncRequest` uses `#[serde(flatten)]` on `MetaData` and `#[serde(rename = \"DocIDs\")]` (PascalCase).
- `DocSyncReply` has the full signed MetaData fields + `#[serde(rename = \"Results\"/\"Heads\"/\"DocID\")]` (PascalCase).
- Routed via the signed two-stream protocol, not pubsub.

### BranchableCollection

Go (`defradb/internal/db/p2p/sync_branchable_col.go:35-50`):
```go
const syncBranchableCollectionTopic = \"sync-branchable\"

type syncBranchableCollectionRequest struct {
    CollectionID string `json:\"collectionID\"`
}
type syncBranchableCollectionReply struct {
    CollectionID string   `json:\"collectionID\"`
    Heads        [][]byte `json:\"heads\"`
    Sender       string   `json:\"sender\"`
}
```

Go uses these over the `sync-branchable` pubsub topic, also unsigned.

Rust (`crates/p2p/src/message/branchable.rs` + `crates/p2p/src/topics.rs`):
- `BranchableSyncRequest` / `BranchableSyncReply` with MetaData + PascalCase fields.
- **No `sync-branchable` topic constant** — `topics.rs` only declares `DOC_SYNC_TOPIC = \"doc-sync\"` and `ENCRYPTION_TOPIC = \"encryption\"`.
- Routed via two-stream.

## Impact

Rust nodes cannot participate in DocSync or BranchableCollection sync with Go peers at all — neither direction decodes.

## Expected Behavior

- DocSync and Branchable requests/replies use the Go wire shape: lowercase camelCase keys, no embedded MetaData, rely on pubsub layer for authentication (`MessageAuthenticity::Signed` is already configured).
- `sync-branchable` pubsub topic is registered alongside `doc-sync` and `encryption`.
- Transport moves from signed two-stream to pubsub for these protocols.

## Acceptance Criteria

- [ ] `DocSyncRequest` / `DocSyncReply` stripped of `MetaData`, field renames changed to lowercase camelCase (`docIDs`, `results`, `sender`, `docID`, `heads`)
- [ ] `BranchableSyncRequest` / `BranchableSyncReply` stripped of `MetaData`, field renames changed to lowercase camelCase (`collectionID`, `heads`, `sender`)
- [ ] `topics.rs`: add `SYNC_BRANCHABLE_TOPIC = \"sync-branchable\"` and a `DefraTopic::BranchableSync` variant
- [ ] DocSync and Branchable wire paths moved from two-stream to pubsub (removing handler registrations from `p2p_host/mod.rs`)
- [ ] Integration test: Rust peer round-trips DocSync / Branchable with Go-generated CBOR fixture

## Files

- `crates/p2p/src/message/docsync.rs` — wire shape rewrite
- `crates/p2p/src/message/branchable.rs` — wire shape rewrite
- `crates/p2p/src/topics.rs` — add sync-branchable topic
- `crates/p2p/src/two_stream/handler/doc_sync.rs` — remove two-stream handling
- `crates/p2p/src/two_stream/handler/branchable_se.rs` — remove branchable two-stream handling
- `crates/p2p/src/host/p2p_host/mod.rs` — stop registering these as two-stream protocols
- `crates/p2p/tests/` — add fixture tests

## References

- Go: `defradb/internal/db/p2p/sync_doc.go`, `sync_branchable_col.go`
- Related: #827 (CBOR flatten issue affects the current request types in the same files)

fredcarle:
Part of #828 — lands the transport-agnostic RPC primitive, Go-shaped wire types, and the `sync-branchable` topic constant. The gossipsub-dispatcher integration and DocSync/BranchableSync handler migration are deliberately deferred so this PR stays reviewable as pure additions without touching live P2P code paths. Follow-up PR to close #828 will rewire the coordinator and retire the two-stream DocSync/Branchable handlers.

## Summary

Three commits on this branch build a standalone `pubsub_rpc` module that ports [`sourcenetwork/go-libp2p-pubsub-rpc`](https://github.com/sourcenetwork/go-libp2p-pubsub-rpc) to Rust, plus the Go-shaped wire types that will flow through it:

**`cf46969d` — primitive foundation.** `envelope` (definite-length-map `InternalResponse` dag-cbor struct, field-order-enforced), `id` (`CIDv1(raw, sha256(data))` matching Go's `cid.NewCidV1(cid.Raw, util.Hash(data))`), `topic` (`<base>/<peer>/_response` formatter + peer-scoped strip), `correlator` (thread-safe `HashMap<Cid, channel>` with single-/multi-response handling, `ignore_response`, late-drop, and cancel).

**`36807073` — wire types + topic constant.** `pubsub_rpc::messages` defines `DocSyncRequest`/`DocSyncReply`/`DocSyncItem` and `BranchableSyncRequest`/`BranchableSyncReply` with Go's lowercase JSON-tag serde renames (`docIDs`, `docID`, `heads`, `results`, `sender`, `collectionID`), no `MetaData` (auth comes from gossipsub's `MessageAuthenticity::Signed`), and a `serde_bytes_vec` adapter so head CIDs serialize as CBOR byte strings not arrays-of-integers. Also adds `SYNC_BRANCHABLE_TOPIC = "sync-branchable"` and `DefraTopic::SyncBranchable` to match Go's `defradb/internal/db/p2p/sync_branchable_col.go:35`.

**`8ae61070` — `TopicHandler`.** Stitches the primitives into a transport-agnostic RPC surface: owns `(base_topic, self_peer)`, a `Correlator` for outbound, and an `Arc<dyn MessageHandler>` for inbound. `deliver_gossip_message` classifies `(topic, from, bytes)` into `DeliveryOutcome::{Forwarded, Respond, Ignored}`; echoed-self is filtered; handler errors propagate as the `Err` field in the response envelope. Swarm-free by design so it stays unit-testable.

## Why split the PR

The full #828 change moves DocSync and BranchableSync off signed two-stream and onto pubsub-RPC, which cascades through ~20 files in `crates/p2p` (host dispatcher, commands, iroh transport, coordinator handlers, two-stream deletions, integration tests). That's a ~1500-line review where the primitive is tangled with the migration. Splitting lets this primitive get a focused review while the integration is developed against it independently.

## Test plan

- [x] `cargo test -p p2p --lib pubsub_rpc` — 29/29 pass: correlator (5), envelope (3), id (3), messages (6), topic (4), topic_handler (8).
- [x] `cargo test -p p2p --lib topics` — 8/8 pass (new `test_sync_branchable_topic` + updated `test_from_str`).
- [x] `cargo clippy -p p2p --tests --no-deps -- -D warnings` — clean.
- [x] `cargo build --workspace` — clean; no existing call sites depend on `pubsub_rpc` yet.
- [x] Go parity anchors:
  - `id::tests::matches_go_known_vector` pins `CIDv1(raw, sha256("hello")).to_string()` against a spec-derived fixture.
  - `envelope::tests::{encodes_as_definite_length_map, field_order_matches_go}` pin the dag-cbor shape against Go's `fxamacker/cbor` output.
  - `messages::tests::*` pin every Go JSON tag key on the wire for both protocols.

## What's still open on #828

The issue stays open until the follow-up PR lands:

- Host-side: generalize the pubsub publish command to raw bytes; extend `host/p2p_host/protocols.rs::handle_gossipsub_event` to route through a `HashMap<base_topic, Arc<TopicHandler>>` before falling through to the PushLog decoder; new `RegisterPubsubRpcHandler` + `PublishRaw` host commands.
- Coordinator-side: replace `sync/coordinator/event_handler/{doc_sync,branchable_sync}.rs` initiation and receipt paths with `TopicHandler::prepare_publish` / `MessageHandler::handle`; register handlers during host init.
- Transport cleanup: delete `two_stream/handler/{doc_sync,branchable_se}.rs`, remove the corresponding methods on `transport.rs` / `host/libp2p_transport.rs` / `iroh/transport.rs`, and drop the `HostCommand::Send{Doc,Branchable}Sync*` + `TwoStreamEvent::{Doc,Branchable}Sync*` variants.
- Tests: two-peer integration test (Rust–Rust) that drives a full `doc-sync` round-trip through real gossipsub, then a Go-fixture byte-parity test that decodes a recorded Go request.


fredcarle:
**Verdict: request changes — one hard parity bug, plus API traps, and unmaintained-dep choice.**

I read this as if it came in cold from a code generator. Several load-bearing claims don't survive cross-checking against Go.

---

### 1. Wire format does not match Go. The test that claims it does is asserting the wrong order.

Go's `go-libp2p-pubsub-rpc/rpc.go:389` encodes the envelope via:

```go
ipld.Marshal(dagcbor.Encode, &res, resType)
```

`github.com/ipld/go-ipld-prime/codec/dagcbor` emits **strict IPLD dag-cbor**, which is canonical: map keys are sorted length-first then bytewise-lex (RFC 8949 §4.2.1). For the envelope keys `{ID, From, Data, Err}` that gives the sorted sequence:

```
ID (2) → Err (3) → Data (4) → From (4, "D" < "F")
```

The PR's test at `crates/p2p/src/pubsub_rpc/envelope.rs`:

```rust
assert!(id_at < from_at);
assert!(from_at < data_at);
assert!(data_at < err_at);
```

asserts `ID, From, Data, Err` — struct declaration order, not canonical order. `serde_cbor::to_vec` does not sort. So:

- Rust emits `ID, From, Data, Err`.
- Go emits `ID, Err, Data, From`.

Consequences:
- **Byte streams from Rust ≠ byte streams from Go** for the same logical response. Any CID computed over the raw envelope bytes (e.g., future dedup, replay protection, gossipsub `message_id_fn`) will diverge.
- `dagcbor.Decode` on Go's receiving side is strict about canonical form in recent go-ipld-prime versions. Depending on the pinned version, Go may **reject Rust's bytes outright**. At minimum the project cannot claim byte parity.
- The test is worse than missing — it *pins the wrong behaviour*.

Fix: switch to a canonical-mode CBOR encoder and add an actual Go fixture. The workspace already depends on `ciborium = "0.2"` (see `Cargo.toml:83`); `ciborium-ll` gives you the control to sort keys before emit, or you can hand-roll a writer that serializes `(ID, Err, Data, From)` directly — it's four fields. Then replace the field-order test with:
1. A byte fixture recorded from `go-libp2p-pubsub-rpc` and checked in under `crates/p2p/tests/fixtures/`.
2. Two asserts: Rust encode → bytes equals fixture; Rust decode(fixture) equals the expected struct.

Until that exists, the "Go wire compatibility" headline of this PR is unverified.

### 2. `serde_cbor` is archived. Workspace has `ciborium`.

`Cargo.toml:81-83`:

```
serde_cbor = "0.11"
ciborium = "0.2"
```

`serde_cbor` has been unmaintained since 2021 and is documented to lack canonical-CBOR support. Adding a *new* module that depends on it — when the non-archived alternative is already a workspace dep — is a maintenance-cost decision being made without justification. Combined with (1) this isn't cosmetic: solving the sort-order problem needs an encoder you have control over, which `serde_cbor` doesn't give you.

### 3. `ignore_response = true` returns a receiver that will never yield.

`correlator.rs:230-247`:

```rust
pub fn publish(&self, data: Vec<u8>, opts: PublishOptions) -> PreparedPublish {
    let id = derive_request_id(&data);
    let (tx, rx) = mpsc::unbounded_channel();
    if !opts.ignore_response {
        self.ongoing.lock().insert(id, Entry { sender: tx, ... });
    }
    // For fire-and-forget we still return the receiver, but nothing
    // will ever send on it — caller drops it immediately.
    PreparedPublish { id, data, responses: rx }
}
```

The comment admits the trap: the returned `responses` receiver for `ignore_response=true` will never produce a value, but the type signature gives no hint. A caller who does `rx.recv().await` hangs indefinitely. Either return `Option<UnboundedReceiver<_>>`, or split into two functions (`publish_fire_and_forget` / `publish_with_responses`), or encode the mode in the return type (two distinct structs). The test `ignore_response_allocates_no_entry` verifies the in-flight count but not the trap.

Also: `derive_request_id(&data)` runs unconditionally, doing a SHA-256 over the full payload even in fire-and-forget where the ID is unused by the caller. Minor but wasteful.

### 4. No automatic correlator cleanup on caller cancellation.

Go's `rpc.go:204-278` is `ctx`-aware: when the caller's context is cancelled, the ongoing entry is removed. The Rust port leaks: if the caller's future is dropped without explicit `cancel(&id)`, the entry sits in `self.ongoing` forever, and late deliveries still succeed for a dead request (the `send()` on the dropped tx will fail and auto-remove via the `is_err()` path, but only on the first stale delivery — if no peer ever replies, the entry is immortal).

For a peer protocol with long tails and adversarial actors, this is a slow memory leak. Fix: track the caller's receiver-drop (oneshot sentinel, or a `Drop` wrapper around `PreparedPublish`) and remove the entry.

### 5. Response channel is unbounded.

`mpsc::unbounded_channel()` in `correlator.rs:232`. Multi-response mode means every peer on the mesh that wants to can flood the caller's channel. Fire-and-forget and single-response are bounded in practice (single-response closes after 1), but multi-response is a memory-amplification vector. Should at minimum have a cap via `mpsc::channel(N)` with configurable N.

### 6. `MAX_DOC_IDS = 1000` is declared but nothing enforces it.

`messages.rs:27`:

```rust
pub const MAX_DOC_IDS: usize = 1000;
```

Commented as "Prevents memory exhaustion from malicious peers" — but serde_cbor will happily deserialize a 10M-element `Vec<String>` into `DocSyncRequest`. No custom deserializer, no bound check. Either enforce (custom `Deserialize` that caps `doc_ids.len()` during deserialization), or delete the constant. Keeping it around advertises a protection that doesn't exist.

### 7. Tautological test.

`topic_handler.rs::correlator_accessor_matches_in_flight`:

```rust
assert_eq!(t.in_flight(), 0);
assert_eq!(t.correlator().in_flight(), t.in_flight());
```

This asserts that a cfg(test) getter returns the same object as the public wrapper that calls it. Zero information value. Delete — the getter's job is to expose the handle for tests that actually exercise correlator state, not to be tested for equality with itself.

### 8. Field-order test uses substring search in a way that hides failures.

`envelope.rs::field_order_matches_go` runs `String::from_utf8_lossy(&bytes).find("ID")`. This relies on the test fixture's `id` / `from` / `data` values not containing the substrings "ID", "From", "Data", "Err". The current sample passes by luck — `"bafkreihdwd...yku"` has no collision, `"hello"` has no collision. Change the fixture to include a payload string like `"Datacenter"` and the test silently passes while being wrong. Use CBOR parsing, not text search — `ciborium_ll::Decoder` exposes tokens directly, or at minimum scan for the two-byte length-prefixed key (`0x62 I D` for "ID", `0x64 D a t a` for "Data", etc.) to avoid substring bleed. *This is moot once you fix (1) — the test will be rewritten anyway.*

### 9. Other things worth fixing before merge

- `strip_response_topic` formats a fresh `String` per delivered message (`topic.rs:34-36`). Cache the suffix `"/<self_peer>/_response"` once in `TopicHandler::new`. Gossipsub can sustain thousands of msg/s; this is in the classifier hot path.
- `TopicHandler::deliver_gossip_message` silently swallows `envelope.to_cbor()` and `InternalResponse::from_cbor()` errors into `DeliveryOutcome::Ignored` (`topic_handler.rs:1157-1162`). An envelope that fails to encode is a bug in the handler's reply bytes; an inbound envelope that fails to decode is a protocol violation. Both deserve at least a `tracing::debug!` with `from` and the error — otherwise you'll never diagnose a wire-format drift.
- `messages.rs` is inside `pubsub_rpc` but is not primitive — `DocSyncRequest` / `BranchableSyncRequest` are specific to two protocols. Consider `crate::sync::messages::pubsub_v2` or similar. Primitive modules shouldn't know about doc-sync and sync-branchable by name.
- The `sender` field on replies (DocSyncReply / BranchableSyncReply) is caller-populated; nothing in the primitive reminds the reply-building code to set it. First place this will break in #828 part 2 is an empty `sender` flowing to a Go peer that expected a valid peer ID. Worth a builder method (`reply.with_sender(self_peer)`) or a required-field wrapper.

---

**Recommendation:** hold this PR until (1) is fixed and a real Go fixture byte-parity test exists. Everything else could land in follow-ups, but the wire-format bug invalidates the entire reason the module exists.

_🤖 AI-assisted review_

fredcarle:
**Second-pass verdict: approve — all blockers from the prior review addressed; two non-blocking follow-ups.**

Walked through the fixup commit `269db138` against the 8 items I raised. Summary:

| # | Prior finding | Status |
|---|---|---|
| 1 | Wire format: serde_cbor in declaration order ≠ Go's canonical dag-cbor (ID, Err, Data, From) | **Fixed correctly.** `envelope.rs:38-55` declares fields in canonical order; switched to `ciborium`. `testdata/gen_pubsub_rpc_fixture/main.go` drives the exact `ipld.Marshal(dagcbor.Encode, ...)` pipeline as `rpc.go:389`; two Go-produced hex fixtures (`GO_FIXTURE_OK_HEX`, `GO_FIXTURE_ERR_HEX`) are baked in, and tests now assert byte-identical encode AND decode round-trip — no more substring heuristic. |
| 2 | serde_cbor archived | **Fixed** for the new module. Pre-existing serde_cbor usage elsewhere in the crate was correctly left out of scope. |
| 3 | `ignore_response=true` returns a receiver that never yields | **Fixed** via API split. `Correlator::publish` returns `PreparedPublish` (channel-bearing), `publish_fire_and_forget(&[u8]) -> Cid` has no channel at all. The footgun is gone at the type level. |
| 4 | Correlator entry leak on caller future-drop | **Fixed.** `PreparedPublish: Drop` (`correlator.rs:99-104`) unconditionally removes the entry. Test `dropping_handle_cancels_in_flight` pins it. |
| 5 | Unbounded response channel | **Fixed.** Bounded `mpsc::channel(buffer)`, configurable `multi_response_buffer` (default 128), single-response always 1-slot. Delivery uses `try_send`; `Full` reports drop without removing the entry, matching Go's `rpc.go:275` non-blocking select. `multi_response_full_buffer_reports_drop_without_panic` covers the overflow path. |
| 6 | `MAX_DOC_IDS` declarative only | **Fixed** for requests. `message/pubsub.rs:57-90` adds a hand-written `Deserialize` with a `BoundedStringVec` helper that rejects oversized payloads before allocating. Tests cover boundary (`accepts_boundary_doc_ids`) and overflow (`rejects_oversized_doc_ids`). |
| 7 | Tautological `correlator_accessor_matches_in_flight` test | **Fixed** — removed. |
| 8 | Hot-path allocation in `strip_response_topic` + silent CBOR error swallowing | **Fixed.** `TopicHandler` pre-computes `self_response_suffix` and `self_response_topic` once at construction (`topic_handler.rs:82-104`); delivery uses `strip_response_topic_with_suffix`. Envelope encode/decode errors now emit `tracing::debug!` with `base_topic`, `from`, `error`, and `payload_size` context. |
| Bonus | `messages.rs` inside the "primitive" module | **Fixed** — relocated to `crate::message::pubsub`. The primitive is now transport-only. |

The Go fixture generator is the right shape: it reuses the Go struct declaration and `ipld.Marshal(dagcbor.Encode, node.Representation(), resType)` path verbatim, which is exactly what `go-libp2p-pubsub-rpc/rpc.go:389` does, so regenerating fixtures after a shape change stays trivially correct.

---

**Non-blocking follow-ups** (file issues if you agree, otherwise drop):

1. **Inner message byte parity is still substring-verified, not Go-fixture-verified.**
   - I checked Go: `defradb/internal/db/p2p/sync_doc.go:19` imports `github.com/fxamacker/cbor/v2` and uses plain `cbor.Marshal` at `:112, :303`. Same for `sync_branchable_col.go:107, :271`. That's fxamacker/cbor v2 **default** — declaration-order keys, definite-length maps, honors `json:` tags.
   - Rust's ciborium with serde does the same (declaration order, definite-length, honors `#[serde(rename)]`), and the Rust fields are declared in the matching Go struct order. So parity *should* hold.
   - But "should" isn't "verified." The `gen_pubsub_rpc_fixture/main.go` tool already exists; adding 2–3 fixtures (`docSyncRequest{DocIDs: ["docA","docB"]}`, `docSyncItem{DocID:"x", Heads:[...]}`, `syncBranchableCollectionReply{...}`) would take ~20 lines of Go and let us delete the substring-scan tests entirely. The substring tests *passed* because the current fixture strings don't have accidental collisions, not because the encoder is right — a future payload like `"Datacenter"` as an ID or sender value would silently defeat them.
   - Given how load-bearing this inner format is for real DocSync traffic, worth doing before #828 part 2 merges.

2. **Reply-side deserialization is unbounded.**
   - `DocSyncReply::results: Vec<DocSyncItem>`, `DocSyncItem::heads: Vec<Vec<u8>>`, `BranchableSyncReply::heads: Vec<Vec<u8>>` — none have the `BoundedStringVec`-style cap that the fix adds to `DocSyncRequest::doc_ids`.
   - A malicious *responder* (a peer we've asked for sync) can send a 10M-element `results` array or a 10M-element `heads` array and OOM the caller. The MAX_DOC_IDS fix is request-side only; the protection is asymmetric, and the attack surface on the reply side is arguably larger because a caller is *expecting* to drain the payload.
   - The same `Bounded*Vec` pattern applies. `MAX_DOC_IDS = 1000` is a reasonable cap for results too, though heads-per-doc and heads-per-collection probably want their own constants (realistic values: `MAX_HEADS_PER_DOC = 64`, `MAX_BRANCHES = 1024`, whatever matches Go's implicit limits).

3. **Minor: `BoundedStringVec` error message uses the sequence `size_hint`, which is 0 for indefinite-length CBOR arrays.**
   - `message/pubsub.rs:106-111`: first check is `if hint > MAX_DOC_IDS` — won't fire for indefinite-length inputs (CBOR `0x9F...0xFF`), which fall through to the per-element overflow branch. The error message `"docIDs exceeds MAX_DOC_IDS (>1000)"` is accurate but not informative (no actual count, since we bail on the 1001st element). Acceptable as-is; only flagging because a malicious peer can force the indefinite-length path to avoid the fast-fail branch.

---

Ship it. Follow-up (1) blocks #828 part 2, not this PR.

_🤖 AI-assisted review_

PR #896, opened 2026-04-22
feat(p2p): Go-compatible DocSync/BranchableSync over pubsub_rpc (#828 part 2)

### LINKED ISSUE #828: P2P: DocSync, Branchable, and sync-branchable topic wire formats incompatible with Go

fredcarle:
Resolves #828 (part 2 of 2).

⚠️ Depends on #895 (part 1). Do not merge until #895 lands — this
branch sits on top of #895's commits and assumes they're on main.

## Summary

Wires the \`pubsub_rpc\` primitive from #895 into the full DocSync /
BranchableSync path, so Rust nodes interoperate with Go DefraDB over
gossipsub byte-for-byte. The previous two-stream implementation is
kept as an iroh fallback.

## What lands

**Coordinator.** \`SyncCoordinator\` now owns a \`PubsubServices\` bundle
with two \`TopicHandler\`s (\`doc-sync\`, \`sync-branchable\`). Inbound
\`GossipRawMessage\` events route to the right handler, which decodes
the request, checks access, serves heads via the existing
\`DocumentHeadProvider\`, and publishes the \`InternalResponse\` envelope
back on the caller's \`<base>/<caller>/_response\` sub-topic.
\`start_pubsub_services()\` subscribes + registers all four topics at
node startup.

**Adapter.** \`P2PAdapter::sync_documents\` and
\`sync_branchable_collection\` call the coordinator's pubsub_rpc path
on libp2p, fall back to two-stream when no coordinator is wired
(light tests). The MergeComplete event-drain loop is unchanged —
\`pubsub_sync_documents\` feeds every reply through
\`handle_doc_sync_reply\` internally, so the DAG-fetch + merge pipeline
behaves exactly like the two-stream path.

**Wire-format safety.** \`DocSyncReply\`, \`DocSyncItem\`, and
\`BranchableSyncReply\` now enforce \`MAX_DOC_IDS\` / \`MAX_HEADS_PER_DOC\`
/ \`MAX_BRANCH_HEADS\` on deserialize (asymmetric cap from #895 that
only covered requests). Empty slice fields serialize as CBOR null to
match Go's \`nil\` default and keep byte parity.

**Parity fixtures.** \`testdata/gen_pubsub_rpc_fixture/main.go\` and
\`testdata/gen_message_fixtures/main.go\` emit Go-produced hex fixtures
that Rust unit tests assert byte-identity against. Covers all 8
wire-format shapes (DocSync{Request,Item,Reply}, BranchableSync{Request,Reply}
including empty/null variants) and one end-to-end exchange
(\`request_bytes → request_id → inner_reply → envelope_bytes\`).

**CI.** \`.github/workflows/ci.yml\` go-compat step now runs \`--test p2p\`
alongside the existing suites. The \`for_each_p2p_topology!\` macro
already generates \`go_rust_*\` variants for every p2p integration test;
this just wires them in.

Locally verified:
\`\`\`
cargo test -p integration-test --test p2p -- --test-threads=1 go_rust_
→ 9 passed, 0 failed (4 unrelated tests remain ignored)
\`\`\`

**Live probe.** \`crates/p2p/tests/live_go_pubsub_rpc.rs\` is a
standalone \`#[ignore]\` probe that dials an external Go defradb and
runs one DocSync + one BranchableSync call. Passed on my machine
against a real defradb built from current main — full exchange over
libp2p gossipsub with byte-identical wire format.

## Test plan

- [x] \`cargo test -p p2p\` — 207 unit tests pass
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --all\` applied
- [x] \`cargo test -p integration-test --test p2p -- go_rust_\` — 9 passed
- [x] \`cargo test -p integration-test --test p2p -- rust_rust_\` — no regressions
- [x] Live probe against real Go defradb: DocSync + BranchableSync
      succeed, replies decoded, heads match
- [x] Reviewer follow-ups from #895 (Go-fixture byte parity, bounded
      reply deserialization) applied


fredcarle:
## Code Review — PR #896: pubsub_rpc integration into DocSync/BranchableSync

Reviewed like codex wrote it. Verified findings directly against the checked-out branch (worktree at `/tmp/defradb.rs-pr896`, HEAD `a654cf73`) rather than trusting agent summaries. Cross-checked wire format against Go (`defradb/internal/db/p2p/sync_doc.go`, `go-libp2p-pubsub-rpc/rpc.go`).

### Overview

Part 2 wires the #895 primitive into the full sync path on libp2p, with two-stream kept as an iroh fallback. The structural choices are right: `PubsubServices` cleanly bundles the two `TopicHandler`s, access checks run server-side before publishing the envelope, received replies are routed through the existing `handle_doc_sync_reply` / `handle_branchable_sync_reply` paths so the DAG-fetch + merge pipeline is unchanged. Go fixtures + round-trip byte-parity tests (including a real end-to-end `request_bytes → request_id → reply_bytes → envelope_bytes` chain at `pubsub_services.rs:344-424`) and the live-probe test are solid.

### Blocker

**1. Access-check parity gap between pubsub_rpc server and the two-stream server.**
The two-stream server’s `SyncCoordinator::check_peer_is_replicator` at `crates/p2p/src/sync/coordinator/access.rs:125-153` is a four-step check:
1. Open mode → allow
2. `is_any_replicator` → allow
3. `peer_state.is_connected` → allow
4. **`transport_reports_connected_peer` → allow** (reads live transport + backfills `PeerState`)

`HandlerContext::peer_may_doc_sync` at `crates/p2p/src/sync/coordinator/pubsub_services.rs:114-120` only implements steps 1-3 — it omits step 4. Same for `peer_may_branchable_sync` at `:126-141`, which also omits the transport `get_replicator` fallback that the two-stream collection path `check_access_str` does at `access.rs:70-82`.

Impact: during the `PeerState` cache-miss window right after a peer connects (before the transport event has populated the cache), the two-stream path accepts the peer while the pubsub path returns `"peer … not authorised for doc-sync"`. Same divergence if a peer is registered as a replicator at the transport layer but not yet mirrored into the in-memory `ReplicatorRegistry` (the new `access_tests.rs` tests — `gossip_access_falls_back_to_transport_*` — prove the two-stream gossip path handles this; the pubsub_rpc path does not). Per CLAUDE.md’s parity rule this is a behavior divergence, not just a theoretical edge case.

Fix: factor the four-step check into a shared helper (e.g. `SyncCoordinator::authorize_peer_for_sync`) and have both `HandlerContext` and the two-stream path call it. Needs an `Arc<T>` on `HandlerContext` or a callback so the handler can reach the transport. Add matching tests in `pubsub_services.rs`.

### Non-blockers

**2. Backpressure drops are silent on the Rust side; Go logs.**
`Correlator::deliver` at `crates/p2p/src/pubsub_rpc/correlator.rs:195-198` returns `false` on `TrySendError::Full` but the topic handler at `topic_handler.rs:220-224` collapses that into `DeliveryOutcome::Ignored` without a log line. Rust code comment at `correlator.rs:179-181` explicitly cites `rpc.go:275` (Go's non-blocking select with a default-log). The drop-rather-than-block parity is correct; the missing telemetry is not. Add a `debug!(base_topic, from, request_id, "pubsub_rpc: response dropped due to full buffer")` at the correlator or in the handler.

**3. Duplicate request-id collision silently replaces the first caller’s entry.**
`Correlator::publish` at `correlator.rs:117-136` uses `HashMap<Cid, Entry>::insert`. If two callers concurrently call `pubsub_sync_documents` with identical `doc_ids` (same bytes → same derived ID), the second overwrites the first; the first’s receiver will never be signalled and its caller will 5-second-timeout. Go's design has the same structure, so it’s a parity match, but the behavior is surprising and undocumented. Either:
- Document it on `Correlator::publish` / `pubsub_sync_documents`, or
- Key on `(caller_peer, request_id)` so concurrent callers coexist (Rust-native improvement).

**4. CI does not regenerate Go fixtures.**
`.github/workflows/ci.yml:229-233` just runs `--test p2p` with `go_` filters; it never invokes `go run testdata/gen_*/main.go`. If upstream Go defradb ever reorders a struct field, the hex constants at `pubsub.rs` and `pubsub_services.rs` go stale silently — the tests keep passing because they assert against the stale constants. Add a CI step that runs both generators and either greps the committed hex or commits them to disk and `git diff --exit-code`.

**5. Shutdown does not explicitly cancel in-flight pubsub requests.**
`SyncCoordinator::shutdown` closes the semaphores (`mod.rs:257-258`) and drains background tasks, but `PubsubServices` is dropped implicitly. In-flight `PreparedPublish` receivers get no signal and wait for their 5s timeout. Low impact since shutdown already blocks on other drains; fix is one `debug!` + iterating `correlator.ongoing` to send close on each sender.

### Nits

**6. `start_pubsub_services` soft-fails on the first `subscribe_raw` error and returns `Ok(())`** (`pubsub_client.rs:66-69`). That leaves whatever topics did subscribe in a partially-registered state. On iroh (the intended target) all four fail identically, but a partial failure on libp2p is silently swallowed. Either fail hard on libp2p or explicitly unsubscribe the ones that succeeded.

**7. `P2PAdapter::sync_documents` binary switch.**
`crates/p2p-adapter/src/libp2p.rs:672` uses `self.sync_coordinator.is_some()` as the selector. If #6 silently degraded, `sync_coordinator` is still `Some` but subscriptions are missing; the subsequent `publish_raw` at `pubsub_client.rs:110-118` then fails. A `ready` flag on `PubsubServices` would make this robust.

**8. `handler_for_topic` uses `concat!("doc-sync", "/")`** (`pubsub_services.rs:86,89`) — the `concat!` is a no-op macro call on two literals; just write `"doc-sync/"` directly.

### Wire-format parity

Verified all eight shapes (`DocSync{Request,Item,Reply}`, `BranchableSync{Request,Reply}`, plus empty-slice / null variants) round-trip against Go hex fixtures. Field order matches Go declaration order for inner messages (fxamacker/cbor default — `pubsub.rs:16-17` comment is accurate). Envelope uses dag-cbor canonical order via ciborium (carried over from #895). Bounded-deserialize is enforced at visitor level, not post-hoc (`pubsub.rs:363-376, 444-457, 521-531`). Empty-slice-as-null is bidirectional: `nil_or_seq` on serialize, `visit_unit` on deserialize. The "both CBOR null and empty array decode to empty vec" asymmetry is fine — Go never emits `0x80` for nil slices.

### Verdict

**Approve with required changes** — the access-check parity gap (#1) needs to land before merge given the parity-first policy. The rest can follow up, but CI fixture regeneration (#4) is worth doing in the same PR so the parity claim has teeth.

fredcarle:
## Re-review — commit 22407e95 ("close pubsub_rpc / two-stream access-check parity gap")

Reviewed the single new commit on top of `a654cf73`. Worktree at `/tmp/defradb.rs-pr896-r2`, HEAD `22407e95`. Cross-checked the authorization claim against Go (`internal/db/p2p/sync_doc.go:279-304`, `sync_branchable_col.go:253-272`).

### Resolved

- **[Blocker #1] Access-check parity gap — fixed.** The new `AccessAuthorizer` trait + `RuntimeAuthorizer<T>` impl (`crates/p2p/src/sync/coordinator/authorizer.rs`) is exactly the right shape. Both `SyncCoordinator::check_peer_is_replicator` / `check_access_str` (`access.rs:44-98`) and `HandlerContext::peer_may_*` (`pubsub_services.rs:102-117`) now funnel through the same four/five-step decision, so drift is structurally prevented. `peer_authorized_for_collection` includes the `transport.get_replicator` cache-miss backfill + registry insert at `authorizer.rs:101-119`, which was the specific path pubsub was skipping. The two new regression tests (`doc_sync_handler_accepts_via_authorizer_transport_fallback`, `branchable_sync_handler_accepts_via_authorizer_collection_fallback`) lock in the delegation. `RecordingAuthorizer` is a clean fake — it proves the handler consults the authorizer rather than shortcutting, which is the correctness claim that matters.
- **[Non-blocker #2] Silent backpressure drop — fixed.** `Correlator::deliver` now emits `debug!` on `TrySendError::Full` (`correlator.rs:196-204`), matching Go's `rpc.go:275` default-log behavior.
- **[Non-blocker #4] CI fixture drift — fixed, different approach.** The new `ci.yml:189-221` step runs both generators, extracts every `(N bytes): <hex>` token from stdout, and greps each against `crates/p2p/src/**`. Strictly weaker than commit-and-diff (the hex just has to _exist_ somewhere in the tree — it doesn't have to be the one the tests compare against), but good enough to catch wire-format reorderings in upstream Go. Acceptable as a pragmatic drift guard; the generator filenames in the `::error::` hint are a nice touch for future maintainers.
- **[Nit #8] `concat!("doc-sync", "/")`** — fixed to literal form (`pubsub_services.rs:79,81`).

### Additional good call

Dropping the `subscribed_collections` RX-side gate in `peer_may_branchable_sync` is correct parity. Verified against Go: `syncBranchableCollectionMessageHandler` does _zero_ peer-auth + zero subscription check — it just fetches heads from the local DB and replies. Rust's `Controlled` mode still layers in a "connected or explicit replicator" check as defense-in-depth, which is stricter than Go but does not break round-trip parity in `Open` mode. The new `branchable_sync_handler_serves_unsubscribed_collection_when_authorized` test pins this.

### Observations on the fix itself

- `authorizer.rs:124-126` — in `peer_authorized_for_collection`, after the transport replicator fallback misses, we fall through to `peer_state.is_connected` + `transport_reports_connected_peer`. That matches `check_access_str`'s old semantics (step 4/5), but note this diverges from `peer_authorized_for_any` in a subtle way: `any` checks `is_any_replicator` then bypasses the transport replicator lookup (there's no "get_any_replicator"). That's fine, just worth internalising — they're not symmetric helpers, and the asymmetry is correct.
- `RuntimeAuthorizer` captures `T: P2PTransport` directly; because `T: Clone` is a transport requirement, `transport.clone()` at `constructor.rs:108` is free. No concerns.
- `pubsub_services: Option<PubsubServices>` kept on the coordinator (`mod.rs:248-250`) — consistent with the iroh-lacks-libp2p-PeerId path. `try_new` returning `None` on unparseable peer id is the right gate.

### Still outstanding from prior review (unchanged, still non-blocker)

- **#3 Correlator request-id collision.** `HashMap::insert` at `correlator.rs:117-136` still silently replaces; the first caller 5-second-timeouts. Not a correctness break, but undocumented. Keying on `(caller, id)` would be a Rust-native improvement — file a follow-up.
- **#5 Shutdown doesn't signal in-flight pubsub waiters.** `SyncCoordinator::shutdown` closes semaphores but doesn't touch `Correlator::ongoing`. Existing callers block on other drains already, so low impact.
- **#6 `start_pubsub_services` partial-failure.** `pubsub_client.rs:66-69` still early-returns `Ok(())` on the first `subscribe_raw` error, leaving any previously-registered topics dangling. On libp2p this never fires in practice (transport is ready by the time `start_pubsub_services` is called); on iroh all four fail identically. Keep as nit, but an `unsubscribe_raw` unwind or a `ready: AtomicBool` flag on `PubsubServices` would harden this.
- **#7 `P2PAdapter::sync_documents` binary switch** (`crates/p2p-adapter/src/libp2p.rs:672`). Same flag-based robustness argument as #6.

### Verdict

**Approve.** The blocker from the prior review is cleanly resolved via a shared authorizer, regression tests are in place for both doc-sync and branchable-sync paths, the Go-fixture drift guard is landed, and the branchable subscription-is-RX-only change is correct Go parity. Remaining items are all follow-up-worthy but not merge-blocking.

PR #897, opened 2026-04-22
fix(p2p): per-peer Bitswap ACP filter (#830)

### LINKED ISSUE #830: Security: Bitswap serves blocks to any peer with no per-peer ACP filter

fredcarle:
## Summary

Closes #830.

Without a filter, a node running in `AccessMode::Controlled` still served every stored block to any connected peer via Bitswap, bypassing the `ReplicatorRegistry` that gates ingress. Any libp2p-connectable peer could fetch collection ciphertext and probe block existence.

This PR wires a per-peer block-request filter at Bitswap construction, mirroring Go's `bitswap.WithPeerBlockRequestFilter(hasAccess)` in `go-p2p/peer.go:146`. The filter reads the requested block, decodes its `collectionVersionID`, and authorizes the requesting peer against the same `ReplicatorRegistry` that `SyncCoordinator` uses.

### Decision flow (`crates/p2p/src/bitswap/filter.rs`)

| Condition                    | Result |
|------------------------------|--------|
| `AccessMode::Open`           | allow  |
| block missing from store     | deny (also avoids existence leak) |
| decodes as `Signature`       | allow (needed to verify sibling blocks) |
| definition-delta block       | allow (schema/collection defs broadcast to all) |
| data-delta block             | allow iff `registry.is_replicator(collection, peer)` |
| decode failure / other       | deny |

### Surface changes

- `crates/p2p/src/bitswap/filter.rs` — new module + 6 unit tests
- `crates/p2p/tests/bitswap_acp_filter_tests.rs` — 4 end-to-end tests over real libp2p/TCP plus 1 `#[ignore]`'d live Go interop test
- `P2PHostConfig` gains `access_mode: AccessMode` (default `Open`); CLI derives it from `config.acp.document_type` identically to `SyncCoordinator`
- `DefraBehaviour::new` signature now accepts `access_mode` + `Arc<ReplicatorRegistry>`
- `bitswap/access.rs` doc comment rewritten to reflect the egress filter
- `iroh/endpoint_commands.rs::handle_dial` args bundled into a `DialContext` struct (drive-by fix for a pre-existing `clippy::too_many_arguments` regression on main that would otherwise block the lint job)

### Upstream dependency

iroh-bitswap's `PeerBlockRequestFilter` trait was synchronous; the Rust check needs an async blockstore read. The beetle-side change (sourcenetwork/beetle#2) has been merged and `Cargo.toml` points back at `sourcenetwork/beetle` main.

## Test plan

- [x] `cargo test -p p2p` → 283 tests pass (157 lib + 126 integration, including the 4 new end-to-end tests; 1 live Go interop test ignored by default)
- [x] `cargo clippy --all -- -D warnings` → clean
- [x] `cargo build -p cli --features iroh` → clean
- [x] `cargo test -p integration-test --test p2p -- --test-threads=1 rust_rust_` → 12 pass (3 resilience-only ignored)
- [x] `cargo test -p integration-test --test p2p -- --test-threads=1 go_` → 21 pass (8 pre-existing ignored cases unchanged)
- [x] `cargo test -p integration-test --test basic -- --skip ::go_` → 12 pass
- [x] `cargo test -p integration-test --test acp -- --skip ::go_` → 105 pass
- [x] **Live Go-Rust parity proof**: ran `bitswap_unauthorized_against_go_defradb` against a real Go defradb v1.0.0-rc1 instance with `--document-acp-type local` serving an ACP-protected User document:
  ```
  connected to Go peer 12D3KooWJn3HaMEjzWDmEXRop1ierA7C5pv61Tvn4ucKgZm6dHdd
  got_block=false completed_success=false
  test bitswap_unauthorized_against_go_defradb ... ok
  ```
  Go denied the Bitswap WANT from the unregistered Rust peer — matching the deny behavior this PR now implements on the Rust side.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6246,7 +6246,7 @@ dependencies = [
 [[package]]
 name = "iroh-bitswap"
 version = "0.2.0"
-source = "git+https://github.com/fredcarle/beetle?branch=fredcarle/async-peer-block-request-filter#e0fa8cd0aab748117d63a8dc1002487db1ee9da8"
+source = "git+https://github.com/sourcenetwork/beetle?branch=main#6441b0ac5f42d466f44963afb782bb56030b1479"
 dependencies = [
  "ahash 0.8.12",
  "anyhow",
@@ -6366,7 +6366,7 @@ dependencies = [
 [[package]]
 name = "iroh-metrics"
 version = "0.2.0"
-source = "git+https://github.com/fredcarle/beetle?branch=fredcarle/async-peer-block-request-filter#e0fa8cd0aab748117d63a8dc1002487db1ee9da8"
+source = "git+https://github.com/sourcenetwork/beetle?branch=main#6441b0ac5f42d466f44963afb782bb56030b1479"
 dependencies = [
  "async-trait",
  "config",
@@ -6477,7 +6477,7 @@ dependencies = [
 [[package]]
 name = "iroh-util"
 version = "0.2.0"
-source = "git+https://github.com/fredcarle/beetle?branch=fredcarle/async-peer-block-request-filter#e0fa8cd0aab748117d63a8dc1002487db1ee9da8"
+source = "git+https://github.com/sourcenetwork/beetle?branch=main#6441b0ac5f42d466f44963afb782bb56030b1479"
 dependencies = [
  "anyhow",
  "cid 0.10.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5029,8 +5029,8 @@ dependencies = [
  "libc",
  "log",
  "rustversion",
- "windows-link 0.2.1",
- "windows-result 0.4.1",
+ "windows-link 0.1.3",
+ "windows-result 0.3.4",
 ]
 
 [[package]]
@@ -5751,7 +5751,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.62.2",
+ "windows-core 0.61.2",
 ]
 
 [[package]]
@@ -6246,7 +6246,7 @@ dependencies = [
 [[package]]
 name = "iroh-bitswap"
 version = "0.2.0"
-source = "git+https://github.com/sourcenetwork/beetle?branch=main#48e70b0345ea842012285e1fafa27406bb14ef83"
+source = "git+https://github.com/fredcarle/beetle?branch=fredcarle/async-peer-block-request-filter#e0fa8cd0aab748117d63a8dc1002487db1ee9da8"
 dependencies = [
  "ahash 0.8.12",
  "anyhow",
@@ -6366,7 +6366,7 @@ dependencies = [
 [[package]]
 name = "iroh-metrics"
 version = "0.2.0"
-source = "git+https://github.com/sourcenetwork/beetle?branch=main#48e70b0345ea842012285e1fafa27406bb14ef83"
+source = "git+https://github.com/fredcarle/beetle?branch=fredcarle/async-peer-block-request-filter#e0fa8cd0aab748117d63a8dc1002487db1ee9da8"
 dependencies = [
  "async-trait",
  "config",
@@ -6477,7 +6477,7 @@ dependencies = [
 [[package]]
 name = "iroh-util"
 version = "0.2.0"
-source = "git+https://github.com/sourcenetwork/beetle?branch=main#48e70b0345ea842012285e1fafa27406bb14ef83"
+source = "git+https://github.com/fredcarle/beetle?branch=fredcarle/async-peer-block-request-filter#e0fa8cd0aab748117d63a8dc1002487db1ee9da8"
 dependencies = [
  "anyhow",
  "cid 0.10.1",
@@ -10446,7 +10446,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "rustix 1.1.4",
- "windows 0.62.2",
+ "windows 0.61.3",
 ]
 
 [[package]]
@@ -14692,8 +14692,8 @@ dependencies = [
  "log",
  "serde",
  "thiserror 2.0.18",
- "windows 0.62.2",
- "windows-core 0.62.2",
+ "windows 0.61.3",
+ "windows-core 0.61.2",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,7 @@ ciborium = "0.2"
 # P2P networking
 # Note: Using 0.53 to match iroh-bitswap's dependency
 libp2p = { version = "0.53", features = ["tcp", "noise", "yamux", "gossipsub", "kad", "identify", "relay"] }
-iroh-bitswap = { git = "https://github.com/sourcenetwork/beetle", branch = "main" }
+iroh-bitswap = { git = "https://github.com/fredcarle/beetle", branch = "fredcarle/async-peer-block-request-filter" }
 libp2p-stream = "0.1.0-alpha.1"
 multiaddr = "0.18"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,7 @@ ciborium = "0.2"
 # P2P networking
 # Note: Using 0.53 to match iroh-bitswap's dependency
 libp2p = { version = "0.53", features = ["tcp", "noise", "yamux", "gossipsub", "kad", "identify", "relay"] }
-iroh-bitswap = { git = "https://github.com/fredcarle/beetle", branch = "fredcarle/async-peer-block-request-filter" }
+iroh-bitswap = { git = "https://github.com/sourcenetwork/beetle", branch = "main" }
 libp2p-stream = "0.1.0-alpha.1"
 multiaddr = "0.18"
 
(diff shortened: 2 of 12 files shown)


fredcarle:
## Review — #897 fix(p2p): per-peer Bitswap ACP filter

### Overview

Closes #830 by installing a per-peer `PeerBlockRequestFilter` on the libp2p Bitswap server, mirroring Go's `bitswap.WithPeerBlockRequestFilter(hasAccess)` wiring at `go-p2p/peer.go:146`. `DefraBehaviour::new` now takes `(AccessMode, Arc<ReplicatorRegistry>)` and builds the filter up front; the CLI derives `AccessMode` from `config.acp.document_type` the same way `SyncCoordinator` does. The filter reads the block, short-circuits on signature/definition kinds, and for data deltas consults the shared `ReplicatorRegistry`. Also drive-by: iroh `handle_dial` args collapsed into `DialContext`, and the beetle pin moves to a commit where `PeerBlockRequestFilter` is async.

### Verdict: **Approve with required changes**

The core fix is correct, scoped, and closes a real security hole. One known parity gap needs either a fix or an explicit carved-out follow-up before merge; the rest are non-blockers.

---

### Required before merge

**1. Parity gap — Lens blocks will be denied in `Controlled` mode.** (`crates/p2p/src/bitswap/filter.rs:87-89`)

Go's `hasAccess` (`internal/db/p2p/p2p.go:335-348`) explicitly allows `lens` / `modules` / `wasmBytes` / `chunks` blocks: decode as `DefraBlock` fails → Go inspects the error string and returns `true`. Rust's filter returns `false` on the same decode path. Rust ships `LensConfigBlock` / `LensModuleBlock` / `LensWasmBlock` in `defra-core/src/lens_block.rs` specifically for Bitswap P2P sync of schema migrations — so with ACP on, lens propagation silently stops.

The filter docstring acknowledges this ("lens/wasm/chunk blocks, decode errors → deny by default"). Under the parity-first policy this is not acceptable as an open-coded TODO. Two acceptable resolutions:

- **Preferred:** try `LensConfigBlock` / `LensModuleBlock` / `LensWasmBlock` deserialization in the "not a `DefraBlock`" branch before denying. Add a test asserting a lens block is served in `Controlled` mode.
- **Acceptable:** file a follow-up issue (e.g. "#XXX lens blocks denied by Bitswap ACP filter in Controlled mode"), reference it in the filter docstring + decision table, and add a negative test locking in the current deny behavior so the regression is tracked.

---

### Non-blocking

**2. Parity gap — no ACP-identity fallback.** Go falls through the replicator check to `CheckDocAccessWithIdentityFunc` (identity-protocol lookup + `acpDB` permission check). Rust's filter only consults the registry. This is a pre-existing architectural gap — `SyncCoordinator::check_access_str` also doesn't do identity-based fallback — not introduced here. Worth a tracking issue if one doesn't exist.

**3. Silent deny.** All deny branches in `check_access` return `false` with no log. Go calls `log.ErrorE("Failed to get doc from block", err)` etc. A `tracing::debug!(?cid, %peer, reason = "...", "bitswap request denied")` on each deny would match Go's observability and make field debugging possible.

**4. `open_mode_serves_all_peers` test cruft.** `crates/p2p/tests/bitswap_acp_filter_tests.rs:372-373` has `let _ = Arc::clone(&producer_replicators); let _ = consumer_peer;` purely to silence unused warnings. Just drop both bindings.

**5. Cargo.lock Windows-crate downgrades.** Lots of `windows-core` / `windows` / `fs_at` / `native-tls-vendored` downgrades from `0.62` → `0.61`. Not explained in the PR description. Appears to be a byproduct of Cargo re-resolving after the beetle bump, but worth a sanity check that nothing on the `0.62` line was intentional.

**6. Filter unit tests.** Good coverage for open/registered/unregistered/cross-collection/missing-block/signature. Missing:
- Definition-delta block (exercises `is_definition` short-circuit).
- Lens block (see #1).

### Things done well

- `ReplicatorRegistry` is constructed up-front in `P2PHost::with_keypair_and_config_and_identity` and shared with the behaviour, so the filter and `SyncCoordinator` genuinely see the same state — not two copies.
- End-to-end tests run over real libp2p TCP loopback, not mocks.
- `bitswap_unauthorized_against_go_defradb` is a real live Go interop test that actually proves deny-parity at the wire.
- `DialContext` drive-by is well-motivated and cleanly factored.
- `missing_block_denies_without_leaking` is the right name — the existence-leak concern is real and the test locks it in.
- CLI `access_mode` derivation is identical to the `SyncCoordinator` path — no accidental divergence between the two enforcement points.

### Security notes

- Signature-block bypass is the same behavior as Go and is necessary for siblings to verify data. No concern.
- Open-mode default preserves pre-#830 behavior for non-ACP deployments.
- No leak of registry state or collection membership through the filter decision (only boolean returned).

Issue #837, opened 2026-04-15
P2P: replicator retry loop and status tracking missing


fredcarle:
## Summary

Go's P2P layer has a full replicator retry machine: a 2-second scheduled loop, per-replicator retry state persisted to the peerstore, exponential backoff up to ~12 hours, and an Active/Inactive status toggle that drives all of it. Rust has none of this. Transient peer failures abandon every queued block until the next manual action, and there is no way to detect that a replicator has been down for seconds vs hours. This is a fundamental parity gap, not a tuning issue.

This issue bundles two findings (P-40 retry loop, P-41 status) because neither is usable without the other.

### P-40 — retry loop missing

Go (`defradb/internal/db/p2p/replicator.go:421-707`) runs `handleReplicatorRetries` on a 2s ticker. For each replicator marked Inactive, it:

1. Pops queued `head` entries for that replicator from the peerstore.
2. Attempts to send the push; on success advances the queue and updates status to Active.
3. On failure records the failure, bumps `retryInterval`, and re-schedules.

`retryIntervals` is an exponential array (see `replicator.go:43-45`) that grows from seconds to hours across ~12 retries.

Rust `crates/p2p/src/sync/replication/` has no equivalent. `handlers.rs` logs errors and continues; `loop_runner.rs` has no retry scheduling; `recovery.rs` is for crash recovery on startup, not for transient-failure retry.

### P-41 — status tracking missing

Ties to prior #832 (`ReplicatorInfo` schema gap). Go's retry loop is driven by `client.Replicator.Status` and `client.Replicator.LastStatusChange` (`defradb/client/replicator.go:18-34`), toggled via `updateReplicatorStatus` at `defradb/internal/db/p2p/replicator.go:495-542`. Rust's `ReplicatorInfo` lacks both fields, so even if the retry loop were implemented it would have nowhere to store state.

## Impact

- Replication is not resilient. A 30-second network blip abandons every queued block for every affected replicator.
- No operational visibility into replicator health (inactive vs never-seen is indistinguishable).
- Interop: Rust and Go replicators on the same network diverge in what gets replicated under load, because Go retries and Rust doesn't.

## Acceptance Criteria

- [ ] `ReplicatorInfo` gains `status: ReplicatorStatus` and `last_status_change` fields (see #832)
- [ ] New retry loop task scheduled on a 2s tick, reading replicators with Inactive status from the peerstore
[...]

PR #902, opened 2026-04-23
fix(p2p): align ReplicatorInfo schema + wire format with Go (#832)

### LINKED ISSUE #832: P2P: `ReplicatorInfo` schema diverges from Go `client.Replicator`
## Summary

Rust's `ReplicatorInfo` (persisted to the Peerstore) is missing two fields that Go's `client.Replicator` uses to track replicator state, and the field naming does not match Go. Rust cannot represent the Active/Inactive status that Go's retry loop depends on, and any shared-storage migration path between Go and Rust will fail.

## Details

Go (`defradb/client/replicator.go:18-34`):
```go
type Replicator struct {
    ID               string
    Addresses        []string
    CollectionIDs    []string
    Status           ReplicatorStatus   // uint8 enum: Active(0) / Inactive(1)
    LastStatusChange time.Time
}

type ReplicatorStatus uint8
const (
    ReplicatorStatusActive ReplicatorStatus = iota
    ReplicatorStatusInactive
)
```

Rust (`crates/p2p/src/replicator.rs:33-44`):
```rust
pub struct ReplicatorInfo {
    #[serde(rename = \"peer_id\")]
    peer_id_str: String,

    pub collections: Vec<String>,   // <-- \"collections\" (Go: \"CollectionIDs\")

    #[serde(default)]
    addresses_str: Vec<String>,     // <-- \"addresses_str\" (Go: \"Addresses\")
}
```

Gaps:
1. No `Status` field. Go toggles this field when a replicator connection fails and drives retry behavior from it (`defradb/internal/db/p2p/replicator.go:713`). Rust cannot represent the inactive state.
2. No `last_status_change` timestamp.
3. Field names differ (`peer_id` vs `ID`, `collections` vs `CollectionIDs`).
4. `addresses_str` has no `#[serde(rename)]` at all — serializes as the raw Rust identifier `addresses_str`.

`ReplicatorInfo::to_bytes` / `from_bytes` (`replicator.rs:117-125`) use `serde_cbor` for peerstore persistence. Any shared-datastore migration fails to decode across implementations.

Also: `ReplicatorInfo::new` accepts empty collections silently; only `try_new` validates. `A-5` from the review — inconsistent enforcement.

## Expected Behavior

`ReplicatorInfo` has the same fields and wire names as Go's `client.Replicator`, and represents the full Active/Inactive state machine.

## Acceptance Criteria

- [ ] Add `status: ReplicatorStatus` field (new `ReplicatorStatus` enum matching Go's `uint8` Active(0)/Inactive(1))
- [ ] Add `last_status_change` field (typed to serialize compatibly with Go's `time.Time`)
- [ ] Rename serialized fields to match Go: `ID`, `Addresses`, `CollectionIDs`, `Status`, `LastStatusChange`
- [ ] Decide on the `peer_id` / `addresses` raw-string ergonomics — keep the getters but fix the wire names
- [ ] Constructor consistency: make `new` reject empty collections or rename it `unchecked_new`
- [ ] Unit test: round-trip CBOR bytes match a Go-produced fixture for the same replicator data

## Files

- `crates/p2p/src/replicator.rs`
- `crates/p2p/src/bitswap/registry.rs` — `load_from_infos` may need to handle the new Status field

## References

- Go: `defradb/client/replicator.go:18-34`
- Go: `defradb/internal/db/p2p/replicator.go:713` — retry loop reads Status

fredcarle:
## Summary

- Rust's `ReplicatorInfo` now matches Go's `client.Replicator` field-for-field: adds `Status` (`Active=0`/`Inactive=1` `uint8` enum) and `LastStatusChange` (`DateTime<Utc>`), renames wire fields to Go PascalCase (`ID`, `Addresses`, `CollectionIDs`, `Status`, `LastStatusChange`).
- Switches `to_bytes`/`from_bytes` from `serde_cbor` to `serde_json` so the on-disk encoding matches Go's `encoding/json` at `defradb/internal/db/p2p/replicator.go:149,525`. Previous CBOR format was silently incompatible with any Go peer.
- `ReplicatorInfo::new` now rejects empty collections (was a silent foot-gun); `try_new` is removed. `from_raw` remains the unchecked escape hatch for storage reads / tests.

## Go parity evidence

New tests in `crates/p2p/tests/replicator_tests.rs` bake **byte-exact** JSON fixtures produced by Go's `encoding/json` on the real `client.Replicator` struct and assert:

- Go → Rust decode for 3 shapes (active+addresses, inactive+`null` addresses, zero `time.Time`).
- Rust → Go encode byte-for-byte match for 2 shapes.
- Round-trip preserves all fields.
- Decode rejects unknown `Status` bytes, truncated JSON, etc.

Go's `nil []string` → JSON `null` vs empty `[]string{}` → JSON `[]` is handled by accepting both on decode; Rust always emits `[]`.

## Scope — write-path gap deferred to #837

`ReplicatorInfo::to_bytes` is called by the libp2p/cli doc-pusher adapters (`p2p-adapter/src/libp2p_doc_pusher.rs:131`, `cli/src/p2p_doc_pusher.rs:93`), but the host command handler itself (`crates/p2p/src/host/command_handler/messaging.rs:280`) still only updates in-memory `bitswap::registry` state. The full peerstore persistence path + Active/Inactive state transitions belong to #837 (replicator retry loop + status tracking) — this PR just aligns the on-disk schema so #837 can wire the writes cleanly.

## Test plan
- [x] `cargo test -p p2p` — passes, 14 new fixture/parity tests
- [x] `cargo clippy --all -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo build --all` — clean
- [ ] CI integration suites

Closes #832.


fredcarle (commented):
## Review — align ReplicatorInfo schema + wire format with Go

### Overview

Makes `ReplicatorInfo` a byte-exact match for Go's `client.Replicator`: renames wire fields to Go PascalCase (`ID`, `Addresses`, `CollectionIDs`, `Status`, `LastStatusChange`), adds `Status`/`LastStatusChange`, and swaps on-disk encoding from CBOR → JSON. Verified against `defradb/client/replicator.go:18-34` — field names, types, and variant ordering all match.

### Strengths

- **Byte-exact parity evidence.** Three real Go-produced JSON fixtures (active+addrs, inactive+null, zero-time) drive decode tests and two round-trip as encode tests. This is the right shape for parity work.
- **Tolerant decode, strict encode.** Accepts Go's `null` and `[]` for list fields on input but always emits `[]` — aligned with how Rust collapses both into an empty `Vec`.
- **Defensive Status decode.** `try_from = "u8"` rejects unknown variants with a typed error (`replicator.rs:28-31, 56-64`).
- **Removes a silent footgun.** The previous CBOR encoding was incompatible with any Go peer.
- **PR description is honest about scope.** Explicitly defers the write-path + state-transition work to #837.

### Issues

**Nanosecond format drift (latent parity gap)**

Verified with a chrono probe: Rust's `DateTime<Utc>` serializes fractional seconds padded to 3/6/9 digits (e.g., `2024-06-15T12:34:56.100Z`), whereas Go's `time.Time.MarshalJSON` uses `RFC3339Nano` which trims trailing zeros (→ `.1Z`). Equal for zero nanos and for `time.Now()` in practice, but a Go-written timestamp that happens to land on `N × 100ms` or `N × 1ms` will not round-trip byte-exact through Rust. Fixtures only cover zero-nano cases, so this is untested. Low practical impact given how `LastStatusChange` is actually set on the Go side (`time.Time{}` or `time.Now()`, per `internal/db/p2p/replicator.go:518-527`), but worth either a custom serde adapter or a comment calling out the edge case.

**`from_raw` now silently bakes in `Active` + zero time**

`from_raw` (`replicator.rs:141-149`) hard-codes `status: Active` and `last_status_change: go_time_zero()`. Doc says "for deserialization or testing," but real deserialization goes through serde directly on the struct — not `from_raw`. The escape-hatch semantics are fine; the doc comment should drop the "deserialization" half or make explicit that callers must set status/last_status_change separately if they care.

**`new()` signature breakage**

`new()` changed from `-> Self` to `-> Result<Self, ReplicatorError>` and now rejects empty collections. Only two callers on this branch were updated (`cli/src/p2p_doc_pusher.rs:93`, `p2p-adapter/src/libp2p_doc_pusher.rs:136`), but any in-flight branches (#897, #896, etc.) that also call `ReplicatorInfo::new` will conflict. Worth a `git grep "ReplicatorInfo::new"` across open branches before merging, or a rebase coordination note.

**Visibility inconsistency**

`collections`, `status`, `last_status_change` are `pub` fields, while `id` and `addresses` are private with getters. Previously everything was private to hide the string-backed PeerId. Now it's mixed — tests rely on `pub status`/`last_status_change` to construct fixtures (`replicator_tests.rs:79-80, 110, 112`). Consider a builder or making all fields either `pub` or `pub(crate)` so the model is uniform.

### Test coverage gaps

- Round-trip with non-zero subsecond `last_status_change` (would surface the nanosecond-format drift above).
- Forward-compatibility: `serde_json` ignores unknown fields by default, but an explicit test documents the contract against future Go additions.

### Verdict

Solid parity fix with strong fixture-based evidence. Nanosecond drift is a latent edge case worth at least a comment even if not fixed now. Everything else is minor polish.

fredcarle:
Thanks for the thorough review. Pushed 6541d0b9 addressing the items:

### Nanosecond format drift
Fixed. Added `go_time_serde` module with a custom `serialize` that trims trailing zeros from the fractional seconds to match Go's `RFC3339Nano` rule (`.100` → `.1`, no-frac when all-zeros). Deserialize stays on chrono's default since the parser already accepts both trimmed and untrimmed forms. New test `timestamp_encodes_like_go_rfc3339nano_with_trailing_zeros` exercises five subsecond shapes — the 100 ms case is the one that silently drifted before.

### `from_raw` doc
Rewritten. Now says explicitly that real decode goes through serde on the struct (not `from_raw`), and that callers wanting non-default `status` / `last_status_change` must set them after construction.

### Visibility inconsistency
Made `id` and `addresses` pub to match the already-pub `collections`/`status`/`last_status_change`. The `peer_id()` / `addresses()` accessors remain for validation-aware callers.

### Test gaps
Both addressed: the nanosecond test above, plus `decode_ignores_unknown_fields` pinning forward-compat against future Go-side schema additions.

### `new()` signature breakage — cross-branch coordination
Confirmed with `git grep` against `origin/fix/828-part2-integration` (#896) and `origin/fix/830-bitswap-per-peer-acp` (#897): both branches use the old `-> Self` form in the same 5 call sites. Once #902 lands, those branches will need a rebase touch — fix is mechanical (append `.unwrap()` in the two registry tests, `?`-propagate in the two doc-pushers) and I'll handle it as part of rebasing those PRs.

fredcarle (commented):
## Follow-up review on `6541d0b9`

Confirmed all three addressed review items land cleanly.

### Nanosecond drift — fixed and verified

The `go_time_serde` module correctly matches Go's `time.Time.MarshalJSON` (RFC3339Nano). Probed independently across 7 edge cases including the drift case and mixed trailing zeros:

| ns | Rust output | Expected (Go) |
|---|---|---|
| 0 | `...T22:13:20Z` | `...T22:13:20Z` |
| 1,000,000 (1ms) | `.001Z` | `.001Z` |
| 100,000,000 (100ms) | `.1Z` | `.1Z` |
| 500,000,000 | `.5Z` | `.5Z` |
| 1,000 (1µs) | `.000001Z` | `.000001Z` |
| 1 (1ns) | `.000000001Z` | `.000000001Z` |
| 123,456,780 | `.12345678Z` | `.12345678Z` |

All byte-exact. The `timestamp_encodes_like_go_rfc3339nano_with_trailing_zeros` test pins five of these — solid coverage. Decode stays on chrono's default, which parses both padded and stripped forms.

Minor: the serializer implicitly relies on `to_rfc3339_opts(AutoSi, true)` producing a `Z` suffix, which holds for `DateTime<Utc>` with `use_z=true`. Fine, but a one-line assumption note above `let z_idx = formatted.len() - 1` would flag it for future chrono upgrades.

### Visibility — fixed

`id` and `addresses` are now `pub`, uniform with `collections` / `status` / `last_status_change`. The `peer_id()` / `addresses()` / `try_peer_id()` accessors stay for validation-aware callers. Consistent model.

### `from_raw` doc — fixed

New wording is accurate — "Unchecked constructor for test fixtures and raw peerstore reads," and explicitly notes that `status` / `last_status_change` defaults must be set by the caller if they care. No more misleading "for deserialization" framing.

### Forward-compat — bonus

`decode_ignores_unknown_fields` is a nice guard against an accidental `#[serde(deny_unknown_fields)]` regression that would break against future Go schema additions.

### Still open

Only the `new()` signature breakage — coordination note across in-flight branches (#897, #896, etc.) before merge. Pure merge-time concern, no code change needed.

Looks ready from the parity side.

PR #915, opened 2026-04-27
fix(p2p): match Go time.Time{} reset on replicator recovery (#914)

### LINKED ISSUE #914: p2p: ReplicatorInfo recovery should reset LastStatusChange to Go time.Time{} (parity gap)
## Parity gap

After #908 (`fix(p2p): persist replicator retry status`), `ReplicatorInfo::set_status_if_changed_now` always stamps `last_status_change` with `Utc::now()`. Go's `updateReplicatorStatus` (`defradb/internal/db/p2p/replicator.go:495`) special-cases the `Inactive → Active` recovery transition and resets `LastStatusChange` to the Go `time.Time{}` zero value (`0001-01-01T00:00:00Z`):

\`\`\`go
case true: // active = true
    if rep.Status == client.ReplicatorStatusInactive {
        rep.LastStatusChange = time.Time{}
    }
    rep.Status = client.ReplicatorStatusActive
case false:
    if rep.Status == client.ReplicatorStatusActive {
        rep.LastStatusChange = time.Now()
    }
    rep.Status = client.ReplicatorStatusInactive
\`\`\`

A Rust node recovering a peer therefore writes a JSON record that diverges from what a Go node would write for the same recovery — regressing the byte-for-byte peerstore contract established in #832 / PR #902.

## Where it bites

- `crates/p2p/src/replicator.rs::set_status_if_changed_now` always writes `Utc::now()`.
- `crates/p2p/tests/replicator_tests.rs::status_update_records_first_transition_only` *pins* the Rust-only behavior — it asserts `info.last_status_change == active_at` after the recovery transition, exactly the wrong post-condition for parity.
- `crates/p2p-adapter/src/replicator_status.rs::set_persisted_replicator_status` and the libp2p + iroh retry loops in `crates/cli/src/commands/start/server_p2p.rs` inherit the wrong timestamp on every recovery.

## Suggested fix

Have `set_status_if_changed_now` use `Utc::now()` only on `Active → Inactive` and the existing `go_time_zero()` helper on `Inactive → Active`. Keep `set_status_if_changed(status, changed_at)` as-is for callers that want to pass an explicit timestamp. Replace the existing parity-pinning test with one that asserts the Go reset.

## Why this matters

`LastStatusChange` is observable through the HTTP API and persisted to the peerstore. No internal logic branches on its value, but #902 established strict byte-exact JSON parity with Go's `client.Replicator` so a shared peerstore round-trips. This is a small regression of that contract.

Discovered while working on the (now-superseded) tight scope of #837.

fredcarle:
## Summary

Closes #914.

`ReplicatorInfo::set_status_if_changed_now` was stamping `last_status_change` with `Utc::now()` for every transition. Go's `updateReplicatorStatus` (`defradb/internal/db/p2p/replicator.go:495`) special-cases the `Inactive → Active` recovery transition and resets `LastStatusChange` to `time.Time{}` (`0001-01-01T00:00:00Z`) so a recovered record serializes identically to a freshly-constructed one. Without this, recovery transitions write JSON that diverges byte-for-byte from a Go-produced peerstore record — regressing the wire-format contract established in #902.

The fix is local to one method: pick `Utc::now()` for `Active → Inactive` and the existing `go_time_zero()` helper for `Inactive → Active`. Both call sites (`crates/p2p-adapter/src/replicator_status.rs` and the inline copy in `crates/cli/src/commands/start/server_p2p.rs`) inherit the corrected behavior automatically.

The previous test (`status_update_records_first_transition_only`) only exercises `set_status_if_changed` (the explicit-timestamp variant) and continues to pass — added a new test that pins the Go-parity rule for the `_now` variant.

## Test plan

- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --all -- -D warnings` — clean (rustc 1.95)
- [x] `cargo test -p p2p -p defra-p2p-adapter` — all green (~382 tests)
- [x] New test `set_status_if_changed_now_matches_go_recovery_rule` confirmed red before fix, green after

### DIFF
--- a/crates/p2p/src/replicator.rs
+++ b/crates/p2p/src/replicator.rs
@@ -237,9 +237,21 @@ impl ReplicatorInfo {
         true
     }
 
-    /// Update status using the current UTC timestamp when the state changes.
+    /// Update status using Go's recovery-aware timestamp rule.
+    ///
+    /// Mirrors Go's `updateReplicatorStatus`
+    /// (`defradb/internal/db/p2p/replicator.go:495`):
+    /// - `Active → Inactive` stamps `Utc::now()`.
+    /// - `Inactive → Active` resets to `time.Time{}` so a recovered record
+    ///   serializes identically to a freshly-constructed one.
+    ///
+    /// Same-status calls leave the timestamp untouched.
     pub fn set_status_if_changed_now(&mut self, status: ReplicatorStatus) -> bool {
-        self.set_status_if_changed(status, Utc::now())
+        let changed_at = match status {
+            ReplicatorStatus::Inactive => Utc::now(),
+            ReplicatorStatus::Active => go_time_zero(),
+        };
+        self.set_status_if_changed(status, changed_at)
     }
 
     /// Format `LastStatusChange` exactly like Go's `time.Time.MarshalJSON`.
--- a/crates/p2p/tests/replicator_tests.rs
+++ b/crates/p2p/tests/replicator_tests.rs
@@ -131,6 +131,34 @@ fn status_update_records_first_transition_only() {
 }
 
 #[test]
+fn set_status_if_changed_now_matches_go_recovery_rule() {
+    // Go's `updateReplicatorStatus` (`defradb/internal/db/p2p/replicator.go:495`)
+    // resets `LastStatusChange` to `time.Time{}` on `Inactive → Active`,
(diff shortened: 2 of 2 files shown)


jackzampolin:
Thanks Fred, appreciate the focused parity fix and the regression coverage.

PR #926, opened 2026-04-28
fix(p2p): harden bidirectional sync recovery

jackzampolin:
## Summary
- register Bitswap retry query IDs so retry completions keep driving pending DAG recovery
- bind DocSync and BranchableSync two-stream replies to pending peer/message IDs
- honor `rebroadcast_on_merge` in batch merges and surface ready-event send failures
- make explicit DocSync integration failures fail the test instead of returning early

## Tests
- cargo test -p p2p --lib
- cargo test -p integration-test --test p2p --no-run

### DIFF
--- a/crates/p2p/src/sync/coordinator/dag_fetcher.rs
+++ b/crates/p2p/src/sync/coordinator/dag_fetcher.rs
@@ -191,7 +191,16 @@ async fn emit_dag_ready(
 ) {
     let mut context = context.clone();
     context.fill_missing_from_block(root_data);
-    let _ = event_tx.send(context.into_dag_ready(root_cid)).await;
+    if event_tx
+        .send(context.into_dag_ready(root_cid))
+        .await
+        .is_err()
+    {
+        warn!(
+            root_cid = %root_cid,
+            "Failed to emit DagReady after DAG fetch"
+        );
+    }
 }
 
 /// Try to fetch an entire DAG via a single CAR request.
--- a/crates/p2p/src/sync/coordinator/event_handler/bitswap.rs
+++ b/crates/p2p/src/sync/coordinator/event_handler/bitswap.rs
@@ -38,18 +38,29 @@ impl<B: Blockstore + 'static, T: P2PTransport> SyncCoordinator<B, T> {
                             providers.push(source_transport_id);
                         }
                     }
-                    if let Err(e) = self
+                    match self
                         .runtime
                         .transport
                         .sync_blocks(root_cid, providers, missing)
                         .await
                     {
-                        tracing::warn!(
-                            query_id = query_id.0,
-                            root_cid = %root_cid,
-                            error = %e,
-                            "Failed to start Bitswap fetch for remaining child blocks"
-                        );
+                        Ok(retry_query_id) => {
+                            self.manager.register_query(retry_query_id, root_cid);
+                            tracing::debug!(
+                                query_id = query_id.0,
+                                retry_query_id = retry_query_id.0,
+                                root_cid = %root_cid,
+                                "Started Bitswap fetch for remaining child blocks"
+                            );
+                        }
+                        Err(e) => {
+                            tracing::warn!(
+                                query_id = query_id.0,
+                                root_cid = %root_cid,
+                                error = %e,
+                                "Failed to start Bitswap fetch for remaining child blocks"
+                            );
+                        }
                     }
                 }
             }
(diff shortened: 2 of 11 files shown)


fredcarle (commented):
## Review — harden bidirectional sync recovery

### Overview

Addresses four sync correctness gaps: Bitswap retry queries weren't registered (retry completions silently ignored), blocking `send_doc_sync_request` returned a hardcoded error, fire-and-forget DocSync/BranchableSync replies weren't bound to a pending request set, and `process_merge_batch` ignored both `rebroadcast_on_merge` and channel-send failures. New tests cover the bitswap-retry path, batch rebroadcast, and pending-request peer-binding semantics; the integration `document_sync_test` no longer silently skips on sync failure.

### Strengths

- Bitswap retry-query registration (`crates/p2p/src/sync/coordinator/event_handler/bitswap.rs:41-65`) closes a real loop-completion gap, exercised by a unit test that checks the second `register_query` actually fires.
- Channel-send failures surface as `Error::ChannelSend` instead of being dropped: `pending_dag.rs:413-435` even re-inserts the pending DAG so the entry isn't lost on failure.
- Splitting `doc_sync_channels` (oneshot) from `doc_sync_requests` / `branchable_sync_requests` (Instant-based pending sets) is the right model. TTL-based `prune_expired` keeps the sets bounded.
- New peer-binding tests in `handler/mod.rs:251-270` verify single-use semantics.
- Removing the silent `if sync_result.is_err() { return; }` early-return in `tools/integration-test/tests/p2p/sync.rs` turns the test into a real assertion.

### Issues

**Major (parity)** — `process_merge_batch` rebroadcast in `replication/handlers.rs:686-738` is gated by `config.rebroadcast_on_merge`, but **every caller in the live codebase sets it to `false`** (`server_p2p.rs:148, :503`, `recovery.rs:49`, `replication/mod.rs:811`, `config.rs:20`). On the Go side, `processPushlogRequest` (`internal/db/p2p/p2p.go:584-621`) rebroadcasts unconditionally after merge. The new code path is dormant in production and doesn't actually restore parity. Either flip the flag default or document a follow-up issue.

**Major** — `handler/doc_sync.rs:53-72`: `send_doc_sync_request` (blocking) returns the response on success but doesn't clean up the `doc_sync_channels` map entry. Slowly leaks memory if `(peer, message_id)` is reused. Note this method has no live callers (everything goes through the fire-and-forget variant via `host/command_handler/messaging.rs:139`); consider deleting it.

**Minor** — `handler/mod.rs:84-90`: `prune_expired` is only called inside `register_*` and `consume_*`. If a peer goes silent and no new requests are issued, expired entries aren't pruned. Bounded by traffic, not by time.

**Minor** — `handler/mod.rs:201-219`: `cleanup_pending_doc_sync` removes both the channel and the request entry, but `cleanup_pending_branchable_sync` only removes from the request set. Worth a one-liner comment so it isn't "fixed" later by mistake.

**Minor** — `bitswap.rs:41-64`: `register_query(retry_query_id, root_cid)` happens after `sync_blocks(...)` returns, leaving a tiny window where a Bitswap-complete event for the new ID arrives before registration. Pre-existing race (also in `replication/handlers.rs:170`), but the new code multiplies how often the path runs.

**Nit** — `bitswap_complete_retries_only_its_registered_root` reuses `QueryId(999)` from `TestTransport::sync_blocks`, making the registration assertion implicit. A more direct test would assert `pending_dag_attempts(&root1_cid) == 2` and that a second `register_query` actually occurred.

### Test coverage

- **Wins**: focused `test_process_merge_batch_rebroadcasts_when_config_enabled` (`replication/mod.rs:1510-1559`) and two peer-binding tests; bitswap retry test verifies the second registration.
- **Gap**: no test for the new "drop unknown DocSync/BranchableSync reply at inbound" path. Easy to add by feeding a synthetic reply with no matching pending entry through `handle_response_stream` and asserting `Ok(None)`.
- **Gap**: no test for the new `Err(Error::ChannelSend)` propagation from `pending_dag.rs:413-435` or `doc_sync.rs:309-322`. A test with a closed event channel would catch regressions.
- **Gap**: `document_sync_test` will now hard-fail on the `go_rust` topology if it was actually flaky. The previous comment ("Rust's sync_documents handler can block for up to 90s with retries") suggests a real underlying issue. PR's test command was `--no-run`; confirm `go_rust_document_sync_test` actually passes end-to-end.

### Verdict

Corrections are real and the new code is internally consistent. DocSync/BranchableSync inbound binding and Bitswap retry registration are clear improvements with appropriate unit tests. Two things to address before merge: (1) confirm `go_rust_document_sync_test` actually passes now that the silent skip is gone, and (2) clarify the `rebroadcast_on_merge` parity status — adding the batch-path implementation while leaving the flag forced to `false` everywhere doesn't actually close the Go-vs-Rust gap; a follow-up issue or a default flip should accompany this PR. Minor issues (memory hygiene, prune-on-idle, race ordering) can land as follow-ups.

Issue #105, opened 2026-01-19
Add pagination support for get_collection_doc_ids REST endpoint


jackzampolin:
## Summary

The `get_collection_doc_ids` REST endpoint (`GET /api/v0/collections/{name}`) returns all document IDs in a collection without pagination. For large collections, this could cause performance issues and excessive memory usage.

## Current Behavior

- Returns all document IDs in a single response
- No limit/offset parameters
- Located in `crates/query/src/rest.rs:148` and handler at `crates/http/src/handlers/collections.rs`

## Proposed Solution

Add pagination support with:
- `limit` parameter (default: 100, max: 1000)
- `offset` or `cursor` parameter for pagination
- Return metadata (total count, has_more, next_cursor)

## Example API

```
GET /api/v0/collections/Users?limit=50&offset=0

Response:
{
  "doc_ids": ["bae-123", "bae-456", ...],
  "total": 1500,
  "has_more": true,
  "offset": 0,
  "limit": 50
}
```

## References

- Go DefraDB pagination patterns (if any)
- REST API best practices for list endpoints

---
**Related upstream:** [sourcenetwork/defradb#4014](https://github.com/sourcenetwork/defradb/issues/4014) — SE pagination. General collection pagination may not have a dedicated Go issue yet.

PR #930, opened 2026-04-29
feat(http): paginate collection doc IDs

iverc:
## Summary

Addresses #105 by adding bounded pagination metadata to `GET /api/v0/collections/{name}` and wiring the missing GET route for collection document ID listing.

## Why

The collection document ID endpoint could return every ID in one response. That makes large collections expensive for clients and servers, and the Rust router was not exposing the existing handler through the expected GET route.

## Changes

| Area | Change |
|---|---|
| HTTP API | Add `limit` and `offset` query params with default `100` and max `1000` |
| Response | Return `doc_ids`, `total`, `offset`, `limit`, `has_more`, and `next_offset` |
| Router | Register `GET /api/v0/collections/{name}` for document ID listing |
| REST layer | Pass pagination through `RestOperations` and fetch `limit + 1` to detect more pages |
| Query/storage | Add paged fetch and count hooks so simple unfiltered reads avoid materializing full collections when safe |
| Tests | Cover handler pagination, invalid limits, route registration, and REST operation pagination/counting |

## Out of scope

- No cursor pagination.
- No pagination for filtered or ordered arbitrary GraphQL queries beyond the existing query support.
- No change to document CRUD routes.

## Test plan

- [x] `cargo test -p query`
- [x] `cargo test -p db`
- [x] `cargo test -p query-plan`
- [x] `cargo clippy -p query-plan -p query -p db -p defra-http --all-targets -- -D warnings`

### DIFF
--- a/crates/db/src/auto_commit_fetcher.rs
+++ b/crates/db/src/auto_commit_fetcher.rs
@@ -75,6 +75,79 @@ impl<S: Store + 'static> DocFetcher for AutoCommitFetcher<S> {
         result
     }
 
+    async fn get_all_page(
+        &self,
+        collection_name: &str,
+        limit: Option<u64>,
+        offset: u64,
+    ) -> query::error::Result<Vec<Document>> {
+        let collection = self
+            .db
+            .get_collection(collection_name)
+            .map_err(|e| query::error::QueryError::execution(format!("db error: {}", e)))?
+            .ok_or_else(|| query::error::QueryError::collection_not_found(collection_name))?;
+
+        let txn = self.db.new_txn(true).await.map_err(|e| {
+            query::error::QueryError::execution(format!("failed to create txn: {}", e))
+        })?;
+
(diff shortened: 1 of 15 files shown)


jackzampolin:
Closing this draft because we want the Rust pagination work to mirror the Go GraphQL cursor-pagination direction, not land a Rust-specific REST limit/offset response shape.

Go reference: sourcenetwork/defradb#4617, "feat: Cursor based query pagination".

Target Rust direction:
- Implement GraphQL cursor pagination via _cursor.
- Support first/after and last/before on the inner collection query.
- Return _pageInfo with hasNext, hasPrev, startCursor, and endCursor.
- Encode opaque cursor tokens using the boundary document ID plus ordered key values when needed for index-backed seeks.
- Keep REST collection-doc-ID pagination out of scope unless we later choose to match Go SSE behavior separately.

The exploration in this PR was useful for route and storage considerations, but the API shape should align with Go before landing.

PR #951, opened 2026-05-08
fix: port Go transaction correctness fixes

jackzampolin:
## Summary
- propagate transaction headers through REST schema-add, migration-set, CLI schema-add, and CLI lens-set so txn-local schema/migration state stays scoped until commit
- remove the separate concurrent transaction endpoint/CLI surface
- serialize top-level actions on a shared transaction handle and wait for in-flight actions before commit/rollback/cleanup
- add txn-scoped update-with-filter regression coverage
- avoid carrying an already-resolved update filter into post-update plan execution, with update/upsert result regressions

## Go SHAs Ported
- 6bc16f7e Fix broken transaction behavior (#4587)
- 6123e549 Remove 'concurrent' transaction concept (#4692)
- 630f4f30 Prevent txn-actions from executing concurrently (#4763)
- aac9b5aa Fix transactional behavior in updateWithFilter (#4643)
- 3c911995 Remove duplicate filter on update/upsert operations (#4283)

## Verification
- cargo fmt --all -- --check
- cargo clippy --all -- -D warnings
- cargo test -p db
- cargo test -p query
- cargo build -p cli
- PATH="/Users/johnzampolin/go/src/github.com/sourcenetwork/defradb/build:$PATH" cargo test -p integration-test --test basic --test acp --test query -- --test-threads=1

## Additional Workspace Check
- Attempted full `cargo test --all` with local `hubd`, `sourcehubd`, and Go `defradb` binaries configured. The run hit an integration harness port collision in `integration-test --test acp`; the exact failed case was rerun and passed: `cargo test -p integration-test --test acp relationship::rust_acp_rel_add_dummy_defined_is_noop -- --exact`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9963,6 +9963,7 @@ version = "0.5.0"
 dependencies = [
  "acp",
  "async-graphql",
+ "async-lock",
  "async-trait",
  "base64 0.22.1",
  "blockstore",
--- a/crates/cli/src/commands/client/collection/mod.rs
+++ b/crates/cli/src/commands/client/collection/mod.rs
@@ -131,7 +131,7 @@ impl CollectionAddArgs {
             .with_auth_token(ctx.auth_token.clone())
             .with_verbose(ctx.verbose);
 
-        let result = client.schema_add(&sdl).await?;
+        let result = client.schema_add(&sdl, ctx.tx_id.as_deref()).await?;
         println!("{}", serde_json::to_string_pretty(&result)?);
         Ok(())
     }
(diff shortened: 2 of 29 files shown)


jackzampolin:
Heads-up from #953: this PR and the CLI rename slice both touch `crates/cli/src/commands/client/tx.rs`. Whichever PR lands second should rebase and confirm the txn slice removes `TxNewArgs.concurrent`, the `tx_begin_concurrent` branch, and the related `test_tx_new_args_default` references after the rename.

PR #952, opened 2026-05-08
Port Go query planner correctness fixes

jackzampolin:
## Summary

Ports the seven Go query-engine correctness fixes from the sync plan:

- `09c050f4` / #4656: multiple filters as AND on inverted joins. Rust already behaved correctly; added regression coverage.
- `9590e105` / #4684: secondary index edge cases for subscriptions with CID filters and inverted 1-to-many ordering. Rust already behaved correctly; added regression coverage.
- `9f9d7de7` / #4633: fall back to scan node for unsupported index filters. Added planner fallback fix plus regression coverage.
- `31a03b77` / #4454: fetch orphan parent docs with `@exhaustive`. Rust already had the needed `@exhaustive` behavior; added regression coverage.
- `aec4717b` / #4657: prevent document writes after collection deletion. Added active-collection write guards plus regression coverage.
- `b486aecd` / #4712: multi-layered self-referencing relations. Fixed SDL relation resolution plus regression coverage.
- `7d011d6e` / #4709: patch new fields on collections with secondary relations. Rust already behaved correctly; added regression coverage.

## Verification

- `cargo fmt --all`
- `cargo clippy --all -- -D warnings`
- `cargo test -p query`
- `cargo test -p db`
- `PATH="/Users/johnzampolin/go/src/github.com/sourcenetwork/defradb/build:/Users/johnzampolin/go/bin:$PATH" cargo test -p integration-test --test query --test basic`

### DIFF
--- a/crates/db/src/auto_commit_mutator/batch.rs
+++ b/crates/db/src/auto_commit_mutator/batch.rs
@@ -11,6 +11,7 @@ use std::sync::Arc;
 use storage::corekv::Store;
 use tracing::warn;
 
+use super::helpers::ensure_collection_is_active;
 use crate::block_builder::{write_collection_block, write_delete_block, write_document_blocks};
 use crate::collection_loader::{get_collection_with_index_manager, get_collection_with_lazy_load};
 use crate::database::DB;
@@ -148,6 +149,7 @@ impl<S: Store + 'static> DocMutator for BatchMutator<S> {
     ) -> query::error::Result<CreateResult> {
         let (collection, datastore, index_manager) =
             get_collection_with_index_manager(&self.txn, collection_name).await?;
+        ensure_collection_is_active(&self.db, collection_name, &collection)?;
         let embedding_config = self.db.options().embedding_config();
 
         db_search::set_embedding(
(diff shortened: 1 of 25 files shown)


jackzampolin:
Addressed the review follow-ups in the latest force-push.

For item 6: `rg -n "mark_collection_created" crates/db/src/` shows the marker is only added in `create_collection_with_txn`, and the non-direct create path (`txn_registry::add_schema_in_txn`) calls that same helper. The JSON patch path does not create collections: unknown-root `add`/`copy`/`replace` returns `adding collections via patch is not supported`, so there is no separate patch-created collection path that needs marking. The `rust_txn_schema_add_and_mutate` regression still covers the supported “create collection inside txn, then write doc inside same txn” path.

For item 7: the Go #4712 fixture’s expected `SelfKind.RelativeID`s match the Rust regression exactly (`Domain -> 1`, `Route -> 0`, `Route.after -> 1`). I also reran `cargo test -p query-parse -p schema`; that includes the existing self-ref collection ID parity tests against Go (`test_self_ref_collection_id_matches_go`, `test_self_ref_complex_collection_id_matches_go`) plus schema relation/property coverage. The full `basic`/`query` integration command had already passed before this review, and I reran the full `query` integration target after the test polish.

Nits handled:
- Tightened the #4633 error assertion to the stable `unexpected type. Property: condition, Actual: bool` text.
- Reworked the #4684 subscription test to wait for the SSE connection and exactly one event with env-tunable timeouts instead of fixed sleeps.
- Left `/api/v0` paths unchanged for now, as noted.

PR #961, opened 2026-05-15
feat: cursor-based GraphQL pagination (ports Go #4617, FFI-validated)

jackzampolin:
## Cursor-based GraphQL pagination

Ports Go DefraDB PR #4617 — `_cursor { Collection(first/after/last/before, …) { … } _pageInfo { hasNext hasPrev startCursor endCursor } }` with opaque, index-seekable cursor tokens.

### Status — DRAFT, kept open intentionally
Held open as draft until the final cursor version lands on the Go side: **Go #4617 is still open** (base `shinzo-develop`; not yet on mainline `develop`). Parity here is anchored to `jsimnz/feat/cursor-pagination-develop-rebase` (`03f720a5`) — note that is *not* the branch GitHub's PR-head pointer shows; the `…-v2` head is a superseded sibling. Will re-validate if `develop-rebase` changes before it merges upstream.

### Validation
Reviewed adversarially, fixed, and **cross-checked against Go #4617's own integration suite via the FFI harness: 74/74 passing, zero regressions** (Go's tests execute against the Rust engine through `rust_ffi`).

### Covered
- Forward/backward pagination; opaque base64url codec **byte-for-byte identical to Go**.
- Index fast-path seek (Go-exact `indexFetches`) + naive slow-path fallback.
- `_pageInfo` canonical field names; Go-exact error strings; SDL generation.

### Fixes from review (vs the initial port)
- `first`/`last` omitted → returns all rows (was 0).
- `after`/`before`: empty-string rejected, `null` treated as no-cursor.
- Slow path rewritten to Go's exact DocID-equality boundary (legacy DocID-only + field-ordered).
- `_pageInfo` aliases discarded → canonical names; fragments expand.
- Cursor index scan bounded to `page_size+1` for Go-exact `indexFetches`.
- Codec emits the `o` direction field; fixtures regenerated from real Go output.

### Known scoped divergence
- `groupBy` not advertised on the cursor field's generated SDL (no `<Type>Field` enum in the Rust generator; functional parsing unaffected, no Go test exercises it).

### Follow-up
- Order-aware/cost-based index selection for *filtered* cursor pagination — #1005 (mirrors Go's deferred #2680).


jackzampolin:
All 5 review findings addressed. Pushed 5 commits on top of `56dfe664`:

| Finding | Commit | What changed |
|---|---|---|
| P1.1 (backward seek direction) | `d400774f` | `RangeIterator::apply_cursor_seek` now takes `reversed` and overrides `self.reverse`; `apply_cursor_seek_to_iterator` passes `seek.reversed` through. New `test_cursor_seek_reversed_param_controls_upper_bound` test. |
| P1.2 (duplicate-key drop) | `6e8babd8` | `build_cursor_seek_key` appends `cursor.doc_id.as_bytes()` to the seek key for non-unique indexes so forward-exclusive boundaries exclude only the boundary doc, not every doc sharing its field values. New integration test `rust_cursor_paginates_through_duplicate_index_keys` (4 users at `age=30`, page 2 was empty before, now returns the remaining 2). |
| P1.3 (false fast-path signal) | `30b0774a` | `IndexScanNode::set_cursor_seek` now pattern-matches `scan_type`: returns `true` only for `PrefixScan`/`RangeScan`; `ExactMatch`/`InScan`/`OrScan` return `false` so `CursorNode` falls back to the slow skip-until-after path. 5 new unit tests cover both paths. |
| P2.1 (silent `limit`/`offset` acceptance) | `8fa19e65` | `extract_cursor_params` explicitly rejects `limit`/`offset` with `Unknown argument "limit" on field "User".`-style error. 2 new parser tests. |
| P2.2 (no type normalization) | `63b401a1` | `build_cursor_seek_key` now applies `normalize_for_index_field` after `json_to_normal_value` so `Float64` from JSON → `Float32` on a Float32-typed index field before encoding. The helper was promoted from `pub(super)` to `pub(crate)` and re-exported. New unit test verifies byte-level equality with the index's encoding. |

Test summary after fixes (all green):
- `cargo test -p storage` — 272 pass (+1)
- `cargo test -p query-plan` — 105 pass (+5)
- `cargo test -p query-parse query_parse::cursor` — 10 pass (+2)
- `cargo test -p query-plan planner::builder::cursor` — 8 pass (+1)
- `cargo test -p integration-test --test cursor` — 13 pass (+1)
- `cargo clippy --all -- -D warnings` — clean

jackzampolin:
Round 2 review findings addressed. 4 commits pushed:

| Finding | Commit | What changed |
|---|---|---|
| P1.1 (reversed flag semantics + post-hoc direction flip impossible) | `d53edd41` | `CursorSeek.reversed` now means "cursor's desired iteration direction": Forward = `order_reversed`, Backward = `!order_reversed`. `IndexScanNode::set_cursor_seek` mutates the underlying `IndexScanType::{PrefixScan,RangeScan}.reverse` field BEFORE the KV iterator is created via `scan_prefix(..., *reverse)`. `RangeIterator::apply_cursor_seek` no longer pretends it can flip direction post-creation — `self.reverse = reversed` removed and replaced with a `debug_assert_eq!`. |
| P1.2 (`before` was inclusive on fast path) | `d53edd41` (same commit) | `CursorSeek.inclusive` is now always `false` — both `after` and `before` are exclusive boundaries. Regression test `rust_backward_before_excludes_boundary` exercises 4-user dataset with `last: 2, before: cursor(c)` and verifies `[a, b]` (boundary `c` excluded). |
| P1.3 (slow path compared only doc_id) | `01f5acf1` + `8bca6a68` | Slow paths now error explicitly with `cursor slow path (no index seek) does not support non-docID ordering` rather than silently using a wrong comparator. Built a full-tuple comparator would require named-field access on the index-based `Doc` type and got punted in favor of explicit failure. The guard is gated on `before.is_some()` so `last: N` (no `before`) still works for any order. Forward path's `SkippingUntilAfter` state is only entered when `after.is_some()`, so it's already correctly gated. |
| P2 (unique composite prefix dropping ties) | `e02a0efc` | `configure_scan_for_cursor` now falls back to slow path when `idx.unique && cursor.keys.len() < idx.fields.len()`. For unique indexes, doc_id lives in the value, not the key, so appending it to seek_key doesn't help. The slow-path fallback is honest. Same fallback handles the unique-null case at `unique.rs:131` pragmatically. Integration test `rust_unique_composite_prefix_falls_back_to_slow_path` verifies. |

### Known limitation

Cursor pagination through ExactMatch / InScan / OrScan scans with non-docID ordering (i.e., the P1.3 + non-docID-order combination on the slow path with a cursor boundary) now errors out instead of silently mispaginating. A full tuple comparator that handles arbitrary order direction + nulls + types would unblock this and is a worthwhile follow-up. Tracked as a known limitation in this PR; the error message is actionable ("ensure the collection has an index covering the ORDER BY fields").

### Test summary after round-2 fixes

- `cargo test -p storage` — clean (debug_assert in place)
- `cargo test -p query-plan` — 108 pass (+3 since round 1: 2 new `set_cursor_seek_propagates_reversed_into_{prefix,range}_scan` tests + 1 `slow_path_supports_no_cursor_with_non_doc_id_order` test)
- `cargo test -p integration-test --test cursor` — **15 pass** (+2 since round 1: `rust_backward_before_excludes_boundary` + `rust_unique_composite_prefix_falls_back_to_slow_path`)
- `cargo clippy --all -- -D warnings` — clean

jackzampolin:
Round 3 review findings addressed in `24bc712d`:

| Finding | Fix |
|---|---|
| P1 (`_docID DESC` silently paginates ASC) | Added `doc_id_order_direction(order_fields)` helper. Forward `SkippingUntilAfter` now compares `row > after` for ASC and `row < after` for DESC. Backward `populate_backward_buffer_slow` mirrors with `>= boundary` for ASC and `<= boundary` for DESC. 2 new integration tests cover both forward and backward DESC pagination. |
| P2 (unique-null rows still drop tier) | Changed seek-key doc_id-suffix condition from `!idx.unique` to `!idx.unique \|\| cursor_has_null_value`. When any cursor key is null, the on-disk unique index appends doc_id (per `unique.rs:228` `has_nil_field`), so the seek key must too. |
| Test gap (unique-composite-prefix only covers page 1) | Added `rust_unique_composite_prefix_after_returns_slow_path_error` which sends page 2 with `after` and asserts the actionable slow-path error message. |

### Test summary

- `cargo test -p query-plan` — 108 pass (unchanged; no new unit tests this round)
- `cargo test -p integration-test --test cursor` — **18 pass** (+3 since round 2: 2 _docID DESC + 1 unique-prefix page-2 error)
- `cargo clippy --all -- -D warnings` — clean

jackzampolin:
Round 4 review findings addressed. 3 commits pushed:

| Finding | Commit | What changed |
|---|---|---|
| P1.1 (seek applied to wrong index) | `f0688aef` | Added `expected_index_name: String` to `CursorSeek`. `IndexScanNode::set_cursor_seek` now rejects (returns `false`) when `self.index_params.index_name != seek.expected_index_name`. New unit test `set_cursor_seek_returns_false_for_mismatched_index_name`. New integration test `rust_cursor_with_filter_and_order_uses_correct_index` exercises filter-index + order-index conflict — seek is rejected, CursorNode falls back to slow path, and (per round 2's actionable error) the slow path errors out for non-docID order with a cursor token. The garbage-rows bug is gone. |
| P1.2 (rustfmt CI red) | `244fdd93` | `cargo fmt --all` cleared the two cursor/smoke.rs offenders. |
| P2.1 (unique-prefix fallback trusts arbitrary keys) | `f0688aef` (same commit as P1.1) | Both `cursor_index_keys_count` and `has_null_value` now intersect cursor keys with the index's field-name set, ignoring untrusted extras. |
| P2.2 (_pageInfo aliases + unknown fields) | `9feb9e9e` | `CursorPageInfoFields` switched from `bool` to `Option<String>` per field, storing the output key (alias or canonical). Parser records aliases and errors on unknown fields with `"Cannot query field \"X\" on type \"PageInfo\"."`. Response shaping emits each field under its stored key. 2 new parser tests + 1 new integration test (`rust_page_info_aliases_preserved_in_response`) verify aliases are preserved AND unknown fields are rejected. |

### Test summary

- `cargo test -p query-types` — 131 pass
- `cargo test -p query-parse` — 173 pass (+2 new alias/unknown tests)
- `cargo test -p query-plan` — 109 pass (+1 new mismatched-index test)
- `cargo test -p integration-test --test cursor` — **20 pass** (+2: multi-index + alias preservation)
- `cargo clippy --all -- -D warnings` — clean

jackzampolin:
Round 5 review findings addressed in `3b0ad7fd`:

| Finding | Fix |
|---|---|
| P2 (`_pageInfo` wrapper alias dropped) | Added `page_info_alias: Option<String>` to `CursorAliases`. Parser records `child.alias.clone()` when matching `_pageInfo`. Response shaping at `nested.rs` now uses `select.cursor_aliases.page_info_alias.as_deref().unwrap_or("_pageInfo")` for the output key. New integration test `rust_page_info_wrapper_alias_emitted_in_response` verifies `info: _pageInfo { hasNext }` surfaces as `info`, not `_pageInfo`. |
| P3 (fragments inside `_pageInfo` silently ignored) | `parse_page_info_selection` now explicitly returns `Err("_pageInfo block does not support fragments")` for `FragmentSpread`/`InlineFragment`, mirroring the outer `_cursor` rejection. New parser test `fragments_in_page_info_rejected`. |

### Test summary

- `cargo test -p query-parse query_parse::cursor` — 14 pass (+2 round-5 tests)
- `cargo test -p query-types` — 131 pass
- `cargo test -p integration-test --test cursor` — **21 pass** (+1: `rust_page_info_wrapper_alias_emitted_in_response`)
- `cargo clippy --all -- -D warnings` — clean

### Note on flake

Saw one transient failure of `rust_page_info_aliases_preserved_in_response` in a full suite run that did not reproduce on rerun (passes in isolation and in subsequent full runs). Stack trace points at `test-infra/src/log_tracker.rs:84` (a transitive harness dep), not cursor code. Not blocking this PR; worth filing a separate harness issue if it recurs.

jackzampolin:
Round 6 fixes pushed in `a681a418`. **CI-red regression resolved.**

| Finding | Fix |
|---|---|
| P1 (CI-red: `exhaustive_orphans_4454::rust_exhaustive_relation_order_includes_orphans` regressed at round-5) | **Root cause was stack overflow on the tokio worker thread**, confirmed via server's `stderr.log`: `thread 'tokio-rt-worker' has overflowed its stack`. Round-5 added `page_info_alias: Option<String>` to `CursorAliases`, which sits on every `Select` — that's +24 bytes propagated through the deeply-nested async future tree built for `@exhaustive` relation-order queries (TypeJoinMany over Book → Publisher). Boxed the field to `Option<Box<String>>` (+8 bytes vs +24) and gated the `plan.page_info()` call on `select.is_cursor` for belt-and-suspenders. Test passes; cursor suite still 21/21. |
| P2 (unique composite-prefix guard trusted cursor.keys) | Switched to `order_fields.len() < idx.fields.len()` — matches what `build_cursor_seek_key` actually encodes. Cursor tokens are client-controlled; the guard now ignores them. |
| P3 (multiple `_pageInfo` siblings overwrite) | Parser errors with `"_cursor block cannot contain multiple _pageInfo selections"`. New test `multiple_page_info_selections_rejected`. |

### How I found P1

Local repro confirmed the test fails on `3b0ad7fd` (round-5 head) and passes at the merge base. Bisected — `3b0ad7fd` first-bad. Diff was contained to cursor paths only, so I suspected stack-frame growth. Set `DEFRA_E2E_KEEP=1` to retain the test artifacts, found `target/e2e/.../rust-0/logs/stderr.log` saying `thread 'tokio-rt-worker' has overflowed its stack; fatal runtime error: stack overflow, aborting`. Cross-referencing the diff size with the failure, the smoking gun was `CursorAliases` growing 24 bytes. Boxing it confirmed the fix.

### Test summary

- `cargo test -p integration-test exhaustive_orphans_4454::rust_exhaustive_relation_order_includes_orphans` — **PASS** (regression fixed)
- `cargo test -p integration-test --test cursor` — 21/21 pass
- `cargo test -p query-parse query_parse::cursor` — 15/15 pass (+1 new multiple-_pageInfo test, see file)
- `cargo clippy --all -- -D warnings` — clean

jackzampolin:
### Upstream Go status snapshot (2026-05-18)

Mirrored Go PR [`sourcenetwork/defradb#4617`](https://github.com/sourcenetwork/defradb/pull/4617) is **still open**, targeting `shinzo-develop` (not `develop`). Latest activity 2026-05-12 — author addressing islamaliev's `CHANGES_REQUESTED` review from 2026-03-24. No new commits in the 6 days before this PR was opened, and no movement since.

**Open Go design concerns that affect us:**

1. **Index requirement on ordered-by field** — islamaliev's top-level question: *"if we just request docs in their natural order why can't we use cursor? We can also specify prefix for such queries after all."* This PR mirrors Go's current behavior (hard error on missing index for explicit non-docID orderings; docID fallback when ordering is empty). If Go relaxes this, we'd need to follow.

2. **`filter` / `groupBy` test coverage** — islamaliev flagged Go has no test covering cursor + filter/groupBy combinations and suggested a test multiplier like secondary-index. This Rust port has the same gap.

**Open inline threads on Go side (no `jsimnz` reply):**

- `internal/db/fetcher/indexer.go` — index iterator excludes `_in`/`_or` (`multiIndexIterator`) and arrays/JSONs (`memorizingIndexIterator`)
- `internal/planner/planner.go` — `validateCursorIndex` may be redundant with `CanBeOrderedByIndex`

**Pushed-back but unresolved:**

- `internal/db/fetcher/wrapper.go` — pointer type ("I'll investigate further")
- `internal/cursor/cursor.go` — typed map vs `map[string]NormalValue` (jsimnz declined)

**Merge consideration:** the longer this Rust PR sits behind a moving Go target, the more drift risk. Could merge now and chase Go-side changes via follow-up issues, or wait until Go #4617 merges — open question.

jackzampolin:
## Dogfooding report: `after` continuation breaks on a reindexed index (forward + backward)

We're trialing this PR downstream (rebased onto `v0.13.1` so it carries the tx-commit Update-event fix; pinned that combined rev in our app) against a real dataset — ~4,540 `CodingSession` docs, ordering by an `@index(direction: DESC)` `created_at` field. First page is great. **The `after`/`before` continuation mis-resolves the boundary**, so you can't actually page through the corpus.

### Repro
Order **DESC**, page size 5:
```graphql
{ _cursor { CodingSession(first:5, order:[{created_at: DESC}]) { _docID created_at } _pageInfo { hasNext endCursor } } }
```
- page1 → 5 rows, last `created_at = 2026-05-29T13:06:28Z`, `hasNext = true`
- decode(endCursor) = `{"d":"bae-d45da628-…","k":{"created_at":"2026-05-29T13:06:28Z"},"o":""}`
- page2 with `after: "<that endCursor>"` → **only 2 rows, then `hasNext = false`** — it abandons the ~4,000 older rows instead of continuing.

Order **ASC**, page size 5 (same query, `created_at: ASC`):
- page1 → last `created_at = 2026-01-05T18:05:40Z`, `hasNext = true`
- page2 with `after: endCursor` → 5 rows, but **first row `created_at = 2026-05-28T23:30:43Z`** — it jumps far past the boundary instead of resuming just after `2026-01-05`.

So forward DESC truncates early and forward ASC seeks to the wrong position. Both point at the `after` boundary/seek, not the codec (the token round-trips and decodes cleanly).

### Two clues
1. The encoded cursor's order-direction field is **always empty** (`"o":""`) regardless of the query's `order:` direction. If the seek reconstructs the scan direction/boundary from the token, an empty `o` could explain it resolving the wrong half-open range.
2. **It's index-state sensitive.** On a *freshly built* `created_at` index this paginated correctly for many pages; the breakage above appeared **after a `maintenance reindex`** of the collection (we'd rebuilt the index after coercing some `created_at` values to the `DateTime` scalar). So the `after` index-seek seems fragile to the reindexed index layout / the index-fetch boundary, where the naive path might not be.

### Ask
Is `after` expected to hold across a `reindex`, and is the empty `"o"` direction in the token intentional? Happy to capture more (full `explain`, the exact index entries around the boundary) if useful — flagging since this blocks using the cursor for anything past page 1 on a non-trivial collection. (We've reverted to offset pagination downstream meanwhile.)

jackzampolin:
Thanks for the fast turnaround on `86204e8f` — and nice catch on the actual root cause. That lines up exactly with what we saw from the outside: the breakage was **index-state-sensitive** (paginated fine on a freshly-built index, broke only after a `maintenance reindex`), which is precisely the reindex path emitting `DateTime` as `String` and producing byte-different index keys from the write path. The two converters drifting is a much better explanation than my "empty `o` direction" guess — that field was a red herring.

Unifying both paths behind one schema-kind converter (`json_to_normal_value_for_kind`) is the right shape, and the byte-equivalence regression guard across both index directions is exactly the test I'd want for this class of bug — it'll catch any future drift deterministically without needing the full lens-reindex e2e harness.

We'll re-pin to the fixed rev downstream and re-run our original repro (the DESC-truncates / ASC-mis-seeks pagination over ~4,540 rows after a reindex) to confirm end-to-end on real data, and report back here. Also noted your flagged follow-up — the seek-key builder (`values.rs::json_to_normal_value`) still inferring type from JSON shape rather than schema kind — agreed that's a latent footgun for a `String` field holding date-like text; worth its own issue. Appreciate the thorough fix. 🙏

PR #988, opened 2026-05-20
fix(db): clean up idle HTTP transactions

### LINKED ISSUE #986: Explicit orphaned-HTTP-transaction cleanup policy
Surfaced by an audit of defra-agent's `config apply` transaction lifecycle. Audit doc (defra-agent side): https://github.com/sourcenetwork/defra-agent/blob/main/docs/superpowers/audits/2026-05-20-defradb-tx-idle-timeout-audit.md

## What the audit found

The defradb.rs transaction registry exposes `cleanup_stale_transactions(max_age)` as a manual API, but no production caller was found in the pinned backend. Today, an orphaned HTTP transaction handle (e.g. client crash between `POST /tx` and `POST /tx/{id}/commit`) is bounded only by per-request HTTP timeout (30s default) and connection drop detection. There is no active GC sweep keyed off transaction idleness.

## Why this matters

Long-lived operator workflows (large config apply, slow remote DefraDB) can legitimately hold a transaction open across many in-tx requests, each well under the per-request timeout. A crashed orchestrator between requests leaves the handle dangling until some future per-request timeout fires for a request that's never coming.

In normal operation this is benign (transaction atomicity makes the apply invisible). But over many orchestrator crashes the registry can accumulate stale handles indefinitely.

## Suggested shape

A. Promote cleanup_stale_transactions to a configurable server-side sweep with a default like max_idle_age = 600s.

B. Track per-tx "last request seen" timestamp and reclaim on idle, not on creation age.

C. Document the current behavior explicitly and treat orphaned-handle accumulation as the operator's responsibility.

## Related

- Downstream audit (defra-agent): docs/superpowers/audits/2026-05-20-defradb-tx-idle-timeout-audit.md
- Downstream issue that surfaced this: sourcenetwork/defra-agent#226

jackzampolin:
Closes #986.

## Summary
- Track explicit transaction last-request timestamps and clean up by idle age instead of creation age.
- Add a periodic stale transaction cleanup worker with a 600s default idle timeout and 60s sweep interval.
- Wire cleanup into CLI and defra-node HTTP startup, with CLI config/flags and tests.

## Verification
- `cargo test -p db test_cleanup -- --nocapture`
- `cargo test -p db test_periodic_cleanup -- --nocapture`
- `cargo test -p cli --test start_tests -- --nocapture`
- `cargo check -p defra-node --features http`
- `cargo test -p defra-node --features http http_config_accepts_extra_routes -- --nocapture`
- `cargo check -p defra-node`
- `git diff --check`

### DIFF
--- a/crates/cli/src/commands/start/mod.rs
+++ b/crates/cli/src/commands/start/mod.rs
@@ -184,6 +184,14 @@ pub struct StartArgs {
     #[arg(long)]
     pub query_timeout: Option<u64>,
 
+    /// Max idle age for explicit HTTP transactions in seconds (0 = disabled, default: 600)
+    #[arg(long)]
+    pub transaction_idle_timeout: Option<u64>,
+
+    /// Interval between explicit HTTP transaction cleanup sweeps in seconds (default: 60)
+    #[arg(long)]
+    pub transaction_cleanup_interval: Option<u64>,
+
     /// Max GraphQL selection nesting depth (0 = unlimited, default: 20)
     #[arg(long)]
     pub query_max_depth: Option<usize>,
@@ -487,6 +495,12 @@ impl StartArgs {
         if let Some(timeout) = self.query_timeout {
             config.api.query_timeout = timeout;
         }
+        if let Some(timeout) = self.transaction_idle_timeout {
+            config.api.transaction_idle_timeout = timeout;
+        }
+        if let Some(interval) = self.transaction_cleanup_interval {
+            config.api.transaction_cleanup_interval = interval;
+        }
         if let Some(depth) = self.query_max_depth {
             config.api.query_max_depth = depth;
         }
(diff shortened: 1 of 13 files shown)


jackzampolin:
Addressed the review polish in ba55f120:

- Unified cleanup timing defaults through db-level constants.
- Changed zero sweep interval handling to a single non-zero policy: CLI validates, defra-node build errors, registry asserts.
- EmbeddedNode now stores the cleanup JoinHandle and awaits aborted cancellation during shutdown.
- Added a cleanup race invariant comment around the final write-lock re-check/remove.
- Restored the orphaned-handle/leaked-guard hint in the cleanup log message.
- Added rustdoc for the defra-node HTTP transaction cleanup builders.
- Relaxed the periodic cleanup test timing.

Re-ran:
- `cargo test -p db test_cleanup -- --nocapture`
- `cargo test -p db test_periodic_cleanup -- --nocapture`
- `cargo test -p cli --test start_tests -- --nocapture`
- `cargo check -p defra-node --features http`
- `cargo test -p defra-node --features http http_config_accepts_extra_routes -- --nocapture`
- `cargo check -p defra-node`
- `git diff --check`

PR #992, opened 2026-05-26
feat(rest): paginate collection doc id listing

### LINKED ISSUE #105: Add pagination support for get_collection_doc_ids REST endpoint

iverc:
## Summary

Closes #105 by adding paginated REST support for listing document IDs from a collection.

## Why

`GET /api/v0/collections/{name}` was not available in Rust, and the REST operations layer only exposed an all-doc-ID listing. Large collections need a bounded response shape so clients can page through IDs without forcing a single unbounded query/result into memory.

## Changes

| Area | Change |
|---|---|
| REST operations | Add `CollectionDocIdsPagination` and `CollectionDocIdsPage` to represent bounded doc-ID listings |
| Query-backed REST | Add a paginated implementation using GraphQL `limit`/`offset` plus a same-identity `COUNT` query for total metadata |
| HTTP handler | Add `GET /api/v0|v1/collections/{name}` with `limit` default `100`, max `1000`, and `offset` default `0` |
| Response shape | Return `doc_ids`, `total`, `has_more`, `offset`, and `limit` |
| Router/docs | Wire the route for both API versions and update REST endpoint docs |
| Tests | Cover route availability, default pagination, paginated metadata, and max-limit validation |

## Out of scope

- No cursor-based pagination.
- No change to GraphQL pagination behavior.
- No change to document CRUD routes.
- No change to REST collection listing at `GET /collections`.

## Test plan

- [x] `cargo fmt --all`
- [x] `git diff --check`
- [x] `cargo test -p defra-http collection_doc_ids`
- [x] `cargo test -p query rest::`
- [x] `cargo clippy -p defra-http -p query --all-targets -- -D warnings`

### DIFF
--- a/crates/http/src/handlers/collections.rs
+++ b/crates/http/src/handlers/collections.rs
@@ -12,6 +12,7 @@ use axum::{
     http::StatusCode,
     Json,
 };
+use query::rest::{CollectionDocIdsPage, CollectionDocIdsPagination};
 use serde::Serialize;
 
 use crate::error::{http_error_from_backend_message, HttpError};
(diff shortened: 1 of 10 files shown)


jackzampolin:
Post-merge follow-ups from review:

**API / forward-compat**
- Consider an opt-in `?include_total=true` (or default-off) — COUNT currently runs on every page, which is expensive on large collections. Issue #105 asked for `total` so it stays available, but an opt-out would help paging workloads.
- When #961 (cursor pagination) lands in GraphQL, this REST endpoint will likely want a `cursor` parameter; response shape is forward-compatible (extra fields), but the request shape isn't easily evolvable without a v2.

**Test gaps**
- `limit=0` rejection
- Negative offset / non-numeric offset
- `offset >= total` returning empty `doc_ids` with `has_more=false`
- Empty collection (total=0)
- `has_more` boundary (last page exact fill vs partial)
- ACP-filtered listing — confirm a non-owner sees a smaller `total`; identity is plumbed but not asserted at integration level
- Unknown collection → 404

**Nits**
- Default trait impl of `get_collection_doc_ids_page` in-memory slices the full doc-id list, defeating pagination. Worth a comment noting it's test-only.
- `query/src/rest/operations.rs` approaching the 400-line guideline.
- Two `identity.cloned()` calls (count then list) could collapse to one.

**Go parity note**
- Go has no `GET /collections/{name}` route — this is a Rust-specific REST surface, not a Go parity item.

PR #993, opened 2026-05-26
fix(p2p): use CBOR for gossip broadcasts

### LINKED ISSUE #876: Investigate and harden Iroh gossip wire compatibility and decode failures
## Summary
Live Iroh logs show repeated gossip decode failures of the form `Failed to decode gossip message (version skew or malformed sender?)` with `Hit the end of buffer, expected more data`. Peers still move data through slower fallback paths, but the fast gossip path remains degraded.

## Evidence
- Iroh gossip receive path decodes `PushLogBroadcast` with `postcard` in `crates/p2p/src/iroh/endpoint_commands.rs`.
- Iroh gossip publish path also serializes with `postcard` in the same file.
- The general message layer is documented as CBOR-oriented for wire compatibility in `crates/p2p/src/message/mod.rs`.
- `PushLogBroadcast` currently includes optional `ACPActorRelationships`, which increases the chance of binary-schema drift causing postcard decode failures across mixed builds.

## Problem
The transport can connect, but gossip remains partially unusable across some peer combinations. This increases replication latency, hides compatibility regressions, and makes restarts/network churn look worse than they are.

## Proposed Work
- Reproduce the decode failures with mismatched binaries or schema drift.
- Decide whether Iroh gossip should remain postcard-based or move to a more compatibility-friendly envelope.
- If postcard remains, add explicit versioning / compatibility guards and tests for schema evolution.
- Add higher-signal diagnostics so operators can see whether failures are due to binary mismatch, malformed payloads, or unsupported fields.
- Add regression tests for mixed-version or intentionally mismatched payload handling.

## Acceptance Criteria
- Root cause of the decode failures is identified and documented.
- The Iroh gossip path either becomes wire-compatible across supported peers or fails with a precise, diagnosable reason.
- Regression tests cover the identified failure mode.

## Motivation
Downstream apps can add reconnect logic, but they cannot compensate for a permanently degraded gossip fast path.

iverc:
## Summary

Closes #876 by making CBOR the canonical PushLog gossip payload encoding across libp2p and Iroh, while keeping legacy postcard decoding for older Iroh peers.

## Why

Iroh gossip was still publishing postcard-encoded `PushLogBroadcast` payloads. Postcard is compact, but it is schema-position dependent, so optional fields such as `ACPActorRelationships` make mixed-build gossip payloads more likely to fail with opaque decode errors like `Hit the end of buffer, expected more data`.

The shared P2P message layer already treats CBOR as the compatibility-oriented wire format. Using it consistently for gossip broadcasts gives us a self-describing payload that tolerates added fields and keeps the fast gossip path healthier across supported peers.

## Changes

| Area | Change |
|---|---|
| PushLog gossip | Add `PushLogBroadcast::encode_gossip_payload()` as the shared canonical encoder |
| Wire format | Prefer CBOR broadcast decoding before request-shaped CBOR and legacy postcard fallback |
| Iroh publish | Publish CBOR gossip payloads instead of postcard payloads |
| Libp2p publish | Route publish encoding through the shared PushLog gossip encoder |
| Compatibility | Continue accepting legacy postcard `PushLogBroadcast` payloads |
| Tests | Add coverage for canonical CBOR gossip encoding and unknown future CBOR fields |

## Out of scope

- No removal of legacy postcard decode support.
- No change to gossip topic naming.
- No change to PushLog request/response semantics.
- No broader Iroh diagnostics redesign beyond the wire-format hardening.

## Test plan

- [x] `cargo test -p p2p message::pushlog`
- [x] `cargo test -p p2p test_encode_gossip_payload_uses_cbor_broadcast --test message_tests`
- [x] `cargo test -p p2p --features iroh-transport message::pushlog`
- [x] `cargo fmt --all --check`
- [x] `git diff --check`
- [x] `cargo clippy -p p2p --all-targets --features iroh-transport -- -D warnings`

### DIFF
--- a/crates/p2p/src/host/command_handler/pubsub.rs
+++ b/crates/p2p/src/host/command_handler/pubsub.rs
@@ -65,7 +65,8 @@ impl<S: Store> P2PHost<S> {
         response: tokio::sync::oneshot::Sender<Result<libp2p::gossipsub::MessageId>>,
     ) {
         let ident_topic = topic.to_ident_topic();
-        let result = serde_cbor::to_vec(&message)
+        let result = message
+            .encode_gossip_payload()
             .map_err(|e| Error::CborSerialization(e.to_string()))
             .and_then(|data| {
                 self.swarm
--- a/crates/p2p/src/iroh/endpoint_commands.rs
+++ b/crates/p2p/src/iroh/endpoint_commands.rs
@@ -817,8 +817,9 @@ fn handle_publish(
     let gossip = gossip.clone();
     let initial_peers: Vec<iroh::EndpointId> = peer_map.lock().endpoint_ids().collect();
     let message_id = MessageId::new(uuid::Uuid::new_v4().to_string());
-    let payload = postcard::to_allocvec(&msg)
-        .map_err(|error| crate::error::Error::Codec(error.to_string()))?;
+    let payload = msg
+        .encode_gossip_payload()
+        .map_err(|error| crate::error::Error::CborSerialization(error.to_string()))?;
 
     let task = tokio::spawn(async move {
         let sender = if let Some(sender) = sender {
--- a/crates/p2p/src/message/pushlog.rs
+++ b/crates/p2p/src/message/pushlog.rs
@@ -324,9 +324,9 @@ pub struct PushLogBroadcast {
 /// Encoding variant accepted when decoding gossip payloads.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum PushLogGossipPayloadEncoding {
-    PostcardBroadcast,
-    CborRequest,
     CborBroadcast,
+    CborRequest,
+    PostcardBroadcast,
 }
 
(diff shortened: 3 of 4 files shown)


jackzampolin:
Post-merge follow-ups from review:

- Stale comment on `PushLogBroadcast::acp_actor_relationships` (`crates/p2p/src/message/pushlog.rs:310-319`) still explains why postcard needs the Option discriminator. Since postcard is now decode-only (never the publish path), reword to "postcard fallback decode still needs this" or remove.
- No tracked deprecation issue for removing the legacy postcard gossip decoder. Worth filing — e.g. "Remove postcard gossip fallback after N releases" or "after `gossip_decode_failure` telemetry shows only CborBroadcast for M weeks".
- The existing `test_decode_gossip_payload_from_cbor_request` in `crates/p2p/tests/message_tests.rs:522` already covers the CborRequest tagging, but a sibling unit test inside `pushlog.rs` would lock in the decode-order reasoning (CBOR request decoded before CBOR broadcast) close to the code it protects.

Issue #1022, opened 2026-06-04
Conformance harness: dedicated CI job + expand behavioral coverage


jackzampolin:
## Background

PR #1014 introduced the conformance harness (`proofs/`) — it binds the TLA+/Lean models to the **real `defra` binary** (Lean-axis vectors + TLA-axis behavioral scenarios driving real nodes via CLI + HTTP), with a Go-parity matrix (rust↔rust / go↔go / rust↔go).

Today it runs **serially inside the unit `Build & Test` job** (a dedicated step `cargo test -p conformance -- --test-threads=1`, added in #1014 because parallel real-node startup flaked with "failed to start rust-0").

## Follow-ups

- **Dedicated CI job/matrix** (like `integration-test`): isolate the heavy behavioral tests, run them against an optimized **release** binary, and run the **Go-parity** variants with a harness-compatible Go binary on PATH (currently those are `#[ignore]`).
- **Expand behavioral coverage**: more model families, the currently-ignored parity/bughunt probes, multi-node (3+) topologies, restart/partition matrices.
- **Lean drift-check**: it currently *skips* when `lake` is absent (CI has no Lean toolchain). Run it in a Lean-equipped job so the Lean↔Rust vocabulary drift-check actually executes in CI.

## Pointers
- Harness + tests: `proofs/tests/`, `proofs/src/`
- Local run: `proofs/verify-all.sh` (TLA + Lean + conformance)
- Binary resolution: `proofs/tests/support.rs::release_binary` (debug default; `DEFRA_CONFORMANCE_BINARY` override)

Issue #1021, opened 2026-06-04
Umbrella: apply the #1041 CRDT proof/conformance pattern to remaining materialized-view paths


jackzampolin:
## Background

#1014 introduced the formal-verification and conformance harness and immediately found real convergence bugs in Rust's CRDT/materialized-view paths. The common architectural shape was a split between:

- an authoritative CRDT store, and
- a materialized document/blob/view derived from that store.

When a local write, merge replay, restart, or derived view update touched only one side of that split, replicas could converge incorrectly even though the individual field merge looked reasonable in isolation.

#1041 is the first completed pattern for fixing and proving this class of bug. It landed:

- the reusable Lean core: `proofs/lean/DefraConvergence/CrdtField.lean`
- worked instances for both CRDT regimes:
  - Counter: op-based, non-idempotent, dedup-required
  - LWW: state-based, idempotent, re-delivery-safe
- a red/green TLA+ model for the two-store counter hazard
- conformance storms and Go-parity probes against the real binary
- the single-store counter fix for this issue's original motivating bug

#1044 then handled the interactive-transaction counter guard follow-up.

This issue now tracks the next phase: apply the #1041 proof + model + conformance pattern to the remaining CRDT/materialized-view areas that can still hide the same class of bug.

## Tracked follow-ups

- [ ] #1047 — composite/document materialization convergence
- [ ] #1048 — mixed-field same-doc convergence
- [ ] #1049 — encrypted LWW restart/replay convergence
- [ ] #1050 — secondary-index reconciliation after merge
- [ ] #1051 — PN-counter as distinct algebra/refinement

## Pattern to apply

Each follow-up should treat #1041 as the template:

1. Find or construct the smallest behavioral probe in `proofs/tests/behavioral/bughunt.rs`.
2. Confirm attribution with parity where possible: Rust↔Rust, Go↔Go, and Rust↔Go.
3. Fix the implementation at the authoritative CRDT/materialization chokepoint, not in a one-off caller.
4. Add the Lean proof or refinement argument for the field/view law.
5. Add the TLA+ red/green model when the bug depends on ordering, restart, replay, dedup, or lock/queue lifecycle.
6. Promote the probe to an asserting regression test.
7. Wire the model into `proofs/src/registry.rs` and update the proof docs.

## Priority notes

[...]

PR #1024, opened 2026-06-04
test(proofs): P2P convergence bug-hunt probes (#1021)

jackzampolin:
## Summary

Adds the multi-node convergence bug-hunt probes from the #1021 two-store audit to `proofs/tests/behavioral/bughunt.rs`. These are `#[ignore]` diagnostic probes (run with `--ignored --nocapture`) that *report* per-node converged state so divergences are visible.

The `port_reuse` fix that originally rode on this branch **already landed in `main`** via #1025 (`2f1fc835`), so it has dropped out of this PR — what remains is just the probes.

## Why keep them

These probes earned their place — between them they surfaced **two real, distinct convergence bugs**, both now fixed in `main`:

1. **libp2p `port_reuse(true)` persistent EADDRINUSE dial failures** → fixed by `2f1fc835` (disable port reuse; DefraDB does no DCUtR hole punching so it's interop-safe).
2. **#894 counter-convergence regression** (the Lark-storage / node-wiring rewrite re-split the materialized-blob vs CRDT-accumulation stores) → fixed by `8204a8fb` ("serialize standalone document merges"). This one was *masked* by bug #1 until the transport flake was removed, then the `counter_3node` probe caught it immediately (`node0=20`, full connectivity, zero Bitswap timeouts).

With both fixes in `main`, `bughunt_counter_3node` converges **20/20**.

## Probes added
- `bughunt_pncounter` (decrement), `bughunt_mixed_fields` (LWW + pcounter)
- `bughunt_encrypted_lww_{convergence,restart}`
- `bughunt_counter_3node` — 3-node full mesh, seed barrier, +10 each → expect 30

## Follow-up (not in this PR)
Promote `counter_3node` to an **asserting** test wired into a conformance CI job (#1022). Had that existed, #894 would have been caught at merge instead of slipping into `main`.

## Test Plan
- [ ] `cargo test -p conformance --test tla_conformance` → 15/15
- [ ] `cargo test -p conformance --test tla_conformance bughunt::bughunt_counter_3node -- --ignored --nocapture` → converges

### DIFF
--- a/proofs/tests/behavioral/bughunt.rs
+++ b/proofs/tests/behavioral/bughunt.rs
@@ -472,10 +472,8 @@ async fn bughunt_indexed_lww_restart() {
 /// both nodes (its key delivered) and both must materialize the same plaintext.
 /// A key that doesn't reach the peer, or a two-store reconcile that mishandles
 /// the ciphertext, shows up as divergence here.
-#[ignore = "bug-hunt probe; run with --ignored --nocapture"]
-#[tokio::test]
-async fn bughunt_encrypted_lww_convergence() {
-    let cluster = TestCluster::builder()
+async fn run_encrypted_lww(label: &str, restart: Option<usize>) {
+    let mut cluster = TestCluster::builder()
         .rust_nodes(2)
         .with_p2p()
         .with_encryption()
@@ -509,6 +507,13 @@ async fn bughunt_encrypted_lww_convergence() {
         .to_string();
     tokio::time::sleep(Duration::from_secs(4)).await;
 
+    if let Some(idx) = restart {
+        cluster
+            .restart_node(idx, Duration::from_secs(30))
+            .await
+            .expect("restart node");
+    }
+
     // Concurrent updates to the encrypted field. LWW tie-break (equal priority) ->
     // lexicographically greater value wins; "zzz" > "aaa".
     cluster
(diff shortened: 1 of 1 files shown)


jackzampolin:
**Heads-up for reviewers — separate regression discovered while rebasing this onto #1025.**

This PR's `port_reuse` fix is sound and isolated. But the `bughunt_counter_3node` probe, which converges **25/25** on this PR's original base (`d8738c9d`, the #1014 merge), drops to **~9-11/15** once stacked on current `main`/#1025 — and it's **not** this fix regressing:

- `EADDRINUSE = 0`, 16 peer connections, **zero Bitswap timeouts** → the transport fix is working.
- Failures are all `node0=20` (creator under-counts one increment) with **full connectivity** → a real CRDT/merge drop, not a transport issue.

Bisected to a single commit: **#894** ("feat(storage): add Lark backend and make it the default"), the only commit between `d8738c9d` and current `main`. The probe forces `--store redb`, so this is **not** Lark-specific — #894 also changed shared code (`crates/storage/src/backends/shared.rs`) and rewrote node/store wiring (`crates/defra-node/src/lib.rs`, +219). Prime suspect: the node-construction rewrite re-split or mis-wired the materialized-blob vs CRDT-accumulation stores (the two-store convergence bug class from #1021).

**This looks like a release-blocker-class convergence regression in `main` that should get its own issue.** It was invisible until the `port_reuse` transport flake (this PR) was removed.

PR #1027, opened 2026-06-05
fix(query): keep filtered cursor pagination on stable index

### LINKED ISSUE #1005: Order-aware index selection for filtered cursor pagination (mirrors Go #2680)
## Summary

For cursor pagination, the performance win comes from the **index seek** — the planner seeks directly to the cursor boundary in index order, so each page costs ~O(page_size) instead of O(matching rows). Today, when a cursor query has **both a filter and an `order` on _separate_ indexes**, index selection picks the **filter** index, the order index is never used, the cursor seek is rejected, and the query falls back to the slow path (materialize all filter matches → sort by the order field → skip to the cursor) on **every page** — i.e. O(n) per page, defeating the point of cursor pagination for filtered queries.

## This currently matches Go (intentional parity)

Go's `selectIndex` (`internal/planner/index_helpers.go`) documents the priority **Filter > RelationID > Ordering** and only records `canSatisfyOrder` for the chosen filter index. So Go behaves identically — it does not seek for filter+order-on-separate-indexes either. Go defers smarter selection to **sourcenetwork/defradb#2680** ("cost-based index optimization"). Our cursor port (PR #961) is at parity here; this issue tracks going **beyond** Go as a defradb.rs-specific perf feature.

## Where it lives (Rust)

- `crates/query-plan/src/planner/builder/index_methods.rs::try_select_index` — tries filter-based selection first and **short-circuits**; the ordering-index fallback is only reached when no filter index matched.
- `crates/query-plan/src/planner/builder/cursor.rs::validate_cursor_index` / `configure_scan_for_cursor` build a seek for the order index; `IndexScanNode::set_cursor_seek` rejects it when the scan's selected index differs → slow path.

## Proposed enhancement

Make index selection **order-aware for cursor queries**: prefer the order-supporting index (enabling the seek) when that's the better plan, ideally via a small cost heuristic rather than an unconditional override.

### Tradeoff to respect (why not an unconditional flip)

A **highly selective** filter can favor the filter index (sort a small candidate set) over seek-scanning a large ordered range while applying the filter as a residual. A naive "always prefer the order index for cursors" can regress those cases. A cost-based decision (filter selectivity estimate vs. expected seek scan distance) is the robust answer — the same direction as Go #2680.

## Notes

- No correctness impact today — results are correct via the slow path; this is purely a performance ceiling.
- The order-only and single-index-covers-both cases already seek correctly (verified: `indexFetches` is Go-exact, Go #4617 FFI suite 74/74).
- Going beyond Go here is a deliberate divergence from the parity North Star, so it should be a conscious product decision.

iverc:
## Summary

- Prefer the cursor-compatible order index for active filtered cursor pages when the filter-selected index cannot provide ordering.
- Keep first cursor pages on the existing filter-first path so highly selective filters do not regress unnecessarily.
- Tighten the filtered cursor regression so page 2 must return the correct ordered, filtered rows instead of accepting the old slow-path error.

Closes #1005

## Testing

- [x] `cargo test -p query-plan --lib`
- [x] `cargo clippy -p query-plan -p integration-test --all-targets -- -D warnings`
- [x] `env RUST_LOG=debug DEFRA_E2E_KEEP=1 cargo test -p integration-test --test cursor rust_cursor_with_filter_and_order_uses_correct_index -- --nocapture`

### DIFF
--- a/crates/query-plan/src/planner/builder/index_methods.rs
+++ b/crates/query-plan/src/planner/builder/index_methods.rs
@@ -28,7 +28,6 @@ impl super::Planner {
         // Extract limit/offset from select for passing to index scan
         let limit = select.limit.as_ref().and_then(|l| l.limit);
         let offset = select.limit.as_ref().map(|l| l.offset).unwrap_or(0);
-        let active_cursor_order_index = select_active_cursor_order_index(select, collection);
 
         // Try filter-based index selection.
         // Relation fields in the filter (e.g., `owner: {name: {_eq: "X"}}`) are
@@ -39,21 +38,6 @@ impl super::Planner {
         // match any index. This means `select_best_index` safely ignores them.
         if let Some(filter) = select.filter.as_ref() {
             if let Some(best_index) = select_best_index(filter, &collection.indexes) {
-                let provides_ordering = select
-                    .order_by
-                    .as_ref()
-                    .map(|o| can_be_ordered_by_index(o, best_index).0)
-                    .unwrap_or(false);
-
-                if !provides_ordering {
-                    if let Some((order_index, needs_reverse)) = active_cursor_order_index {
-                        return Some((
-                            order_index_scan_params(order_index, needs_reverse, None, 0),
-                            true,
-                        ));
-                    }
-                }
-
                 if let Some(params) = filter_to_index_scan(
                     filter,
                     best_index,
@@ -62,15 +46,13 @@ impl super::Planner {
                     limit,
                     offset,
                 ) {
+                    let provides_ordering = select
+                        .order_by
+                        .as_ref()
+                        .map(|o| can_be_ordered_by_index(o, best_index).0)
+                        .unwrap_or(false);
                     return Some((params, provides_ordering));
                 }
-            }
-
-            if let Some((order_index, needs_reverse)) = active_cursor_order_index {
-                return Some((
-                    order_index_scan_params(order_index, needs_reverse, None, 0),
-                    true,
-                ));
             }
         }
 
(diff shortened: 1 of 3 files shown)


jackzampolin:
Hi @iverc — thanks for this. We ran a deep adversarial review on the cursor-pagination batch and this one needs rework before it can land. Two correctness-class issues plus a premise concern. Flagging here rather than merging so #1005 stays open.

### 1. The Go-parity premise is backwards
The PR is framed as mirroring Go #2680, but Go does the opposite of "prefer order index for filtered cursors." Go selects a single scan index **filter-first**, and only falls back to the ordering index when *no* filter index exists (`internal/planner/select.go:340-345`), then `validateCursorIndex` reads that same `scan.index` (`internal/planner/planner.go:313-337`). Go never switches to a separate order index mid-pagination, so it never hits the split-tiebreak problem below. Go #2680 is itself an explicit "take the first candidate, optimize later" TODO (`select.go:380-382`), not a directive to prefer order indexes. So this is a Rust-only behavior layered on the Rust-specific skip-by-docID slow path — it shouldn't be described as parity, and its correctness can't lean on "Go does this."

### 2. (Blocking) Cross-page tiebreak inconsistency drops/duplicates rows on non-unique order fields
Page 1 and page ≥2 of the *same* query end up sorting ties differently:
- **Page 1** (no token): filter index selected, `provides_ordering=false`, so an `OrderByNode` sorts. That sort has **no docID tiebreak** — `compare_docs_static` returns `Equal` for equal order-fields and keeps scan input order (`crates/query-plan/src/plan/orderby.rs:260-262, 134-138`).
- **Page 2+** (active token): `select_active_cursor_order_index` picks the **order index**, returns `provides_ordering=true`, so the `OrderByNode` is skipped (`groupby.rs:308`). Ties now come out in index-native **docID-byte order** (`cursor.rs:336-343`).

With a range filter where two rows share the order-field value but differ on the filter field, page 1's tiebreak (scan order) disagrees with page 2's (docID order) → the page boundary lands between two rows whose relative order flipped, and one row is **dropped or repeated**. Repro shape: indexes `idx_age(age)` + `idx_score(score)`, rows `p(age=50,score=80)` / `q(age=50,score=85)`, query `filter:{score:{_gte:80}}, order:[{age:ASC}], first:N`. No test currently exercises ties-on-order-field across a page boundary.

### 3. (Blocking) Scan-index vs seek-index selectors use opposite field-count rules
`select_active_cursor_order_index` requires `order_by.conditions.len() >= index.fields.len()` (`index_methods.rs:167-169`), but `validate_cursor_index`/`find_matching_index` requires `idx.fields.len() >= order_fields.len()` (`cursor.rs:186, 171-178`) — the opposite relation. With `idx_a=(age,name)` and `idx_b=(age)` and `order:[{age}]`, the scan picks `idx_b` while the seek key is built for `idx_a`; `set_cursor_seek` then rejects it (`index_scan.rs:284-286`) and the query silently falls to the slow path — so the optimization never fires, and combined with #2 (OrderByNode already suppressed) the slow path reorders/skips.

### Suggested direction
Either (a) match Go: keep filter-first selection and let the filtered cursor error/slow-path when the filter index can't satisfy the order (no behavior change vs Go), or (b) if we intentionally extend beyond Go, make **page 1 use the same order index as page ≥2** (so both share the docID tiebreak), append an implicit `_docID` tiebreak so `OrderByNode` and the index agree, and **unify the scan/seek index selectors** into one helper so they can't diverge (fixes #3). Either way, please add tests for ties-on-order-field across a page boundary and for the composite scan/seek-index case.

Note this is logically coupled to #1026 (seek-key building) — landing this widens the set of queries that depend on #1026's per-field encoding + docID-suffix correctness, so they should be reasoned about together.

Happy to pair on the approach if useful.

jackzampolin:
Re-reviewed @iverc — the filter-first rework resolves all three findings: B1 (both pages now stay on the filter index + OrderByNode, so the tiebreak is consistent across the page boundary), B2 (the divergent scan/seek selector is gone with `select_active_cursor_order_index` deleted), and B3 (now matches Go's `findIndexByFilteringField`-first contract). The two previously-untested adversarial cases (ties across a page boundary, composite scan/seek) now have passing tests, and CI is green incl. the conformance job. Nice. Merging.

Note for tracking: this consciously trades the order-aware *seek* optimization for Go-parity correctness — page ≥2 of a filtered+ordered cursor now uses the O(n) drain slow path. So #1005's original perf intent isn't delivered here; we'll keep #1005 open (re-scoped to the optimization done correctly) rather than close it on this merge.

PR #1042, opened 2026-06-15
feat(defra-node): retry embedded execute conflicts

### LINKED ISSUE #1031: EmbeddedNode::execute: opt-in built-in retry on transaction conflict
## Request

`EmbeddedNode::execute()` is auto-commit (each call is its own implicit transaction) and, on optimistic-concurrency races, returns:

```
commit error: datastore error: storage error: transaction conflict. Please retry
```

The error explicitly instructs the caller to retry, but `execute()` offers no built-in retry — so every downstream consumer must reimplement bounded retry + conflict-string classification themselves.

Requesting an **opt-in** retry mechanism, e.g.:
- `execute_with_retry(query, RetryPolicy)`, or
- a retry policy on an `ExecuteOptions`/builder,

that retries the prescribed transient transaction-conflict a bounded number of times with backoff. This would let consumers rely on the DB's own "please retry" contract instead of string-matching the error message and hand-rolling backoff.

## Context

`sourcenetwork/defra-agent` currently has to wrap this manually (concurrent writers to the same documents — runtime + tests). Tracking issue: sourcenetwork/defra-agent#440.

Pinned version in use: `v0.14.2`.

iverc:
## Summary

Closes #1031.

This PR adds an opt-in retry path for `defra_node::EmbeddedNode::execute` callers that hit transient optimistic-concurrency conflicts.

The existing `execute()` method remains unchanged and still performs a single attempt. Callers that want bounded retries can now use `execute_with_retry(query, ExecuteRetryPolicy)` or `execute_request_with_retry(request, ExecuteRetryPolicy)` instead of string-matching the "transaction conflict. Please retry" response themselves.

## Details

- Adds `ExecuteRetryPolicy` with bounded exponential backoff.
- Adds `EmbeddedNode::execute_with_retry`.
- Adds `EmbeddedNode::execute_request_with_retry` for callers that need variables, operation names, or identity on a prepared `QueryRequest`.
- Reuses the same one-shot execution path as `execute()`, including node signing / ambient identity behavior.
- Retries only single-error responses with no data that contain the storage transaction-conflict retry message.
- Stops immediately on success, non-conflict errors, or retry exhaustion.
- Adds unit coverage for default policy values, backoff capping, conflict classification, and retry-until-success behavior.

## Validation

- [x] `cargo fmt --all --check`
- [x] `git diff --check`
- [x] `cargo test -p defra-node execute_retry`
- [x] `cargo test -p defra-node`
- [x] `cargo clippy -p defra-node --all-targets -- -D warnings`

## Notes

This is intentionally opt-in so existing embedded callers keep one-shot semantics. It also avoids broad retrying of arbitrary GraphQL errors; only the explicit transaction-conflict retry contract is retried.

### DIFF
--- a/crates/defra-node/src/lib.rs
+++ b/crates/defra-node/src/lib.rs
@@ -51,6 +51,40 @@ pub use schema::CollectionVersion;
 #[cfg(feature = "otel")]
 pub use telemetry::{TelemetryConfig, TelemetryHandle};
 
+const TRANSACTION_CONFLICT_RETRY_MESSAGE: &str = "transaction conflict. Please retry";
+
+/// Retry policy for [`EmbeddedNode::execute_with_retry`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ExecuteRetryPolicy {
+    /// Number of retries after the initial attempt.
+    pub max_retries: u32,
+    /// Backoff before the first retry.
+    pub initial_backoff: Duration,
+    /// Maximum backoff between retries.
+    pub max_backoff: Duration,
+}
+
+impl ExecuteRetryPolicy {
+    /// Create a retry policy with bounded exponential backoff.
+    pub const fn new(max_retries: u32, initial_backoff: Duration, max_backoff: Duration) -> Self {
+        Self {
+            max_retries,
+            initial_backoff,
+            max_backoff,
+        }
+    }
+}
+
+impl Default for ExecuteRetryPolicy {
+    fn default() -> Self {
+        Self {
+            max_retries: 3,
+            initial_backoff: Duration::from_millis(10),
+            max_backoff: Duration::from_millis(250),
+        }
+    }
+}
+
 /// Type-erased schema operations so we can store DB<S> without leaking the Store generic.
 #[async_trait::async_trait]
 trait SchemaOps: Send + Sync {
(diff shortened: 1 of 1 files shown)


jackzampolin (approved):
Approving — clean, bounded, opt-in change that closes #1031.

The refactor is behavior-preserving: `execute()`'s body is extracted verbatim into `execute_request_once()` and `execute()` just delegates, so existing one-shot callers are untouched. Everything else is additive (`ExecuteRetryPolicy` + the two `*_with_retry` methods).

A few things I verified:
- The conflict string is real — `crates/storage/src/corekv/errors.rs:57` emits `"transaction conflict. Please retry"`, and the `.contains()` classifier matches the wrapped form correctly.
- The classifier is conservative in the right way: it only retries single-error, no-data responses and does **not** retry partial responses, which avoids double-applying a partially-committed mutation.
- This mirrors an established Go pattern — `internal/db/merge.go` retries `corekv.ErrTxnConflict` in a bounded loop, and the test helper `withRetryOnNode` (tests/integration/utils.go) classifies by `strings.Contains(err, ErrTxnConflict)` precisely because the type is lost crossing the C-binding boundary. Same situation here at the embedded-node layer.

One minor note (non-blocking): Go's default `MaxTxnRetries` is **5** (`internal/db/config.go:25`, `defaultMaxTxnRetries = 5`), whereas this defaults to 3. Not a parity concern since this is a Rust-native embedded convenience rather than wire/protocol behavior, and the exponential backoff here is arguably better than Go's flat sleep — but if you want to match Go's tuning exactly, bumping the default to 5 would do it.

PR #1056, opened 2026-06-19
spike(acp): cross-object / collection-level ACP seam (TTU read inheritance)

jackzampolin:
## What

Proves — and minimally unblocks — **cross-object (collection-level) ACP** on the live Rust document-ACP path. A `directory -> file` read cone (`file.read = reader + parent->read`) now resolves end-to-end through `ZanzibarDocumentACP::check_doc_access`.

## Why

Diagnosis showed the live node cannot *seed* cross-object / userset relationships on documents (e.g. `file:report#parent@directory:teamdir`), even though the Zanzibar engine + store already model and resolve them (the TTU evaluator is even Lean-proven). The gap is purely the **seeding seam**: `add_actor_relationship` flattens every target to an actor DID via `actor_subject()`. This is a shared limitation with Go DefraDB — not a Rust-behind-Go regression — and it is a seam-widening, **not** a V2 rebuild.

## Changes

- **`parse_target_subject(s)`** — maps a target string to the full subject language:
  - `*` → `Subject::Wildcard`
  - `did:...` → `Subject::Entity` (actor)
  - `resource:id#relation` → `Subject::EntitySet` (userset)
  - `resource:id` → `Subject::EntitySet` object / TTU edge
- **`ZanzibarDocumentACP::add_subject_relationship(...)`** — stores a relationship with an arbitrary `Subject` under the **same owner/manager authority check** as an actor grant.
- Re-exported both from the `acp` crate.
- **`tests/spike_cross_object_ttu.rs`** — seeds the cross-object `parent` edge through the real seam (no store backdoor) and asserts alice reads the file via `parent->read -> directory#read -> reader` (before = denied, after = granted; bob stays denied).
- 6 parser unit tests covering all four subject forms + malformed input.

Engine and store required **zero changes** — the TTU evaluator already reads parent subjects and the memory store already maps an `EntitySet` subject to a cross-object `ObjectRef`. Full `acp` suite green.

## Scope / not yet done

This is the proving seam. Making it the *full* live path is follow-up plumbing (not invention): the `DocumentACP` trait API (`&Did`-typed), the CLI `doc_acp_adapter.rs` DID-parse gate, the HTTP `target_actor: String`, routing the default backend through `ZanzibarDocumentACP` (the `Local` backend stores a bare `Did`), and multi-resource collection↔policy registration for cross-collection edges. Userset-subject-on-a-direct-relation (the `This`-path `check_permission_direct` EntitySet expansion) is a separate, deferred blocker; TTU-reached usersets already work.

**Open question this spike cannot answer:** whether a cross-object grant behaves correctly through a live *query* (planner / sync / indexing). That should be probed before widening further.

### DIFF
--- a/crates/acp/src/lib.rs
+++ b/crates/acp/src/lib.rs
@@ -53,6 +53,7 @@ pub use store::AcpStore;
 pub use validation::{validate_resource_interface, REQUIRED_DOCUMENT_PERMISSIONS};
 
 // Re-export key zanzibar engine types from the standalone zanzibar crate
+pub use zanzibar::parse_target_subject;
 pub use zanzibar::PersistentZanzibarStore;
 pub use zanzibar::ZanzibarDocumentACP;
 
--- a/crates/acp/src/zanzibar/acp/mod.rs
+++ b/crates/acp/src/zanzibar/acp/mod.rs
@@ -9,12 +9,57 @@ use identity::Did;
 use zanzibar::engine::PermissionEngine;
 use zanzibar::expression::RelationExpression;
 use zanzibar::store::ZanzibarStore;
-use zanzibar::types::{Policy, Relation, Resource};
+use zanzibar::types::{Policy, Relation, Relationship, Resource, Subject};
 
 use crate::error::{Error, Result};
 use crate::permission::DocumentPermission;
 
 pub const OWNER_RELATION: &str = "owner";
+
+/// Parse a relationship-target string into a [`Subject`], supporting the full
+/// document-ACP language rather than only actor DIDs. This is the collection-
+/// level-ACP seam: it lets a document relation point at another object (a TTU
+/// parent edge) or at a userset, mirroring what the Zanzibar engine and store
+/// already resolve.
+///
+/// Forms:
+/// - `*`                     → [`Subject::Wildcard`] (all actors)
+/// - `did:...`               → [`Subject::Entity`] (a single actor)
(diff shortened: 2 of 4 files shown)


jackzampolin:
**Merge as a spike only — do not build on this as if collection-level ACP is de-risked.**

Two reviews + code verification confirm this proves the *local* engine seam (~10%) and not the path that matters for distributed consumers:
- Cross-node ACP rides **SourceHub**, whose `SubjectRef` is `Actor(String) | AllActors` only (`crates/sourcehub/src/provider.rs:5`) — it cannot represent a cross-object/userset subject. This seam widens the node-local Local engine.
- The seam **bypasses `Relationship::validate`**, which would reject the empty-relation `EntitySet` object-edge (`crates/zanzibar/src/types/policy.rs:197`).
- Relation `types:` are dropped by serde (`crates/acp/src/policy_yaml/mod.rs:82`), so target-type restrictions are unenforced (soundness gap).

Keep this **off the `DocumentACP` trait**. The honest scope + the Phase 0 distributed gate (two-node Iroh, live query under SourceHub, parent-after-child sync ordering per Go #4789) are tracked in #1057.

jackzampolin:
Superseded by #1060 (step 2a). The cross-object resolution test is carried forward there, now driven through the real `DocumentACP::add_relationship(Subject)` API and gated by the soundness floor (#1059) — rather than the inherent `add_subject_relationship` + `parse_target_subject` seam this spike used, which is intentionally not merged. Closing without merge.

Issue #1057, opened 2026-06-19
Collection-level ACP: cross-object/userset relationships (engine seam proven; distributed + soundness work open)


jackzampolin:
## Status

**Engine seam PROVEN; the feature is NOT de-risked.** The Rust Zanzibar engine already resolves cross-object TupleToUserset once the tuple exists, and PR #1056 adds a minimal *local* seeding seam + a passing test. But two independent reviews + code verification show the real cost is **downstream of the seam**, and it is **not** on the path most consumers (e.g. defra-agent, cross-node) actually use.

Do **not** schedule the items below as if PR #1056 de-risked them. Run the Phase 0 gate first.

## What PR #1056 proves (the cheap ~10%)
- The local in-memory engine evaluates `file.read = reader + parent->read` and a `file#parent@directory` edge yields inherited read. (`crates/zanzibar/src/engine/evaluate.rs`, `crates/acp/tests/spike_cross_object_ttu.rs`.)
- The only *local* missing piece was seeding non-actor subjects: `parse_target_subject` + `ZanzibarDocumentACP::add_subject_relationship` (no non-test callers; off the production trait path).

## What it does NOT prove (the expensive ~90%)
Live query behavior; multi-node sync; SourceHub/on-chain subject support; branchable-collection object lifecycle; correctness when parent/collection objects arrive after children; enforcement of `types:` restrictions; safe public API shape; delete/revoke of object subjects; Go/Rust parity.

---

## Phase 0 — GATE: prove the distributed path FIRST (do this before any widening)
The single highest-value action. Two-node Iroh cluster; seed a cross-object edge; check a grant through a **live query under SourceHub enforcement**; include the **sync-ordering** case where the parent/collection object arrives *after* the child (mirrors Go #4789 — authorization can depend on a collection-version whose DAG is still in flight). Verify **fail-closed** behavior.

Outcome decides whether collection-level ACP is a ~2-week or ~2-month effort. If the distributed authorization-vs-sync model doesn't hold, Phases 1–3 are wasted.

Why this matters for cross-node consumers: cross-node ACP does **not** ride the Local/Zanzibar engine — it rides **SourceHub**, and `SubjectRef` is `Actor(String) | AllActors` only (`crates/sourcehub/src/provider.rs:5`) — it cannot represent a cross-object/userset subject at all. The seam PR #1056 widened is the in-memory Local path, which is node-local only.

## Phase 1 — Subject grammar + validation + type soundness (authz-critical)
[...]

PR #1069, opened 2026-06-24
Enable SourceHub cross-object ACP relationship subjects

iverc:
## Summary

- add SourceHub support for structured relationship subjects (`object` edges and `actor_set` usersets)
- pass bearer tokens through SourceHub structured-subject relationship writes/deletes
- encode SourceHub protobuf `Subject.object` and `Subject.actor_set` in transaction payloads
- add SourceHub cross-object ACP regression coverage for `file.parent -> directory.read` inheritance

Part of #1057

## Testing

- [x] `cargo test -p sourcehub`
- [x] `cargo check -p cli`
- [x] `cargo clippy -p sourcehub --all-targets -- -D warnings`
- [x] `cargo clippy -p cli -- -D warnings`
- [x] `RUST_LOG=info SOURCEHUB_BINARY=/Users/ivan/Source/sourcehub/build/sourcehubd cargo test -p integration-test --test sourcehub -- cross_object --nocapture`
- [x] `RUST_LOG=info SOURCEHUB_BINARY=/Users/ivan/Source/sourcehub/build/sourcehubd cargo test -p integration-test --test sourcehub -- rust_sourcehub_smoke --nocapture`

### DIFF
--- a/crates/sourcehub/src/cosmos/cosmos_provider.rs
+++ b/crates/sourcehub/src/cosmos/cosmos_provider.rs
@@ -86,6 +86,44 @@ fn subject_to_json(subject: &SubjectRef) -> serde_json::Value {
     }
 }
 
+fn structured_subject_to_json(
+    kind: u8,
+    subject_resource: &str,
+    subject_object_id: &str,
+    subject_relation: &str,
+) -> Result<serde_json::Value, ProviderError> {
+    match kind {
+        2 if !subject_resource.is_empty()
+            && !subject_object_id.is_empty()
+            && subject_relation.is_empty() =>
+        {
+            Ok(serde_json::json!({
+                "object": {
+                    "resource": subject_resource,
+                    "id": subject_object_id,
+                }
+            }))
+        }
+        3 if !subject_resource.is_empty()
+            && !subject_object_id.is_empty()
+            && !subject_relation.is_empty() =>
+        {
+            Ok(serde_json::json!({
+                "actor_set": {
+                    "object": {
+                        "resource": subject_resource,
+                        "id": subject_object_id,
+                    },
+                    "relation": subject_relation,
+                }
+            }))
+        }
+        _ => Err(ProviderError::Unsupported(format!(
+            "unsupported structured subject tuple kind={kind} resource='{subject_resource}' object_id='{subject_object_id}' relation='{subject_relation}'"
+        ))),
+    }
+}
+
 fn resolve_cosmos_bearer_token(
     did: &str,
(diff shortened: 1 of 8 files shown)


jackzampolin (changes requested):
Thanks Ivan — the SourceHub encoding work here is solid: the `Subject` oneof slots (`object`→4, `actor_set`→2) match the proto, the kind-2 object-edge path is validated end-to-end against live `sourcehubd` including delete/revoke, and `structured_subject_to_json` is properly fail-closed on malformed tuples. No concerns with the production code itself.

I'm holding the merge, though, because as written this PR claims **`Fixes #1057`** and #1057 makes the distributed path a hard **Phase 0 GATE** that this doesn't yet clear: *"prove the distributed path FIRST ... before any widening."* What's needed before this lands:

**1. Phase 0 distributed gate (blocking).** A two-node Iroh cluster that:
- seeds a cross-object edge and confirms inherited read through a **live query under SourceHub enforcement** across nodes (not single-node `rust_nodes(1)`);
- covers the **sync-ordering** case where the parent/directory object arrives *after* the child file (mirrors Go #4789);
- asserts **fail-closed** behavior while propagation is incomplete (access denied until the edge is fully visible).

This is the acceptance criterion the issue calls "the single highest-value action," and CodeRabbit's two Major comments are pointing at the same gap.

**2. Live kind-3 userset coverage (blocking).** kind-3 / `actor_set` (field 2) is currently the *only* wire path proven solely by unit tests — every other path is validated against live SourceHub. Please exercise a real userset relationship through query + revoke against `sourcehubd`, not just the encoder round-trip.

**3. Retitle `Fixes #1057` → `Part of #1057`.** Even with the above, #1057's Phase 1 type-soundness work (the `types:` restriction is still a decorative no-op → a `file#parent` edge can point at any object of any resource) stays open, so this shouldn't auto-close the issue.

**4. Trivial:** `file_count`'s `unwrap_or(0)` makes a malformed/missing `File` array indistinguishable from "access denied," so a broken response shape can pass the negative assertions. Assert the array is present before counting (CodeRabbit left a committable suggestion).

Happy to re-review as soon as the Phase 0 cluster test is in.

jackzampolin:
Closing — superseded by #1073 (#4990 branchable-collection ACP). The #4990 model uses collection-as-object + actor-DID relationships, not SourceHub cross-object subjects, so this path is no longer the direction. The Phase-0 distributed concern it was held on is now covered by the #4990 P2P-sync enforcement (A3) + the two-node harness (B1).

PR #1071, opened 2026-06-24
Add indexed CRDT merge conformance

### LINKED ISSUE #1050: proof+conformance(crdt): secondary-index reconciliation after merge
Parent: #1021
Pattern to apply: #1041
Related foundation: #1014

## Background

#1041 focused on CRDT field values, but DefraDB also keeps secondary indexes as materialized views of those values. A correct CRDT merge is not sufficient if the index view retains stale winners/losers after re-materialization, restart, or a clobbered merge.

## Scope

Model and test secondary-index reconciliation after CRDT merges, especially:

- LWW indexed value changes across partition/reconnect
- restart after a stale materialized blob was clobbered and repaired
- index cleanup for losing values
- counter/index interactions if indexed CRDT values are supported or later added
- queries observing index state before and after merge replay

## Method

Use the #1041 approach, treating the index as a derived materialized view:

- Lean/TLA+: prove or model that the index is a function of the converged CRDT state, not of stale intermediate blobs.
- Conformance: promote indexed LWW restart/partition probes to asserting tests that query through the index.
- Parity: compare Go and Rust behavior where the index feature overlaps.
- Registry/docs: connect this to the existing index-maintenance proof family and the CRDT-field registry row.

## Acceptance

- A model states the invariant: after merge/replay, index entries match the winning materialized CRDT value and stale entries are absent.
- Behavioral tests query through the index and assert no stale losing value remains.
- Restart/replay coverage is included.
- Any implementation fix is tied back to the model and regression test.

iverc:
## Summary

Closes #1050.

This adds proof/conformance coverage for secondary-index reconciliation after CRDT merge/replay. The new slice verifies that once an indexed LWW field converges, indexed queries expose exactly the winning materialized value and do not retain stale seed or losing-value keys.

## What changed

- Added `IndexReconciliation.tla` with red/green configs for stale index-key retention vs delete-old-then-save-new reconciliation.
- Consolidated the indexed LWW restart/merge coverage into a single asserting conformance test in `behavioral/index.rs`, removing the overlapping `partition.rs::convergence_indexed_lww_restart_merge` (and its now-orphaned helpers) so the suite no longer pays for duplicate restart/replication coverage. The surviving test carries the stronger structured `@explain` oracle plus the two-store regression rationale and a stricter seed-index precondition (the seed must be index-resolvable on node1, not merely materialized).
- Added an `@explain` guard so the regression verifies the filtered query actually uses the secondary index.
- Wired the temporal model into the existing `Index-maintenance consistency` registry/README entries.

## Unrelated change (called out per review)

`crates/defra-node/src/lib.rs` adds a test-only (`#[cfg(test)]`) `SIGNING_STORE_GUARD` mutex that serializes `node_identity_did_requires_registered_signer` and `node_identity_did_accepts_registered_remote_signer`. Both call `clear_identity_store()` on a process-global signing store and flaked when run concurrently. This is independent of #1050 and is kept here only to keep this PR's identity-test CI green.

## Testing

- [x] `cargo test -p conformance --lib matrix::`
- [x] `cargo clippy -p conformance --all-targets -- -D warnings`
- [x] `env RUST_LOG=info cargo test -p conformance --test tla_conformance index:: -- --test-threads=1 --nocapture`
- [x] `git diff --check`

Note: targeted TLC runs for the new configs were attempted locally, but this machine has no usable JDK installed, so TLC could not start.


jackzampolin:
Conformance coverage looks good and CI is green. Before merge, I'd like the `@explain` oracle tightened per CodeRabbit's point: right now `index_filter_uses_index` only asserts the substring `"index"` appears in the explain output, and only on node0 *before* the restart. If the post-replay filtered query silently falls back to a document scan, `count_by_age()` still returns the expected rows and the regression passes without proving the secondary index was actually consulted — which defeats the purpose of this conformance slice.

Can you re-run `@explain` for the same filtered query *after* reconnection/convergence (ideally on both nodes) and assert the concrete index-scan plan shape rather than a substring match? Once that's in I'm happy to merge.

jackzampolin:
The new index conformance slice looks directionally right, and I verified the focused test passes locally after building `target/debug/defra`:

`cargo test -p conformance --test tla_conformance index_reconciles_lww_merge_after_restart -- --test-threads=1 --nocapture`

One cleanup ask before merge: this adds another long two-node restart/P2P indexed-LWW test in `behavioral/index.rs`, but `behavioral/partition.rs::convergence_indexed_lww_restart_merge` already exercises essentially the same scenario. The new structured `@explain` oracle is better, but keeping both means the conformance suite pays for duplicate restart/replication coverage. Can we consolidate by moving/tightening the existing test into `index.rs`, or otherwise remove the older overlapping copy?

Small scope note: the signing-store mutex change in `crates/defra-node/src/lib.rs` looks benign and probably addresses the flaky identity test, but it is unrelated to #1050. Can that be split out or called out explicitly if it needs to stay in this PR?

iverc:
Thanks — both addressed in fd1e73a5:

**Consolidation:** removed `partition.rs::convergence_indexed_lww_restart_merge` and its now-orphaned helpers, folding its two-store regression rationale into the surviving `behavioral/index.rs::index_reconciles_lww_merge_after_restart` (the one with the structured `@explain` oracle). While consolidating I also tightened the seed gate there to require the seed be *index-resolvable* on node1 before the partition (the strictness the old test had), and fixed the stale cross-reference in `parity.rs`. Net −201 lines in `partition.rs`; no more duplicate restart/replication coverage.

**Scope note:** the `defra-node` signing-store mutex is test-only (`#[cfg(test)]`) and serializes the two `node_identity_did_*` tests that both `clear_identity_store()` on a process-global store. Opted to keep it here (rather than split) to keep this PR's identity-test CI green, and called it out explicitly under a new "Unrelated change" section in the PR description.

Issue #1043, opened 2026-06-16
Mixed Rust↔Go counter storm: Go peers intermittently double-apply (timing-sensitive); Rust materializes identical DAG correctly


jackzampolin:
## Summary

The mixed Rust↔Go same-document counter storm (`parity_counter_storm_mixed` in `proofs/tests/behavioral/parity.rs`) intermittently **over-counts on the Go peers** while the Rust node stays exact — e.g. `hits=[24,25,25]` (expected 24), also seen `[12,15,15]`, `[36,40,40]`. Investigation shows the miscount is a **timing-sensitive double-apply on the Go reference side**, not a Rust value bug.

## Evidence

- **Identical DAGs, divergent materialization.** At a divergence, all three nodes hold a byte-identical commit DAG (same commit count, CIDs, heights). Link-level dump confirms each counter (`hits`) field block is linked by exactly one composite. The DAG contains exactly N `+1` blocks for the N increments issued. **Rust materializes the exact sum (N); the Go peers materialize N+k.** So no lost deltas, no extra/phantom blocks, no duplicate CID — Go applies one (or more) counter field delta **twice** during merge.

- **Go applies counter deltas unconditionally.** `internal/core/block/store.go::ProcessBlock` calls `crdtData.Merge` → `internal/core/crdt/counter.go::incrementValue` (a pure read-modify-write with no nonce/CID dedup). There is **no per-block `IsMerged` guard inside the apply**; Go's only counter dedup is structural — `internal/db/merge.go::executeMerge`→`loadComposites` collects new composites (walking back to current heads, no `IsMerged` check) and `processBlock` applies each composite's linked field deltas. By contrast, the Rust counter merge guards on `is_merged(cid)` before applying (`crates/db-merge/src/merge_handler/counter.rs`).

- **Pure go↔go never reproduces.** `parity_counter_storm_go_go` (3 Go nodes, same burst pattern): 0 over-counts in 15 stress runs.

- **Heisenbug.** Building an instrumented Go binary that logs every `ProcessBlock` apply (one `fprintf` to stderr per call) **suppresses the race entirely** (12/12 converged) — the per-call latency closes the timing window. The race is acutely timing-sensitive and is triggered by the Rust node's delivery timing (`RUST_LOG=info` reproduces; `RUST_LOG=error`/idle usually converges).

## Reproduction

```
PATH=~/.cache/defra-harness/<GO_COMPAT_COMMIT>:$PATH \
[...]

PR #1096, opened 2026-07-07
fix(p2p): heal dead iroh-gossip send paths instead of retrying a corpse forever (#1092)

### LINKED ISSUE #1092: p2p gossip: send-loop death is retried forever with no backoff — dead connections never heal without process restart
## Summary

When an iroh-gossip connection's send loop dies, the p2p layer retries the dead connection forever with **no backoff and no reconnect/heal** — the paths never recover without a full process restart. On the amygdala fleet hub (16-peer topology, defra-agent embedding defradb.rs rev `1b78bd21`), the hub retries dead gossip connections to ~14 peers **in the same millisecond** every few minutes, indefinitely:

```
2026-07-07T13:49:40.627011Z WARN gossip{me=775ad8b54c}: iroh_gossip::net: failed to send: connection task send loop terminated peer=a88aaf7f0d
```

## Numbers (48h, 2026-07-05→07)

- Hub: **452,620** `send loop terminated` warns; spokes 3k–127k each.
- 24h after a clean hub restart, still ~144/h steady at idle — the dead paths persist; only a restart heals them.
- During the 07-05/06 PushLog fan-in storm (defra-agent#630) this retry churn amplified into fleet-wide CPU waves.
- Operational blast radius today: the hub process was found with its HTTP control plane unresponsive while the gossip loop churned (runtime-side control-plane isolation is being tracked in defra-agent; the underlying non-healing connection state is this issue).

## Suggested direction

Treat `send loop terminated` as connection death: tear down the gossip connection state and re-dial with exponential backoff (or drop the peer until next discovery), instead of re-queuing sends onto a corpse. An error every few ms/min per dead peer forever is the worst of both — no recovery, constant churn, log/CPU load scaling with peer count.

## Acceptance

On a healthy 16-peer topology, kill one peer's transport: the survivor's gossip warns decay under backoff and the connection heals on peer return, with zero warns at steady state (current steady state: ~144/h at idle, permanent).

jackzampolin:
Closes #1092.

## Problem

When an iroh-gossip send loop dies, the peer stays in `PeerState::Active` with a closed send channel forever: gossip only prunes a neighbor when its connection task finishes, but a stream-level write error kills only the send half while iroh keep-alives hold the QUIC connection open. Every send then warns `connection task send loop terminated` and is dropped — 452k warns/48h on the fleet hub, ~144/h at idle, healed only by process restart. Even our existing `join_peers` rejoin on reconnect was blackholed through the same dead channel.

## Why the obvious fix doesn't work

iroh-gossip owns **separate QUIC connections on `GOSSIP_ALPN`**, disjoint from our RPC connections — force-closing our connections never touches gossip's path. And there is no per-peer signal at our layer for a dead send loop behind a live connection: `broadcast()` returns Ok, `NeighborDown` never fires in the wedged state.

## Fix (no iroh/iroh-gossip forks or patches)

The heal lever is `Gossip::handle_connection`: dial `GOSSIP_ALPN` ourselves and hand gossip the fresh connection — its own duplicate-connection handling atomically swaps the active send path and the old send loop drains out. Wrapped in:

- **Event-driven 0→1 heal** (accept and dial paths): a returning peer heals immediately. Dial failure falls back to the pre-fix plain `join_peers`.
- **Periodic sweep** (default 60s, `DEFRA_P2P_GOSSIP_HEAL_INTERVAL_SECS` override, 0 disables): since a successful dial IS the only possible probe, probe and heal are the same cheap action — the silent half-dead state is cured within one interval, giving zero warns at steady state.
- **Per-peer exponential backoff** (2s base, doubling, 60s cap, 5 attempts, in-flight guard) as a pure time-injected state machine.
- **Bounded give-up**: if attempts exhaust *and* we had previously injected a gossip connection for that peer (dialability regressed → its state is stale), force-close its RPC connections so `PeerDisconnected` fires and the peer is dropped until rediscovery — driving the existing 0→1 machinery on return. Peers we could never dial back (NAT'd inbound, no relay) keep their healthy inbound connections and just cool down.
- Superseded injected connections close after a 10s grace (both sides must adopt the replacement first; iroh keep-alives mean they'd otherwise leak one per refresh).

## Tests

- 10 deterministic unit tests on the backoff/schedule state machine (injected time).
- 2 new integration tests: `remesh_after_peer_restart` (kill a peer's transport, restart under the same identity, assert the survivor re-establishes gossip and raw delivery — the issue's acceptance shape) and `sweep_refresh_preserves_gossip_delivery` (~6 sweep cycles at 500ms, asserts bidirectional delivery — guards the assumption that injection is non-disruptive on a healthy mesh). Re-run 5× consecutively, stable.
- True half-dead injection (send loop dead behind live QUIC) is not deterministically reproducible from outside iroh-gossip without faking its wire protocol; the sweep path that cures it is exercised by the tests above.

Results: `cargo test -p p2p --features iroh-transport` 367 passed; `cargo test -p integration-test --test p2p_iroh` **229 passed, 0 failed**; workspace tests green; clippy `-D warnings` clean; fmt clean.

## Operational notes

- Acceptance check on the fleet after a pin bump: warns decay under backoff, heal on peer return, zero warns at steady state within one refresh interval; watch `gossip path refreshed` / `heal exhausted` lines.
- Sweep cost: ~1 QUIC dial per connected peer per interval per side (trivial at 16 peers); `DEFRA_P2P_GOSSIP_HEAL_INTERVAL_SECS` is a no-deploy knob if telemetry shows churn.
- defra-agent embeds via `crates/embedded`, which reads the env override automatically — no P2PConfig change needed for defaults.

### DIFF
--- a/crates/cli/src/commands/start/server_p2p.rs
+++ b/crates/cli/src/commands/start/server_p2p.rs
@@ -765,6 +765,7 @@ impl Node {
                 discovery: Self::iroh_discovery(config)?,
                 bind_port: config.net.iroh_bind_port,
                 bind_addr: config.net.iroh_bind_addr,
+                gossip_heal: p2p::iroh::GossipHealConfig::from_env(),
             })
             .await
             .map_err(Error::P2P)?;
--- a/crates/defra-node/src/lib.rs
+++ b/crates/defra-node/src/lib.rs
@@ -1346,6 +1346,7 @@ impl NodeBuilder {
             discovery: config.discovery.clone(),
             bind_port: Some(config.port),
             bind_addr: config.bind_addr,
+            gossip_heal: p2p::iroh::GossipHealConfig::from_env(),
         };
         let (command_tx, iroh_events, replicator_registry, endpoint_task) =
             p2p::iroh::spawn_endpoint(iroh_config)
--- a/crates/embedded/src/node_p2p.rs
+++ b/crates/embedded/src/node_p2p.rs
@@ -398,6 +398,7 @@ where
         discovery: config.discovery.clone(),
         bind_port: config.bind_port,
         bind_addr: config.bind_addr,
+        gossip_heal: p2p::iroh::GossipHealConfig::from_env(),
     };
     let (command_tx, event_rx, replicator_registry, endpoint_task) =
         p2p::iroh::spawn_endpoint(iroh_config)
--- a/crates/ffi/src/collection/view.rs
+++ b/crates/ffi/src/collection/view.rs
@@ -96,7 +96,7 @@ pub unsafe extern "C" fn add_view(
 
             // Parse the GQL query into a Select, matching Go's `addView` behavior:
             // Go wraps query as `query { <input> }` then parses to request.Select
-            let wrapped_query = format!("query {{ {} }}", &query_str);
+            let wrapped_query = format!("query {{ {} }}", query_str);
             let selects = query::parse_query_with_limits(&wrapped_query, None, query_limits)
                 .map_err(|e| format!("failed to parse view query: {}", e))?;
             if selects.is_empty() {
--- a/crates/p2p-adapter/src/iroh.rs
+++ b/crates/p2p-adapter/src/iroh.rs
@@ -1056,6 +1056,7 @@ mod tests {
             discovery: IrohDiscoveryConfig::Disabled,
             bind_port: None,
             bind_addr: Some(IpAddr::V4(Ipv4Addr::LOCALHOST)),
+            gossip_heal: Default::default(),
         }
     }
 
--- a/crates/p2p/src/iroh/endpoint.rs
+++ b/crates/p2p/src/iroh/endpoint.rs
@@ -60,7 +60,11 @@ pub(super) struct ActiveSync {
 pub(super) type SpawnedTasks = Arc<parking_lot::Mutex<Vec<JoinHandle<()>>>>;
 
 pub(super) fn track_task(spawned_tasks: &SpawnedTasks, task: JoinHandle<()>) {
-    spawned_tasks.lock().push(task);
+    let mut tasks = spawned_tasks.lock();
+    // The heal sweep (#1092) pushes tasks on a timer, so finished handles must
+    // be dropped here or the vec grows without bound over a node's lifetime.
+    tasks.retain(|task| !task.is_finished());
+    tasks.push(task);
 }
 
 async fn shutdown_tracked_tasks(spawned_tasks: SpawnedTasks) {
--- a/crates/p2p/src/iroh/endpoint.rs
+++ b/crates/p2p/src/iroh/endpoint.rs
@@ -25,10 +25,23 @@ use super::endpoint_config::{
 };
 use super::endpoint_rpc::ConnectionCache;
 use super::endpoint_streams::handle_incoming;
+use super::gossip_heal::{self, GossipHealer};
 use super::peer_map::{parse_endpoint_id, PeerMap};
 use super::protocols;
 
 const MAX_COMMAND_BATCH: usize = 16;
+
+/// Shared handles to the endpoint's long-lived state, cloneable into spawned
+/// tasks (dials, incoming connections, gossip heals).
+#[derive(Clone)]
+pub(super) struct EndpointResources {
+    pub(super) endpoint: Endpoint,
+    pub(super) gossip: Gossip,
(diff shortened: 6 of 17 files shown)


jackzampolin:
**Follow-up push (`36e191fd`):** review of the heal lifecycle found one retention issue — `track_task` never pruned finished `JoinHandle`s until shutdown, and the periodic sweep now pushes tasks on a timer (~2 handles/peer/min at idle, unbounded on a long-lived node). Fixed by retaining only unfinished handles at the push chokepoint. All other lifecycle paths checked out: in-flight guard prevents duplicate heal loops, backoff caps and resets only after a verified-healthy path, shutdown aborts + drains heal tasks, departed peers are pruned and their injected connections closed.

**On the failing `Integration (go-compat)` check — pre-existing flake, unrelated to this branch:**
- The failing test (`parity_mixed_fields_3node_mixed`, views=7 vs 17: one Go node's counter delta lost) runs Go nodes over **libp2p**; this branch touches only `crates/p2p/src/iroh/` plus iroh-config plumbing.
- Main failed the **same job** on the sibling test `parity_indexed_lww_mixed` at `1b78bd21` ([run 28805022117](https://github.com/sourcenetwork/defradb.rs/actions/runs/28805022117)) — before this branch existed.
- Locally on this branch with the pinned Go binary (`6c874754`): the failing test passes **5/5**, and the full CI parity selection passes **11/11**.
- The CI workflow already documents this parity family as timing-sensitive (`parity_counter_storm_mixed` excluded for #1043); the failure shape matches the known load-dependent lost-delta class over libp2p gossip.

Local validation on `36e191fd`: `cargo test -p p2p --features iroh-transport` green, `p2p_iroh` integration 229/229 (incl. both gossip-heal tests), fmt clean, `cargo clippy --all -- -D warnings` clean.

jackzampolin:
**Review findings addressed (`1e638931`):**

1. **Give-up path now closes every RPC connection.** `PeerMap` retained only the most recently stored handle (overwritten per connection), so with multiple live connections the close couldn't drive the count to zero and `PeerDisconnected` never fired. `ConnectionInfo` now holds all live handles (already-closed ones pruned on insert), and `close_peer_connections` closes each. Regression test `close_peer_connections_closes_every_retained_handle` dials two real connections and asserts both are closed — it fails against the old single-handle behavior.
2. **Publish-only nodes heal.** The connect-time heal and the sweep no longer skip when there are no persistent subscription senders — ephemeral publishers (`publish`/`publish_raw` without subscribe, the KMS `_response` shape) ride the same per-peer gossip send path. The topic rejoin is simply a no-op with an empty sender list. New integration test `sweep_refresh_covers_publish_only_node`.
3. **Superseded-connection cleanup is now exercised.** The close grace moved into `GossipHealConfig` (`superseded_close_grace`, default unchanged at 10s); the sweep test runs with a 750ms grace so several delayed closes fire mid-test, and it asserts no `PeerUnsubscribed` churn plus continued bidirectional delivery.
4. **Sub-100ms refresh intervals no longer panic.** `tick_period` raises the clamp's upper bound to the floor, keeping the bounds ordered; unit-tested.

Also in `4435f0e7`: the red **Lint** job on the previous push was CI's stable toolchain moving to 1.97, whose new `useless_borrows_in_formatting` lint flags a pre-existing line in `crates/ffi/src/collection/view.rs` (untouched by this branch). Fixed the one instance the Lint job checks; verified `cargo +1.97.0 clippy --all -- -D warnings` clean locally.

Validation: p2p unit tests 369/369, full `p2p_iroh` suite 230/230, gossip-heal tests stable across 5 consecutive runs, fmt clean, clippy clean on both 1.95 and 1.97.

jackzampolin:
**Test-gap note addressed (`9c7217ce`):** `sweep_refresh_covers_publish_only_node` previously passed regardless of the empty-subscription guard — node 0's healing injects fresh connections into node 1 either way, and the half-dead state can't be induced from outside iroh-gossip. Took the suggested probe route: `GossipHealConfig` gains an optional `refresh_probe` counter (`None` in production), and the test asserts the publish-only node performs **≥ 2** refreshes — the connect-time (0→1) refresh plus at least one sweep-driven refresh — so a regression in either empty-subscription path fails it.

Verified negatively by temporarily restoring the old guard: full guard → probe = 0, fails; sweep-only guard (connect-time heal active) → probe = 1, still fails.

Validation: gossip-heal tests green 3× consecutively, p2p unit tests 369/369, fmt clean, `cargo +1.97.0 clippy --all -- -D warnings` and `-p p2p --features iroh-transport --all-targets` clean.

Issue #1094, opened 2026-07-07
test-infra: plain and --features iroh workspace builds race on the same target/debug/defradb path (last-build-wins)


jackzampolin:
## Summary

The integration-test harness builds the workspace binary for both feature sets to the **same** output path (`test-infra/src/binary.rs` `build_from_workspace` → `target/debug/defradb`). When a single test invocation mixes iroh and non-iroh tests (e.g. `cargo test -p integration-test --test p2p` with no filter), whichever build ran last wins, and tests for the other feature set run against the wrong binary.

## Symptom

In a combined `--test p2p` run, the 5 `filtered_replication::*_iroh` tests fail with "iroh transport not enabled". This reproduces on unmodified `main` — it is a harness race, not a product regression. The same tests pass when run in a split invocation:

```
cargo test -p integration-test --test p2p -- --skip go_ --skip _iroh   # 66 passed
cargo test -p integration-test --test p2p -- _iroh                     # 5 passed
```

Collateral: unrelated tests in the same broken run (e.g. `document::rust_p2p_document`) can fail as fallout and pass standalone, which makes real failures harder to triage.

## Suggested fix

Give each feature set its own artifact path (e.g. `target/debug/defradb-iroh` vs `target/debug/defradb`, or a per-feature-set cache dir like the external-ref builds use) and have the harness select by required features. Found while validating #1093.

PR #1100, opened 2026-07-09
fix(p2p): bound outbound push backlog, retire task handles, persist pending-DAG acks (#1099)

### LINKED ISSUE #1099: P2P: bound outbound push backlog and retire coordinator task handles
## Summary

The outbound PushLog semaphore bounds active network sends, but it does not bound resident push work. The coordinator currently materializes a request/DAG batch and spawns one task per `(write, peer)` before the task acquires `push_semaphore`. `SyncShutdownHandle` then retains every registered `JoinHandle` until shutdown.

Under sustained multi-wave fan-out, waiting futures can therefore retain request/DAG buffers proportional to total writes × peers, while completed task handles accumulate for the lifetime of the node. Lowering `max_concurrent_push_tasks` can increase backlog residence time. This leaves a resource-exhaustion path capable of shutting down a node even when in-flight sends respect the configured semaphore bound.

This is intentionally separate from, and complementary to, #1088:

- #1088 covers inbound PushLog admission, pending-DAG capacity/fairness, and retry-walk CPU.
- This issue covers outbound spawned-work admission/lifetime and the missing runtime contract needed to prove sustained-load resource bounds.

Downstream incident and operator mitigation: sourcenetwork/defra-agent#630 and sourcenetwork/defra-agent#670.

## Current shape

At the current downstream pin (`63c0be62`):

1. `push_dag_to_replicators`, `push_to_replicators`, and `push_document_to_replicators` build work and call `spawn_background_task` before acquiring the semaphore.
2. The spawned future waits on `push_semaphore.acquire()` while owning its request payload.
3. `SyncShutdownHandle::register_task` appends every handle to `background_tasks`; completed handles are only drained during shutdown.
4. `max_concurrent_push_tasks` consequently bounds active sends, not queued work, queued bytes, spawned tasks, or retained handles.
5. Pending-DAG registrations that justify a successful PushLog reply remain process-local; crash/restart recovery depends on unmodeled anti-entropy/re-drive behavior.

## Proposed direction

- Admit work into a bounded queue before spawning per-item work or expanding/capturing full DAG payloads.
- Prefer a fixed worker pool with compact work identifiers; reload blocks from the blockstore when a worker owns the item.
- Define an explicit full-queue outcome: nack/backoff, coalesce, or durable retry. Full capacity must not allocate another waiting task.
- Remove completed handles promptly, or keep only the fixed worker handles.
- Add per-peer scheduling/fairness so slow peers cannot monopolize global capacity.
- Expose a stable admission/diagnostic snapshot containing at least:
  - configured queue item/byte limits
  - queued items and queued bytes
  - active pushes
  - retry depth
  - pending DAG depth
  - live worker/task count
  - overload nack/coalesce counters
- Resolve the success-ack/restart contract: persist and recover pending registrations, delay success until merge, or specify and test the anti-entropy path that guarantees re-drive after restart.

## Formal safety contract

For arbitrary-length input traces, P2P-owned resident state should satisfy bounds independent of total arrival count:

- `queued_items <= queue_capacity`
- `queued_bytes <= queue_byte_capacity`
- `active_pushes <= worker_capacity`
- `retry_items <= retry_capacity`
- `pending_dags <= pending_dag_capacity`
- `available_permits + active_pushes = worker_capacity`
- registered task handles are fixed-worker handles or otherwise bounded
- every success ack is backed by merged state or restart-recoverable pending work
- overload produces an explicit observable outcome without increasing resident work

Liveness should be conditional: admitted work for responsive peers eventually succeeds or exhausts retry policy; excess work may be rejected/coalesced, but healthy peers must not be starved by slow peers.

## Acceptance

- [ ] Fault-injection test with at least one nonresponsive peer, healthy peers, and sustained writes well beyond every configured cap.
- [ ] Queue depth, queued bytes, active pushes, retry state, pending DAGs, and task handles stay within their configured bounds throughout the test.
- [ ] Healthy peers continue to make progress while the slow peer times out/retries.
- [ ] Excess work receives an explicit nack/coalesce/durable-retry outcome.
- [ ] Completed tasks do not accumulate handles for the process lifetime.
- [ ] Crash after success-acking a missing DAG, then restart; the DAG merges through a specified recovery path or the pre-crash request was not success-acked.
- [ ] Diagnostics are exported through the P2P operations surface so downstream runtimes can conformance-test and alert on the effective state.
- [ ] Add a formal/model-checking fence with green and intentionally failing configurations for queue growth, permit leak, handle retention, starvation, and restart loss.

## Related

- #1088 — inbound hub PushLog admission meta
- #885 — pusher replay/spin against unhealthy transport
- sourcenetwork/defra-agent#630 — Amy hub saturation incident and fleet acceptance
- sourcenetwork/defra-agent#670 — operator admission controls and explicit obligation-model boundary

jackzampolin:
Closes #1099. Downstream tracker: sourcenetwork/defra-agent#673. Incorporates the Amy v0.6.7 canary requirements from sourcenetwork/defra-agent#630 (per-peer isolation, no pre-admission allocation, per-peer retry backoff, live occupancy diagnostics).

## Outbound bounding (the #1099 core)

- `broadcast.rs` no longer spawns a task per `(write, peer)` or materializes DAGs at enqueue time. Pushes are admitted as compact job specs (head block + ids) into a new **`PushBacklog`** bounded by item count *and* resident bytes, drained by a **fixed worker pool** sized by `max_concurrent_push_tasks`. Workers own DAG expansion + signing, so resident payloads are bounded by the worker count. `push_semaphore` is deleted — workers are the permits, so permit conservation holds by construction.
- Queue-full is an explicit outcome: counted, WARN-logged, and handed to the persisted retry ladder. Identical `(peer, cid)` jobs coalesce (never `(peer, doc)` — dropping a distinct field-block head would break the #1043 encrypted-LWW KMS DEK trigger).
- `SyncShutdownHandle::register_task` prunes finished handles on every registration; the only long-lived push handles are the fixed workers.

## Per-peer isolation (canary reqs 1 & 3 — `max_concurrent_push_tasks=1` must stay safe)

- Per-peer FIFO queues, round-robin dispatch, per-peer active cap (`max_active_pushes_per_peer`, default 4).
- Escalating per-peer **failure cooldown** (1s→64s, reset on success): a dead peer costs one probe per window instead of a full send timeout every rotation.
- Per-peer **queued-share quota** (¼ of the item budget): a dead peer's parked jobs cannot squat the global queue.
- The persisted retry loops (cli libp2p/iroh, defra-node, embedded ×2) previously ground serially through a nonresponsive peer's docs with untimed sends — fault injection showed this starving healthy peers harder than the queue itself. Each retry send is now bounded (15s) and a peer's pass stops on first failure.

## Success-ack/restart contract (resolved, not fenced)

A hub success-acks a missing-links push on the strength of a process-local pending-DAG registration, and the pusher then deletes its persisted retry record — a hub crash in that window silently lost the document forever (no automatic recovery path existed; verified). Push-originated registrations are now **persisted before the success reply** (`/p2p/pending_dag/` systemstore keys; a failed persist nacks the push fail-closed), deleted on resolve/expiry, **restored and re-driven at startup**, with a peer-connect re-drive healing entries whose providers had not reconnected at restore time. Nack-on-pending was rejected: pending registration is the *normal* path for every unfiltered live push, so nacking it would error the happy path and flap replicator health.

## Diagnostics (canary req 4 — occupancy, not just configured limits)

`GET /api/v0/p2p/sync/status` (P2pPeerInfo permission; `P2POperations::sync_status` on all four adapters) returns `p2p::sync::SyncStatus`: configured/effective capacities, queued items/bytes, active workers, **per-peer backlog rows with consecutive-failure and cooldown state**, overload counters, pending-DAG depth/capacity, retained task handles, retry counters. New CLI knobs: `DEFRA_P2P_PUSH_QUEUE_CAPACITY`, `DEFRA_P2P_PUSH_QUEUE_BYTES`, `DEFRA_P2P_MAX_ACTIVE_PUSHES_PER_PEER`.

## Formal fences

- `proofs/tla/PushBacklog.tla` — green config + intentional reds for per-item spawn growth (`INV_QueueBounded`), permit leak (`INV_PermitConservation`), handle retention (`INV_HandlesBounded`), and slow-peer starvation (`LIVE_HealthyProgress`).
- `proofs/tla/PendingDagRestart.tla` — `INV_AckBacked` (a success ack implies merged ∨ pending ∨ persisted); process-local-only registration is the red (`Send → HubAdmit → Crash`).
- Registered in `run-all.sh`: all 91 runs match their verdicts.

## Tests

- New deterministic fault-injection unit tests: nonresponsive mock peer (stalled sends), fairness under per-peer caps, cooldown escalation/reset, quota, handle pruning, persistence roundtrip/restore/fail-closed-nack, re-drive selection.
- `tests/p2p_backlog.rs` (own env-injected binary): 1 pusher → 1 healthy + 1 **SIGSTOPped** target, 120 writes ≫ caps; asserts every sampled bound holds throughout, overload counters advance, the dead peer is visible in per-peer diagnostics, the healthy target converges during the freeze, and the frozen target converges after SIGCONT (durable overflow outcome).
- `tests/p2p_admission_restart.rs`: pushers frozen once a success-acked registration provably occupies the 1-slot pending map, hub hard-killed and respawned on its rootdir (redb + file keyring — the harness defaults of memory store + ephemeral peer key cannot survive restart); gates on the restore log and full completeness.

## Gates

`cargo test` workspace 4228/0 · `clippy --all --all-targets -D warnings` clean · `fmt --check` clean · TLA 91/91 · `p2p_admission` pass (50s, was 465s — the retry-loop fix) · `p2p` suite 66 default + 24 go_* (Go binary at `GO_COMPAT_COMMIT` on PATH) + 5 iroh (after the known #1094 feature-race rebuild) all pass · `p2p_iroh` 227/227.

Notes: no overlap with #1095/#1096 files. defra-node/embedded deliberately keep `SyncConfig` defaults for the new knobs (their configs have no `Default` impl; adding required fields would force downstream churn) — exposing them plus `sync_status` through the embedded API is the defra-agent#673 follow-up.


jackzampolin:
All four findings were valid — addressed in 76e85d0d. Thanks for the careful pass; #2 and #3 in particular were real re-introductions of the loss class this PR exists to close.

**1 (P1) — record deleted at DagReady, before the merge.** Agreed, and per the follow-up guidance the bar is crash-resumability. Records are now deleted **only when the root is successfully marked merged** — the hook lives in `SyncManager::mark_as_merged`/`mark_batch_as_merged` (every DagReady-driven merge, including terminal skips and `MergedButNotMarked` retries, marks through these), gated by an in-memory persisted-roots set so merged field blocks don't pay a delete transaction. Crash after DagReady: the record survives, and startup resync re-registers the complete-but-unmerged DAG and re-drives it straight through `retry_pending_dag`. The unit test that codified deletion-at-DagReady is rewritten to assert the record survives DagReady and clears only at merge-mark (`pending_registration_persists_until_marked_merged`).

**2 (P1) — `try_send` drops overflow handoffs.** `report_push_failure` is now an async `send().await`: a full failure channel backpressures the reporter (worker or enqueue path) instead of discarding the retry record; only a closed channel (recorder gone at shutdown) logs and releases. New test `report_push_failure_backpressures_instead_of_dropping` pins the behavior.

**3 (P1) — TTL eviction deleted the durable obligation.** TTL eviction (and `clear_pending_dag`, and the event-channel cleanup path) now frees the in-memory slot only. Two additions make that sound: durable records carry their own cap (4× `max_pending_dags`) at which new registrations are **nacked** — the hub refuses an obligation it can't keep rather than accepting and later dropping it, and the pusher still owns retry state — and a single-flight resync sweep (startup + peer connect, with a cheap steady-state early-exit) deletes merged records and re-registers + re-drives evicted ones, so recovery does not require a restart. Tests: `clear_keeps_durable_record_until_merge`, `resync_re_registers_cleared_records`, `durable_cap_nacks_new_registrations`.

**4 (P2) — permanent head-of-line blocking in retry passes.** All five loops (cli libp2p/iroh, defra-node, embedded ×2) now distinguish: a **timeout** still ends the peer's pass (unreachable peer, bounded cost), while **fast rejections** consume a budget of 3 before stopping — so one permanently rejected doc at the head of the key order can no longer starve every later retry for that peer.

Re-verified: p2p crate suite, workspace tests (4232/0), clippy `-D warnings` clean, fmt clean, and all three fault-injection integration binaries green — `p2p_admission_restart` 92s, `p2p_backlog` 69s, `p2p_admission` 17s. `PendingDagRestart_DESIGN.md`'s conformance-fence section now states the record-lifetime refinement (the model's atomic `Resolve` = Rust's marked-merged, not DagReady).

jackzampolin:
Round-2 findings addressed in 51d46296.

**1 (P1) — resync could stall after capacity frees.** The resync is now a **periodic bounded drain**: every runtime (cli libp2p/iroh, defra-node, embedded ×2) replaces its one-shot restore spawn with `run_pending_dag_resync(60s)`, which sweeps until shutdown — records skipped at capacity are re-driven as slots free, with no dependency on peer reconnects or restarts (single-flight guard + steady-state early-exit keep idle ticks free). Both masking bugs you called out are fixed: the steady-state early-exit and the per-record skip are now **TTL-aware**, so an expired-but-not-yet-evicted in-memory entry no longer hides its record — the sweep evicts it and re-registers the record with a fresh TTL and recomputed missing set. New test `resync_drains_records_skipped_at_capacity_once_slots_free` pins the exact scenario (record held at capacity → slot frees → next sweep re-drives it; at map cap 1 the drain progresses one freed slot per sweep, which is the intended bounded behavior).

**2 (P2) — starvation shifted from one to three.** Agreed — the budget alone just moved the horizon. Retry passes now **rotate their starting document each pass** (advancing counter in the inline cli/defra-node loops, a per-fn atomic in the embedded helpers, applied to all five), so the rotation walks the key order over successive passes and every document eventually gets attempts regardless of how many permanently rejected ones sit at the head. Combined with the budget, a pass costs at most 3 fast rejections or 1 timeout.

**3 (P2) — durable backlog invisible.** `SyncStatus` (and therefore `GET /api/v0/p2p/sync/status`) now reports `persisted_pending_dags`, `persisted_pending_dag_capacity`, and `pending_resync_in_flight` alongside the in-memory `pending_dags`, so a node holding or nacking durable obligations can't present as idle. Manager accessors are covered in the durable-cap unit test.

Re-verified: p2p suite green (9 persistence tests incl. the new drain case), workspace 4233/0, clippy `-D warnings` clean, fmt clean, and the three fault-injection binaries re-run green — `p2p_admission_restart` 84s, `p2p_backlog` 69s, `p2p_admission` 41s.

jackzampolin:
Round-3 findings addressed in 43a521a4 (note: the PENDING inline review hasn't been submitted, so I'm responding to the summary table).

**bug — retry break-after-3 starves later docs.** This is fixed at HEAD by 51d46296, which the table may predate: every pass **rotates its starting document** (advancing counter in the inline cli/defra-node loops, per-fn atomic in the embedded helpers), so the head of the stable key order changes each pass. With rotation, a doc at index *i* is first-in-order within *len* passes regardless of how many permanently rejected docs exist — the budget bounds per-pass cost (3 fast rejections or 1 timeout), the rotation guarantees eventual coverage. If the inline comment identifies a residual hole in the rotated variant, happy to dig further once the review is submitted.

**suggestion — durable-cap TOCTOU.** Fixed: admission is now an atomic check-and-reserve on the persisted-roots set under the write lock (`Reserved`/`AlreadyPresent`/`AtCapacity`), with the reservation released if the subsequent `put` fails. The cap is now hard under concurrent PushLogs.

**suggestion — `send().await` can stall the pool.** Partly by design: backpressure is where the losslessness the round-1 review required comes from — the alternatives are dropping (the original P1) or unbounded buffering. What was missing is observability, so the handoff now waits via cancel-safe `reserve()` with a 5-second warning tick (`"Push failure recorder is backlogged"`): still lossless, but a backed-up recorder is visible in logs rather than presenting as a silent stall. The recorder's work is small KV transactions, and no cycle exists from recorder back to the workers, so sustained saturation indicates a storage problem the operator should see anyway.

**suggestion — double `job_done` desync.** Hardened: a `job_done` with no matching active slot is now ignored (plus `debug_assert` in dev builds), so a caller bug can't desync `active_jobs` vs the per-peer map or double-charge a cooldown. The contract is documented on the method and pinned by `spurious_job_done_is_ignored`; the one test that previously abused the API to force cooldown escalation now escalates through real job cycles.

**suggestion — TLA fidelity (Resolve collapses resolve+delete).** Acknowledged as a documented refinement rather than a model change: `PendingDagRestart_DESIGN.md` states that the model's atomic `Resolve` corresponds to Rust's *marked-merged*, not DagReady. Since the implementation deletes strictly later than the model, every reachable implementation state maps to a model state where the record is still held — `INV_AckBacked` (`acked ⇒ merged ∨ pending ∨ persisted`) can only be easier to satisfy, so the abstraction is sound in the safe direction. Splitting Resolve into emit+merge steps would grow the state space without adding a checkable failure mode.

**nits.** Both fixed: the `pending_store` module docs now state the merge-time deletion contract, and `retained_task_count` prunes finished handles before counting so the status endpoint can't overstate live work after a completion burst.

Re-verified: workspace 4234/0, clippy `-D warnings` clean, fmt clean, p2p suite green (15 backlog + 9 persistence tests), and the three fault-injection binaries re-run green — restart 99s, backlog 129s, fan-in 40s.

jackzampolin:
Round-4 items addressed in 4b2ee03b.

**[P2] Global rotation aliasing.** Correct — with N due peers a shared cursor advances by N per sweep, so a peer whose `docs.len()` divides N gets the same offset every pass. Adopted your suggestion exactly: all five loops now rotate by the peer's own persisted `retry_info.num_retries % docs.len()`. It's per-peer, survives restarts, and advances precisely on that peer's failed passes (a successful pass clears the record anyway), so every document eventually leads a pass with zero new state.

**Nit 1 — soft cap after restart.** Fixed: `install_pending_dag_store` is now async and hydrates the persisted-roots accounting from the store *before returning* (all five runtimes await it before serving), so the durable cap is hard from the first PushLog after restart instead of after the first spawned sweep. Pinned by `install_hydrates_durable_accounting_before_first_push`.

**Nit 2 — reserve/put race orphans.** Made the sweep self-healing in both desync directions rather than trying to close the race window itself: every 10th sweep is forced past the steady-state early-exit, so an orphan record (in the store, missing from the set) is rediscovered within ~10 minutes even under permanent steady-state traffic; and the sweep prunes accounting entries with neither a record nor a live in-memory entry, so the reverse leftover can't hold durable-cap headroom forever. In-flight reserve→put windows always hold an in-memory pending entry, so they're never pruned; the reconcile block acquires roots-before-pending to match the early-exit's lock order. Pinned by `forced_sweep_rediscovers_orphan_records` and `sweep_prunes_stale_accounting_entries`.

Re-verified: workspace 4237/0, clippy `-D warnings` clean, fmt clean, 12 persistence + 15 backlog + 4 worker tests green, and the three fault-injection binaries green — restart 84s, backlog 129s, fan-in 17s.

Issue #1112, opened 2026-07-12
Cold-join/catch-up thundering herd: PushLog rejection has no sender backpressure — persisted retry turns admission rejects into a sustained distributed storm


jackzampolin:
## Summary

PushLog admission control (#1088/#1100) bounds the receiver but provides **no
sender-side flow control**: a `RejectedItems`/at-capacity rejection carries no
backoff signal, and the sender's persisted-retry (#1110) durably re-offers the
rejected push. When a node that is behind (fresh store, restored backup, or
hours of skew) joins a fleet of N peers, all N push their backlogs
simultaneously; the receiver's pending-DAG registry pins at capacity and every
subsequent PushLog is reject→re-offer→reject — a self-sustaining distributed
retry storm in which *processing rejections* becomes the DoS. Bounding held;
the system around it thrashed.

## Live evidence (2026-07-12 fleet bring-up abort, defra-agent#630 comment 4949290217)

Hub node ("Amy", 16 pairings), THREE scenarios, all saturated:

1. **v0.6.11 (defradb 99a4626d), FRESH store**: 792.7% CPU,
   `pending_dags=1000/1000`, 904 persisted DAGs, 1,793 retained background
   tasks. Log: 1,725 × `Pending DAGs at capacity, rejecting PushLog DAG
   registration … missing_count=N max=1000` within minutes (867 of them with
   an EMPTY doc_id — worth a look on its own), plus 132 × outbound
   `backlog full; deferring push to persisted retry`, plus ~750 ×
   `iroh_gossip::net: failed to send: connection task send loop terminated`
   across ~8 peers (send loops dying under load).
2. **v0.6.5 rollback with its original (12h-stale) store**: same signature,
   2,400+ pending-DAG rejections, 536% CPU, 5.8–6.7 GiB RSS.
3. **Seven sibling agents on the same host** (v0.6.11): saturated to a
   combined 1,088% CPU / 35.1 GiB RSS; their v0.6.5 rollback failed within
   2m24s.

Symbolized profile (dSYM matched): hot in `noq::connection::Connecting::new`
poll churn via `iroh::runtime`/tokio blocking pool + malloc/memcmp stubs —
connection/admission processing, not productive sync.

Sender side: N=15 peers on defradb `99a4626d` whose persisted-retry re-offers
rejected items; no observed decay in rejection rate over the window.

## Asks (protocol-level, in priority order)

1. **Rejection carries backpressure**: at-capacity PushLog rejection returns a
[...]

Issue #1088, opened 2026-07-06
P2P sync meta: hub PushLog admission control — capacity drop replies success (silent divergence); backpressure nacks regressed by fa4a84f7 (re-land #592)


jackzampolin:
## Summary

Under N-pusher fan-in (16 subagent-host replicator legs pushing to one coordinator — the defra-agent Amy-hub topology), the hub's sync manager saturates into a **restart-resistant, non-converging drop/re-push loop**: pending-DAG queue pinned at its 1000 cap, ~10 cores burned on DAG re-walks, QUIC paths dying within minutes of each restart. Observed fleet-wide 2026-07-01→07-06; downstream tracker with incident data: sourcenetwork/defra-agent#630.

Tracing the mechanism through the code shows this is not one defect but a cluster, anchored by a **regression**: the hub-side backpressure replies delivered for #592 were removed in `fa4a84f7` ("align iroh replay and shutdown with go model", 2026-04-18), while the pusher-side backoff that consumes those replies (`2630b24d`, #843/`ab18f758`) stayed. The pusher half of the flow-control loop is live, tested, and listening for a signal no production code path emits anymore.

All references below are at `1b78bd21` (current defra-agent pin).

## Mechanism

**M1 — Pending-DAG capacity drop replies *success* (silent divergence, worst finding).**
When a PushLog arrives with missing links and the pending map is at `MAX_PENDING_DAGS = 1000`, the hub logs a WARN, drops the registration, and returns `Ok(())` — which becomes `PushLogReply::success`:
- drop site: `crates/p2p/src/sync/manager/process/pushlog.rs:378-388`
- success reply: `crates/p2p/src/sync/coordinator/event_handler/pushlog.rs:117-120` (gossip-token path) and `:203-206` (two-stream path)

The pusher believes the push landed. Worse: the pusher's persisted retry loop treats the success reply as terminal and **clears the doc from its retry queue** (`crates/defra-node/src/lib.rs:1729` `remove_retry_doc`), then marks the replicator Active. The failure record is destroyed; the doc never merges on the hub and nothing will ever re-push it (until an unrelated later update). That is data divergence with the evidence laundered, and it is strictly worse than the CPU storm.

**M2 — The backpressure nack is dead code on the emit side (the #592 regression).**
[...]

Issue #1116, opened 2026-07-12
Design: converge sync ownership on the Go reference — head hints + receiver-driven pull; retire sender delivery state


jackzampolin:
## Summary — design issue

Three fleet incidents in five days (defra-agent#630 hub saturation, cold-join
herd, defra-agent#696 empty-store genesis storm) all live in the same seam:
**this implementation gave the SENDER ownership of sync state (push worker,
persisted delivery/retry ledger, per-peer delivery timeouts, pending-DAG-as-
admission), while ALSO running a receiver-side dag_fetcher — two ownership
models fighting.** The Go reference gives the receiver sole ownership: a
PushLog is one head block treated as an idempotent HINT (CID-verified,
per-CID single-flight, already-merged fast-path), and the receiver pulls the
missing DAG at its own pace (`processPushlogRequest`/`syncDAG`,
`internal/db/p2p/p2p.go` v1.0.0); the only sender persistence is a scheduled
retry coordinator with capped exponential backoff
(`internal/se/coordinator_retry.go`, 2s→4s→8s→16s cap).

Proposal: **converge on the reference ownership model** — hint + receiver-
pull — and retire the sender-side delivery machinery.

## Why each recent bug is a symptom

- Same-CID timeout parade (#630): impossible under single-flight + merged
  fast-path.
- Pending-DAG registry overflow (#696): the registry exists to track SENDER
  pushes; under receiver-pull it becomes the receiver's own bounded fetch
  queue.
- Retry ledger cannot replay collection commits (#1113): under hint+pull
  there is no delivery ledger to mis-key — #1113 dissolves rather than needs
  fixing.
- Sender re-offers rejected pushes forever (#1112): senders re-HINT on capped
  backoff; payloads move only by pull. The backpressure ask largely
  evaporates; the catch-up-mode ask becomes the only mode.

## Scope sketch (staged)

1. Kernel (separate issue, do first): single-flight + merged fast-path +
   cheap shed.
2. Make receiver pull authoritative: dag_fetcher (with #1095 rotation) owns
   all block acquisition; decide transport — extend the CAR request path vs
   wire the already-present iroh-bitswap dependency (want-list pacing for
   free). Comparative research landing as comments on this issue.
3. Demote sender: push worker sends head hints; retries = per-(peer,doc)
   scheduled re-hint with capped exponential backoff (mirror Go's
   coordinator); DELETE the CID-scoped delivery replay machinery.
[...]

PR #1119, opened 2026-07-12
fix(p2p): stop the empty-store genesis storm (#1112, #1113)

### LINKED ISSUE #1113: Persisted retry ledger cannot replay collection-commit (doc-less) CIDs — failed collection pushes are permanent
## Summary

The #1110 persisted retry ledger is **document-keyed** and cannot replay
**collection-commit** (doc-less) CIDs. When a collection-level push fails, it
fails permanently:

```
WARN p2p::sync::coordinator::push_worker: Collection-commit push failed;
document retry ledger cannot replay CID-scoped work
peer_id=<id> collection_id="<cid>" cid=Some(Cid(<cid>))   ×122 in minutes
```

Receivers never obtain those blocks → heads with missing parents → pending-DAG
registrations that can never complete. This explains the persistent
**empty-`doc_id`** pending-DAG rejections observed in BOTH fleet incidents:
1,760 of 3,517 rejects in the empty-store genesis run (defra-agent#696), and
867 in the earlier hub saturation (defra-agent#630) — it was visible then too.

## Repro

Trivial: the 19-node empty-store genesis in defra-agent#696 — only network/
membership/endpoint docs + collection registrations exist, and the storm forms
from this gap (plus the gossip direction filter, filed separately). A 2-node
version: register a collection, fail its first collection-commit push (peer
briefly at capacity), observe the ledger's inability to replay.

## Ask

Retry ledger keys collection-scoped work (collection_id + cid) alongside
document-scoped work; a failed collection-commit push is replayable and honors
the same latest/coalescing + backoff semantics. Regression test: collection
commit push rejected once → replayed → converges.

## Relations

#1110 (the ledger), defra-agent#696 (repro), #1112 (backpressure family),
defra-agent#630 (fleet thread).

### LINKED ISSUE #1114: GossipSub direction filter drops subscribed replication traffic in symmetric meshes
## Summary

The GossipSub direction filter drops messages that should be accepted, forcing
replication onto the (currently fragile) fetch path. On a symmetric 19-node
control mesh with EMPTY stores (defra-agent#696):

```
WARN p2p::sync::coordinator::event_handler::gossip: Dropping GossipSub message
outside accepted replication direction
peer_id=<id> topic=<cid> collection_id=<cid>
topic_matches_collection=true topic_matches_document=false
is_subscribed=true is_outbound_replicator_target=true          ×~330 in minutes
```

The node is SUBSCRIBED to the topic and the peer is a replicator target, yet
the message is discarded because the direction policy treats the edge as
outbound-only. In a mesh where control-plane collections (network/membership/
endpoint) replicate symmetrically, this silently halves propagation and
manufactures missing-link pressure (4,289 missing-link retries in the same
window).

## Ask

Either accept gossip for any topic the node is genuinely subscribed to
(subscription = intent to receive), or make direction rules per-collection so
symmetric control-plane collections are exempt from outbound-only filtering.
Add a counter for direction-filtered drops (it was only visible as log volume).

## Relations

defra-agent#696 (repro), #1112, defra-agent#630.

iverc:
Fixes #1113 and #1112. Rebased onto main, which now carries #1118 — see "Scope change" below.

Reported downstream as sourcenetwork/defra-agent#696: 19 fresh nodes forming a mesh on empty stores never converge. The three root causes compose into a **fixed point**, which is why bounding the receiver (#1088/#1100) held while the mesh around it thrashed.

## Scope change: #1114 is now main's (#1118)

This branch originally fixed #1114 by removing the outbound-target veto outright and making `add_replicator` stop auto-subscribing, so that "who receives" was expressed only by who subscribes. Main has since landed #1118, which fixes the same defect more conservatively: the veto stays, but is narrowed to fire only when the collection is *not* locally subscribed.

I dropped my version in favour of main's. It fixes the storm (both peers in a symmetric mesh are subscribed, so both accept), and it does so **without changing `add_replicator` semantics** — which is the safer trade for existing consumers who rely on auto-subscribe. What remains here is #1113 and #1112, which are still unfixed on main and which #1118 alone does not address.

Kept from the original work: a regression test for the payload that actually broke, a **doc-less collection commit**. Every gossip test upstream uses a non-empty `doc_id`, so they all have a document-topic fallback. Collection commits do not — the collection topic is their *only* delivery path, which is exactly why the veto was fatal for them and merely inconvenient for document updates.

## #1113 — the persisted retry ledger could not replay collection commits

The ledger is document-keyed, so a doc-less collection-commit CID was dropped at four independent points: it could not be recorded, scanned, swept, or replayed. A failed collection-commit push therefore failed **permanently**. Receivers kept heads whose parents never arrived, so pending-DAG registrations could never complete — the empty-`doc_id` reject class seen in both fleet incidents (1,760/3,517 here; 867 in defra-agent#630).

- new CID-scoped keyspace `/rep/retry/commit/{peer}/{collection}/{cid}` alongside the document keyspace
- `RetryScope { Document, CollectionCommit }` on the persisted record (`#[serde(default)]`, so existing ledgers deserialize unchanged)
- a CID replay executor (`retry_collection_commit`, both libp2p and transport), dispatched from every replay loop
- a missing block is an **error**, not a no-op: the document executor returns `Ok(())` when a doc has no heads, and reusing it here would have ack'd the retry and deleted the ledger record — silently dropping the very obligation being replayed

## #1112 — a capacity rejection was answered with rate-limit policy

Rate limiting is a *pacing* condition; a saturated receiver is *peer-wide and structural* — it cannot accept any new root until it drains. Conflating them turned one logical push into 11 resends in ~3.3s, each costing the receiver a block write plus a full DAG traversal, all guaranteed to fail. The cooldown was keyed per-CID while the condition is per-peer, so every other CID got a fresh burst, and 19 senders re-fired in lockstep.

- a distinct `AT_CAPACITY_MESSAGE` nack sentinel (byte-exact, like the rate-limit one)
- the sender stops at the first capacity nack instead of walking the rest of its queue into a full receiver
- the whole **peer** is parked with an escalating, jittered cooldown, and `peer_capacity_parks_total` is exported so backpressure engaging is observable rather than inferred

Main's capacity tests asserted the rate-limit sentinel for capacity rejections — that assertion *was* the bug, so they now assert the capacity sentinel and are renamed accordingly.

## Gates

`cargo fmt --all --check`, `cargo test -p p2p` (409), `cargo test --workspace --exclude integration-test --exclude ffi-test`, `cargo check --workspace --all-targets`, and `cargo check -p defra-node --features p2p` — all green.

That last one is not currently in CI: the workspace check builds `defra-node` with `p2p` **off**, so a missing dependency under the feature only surfaces in a downstream consumer that enables it (it did — hence 2bc8d70f). Worth adding a feature-gated check to CI regardless of this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,14 @@ jobs:
       - run: cargo clippy -p cli --features otel --all-targets -- -D warnings
       - run: cargo clippy -p defra-node --features otel --all-targets -- -D warnings
 
+      # `p2p` is off by default on both crates that gate on it, so no other job
+      # ever builds that dependency set. A dependency missing from the feature
+      # list therefore compiles fine here and only breaks in a downstream
+      # consumer that turns the feature on (which is how a missing `cid` on
+      # defra-node reached a release).
+      - run: cargo clippy -p defra-node --features p2p --all-targets -- -D warnings
+      - run: cargo clippy -p db --features p2p --all-targets -- -D warnings
+
       - name: Cleanup
         if: always()
         run: |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3697,6 +3697,7 @@ dependencies = [
  "async-trait",
  "axum",
  "blockstore",
+ "cid",
  "crypto",
  "db",
  "db-merge",
--- a/crates/cli/src/commands/start/server_p2p.rs
+++ b/crates/cli/src/commands/start/server_p2p.rs
@@ -1313,7 +1313,7 @@ impl Node {
                     if let Ok(rep_info) = p2p::ReplicatorInfo::from_bytes(&data) {
                         let pid = p2p::transport::PeerId::new(rep_info.peer_id_str().to_string());
                         let _ = coordinator
-                            .create_replicator_info(&pid, rep_info.clone(), false)
+                            .create_replicator_info(&pid, rep_info.clone(), true)
                             .await;
                     }
                 }
--- a/crates/cli/src/commands/start/server_p2p.rs
+++ b/crates/cli/src/commands/start/server_p2p.rs
@@ -622,16 +622,38 @@ impl Node {
                         // fast rejection only consumes a bounded budget so
                         // one permanently rejected doc at the head of the
                         // key order cannot starve the rest forever.
-                        match tokio::time::timeout(
-                            std::time::Duration::from_secs(15),
-                            retry_pusher.retry_doc(
-                                &retry_handle,
-                                peer_id,
-                                &retry.doc_id,
-                                &retry.collection_id,
-                            ),
-                        )
-                        .await
+                        // Collection commits are doc-less and replay by CID
+                        // (defradb#1113).
+                        let replay = async {
+                            if retry.is_collection_commit() {
+                                match retry.cid.parse::<cid::Cid>() {
+                                    Ok(cid) => {
+                                        retry_pusher
+                                            .retry_collection_commit(
+                                                &retry_handle,
+                                                peer_id,
+                                                &retry.collection_id,
+                                                &cid,
+                                            )
+                                            .await
+                                    }
+                                    Err(error) => Err(format!(
+                                        "unparseable collection-commit CID {}: {error}",
+                                        retry.cid
+                                    )),
+                                }
+                            } else {
+                                retry_pusher
+                                    .retry_doc(
+                                        &retry_handle,
+                                        peer_id,
+                                        &retry.doc_id,
+                                        &retry.collection_id,
+                                    )
+                                    .await
+                            }
+                        };
+                        match tokio::time::timeout(std::time::Duration::from_secs(15), replay).await
                         {
                             Ok(Ok(())) => {
                                 let _ =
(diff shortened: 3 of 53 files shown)


jackzampolin:
**Reviewed — strong PR, approve pending a small rebase.** Verified the three highest-risk points against the diff.

### Correctness (spot-checked)
- **#1113 replay honesty** — `retry_collection_commit` returns `Err` on a missing block instead of `Ok(())`, so the ledger never deletes an unfulfilled obligation. This is the exact ack-honesty principle (sibling of #1120) applied to collection-commit replay. Correct.
- **#1112 peer-park** — peer-wide, escalating (2s→64s), jittered cooldown keyed per-peer (not per-CID), with `peer_capacity_parks_total` exported. Directly fixes the per-CID-fresh-burst root cause; won't lockstep. Good.
- **`AT_CAPACITY_MESSAGE` vs `RATE_LIMITED_MESSAGE`** — the pacing-vs-structural distinction is the right call and byte-exact per the #1088 W1 matching contract.

### Blocker: rebase (mechanical)
`mergeable=CONFLICTING` against current main — main moved (post-#1117). Conflicts are localized to **3 files**: `crates/cli/src/commands/start/server_p2p.rs`, `crates/defra-node/src/lib.rs`, and `crates/p2p/src/sync/coordinator/push_worker.rs` (around `retry_push_txn_conflicts`). A rebase onto current `origin/main` should resolve them; no design change implied.

### Coordination notes (non-blocking)
- **error.rs vs #1121 (open):** both edit `backpressure_reply_message()` but add different arms (`PendingDagCapacity→AT_CAPACITY` here; `PushLogInFlight→RATE_LIMITED` there). Compatible; whichever lands second takes a one-line combine.
- **#1116 interaction (acknowledged):** this is the tactical genesis-storm unblock; the #1113 CID-scoped ledger (`/rep/retry/commit/…`) is on #1116 stage-3's deletion list (marker-plus-rederive replaces it) — noted on #1116. The #1112 peer-park is expected to survive the convergence.

Rebase + green → good to merge as the third leg of the genesis-storm fix (with #1117 kernel + #1118 direction filter).

jackzampolin:
Merged current `main` into this branch to clear the conflict (pushed the merge commit). **One semantic resolution in `push_worker.rs`:** main removed `pushed_cids` (it switched the receiver recovery grant to `collect_dag_cids`/`grant_cids` — the DAG-walk approach, which also decouples recovery from payload expansion for #1116 stage 3); your branch had renamed the result to the `send_outcome` `PushSendOutcome` struct for the peer-park logic. Kept `send_outcome` (downstream `send_outcome.at_capacity`/`.failed` depend on it) and dropped the now-unused `pushed_cids`. fmt + `clippy -D warnings` clean on `p2p`; the merged else-branch uses main's `grant_cids` + your `at_capacity` park together. Verify you're happy with the resolution. Will squash-merge once CI re-greens.

jackzampolin:
Re-merged main after #1121 landed — one conflict in `error.rs backpressure_reply_message()` (the overlap flagged earlier). Combined both: `PendingDagCapacity` keeps this PR's `AT_CAPACITY_MESSAGE` (structural saturation → sender parks the whole peer), and #1121's `PushLogInFlight` joins `RATE_LIMITED_MESSAGE` (transient in-flight = a short retry, not a park — lumping it into AT_CAPACITY would wrongly peer-park on a momentary single-flight suppression). clippy -D warnings + fmt clean on p2p. Merging on green.

Issue #1115, opened 2026-07-12
Receive-path convergence kernel: per-CID single-flight + already-merged fast path (Go reference guards)


jackzampolin:
## Summary

Adopt three receive-path guards from Go DefraDB's `processPushlogRequest`
(`internal/db/p2p/p2p.go`, v1.0.0) that make announcement floods structurally
cheap. These are the minimal convergence kernel toward the reference sync
model (capstone design issue filed separately) and the direct unblocker for
fleet bring-up.

## The guards (Go reference, code-verified)

1. **Per-CID single-flight**: Go wraps DAG sync in `processQueue.add(headCID)`
   / `doneOnce(headCID)` — overlapping syncs for one CID cannot run
   concurrently. Fleet evidence of its absence here: the same CID pushed/
   processed for 12 peers within 60µs, each attempt independently holding
   resources through a 30s timeout (defra-agent#630).
2. **Already-merged fast path**: Go checks `Blockstore().IsMerged(headCID)`
   and exits before any sync work. Re-announcements cost ~one lookup. Fleet
   evidence: 3,517 pending-DAG registrations rejected at capacity in minutes
   on an EMPTY-store mesh (defra-agent#696) — repeated announcements are
   currently full-cost.
3. **CID/content verification before processing** (Go verifies the advertised
   CID matches the pushed head block) — confirm the Rust path does this;
   add if absent.

## Asks

- Single-flight keyed on head CID across the entire receive path (pushlog
  registration + fetch trigger), receiver-side.
- Merged/known fast-path exit BEFORE pending-DAG registration or any
  allocation-heavy decode; at-capacity rejection should also happen before
  full decode (shed cheap).
- Counters: single-flight suppressions, fast-path exits (observability for
  the #630-class canaries).

## Acceptance

- Extend the #1108 storm harness: N peers announce the same head CID
  concurrently → exactly one sync executes, N−1 exits are cheap (assert
  counters + bounded CPU); repeat-announcement of a merged head is ~free.
- 19-node empty-store genesis scenario (defra-agent#696 repro) stays calm on
  the control mesh with this kernel + #1114 fixed.

## Relations

Reference: Go `processPushlogRequest` + `syncDAG`. Capstone: sync-ownership
convergence (filed separately). Siblings: #1114 (direction filter), #1112
(recast under capstone), defra-agent#696 (repro).

jackzampolin:
[...]

Issue #1120, opened 2026-07-12
Single-flight suppressed PushLog acks success optimistically — owner failure before registration can silently drop the suppressed pusher's copy


jackzampolin:
## Summary

Followup to #1117 (merged). The single-flight receive path suppresses a
duplicate PushLog for an in-flight CID and returns `Ok(())`, which the
request/response handler turns into a **success ack** to the pusher
(`build_pushlog_reply`: `Ok(())` → `PushLogReply::success`,
`event_handler/pushlog.rs`). That ack is **optimistic**: it is sent before the
owner has durably registered the head, so if the owner fails in the window
between `try_acquire_nowait` and pending-DAG registration
(`is_merged` → capacity shed → `verify_cid` → store → register), the head is
neither merged nor tracked — yet the suppressed pusher was told success and
will clear its retry-ladder entry.

This narrows the no-silent-divergence guarantee that #1116's **register-then-ack**
model rests on (ack = "durably registered or already have it", not "an owner is
probably handling it"). Recovery then depends solely on a future gossip
re-announcement, not the suppressed pusher's retry.

The pre-#1117 path avoided this by having the duplicate wait and re-check
`is_merged`, reprocessing on owner failure — so its ack reflected the real
outcome. #1117 traded that (correctly, for CPU) but did not preserve the ack
honesty.

## Repro gap

`conformance_same_cid_concurrent_announcements_are_idempotent` only fires
suppressed duplicates AFTER `pending_dag_count() == 1` (owner already
registered) — the safe ordering. The unsafe window (duplicate suppressed while
owner is pre-registration, owner then fails) is untested.

## Ask

On suppression, ack `success` only when the head is already registered/merged
(i.e. an owner has reached durable pending-DAG registration or the block is
merged). Otherwise return the same retriable/backpressure nack an at-capacity
push receives, so the suppressed pusher keeps the head in its ladder and
re-pushes. This preserves the register-then-ack invariant while keeping
#1117's cheap suppression (no reprocessing). Add a test for the
owner-fails-before-registration window.

## Relations

Followup to #1117 (merged). Invariant owner: #1116 (stage-1 acceptance note
added). Same family as the ack-honesty guards in the p2p_admission
no-silent-divergence conformance test.

jackzampolin:
[...]

PR #1121, opened 2026-07-12
fix(p2p): preserve retry for suppressed PushLogs

jackzampolin:
## Summary

- return a retriable backpressure nack when same-CID suppression happens before the owner establishes recovery state
- distinguish recovery-backed pending registrations from entries still being persisted or enqueued
- preserve cheap success acks for already-merged heads and established pending registrations
- remove the unreachable `BlockAlreadyMerged` sync event and update the receive-flow documentation/logging

## Root cause

PR #1117 suppressed concurrent same-CID PushLogs with an immediate success result. If the owner then failed before persisting or registering recovery work, the suppressed pusher could delete its retry record while the receiver held neither a merged block nor a recovery obligation.

This follow-up marks a pending entry recovery-backed only after its durable write succeeds, or after its fetch event is successfully enqueued when no durable store is installed. Earlier duplicates receive the standard rate-limited reply and retain their retry state.

## Tradeoff

The cheap capacity shed introduced in #1117 intentionally rejects before storing the pushed block. A retry therefore resends the block bytes, trading network bandwidth for bounded receiver CPU and storage work during overload.

## Validation

- `cargo test -p p2p`
- `cargo test -p integration-test --test p2p_admission -- --nocapture`
- `cargo check --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings`

Follow-up to #1117 and #1115.

### DIFF
--- a/crates/p2p/src/error.rs
+++ b/crates/p2p/src/error.rs
@@ -213,6 +213,10 @@ pub enum Error {
         max: usize,
     },
 
+    /// Another receive task owns this CID and has not established recovery state yet.
+    #[error("PushLog for CID {cid} is already being processed, retry later")]
+    PushLogInFlight { cid: String },
+
     /// Request rejected because the caller is not authorized.
     #[error("unauthorized: {0}")]
     Unauthorized(String),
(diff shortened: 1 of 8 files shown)


jackzampolin:
**Approve** — correct and complete fix for #1120. Verified the two crux points against the diff.

### Why it closes the window
- **`is_recovery_registered` flips true only after durable registration** — set via `mark_pending_dag_recovery_registered` after `has_durable_registration` is confirmed against `pending_store()` (durable path), or after the in-memory insert when no durable store is configured (the registration boundary in that config). A duplicate suppressed *before* that point sees `false` → falls to `is_merged` → not merged → **retriable nack**. Exactly the pre-registration window from #1120, closed.
- **`PushLogInFlight` is a genuine backpressure nack** — added to `backpressure_reply_message()` alongside `PendingDagCapacity`, returning `RATE_LIMITED_MESSAGE`. So `build_pushlog_reply` emits the same nack an at-capacity push gets, which the pusher already treats as keep-in-ladder-and-retry (the no-silent-divergence contract). The suppressed copy is preserved; no new pusher behavior needed.
- **Both branches tested** — a duplicate gets `Err(PushLogInFlight)` before recovery state exists ("must not ack before the owner establishes recovery state"), and an established registration *can* ack a duplicate cheaply. That is precisely the repro gap #1120 named.
- **Cleanup done** — dead `BlockAlreadyMerged` event removed (events.rs + handlers.rs); flow docs/logging fixed.

### Residual (non-blocking, correct-by-design)
A legitimate concurrent second push during the short pre-registration window now gets nacked and re-pushes on its backoff ladder — bounded, backed-off extra retry traffic in exchange for ack honesty. Right trade; it can't storm (the window is registration, not the full DAG fetch).

Marks the register-then-ack invariant intact, satisfying the #1116 stage-1 gate. Ready to mark non-draft + merge.

PR #1122, opened 2026-07-12
Resolve root-level __typename as the Query type

### LINKED ISSUE #697: GraphQL executor rejects { __typename } root introspection with 'collection not found: __typename'
Standard GraphQL liveness probes send `{ __typename }`; the executor treats it as a collection query and fails (`Query execution failed … error=collection not found: __typename`, ×38 during the #696 genesis window — pure noise + breaks generic health tooling). Ask: handle root-level `__typename` (and ideally basic introspection) before collection resolution. Evidence: runs/agent-log-bundles/20260712T033951Z-genesis-abort (amy.err tail).

iverc:
Fixes sourcenetwork/defra-agent#697.

`{ __typename }` — the standard GraphQL liveness probe — was parsed into a root `Select` with `collection_name == "__typename"` and fell all the way through `execute_select_internal` to `get_collection()`, which raised `collection not found: __typename`. Generic GraphQL tooling that probes a node before doing anything else got a hard error from a healthy node (38 occurrences in one empty-store window during the defra-agent#696 investigation).

`__typename` was already resolved *within* a selection set (`plan.rs`, `commits.rs`, `version.rs`); only the root case was missing. It is a meta-field naming the operation's root type, so it is now answered as `"Query"` before any collection resolution.

Handled in the data path rather than routed to the introspection engine, so it keeps composing with real data fields (`{ __typename Users { name } }`) and aliases keep working through the existing `output_name()` keying.

Scope: `__schema`/`__type` already work via the async-graphql dynamic schema. `mutation { __typename }` still returns a clean `Cannot query field "__typename" on type "Mutation".` parse error rather than `"Mutation"` — untouched here, since no client probes that way.

## Tests

Three in `runner::executor::tests`: the bare probe, an alias mixed with a data field, and the probe against a node with **zero collections registered** (the empty-store case from defra-agent#696). All three fail with the reported `collection not found: __typename` when the guard is removed.

`cargo fmt --all --check`, `cargo test -p query`, `cargo clippy -p query --all-targets`, `cargo check --workspace --all-targets` — green.

### DIFF
--- a/crates/query/src/runner/executor.rs
+++ b/crates/query/src/runner/executor.rs
@@ -795,4 +795,58 @@ mod tests {
             }))
         );
     }
+
+    #[tokio::test]
+    async fn root_typename_resolves_to_the_query_type() {
+        let collections = crate::parse_sdl("type Users { name: String }").expect("schema");
+        let runner = QueryRunner::new(MockFetcher::new(), collections);
+
+        let response = runner.execute(QueryRequest::new("{ __typename }")).await;
+
+        assert!(
+            !response.has_errors(),
+            "unexpected errors: {:?}",
+            response.errors
+        );
+        assert_eq!(response.data, Some(json!({ "__typename": "Query" })));
+    }
+
+    #[tokio::test]
+    async fn root_typename_is_aliasable_and_composes_with_data_fields() {
+        let collections = crate::parse_sdl("type Users { name: String }").expect("schema");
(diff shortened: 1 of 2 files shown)


jackzampolin:
Superseded by #1125 (merged to main as f750e009 on 2026-07-13), which resolves root-level `__typename` via the same introspection routing. Closing as redundant — thanks for the parallel fix.

Issue #700, opened 2026-04-10
Generalize zeorize Issue


JesseAbram:
### Summary 
* Sensitive materials are not zeroized across the codebase. This is not an immediate issue but is worthwhile to track as a group 

### Expected Behaviour 

* When sensitive material is loaded into memory it should be zeroized on drop. This can be done by wrapping the material in zeroized, but the best handling would be to create a type for the material that implements the drop trait and zeroizes it in there 

rough examples

Zeroize on drop 
```
use zeroize::Zeroize;

impl drop for SensitiveMatieralWrapper {
    fn drop(&mut self) {
     sensitiveMaterial.zeroize();
  }
}
```

wrapping the type in zeorize
```
zeroize::Zeroizing::new(sensitiveMaterial);
```

If any problem areas are found here please edit the table in the next comment just to keep track

JesseAbram:
Area | Location | Variable | Secret? | Why track it | Current concern
-- | -- | -- | -- | -- | --
encryption/aes | aes.rs | key | Yes | Symmetric encryption key | No explicit zeroization in module
encryption/aes | aes.rs | plaintext | Yes | Decrypted secret output | Returned as plain Vec<u8>
encryption/ecies | ecies.rs | private_key | Yes | Long-lived X25519 private key input | No explicit wipe here
encryption/ecies | ecies.rs | ephemeral_private | Yes | One-time sender secret | No explicit wipe
encryption/ecies | ecies.rs | shared_secret | Yes | ECDH output | No explicit wipe
encryption/ecies | ecies.rs | keys | Yes | HKDF-expanded key material | No explicit wipe
encryption/ecies | ecies.rs | aes_key | Yes | Derived encryption key | No explicit wipe
encryption/ecies | ecies.rs | hmac_key | Yes | Derived MAC key | No explicit wipe
keys/ed25519 | ed25519.rs | bytes | Yes | 64-byte private key input (`seed |  
keys/ed25519 | ed25519.rs | seed | Yes | Extracted private seed during parse | Explicitly zeroized after use
keys/ed25519 | ed25519.rs | self.key | Yes | Stored Ed25519 private signing key | No ZeroizeOnDrop / explicit wipe on struct drop
keys/ed25519 | ed25519.rs | self.key.to_bytes() | Yes | Seed bytes materialized during equality/raw operations | Temporary secret copies not explicitly wiped
[...]

PR #1126, opened 2026-07-13
fix(index): define unique-index semantics at the merge and tombstone boundaries

### LINKED ISSUE #1111: P2P merge bypasses unique indexes — replicated writes mint duplicate docs on unique-indexed fields; index constraints unpatchable after the fact
## Summary

Unique indexes are enforced only on the local mutation path; **P2P merge
performs no uniqueness check**, so replicated writes can mint duplicate
documents on a unique-indexed field. Additionally, index constraints are
immutable to schema patches, so a store that has been wounded this way cannot
be re-fenced afterward.

This is not hypothetical: four production stores accumulated duplicate
`AgentConversation` docs sharing a `session_id` (a field with a unique index)
via P2P merge during a fleet incident window. Downstream, every
filter-addressed `upsert_` on the duplicated value fails with
`cannot upsert multiple matching documents` — in the affected app this bricked
both startup recovery and the live request path on wounded sessions
(sourcenetwork/defra-agent#693, fixed app-side by canonical-pick + _docID
addressing in sourcenetwork/defra-agent#694).

## The honest design question

Cross-replica uniqueness is fundamentally coordination-shaped; a CRDT merge
cannot always preserve it. So the ask is to pick and implement a defined
semantic, rather than the current silent violation:

1. **Enforce at merge/ingest** — reject or quarantine an incoming doc that
   would violate a unique index (with a durable, queryable conflict record); or
2. **Document unique indexes as local-write constraints only**, and provide a
   first-class conflict surface: deterministic canonical-pick rule (e.g.
   latest-writer by commit height with docID tie-break), duplicates flagged and
   queryable, so applications can reconcile instead of discovering twins via
   upsert failures.

Either way:
- **Index mutability**: allow adding/repairing index constraints via schema
  patch (or a rebuild affordance), so wounded stores can be re-fenced.
- A two-node regression test: create docs with the same unique-field value on
  both sides, merge, assert the defined semantic (no silent twins).

## Evidence

- Fleet: 4 stores each with exactly one duplicated `session_id`, minted during
  a period of write-path failures + P2P churn; 10+ sibling stores clean.
- App-side reproduction in sourcenetwork/defra-agent#694's test support: an
  era-faithful SDL (non-unique index) is required to seed the duplicate state,
  because the current mutation path correctly refuses — demonstrating the
  asymmetry between local writes and merge.

iverc:
Fixes #1111. Root-causes and unblocks sourcenetwork/defra-agent#700.

## What the mapping actually found

#1111's literal claim ("merge performs no uniqueness check") is stale: the merge path routes through the checking `UniqueIndex::save` today. The live defects are sharper:

1. **The on-violation semantic is a divergence wedge.** A unique conflict at merge returns `MergeFailed`, the txn is discarded, the block is never marked merged — so the receiver retries it forever, and every later composite of that document re-walks the failed parent and fails too. Two nodes that each locally accepted the same unique value can never sync that document again, in either direction. (Go has the same outcome — its merge aborts atomically and peers silently diverge; the semantic is code-defined only, undocumented, untested.)
2. **Tombstones can occupy unique slots forever.** Two merge orderings mint a stale entry pointing at a dead document: a replicated update merged after a delete is classified as a fresh CREATE (`get_with_datastore` hides tombstones) and re-mints the entry while the `/del/` marker stays; and an out-of-order delete cleans nothing when the doc isn't materialized yet. Four production stores were wounded this way, and with index constraints unpatchable there was no repair — recreating the value fails forever with `can not index a doc's field(s) that violates unique index` (defra-agent#700).
3. **`if let Ok(index_manager)` swallowed schema errors silently**, skipping all index maintenance with no trace.

## The defined semantic

**Enforce locally, converge deterministically at merge, tombstones never hold slots.**

- **Local writes: unchanged.** A live conflict still rejects hard (Go parity where behavior is user-facing and defined).
- **Merge conflicts: deterministic canonical pick.** The merged document always persists; on a unique-slot collision the slot goes to the **lexicographically smallest docID** — immutable, content-addressed, totally ordered, so every replica computes the same winner with no clock or height state, and converged replicas have identical indexes. The losing document stays fully readable via non-index queries and the resolution is logged at error level with both docIDs. This deliberately diverges from Go's abort: Go's own outcome is permanent silent divergence, which the CRDT convergence constraint rules out; no Go test pins its behavior.
- **Stale entries self-heal.** When the unique check finds an existing entry, the layer above now asks whether the holder is still alive (body present, no `/del/` marker — deletion is terminal, so a dead holder can never return). A stale entry is reclaimed and the write proceeds: era-wounded stores repair themselves lazily on the exact operation that used to fail, with no migration and no manual rebuild.
- **Merges onto tombstoned docs skip index maintenance** (mirrors Go's nil-doc skip in `syncIndexedDoc`), closing both minting orderings.
- **Unique-entry deletion is owner-checked**: with canonical-pick, an unindexed loser can exist; deleting it must not free the winner's slot.
- Index-manager construction failures now log at error level instead of vanishing.

Also deduped the document/deletion-marker key layouts into `storage::keys` (db-merge hand-rolled `/del/`; the index layer needed them for the liveness probe).

## Not in scope, deliberately

- **Index create/drop/rebuild exposure** (CLI/HTTP): Go parity supports it and the FFI + `bulk_index` machinery already exist, but the self-heal removes the operational urgency (wounded stores repair lazily). Worth its own issue.
- **P2P-synced collection definitions carry no indexes** — that is Go parity (indexes are per-node, created via client API), documented here for the record.

## A finding left as a flagged decision: create onto a tombstoned docID

Testing downstream surfaced a fourth behavior: `create_` onto a **tombstoned docID** (docIDs are content-addressed, so an identical-content recreate regenerates the tombstoned id) silently writes a "zombie" — success is reported, the body is rewritten, but the `/del/` marker keeps the row invisible to normal queries. Go errors here (`ErrDocumentDeleted`, deletion is terminal).

I initially added the Go-parity guard (commits 721f5b1f, reverted in 67ca1c6c): it immediately broke defra-agent's pairing reconciler, whose applied-state flow legitimately delete-then-identically-recreates `PeerPairingApplied` rows and converges against today's semantics. So the zombie is load-bearing downstream, and hard-rejecting is an ecosystem semantic change that should be its own decision — options are (a) Go-parity terminal error + migrating downstream flows off delete-then-identical-recreate, or (b) defining local resurrection (create clears the marker), which needs a merge-side answer to stay convergent. Left out of this PR deliberately; both commits are in the history for reference.

## Tests

- **Unit** (`index_manager_tests::unique_boundaries`): delete-then-recreate frees the slot (mirrors Go's regression test); stale entry pointing at a tombstoned holder reclaimed; stale entry pointing at a missing holder reclaimed; live conflict still rejected locally; merge conflict resolves to the smallest docID in **both** arrival orders, and deleting the unindexed loser does not free the winner's slot. Reverting the heal fails the two stale-entry tests.
- **Two-node integration** (`p2p_iroh::schema::index`): both peers locally accept the same unique value, then sync — the merge completes (the old semantic never converges; reverting the fix fails with *"conflicting doc never merged: unique violation is still wedging the merge"*), both docs are present via scan, and the unique-indexed lookup returns the same deterministic winner. Plus a replicated delete-then-recreate test mirroring Go's #700-shaped regression at the replication level.

## Gates

`cargo fmt --all --check`, `cargo clippy --all -- -D warnings`, `cargo test --workspace --exclude integration-test --exclude ffi-test --exclude conformance`, `cargo check --workspace --all-targets`, and the full `p2p_iroh schema::index` suite (5/5) — all green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3487,6 +3487,7 @@ dependencies = [
  "schema",
  "storage",
  "thiserror 2.0.18",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/db-index/Cargo.toml
+++ b/crates/db-index/Cargo.toml
@@ -14,6 +14,7 @@ datastore = { path = "../datastore" }
 defra-core = { path = "../defra-core" }
 document = { path = "../document" }
 schema = { path = "../schema" }
+tracing = { workspace = true }
 storage = { path = "../storage" }
 
 # Error handling
--- a/crates/db-index/src/index_manager/mod.rs
+++ b/crates/db-index/src/index_manager/mod.rs
@@ -16,6 +16,19 @@ use storage::corekv::Key;
 use storage::index::{FullTextIndex, IndexType};
 use storage::keys::IndexIDSequenceKey;
 
+/// How a live unique conflict was resolved during merge (#1111).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MergeConflictOutcome {
+    /// No conflict, or the stale-entry heal handled it.
+    Clean,
+    /// The incoming document won the deterministic pick and now holds the
+    /// unique entry; the previous holder is no longer indexed.
+    IncomingIndexed,
+    /// The existing holder won; the incoming document is persisted but not
+    /// indexed for the conflicting values.
+    IncomingUnindexed,
+}
+
 /// Result of a bulk index operation.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct BulkIndexResult {
@@ -79,6 +92,11 @@ pub fn fulltext_index_name(field_name: &str) -> String {
 pub struct IndexManager {
     /// Collection short ID for index key generation
     collection_short_id: u32,
+    /// Collection ID for document/deletion-marker key generation. Unique
+    /// enforcement needs it to ask whether the document a stale index entry
+    /// points at is still alive (#1111/#700). Empty when constructed without a
+    /// schema, in which case stale-entry healing is disabled.
+    collection_id: String,
     /// Active index instances keyed by index name
     indexes: HashMap<String, IndexType>,
 }
@@ -88,6 +106,7 @@ impl IndexManager {
     pub fn new(collection_short_id: u32) -> Self {
         Self {
             collection_short_id,
+            collection_id: String::new(),
             indexes: HashMap::new(),
         }
     }
@@ -100,6 +119,7 @@ impl IndexManager {
     /// (e.g., empty fields list).
     pub fn from_collection(collection_short_id: u32, schema: &CollectionVersion) -> Result<Self> {
         let mut manager = Self::new(collection_short_id);
+        manager.collection_id = schema.collection_id.clone();
         for desc in &schema.indexes {
             if desc.fields.is_empty() {
                 return Err(Error::Other(format!(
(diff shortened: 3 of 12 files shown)


jackzampolin:
Fleet evidence escalating the severity here (2026-07-14 return-leg rollout): the twin `session_id` documents this PR addresses are no longer just a data-quality curiosity — under hub fan-in they became a **durable P2P wedge**. Amy retained 6 stuck pending-DAG heads (mini-1 `4a217682`, mini-2 `2c82b07e`, 3 heads each) because the incoming twin docs deterministically fail the unique-index merge and the receiver ledger re-drives them every ~60s forever. This PR fixes the *trigger*; the ledger-side terminal-failure quarantine is tracked separately and folds into #1116 stage-2. Cross-linking so both halves land.

Issue #1081, opened 2026-07-01
Retarget go-parity to Go v1.0.0 (from develop 6c874754)


jackzampolin:
## Goal

Move the Rust port's go-parity target from develop `6c874754` to **Go v1.0.0** (the release the branchable-ACP epic and current work track). This was scoped OUT of #1072 to keep that PR green against the current target, but the groundwork exists and should land as its own PR(s).

## Sub-tasks

- [ ] **Bump `GO_COMPAT` → v1.0.0** in `crates/defra-version/src/lib.rs` (set `GO_COMPAT_TAG = \"v1.0.0\"`; CI's go-compat job then downloads the v1.0.0 release binary instead of building develop).
- [ ] **Bump the harness dep** to a backbone commit carrying the v1.0.0 CLI fixes. **These are already on backbone `main` at `adaf59d`** (`fix(harness): DefraDB v1.0.0 CLI compatibility`): `--url`/keyring flags after `start`, and `--collection-name` for collection `truncate`/`describe` + `view refresh`. (The `feat/shared-se-key-harness` branch was reset back to `68838b3` for #1072, so cut a fresh branch/rev off `main` or repoint the dep.)
- [ ] **Port Go #4838 (DocID from genesis CID)** — see #1080. v1.0.0 derives the public DocID from the genesis composite CID; without it, Rust and v1.0.0 produce different document identities/commit CIDs.
- [ ] **Refresh hardcoded CID vectors** in `tools/integration-test/tests/query/multi_cid_vectors.rs` (the `*_CID` constants) to v1.0.0 values after #4838 lands.
- [ ] Sweep for any other v1.0.0 CLI/HTTP/API diffs surfaced by the integration suite (the harness fixes above were the two found so far).

## Context

Discovered while validating #1072 against a local v1.0.0 binary: the harness/CLI and DocID/CID vectors diverge from v1.0.0. Reverted from #1072 (which stays on `6c874754`) to keep it mergeable; tracked here for a follow-up.

iverc:
[...]

PR #1131, opened 2026-07-15
feat: DocID based on genesis CID + retarget go-compat to v1.0.0 (Go #4838)

### LINKED ISSUE #1080: Port Go #4838: DocID based on genesis CID (Rust↔Go v1.0.0 docID/CID parity)
## Summary

Go DefraDB v1.0.0 includes **#4838 "DocID based on genesis CID"**, which derives the public DocID from the genesis composite block CID. The Rust port has **not** implemented this — there is no genesis-CID-based DocID in `crates/document` or `crates/defra-core`. As a result, Rust and Go v1.0.0 produce **different document identities (DocIDs and therefore commit CIDs)** for the same content.

## How it surfaced

Running the go↔rust cross-impl parity suite against a Go **v1.0.0** binary:
- `rust_multi_cid_simple_*` pass (Rust is self-consistent, produces the pre-#4838 CIDs baked into the test vectors).
- `go_multi_cid_simple_*` fail: querying `Users(cid: [<hardcoded pre-#4838 CID>])` returns `null` on a v1.0.0 Go node because v1.0.0 computes a different (genesis-CID-derived) DocID/CID.
- Dynamic-CID tests (`go_multi_cid_commits_*`) pass, since they compute CIDs per-node rather than comparing baked constants.

This is orthogonal to the branchable-collection ACP epic (#1072) — it affects all content-addressing.

## Scope

Implementing #4838 in Rust is a core content-addressing change (DocID generation, genesis-block DocID omission, short-ID internal remapping, `_docID` index/query behavior). It should be scoped and landed independently, then the pre-#4838 CID test vectors in `tools/integration-test/tests/query/multi_cid_vectors.rs` refreshed.

## Status

The 3 affected `go_multi_cid_simple_*` tests are **intentionally left FAILING** (not `#[ignore]`d) as a visible parity signal until #4838 is ported. Their source carries a NOTE explaining the expected failure and pointing here. The `rust_multi_cid_simple_*` variants pass.

Ref: Go commit `0783f3f5` (feat: DocID based on genesis CID (#4838)).

iverc:
Ports Go DefraDB **#4838 "DocID based on genesis CID"** to defradb.rs and retargets the go-parity suite from develop `6c874754` to **Go v1.0.0**. Closes #1080; completes the core of #1081.

## What changed

Document identity is now derived from the **genesis composite block CID** (as in Go v1.0.0), not from a pre-write content hash. Storage keys move to a node-local `u64` short ID, and the public `bae-…` DocID is assigned only after the first composite block is written.

- **Wire format** (`defra-core`): `docID` is removed from every delta payload (LWW, counter, composite) and from the encryption block (`{key}` only). Golden vectors regenerated from a Go v1.0.0 test-harness generator. Deltas carrying identity is the divergence #4838 removes.
- **Identity assignment** (`document`, `db`, `db-blocks`): client-side generation is gone. Create allocates a short ID from `/seq/doc`, writes the CRDT DAG first, derives the DocID from the genesis CID, dup-checks against the mapping (tombstoned → deleted error), then registers the short-ID/DocRef/block-ownership mappings before the blob + indexes. Unsigned identical creates on two nodes converge to the same DocID.
- **Storage keys** (`storage`): new `/d/{s,p,r,b}` systemstore keyspace (short-ID↔DocID, DocID→DocRef, alias set, block-CID→DocID ownership set) + `/seq/doc`. `DataStoreKey`, `HeadstoreDocKey`, `HeadstorePriorityKey`, `PrimaryDataStoreKey` and the secondary-index keys are rekeyed to the short ID, so collection scans return documents in insertion order like Go.
- **Merge** (`db-merge`): resolves a composite's owning DocID the way Go's `resolveCompositeBlockDocRef` does (recorded owner → genesis-CID derivation → head-walk), mints the receiver-local short ID, and records block ownership. Field blocks inherit identity from their composite.
- **KMS/encryption**: the encryption block carries only the key; ownership is resolved through the block-CID→DocID index. Key release checks any owning document. A DEK request for a block whose ownership isn't recorded locally yet (a replication fetch before merge) defers to the owner instead of self-denying.
- **Backup**: export writes the current identity as `_docIDNew`; import derives fresh IDs and registers the file's identities as aliases, remapping relations through them (Go's alias model).
- **go-compat retarget**: `GO_COMPAT` → v1.0.0 (CI downloads the release binary), harness rev bumped to the v1.0.0-CLI-compatible backbone, multi-CID vectors regenerated from a live v1.0.0 node, and the `--collection-name` CLI flag renamed to match #4946.

## Validation

Cross-runtime parity confirmed against a real Go **v1.0.0** binary:

| Suite | Result |
|-------|--------|
| workspace unit tests | green; `fmt`/`clippy -D warnings` clean |
| integration `basic` (incl. Rust↔Go v1.0.0) | 31/31 |
| integration `query` (incl. refreshed CID vectors) | 57/57 |
| integration `acp` | 198/198 |
| integration `encryption` + SE cross-runtime (Go↔Rust) | all pass vs the v1.0.0 binary |
| integration `p2p` (libp2p) | green (iroh variants are pre-existing flakiness under parallel load; pass solo) |

DocID derivation was cross-checked byte-for-byte against a live v1.0.0 node (`bae-607006ec-…` from genesis CID `bafyreig4cg2…`).

## Notes / follow-ups

- **Signed** creates fold the signature CID into the genesis block, so their DocID is per-identity (not cross-node-deterministic) — by design, matching Go.
- The **Action-system** response envelope (`#4922`, `{Description, Execution}`) that Go v1.0.0 wraps index listings in is **not** ported here; the index integration test accepts both shapes. Tracking as separate work under #1081, alongside the other category-A v1.0.0 surface diffs (GQL operation ordering #5013, `@default(value:)` #4938, etc.) which the go-compat suite will gate.
- Standalone field-block merges only proceed when the ownership index names a single owner; otherwise they're merged via their composite (Rust-transport behavior, outside Go's per-doc PushLog envelope).


iverc:
## CI fix + review pass

**CI (go-compat):** the `parity::*` failures were "failed to start go-0" — `proofs/Cargo.toml` still pinned `defra-harness` to the old `feat/shared-se-key-harness` branch, so it launched Go nodes with pre-v1.0.0 CLI flags. Bumped it to `rev = adaf59d` (same as `tools/integration-test`), which starts v1.0.0 Go nodes correctly (verified: the `go_*` tests in the integration suites already start v1.0.0 nodes fine with this rev).

**CI (P2P Linux):** the two `filtered_replication::rust_*` timeouts (`_or`, `_encrypted_respects_filter`) both pass deterministically locally and in the earlier local p2p run; they're the known Linux poll-timeout flakiness under parallel load, not a regression — should clear on re-run.

**Review comments addressed** (commit `4f474ce`):
- **downsample validation** (major): a first-time create no longer skips the timestamp/numeric/paired-field validation — only the closed-bucket lookup that needs the not-yet-assigned series id is skipped.
- **update / delete_many ownership registration** (3× major): now fatal — a systemstore or mapping-write failure rolls back the txn instead of committing a document whose blocks have no ownership entry (matches the create path and Go's `SetBlockDocIDMapping`).
- **doc_mutator pre-delete read** (major): propagates instead of `.ok()`-swallowing the payload filtered replication needs.
- **doc_id_map alias purge** (minor): recovers each alias DocID from the key, so a malformed value can't orphan the reverse mapping.
- **merge doc-identity resolution** (major): blockstore/recursive infra errors propagate; only a genuine "unresolved" falls through to the next head.

**Intentionally deferred / not changed, with rationale:**
- `doc_id_map` cross-collection reassignment: Go-faithful (`resolveOrAllocateDocShortID` overwrites the same way), and structurally near-impossible — the genesis CID commits to `collectionVersionID`, so the same public DocID cannot arise in two collections.
- `versioned_fetcher` / `commits_fetcher` first-owner pick: composite blocks (the commits case) have exactly one owner; the multi-owner case only arises for a byte-identical **field** block shared across docs, where any owner is a valid attribution.
- Truncate full-DAG block cleanup: a pre-existing truncate limitation (main only clears head blocks too), orthogonal to this identity port and self-flagged heavy-lift — tracked as follow-up.

Verified after the fixes: workspace unit tests green, `fmt`/`clippy -D warnings` clean, and integration acp 198/198, basic 31/31, query 57/57, encryption + SE cross-runtime all pass against the Go v1.0.0 binary.

jackzampolin:
Diagnosis of the failing `P2P Integration (Linux)` check (run 29450601154), from the log:

```
test filtered_replication::rust_filtered_replication_encrypted_respects_filter ... FAILED
test filtered_replication::rust_filtered_replication_encrypted_rich_filter ... FAILED
(both: poll timeout panic in defra-harness poll.rs:16; 71 passed / 2 failed)
```

Two observations before writing this off as the known parallel-load flake (#1132):

1. It does **not** match #1132's signature (random, different subsets per run, non-encrypted tests included). Here it's exactly and only the two **encrypted** filtered-replication tests — the intersection of the two things this PR reworks: DEK/ownership resolution through the new block-CID→DocID index, and (per the notes) deferral when a block's ownership isn't recorded locally yet. Filtered replication is precisely the path that delivers **partial DAGs**, so receiver-side ownership-index gaps are expected there — if the DEK request defers to the owner for blocks whose composite never arrives (filtered out), the poll would time out exactly like this. Worth checking whether the deferral path terminates when the ownership record can never be completed locally.

2. If it does turn out to be load-flake, note both tests pass on `main` under the same CI matrix as of `e74733a6`, so a rebase run would disambiguate cheaply.

Related heads-ups for the rebase (three things landed on main after this branched, all today): #1133 (terminal-failure quarantine — adds `MergeOutcome::Rejected` consumers in the merge path your `resolveCompositeBlockDocRef` port touches), #1135 (`parity_unique_twins_*` probes — compute docIDs in-test so they should survive the derivation change; the go_go probe's #5058 silent-ack signature is verified present at v1.0.0 `message.go:175`), and the mixed characterization test is deliberately pinned to pre-#4838 behavior and NOT CI-allowlisted, so your retarget shouldn't trip it. Also: `~/.cache/defra-harness/` on the shared box now has Go binaries built at both `6c874754` and `68eafac2` if useful for A/B runs.

jackzampolin:
Taking the remaining `P2P Integration (Linux)` failure (the two `filtered_replication::*encrypted*` poll timeouts) to land this — will merge current main (picks up #1135's parity probes), root-cause locally, and push the fix to this branch. Shout if you're mid-flight on the same thing.

PR #1157, opened 2026-07-19
fix(p2p): recover saturated pushes after restart

### LINKED ISSUE #1154: P2P: restart admission acceptance loses success-acked documents on v0.16.0
## Summary

The existing `hub_restart_recovers_success_acked_pending_dags` acceptance test fails on the released v0.16.0 SHA after its 240-second post-restart convergence deadline. This is the restart-loss acceptance item from #1099, which is closed.

## Reproduction

On untouched v0.16.0 `main` at `4d7a2f42f2e7a675acdbae0d82a339e64b63dbc1`:

```sh
cargo test -p integration-test --test p2p_admission_restart -- \
  --exact hub_restart_recovers_success_acked_pending_dags --nocapture
```

Observed:

```text
765 of 2424 documents lost across hub restart
test result: FAILED
finished in 296.09s
```

The same test also failed during validation of an unrelated store-migration branch:

```text
716 of 1383 documents lost across hub restart
test result: FAILED
finished in 267.93s
```

The broader P2P matrix immediately before it was green (97 passed, 11 ignored), including the non-restart admission test.

## Expected

Every success-acked document should merge on the restarted hub through persisted pending-DAG recovery or the pusher retry path, satisfying the restart acceptance criterion from #1099.

## Notes

This was reproduced independently on untouched v0.16.0 to exclude the unrelated store-migration patch as the cause.

iverc:
Fixes #1154.

## Problem

When a receiver reported that its pending-DAG registry was full, the sender parked that peer but only handed the active failed push to the durable retry ledger. Up to 256 already-queued heads remained in the volatile outbound backlog with dormant retry watermarks. As the peer-wide cooldown escalated, an old document could wait minutes before its first delivery attempt, leaving hundreds of success-acknowledged writes unmerged after a hub restart.

Recovery was also impeded by two adjacent behaviors: a full receiver rejected blocks that were dependencies of an existing pending root, and the libp2p replay path replaced the receiver's exact capacity response with a generic error before retry scheduling.

## What changed

- Drain a saturated peer's queued outbound work into the existing lossless failure recorder on the first capacity rejection, immediately activating durable retry ownership.
- Preserve exact libp2p replay failures so the capacity sentinel reaches the persisted retry scheduler.
- Reschedule capacity-limited retries at the normal sweep cadence without advancing their exponential delivery-failure ladder.
- Admit exact missing dependencies and retransmitted roots while the pending-DAG registry is full, allowing existing roots to complete and free capacity.
- Keep persisted retries ordered by next-attempt time so bounded sweeps make fair progress.
- Promote the original at-scale restart reproduction from an ignored guard to a permanent passing regression test, alongside focused queue-demotion and admission tests.

## Validation

- [x] `git diff --check`
- [x] `cargo clippy -p storage -p p2p -p db-merge -p cli -p embedded -p defra-node -p integration-test --all-targets -- -D warnings`
- [x] `cargo test -p storage -p db-merge -p p2p`
- [x] `cargo test -p integration-test --test issue1154_repro -- --nocapture` (all 3,155 documents merged 145.7s after restart)
- [x] Independent diagnostic run of the same workload (all 1,057 documents merged 103.1s after restart)


jackzampolin:
Reviewed. The core fix is sound: sorting `get_retry_documents` by `next_retry_unix` (peerstore.rs:475) is provably starvation-free given `bump_for` always sets a strictly-future timestamp — hot failers rotate behind every currently-due record on each sweep — and it's centrally placed so all five sweep copies benefit. The unit test pins the ordering, and the rewritten restart test is a genuinely better deterministic check of the durable pending-DAG contract. push_backlog quota/cooldown is untouched, and nothing load-bearing was removed.

One gap before merging:

1. **No on-branch evidence the sort fixes #1154's observed failure.** The new acceptance test recovers through the pending-DAG re-drive path — the pusher retry ladder (the thing the sort fixes) is never engaged, so it passes with or without the peerstore change. Meanwhile the PR deletes the only reproduction of the at-scale symptom ("765 of 2424 documents lost"), so the issue's criterion — every success-acked document merges — is no longer asserted anywhere. Please run the pre-rewrite workload (or a bounded variant) once on this branch to confirm the sort resolves the drain failure; if the diagnosis is wrong, the symptom is now unguarded.

Two nits, non-blocking:
2. p2p_admission_restart.rs:152-156 — the `pending_dags == 1` assertion message says "source unexpectedly served the pending DAG" while the check reads the hub's endpoint; naming both would end the confusion.
3. p2p_admission_restart.rs:112-121 — the process-global `set_var` toggling of `DEFRA_P2P_RATE_LIMIT_BURST`/`RUST_LOG` is only safe while this binary has exactly one test; extend the header comment's own-binary constraint to cover these vars too.

With a validation run of the original workload posted here, this is good to merge.

jackzampolin:
Ran the requested validation — **the pre-rewrite workload fails on this branch**, so this should not merge in its current form.

**Setup:** branch + origin/main merge (clean), with main's `p2p_admission_restart.rs` resurrected verbatim as a temp test target. Only deltas from the original: a scale floor (writers must produce ≥600 docs before the freeze hunt, guaranteeing hundreds of nacked pushes in the pusher retry ladders — the mechanism the sort targets) and doc-count/timing instrumentation. Mechanism unchanged: 4 writer threads @25ms, 1-slot hub, SIGSTOP freeze, hard kill, restart on rootdir, 240s convergence deadline.

**Result:**
```
issue1154_repro: 2546 success-acked documents expected on restarted hub
691 of 2546 documents lost across hub restart (success-acked pending registration not recovered)
test result: FAILED (360.51s)
```
Statistically indistinguishable from #1154's baselines (765/2424, 716/1383). 

**Characterization from retained logs:** the durable-registration path worked (restart logged `restored persisted pending DAG registrations restored=1` and that root completed immediately). The loss is entirely in the pusher retry path: the hub was still receiving PushLogs when the deadline expired, with intake collapsing per minute post-restart — 3323 → 2130 → 447 → 331 → 8. The ladder isn't wedged; it drains at a decaying rate that leaves ~27% of docs unmerged at the deadline. Ordering retries by `next_retry_unix` is correct for fairness (the unit test stands) but doesn't change drain throughput at scale, so the diagnosis for #1154 is incomplete.

**Suggested path forward:** keep the sort + the new deterministic durable-registration test (both are good), but (1) retain an at-scale variant of the old workload as the regression guard for #1154, and (2) investigate the post-restart drain collapse — the decay pattern suggests backoff ladders climbing while the hub recovers, so success-acked-but-unmerged docs wait behind exponentially-spaced retries rather than being re-driven. #1154 should stay open either way.

jackzampolin:
Pushed the validation tooling onto the branch so it's all in one place (7d4f9466, on top of a clean origin/main merge in b9911a4c — no force-push):

- `tools/integration-test/tests/issue1154_repro.rs` — the resurrected pre-rewrite workload as its own `[[test]]` binary, committed **`#[ignore]`d** since it currently fails (691/2546 lost, per the run above). It keeps #1154's symptom pinned until the root cause lands; run it with `cargo test -p integration-test --test issue1154_repro -- --ignored`. The header documents the failure numbers and the drain-collapse characterization.
- The two review nits: the `pending_dags == 1` assertion message now names both the hub's count and the source's serving, and the own-binary header comment covers the process-global `DEFRA_P2P_RATE_LIMIT_BURST`/`RUST_LOG` toggling.

Failure logs from the validation run are also retained locally if useful: `defradb.rs-1157/target/e2e/20260722-180720-3f0da3a4/rust-{0..4}/logs/` (hub restore line, per-minute PushLog intake decay). Happy to pair on the drain-collapse investigation — the decay pattern (3323 → 2130 → 447 → 331 → 8/min) points at pusher backoff ladders climbing while the hub recovers, so success-acked-but-unmerged docs sit behind exponentially-spaced retries instead of being re-driven on reconnect.

iverc:
Confirmed independently: the at-scale guard still left 687 of 2,719 documents missing on this branch, matching the reported 691 of 2,546 result. Retry ordering improves fairness, but it does not resolve the post-restart drain collapse.

I have updated the PR description accordingly:

- changed `Fixes #1154` to `Related to #1154`
- retained the ignored at-scale reproduction as the regression guard
- scoped this PR to retry fairness and the deterministic durable-registration contract
- documented that #1154 remains open and that the broader sender retry ownership work is tracked by #1116

I also tested a peer-level scheduling follow-up locally, but it produced the same loss ratio, so none of that unproven change was pushed.

iverc:
The at-scale gap is now fixed in `6c05fa38`.

The retained trace exposed the ownership bug: a missing document was created early but did not receive its first push attempt until more than three minutes later. A capacity rejection parked the peer, but only the active job entered the durable retry ledger; the peer's queued heads remained in volatile memory while the cooldown escalated.

The first capacity rejection now drains that queued peer work through the existing lossless failure recorder, so persisted retries own it immediately. The change also preserves the receiver's exact capacity response, avoids advancing the delivery-failure ladder for saturation, and lets exact dependencies complete an already-admitted pending DAG.

At-scale validation now passes twice:

- all 1,057 documents merged 103.1s after restart
- all 3,155 documents merged 145.7s after restart under the final CI-shaped invocation

The original workload is no longer ignored and now serves as the permanent regression test for #1154.

PR #1162, opened 2026-07-20
fix(storage): retain conflicts for active snapshots

### LINKED ISSUE #1158: Redb: successful AgentResponse create is intermittently invisible to immediate update
## Summary

On DefraDB v0.16.1 (1c38ab0c397437beb2664b212f85abb6ea6cd841), a fresh persistent Redb-backed defra-agent test intermittently observed a successful GraphQL create followed by an immediate update/query that could not find the created document.

This is after #1139 / bee90f5f, so it may be a remaining commit-acknowledgement, snapshot, or index-visibility race rather than the exact stale-snapshot case fixed there.

## Observed sequence

1. create_AgentResponse returned document ID bae-167af222-2ce8-57be-8c18-b8698ba2fead with status streaming and progress_seq 0.
2. The next operation updated AgentResponse by that _docID plus status == streaming to set progress_seq 1.
3. The mutation returned no matching documents.
4. Immediate status lookup by _docID, then by the unique request_id/response_key, reported the response as missing.
5. defra-agent failed the request closed. Its error-response recovery also left no response visible to the submitter.

Exact runtime error:

    cannot advance progress for request_id=f2835d98-b8d8-4cd2-a20d-8cb868f327f6 doc_id=bae-167af222-2ce8-57be-8c18-b8698ba2fead: response is missing

## Environment and evidence

- DefraDB release: v0.16.1
- DefraDB SHA: 1c38ab0c397437beb2664b212f85abb6ea6cd841
- Consumer: source-inc/defra-agent PR #810, SHA 39065a28ead9e05abea630a8d19fbfe24c400d37
- macOS arm64 self-hosted runner
- Fresh temporary persistent store; no migration involved in this test
- CI run: https://github.com/source-inc/defra-agent/actions/runs/29700654381
- Failed job, attempt 1: https://github.com/source-inc/defra-agent/actions/runs/29700654381/job/88229763246
- Exact consumer test passed 10/10 consecutive runs locally after the CI failure, so this is intermittent.

The failing consumer test is:

    cargo test -p defra-agent-cli --test cli_reconciliation reconciled_runtime_sends_generation_two_tools_and_completes_tool_loop -- --nocapture --test-threads=1

## Expected

Once create_AgentResponse returns a document ID without GraphQL errors, a subsequent transaction on the same EmbeddedNode must observe that document. Either the update should match it, or an explicit conflict/error should be returned; successful create followed by missing is not a valid outcome.

iverc:
Fixes #1158.

## Root cause

The shared optimistic conflict tracker retained only the latest 1,000 committed read/write sets, regardless of which write snapshots were still active. Under sustained commit activity, a long-lived write transaction could therefore outlive the committed record that should make it conflict.

For concurrent creates, both transactions can read the same `/seq/doc` value. If the newer create commits and its conflict record is pruned before the older create commits, the older transaction can be accepted and reuse the same short ID. That replaces the canonical mapping and document body after the newer create already returned success, making the acknowledged document unqueryable by its returned DocID and indexes.

## What changed

- Track active write snapshots per conflict version and prune committed records only once they are no longer needed by the oldest active snapshot.
- Register snapshot guards for Redb, Lark, Fjall, RocksDB, and memory transactions while preserving the lock-free read-only version path.
- Transfer Redb guards into queued group commits and blocking direct commits so cancellation cannot release required history while a commit remains in flight.
- Add a deterministic regression with 1,000 intervening commits that verifies a stale create conflicts instead of overwriting the acknowledged document.

## Validation

- [x] `cargo test -p storage --all-features`
- [x] `cargo clippy -p storage --all-features --all-targets -- -D warnings`

### DIFF
--- a/crates/storage/src/backends/fjall/store.rs
+++ b/crates/storage/src/backends/fjall/store.rs
@@ -172,7 +172,11 @@ impl Store for FjallStore {
         }
         let mut guard = NewTxnGuard(&self.active_txn_count, false);
 
-        let read_version = self.conflict_tracker.current_version();
+        let conflict_snapshot = (!readonly).then(|| self.conflict_tracker.begin_snapshot());
+        let read_version = conflict_snapshot.as_ref().map_or_else(
+            || self.conflict_tracker.current_version(),
+            |snapshot| snapshot.version(),
+        );
         let snapshot = self.db.snapshot();
 
         // Defuse guard — transaction will manage its own count via Drop
@@ -182,6 +186,7 @@ impl Store for FjallStore {
             db: self.db.clone(),
             keyspace: self.keyspace.clone(),
             conflict_tracker: Arc::clone(&self.conflict_tracker),
+            _conflict_snapshot: conflict_snapshot,
             active_txn_count: Arc::clone(&self.active_txn_count),
             read_version,
             snapshot,
--- a/crates/storage/src/backends/fjall/transaction.rs
+++ b/crates/storage/src/backends/fjall/transaction.rs
@@ -7,7 +7,9 @@ use std::sync::Arc;
 use super::compute_range_bounds;
 use super::iterator::MergingIterator;
 use crate::backends::shared::DurabilityMode;
-use crate::backends::shared::{CallbackCounts, CallbackManager, ConflictTracker, ReadSet};
+use crate::backends::shared::{
+    CallbackCounts, CallbackManager, ConflictSnapshot, ConflictTracker, ReadSet,
+};
 use crate::corekv::{
     AsyncTxnCallback, Error, IterOptions, Iterator, Reader, Result, Txn, TxnCallback, Writer,
 };
@@ -17,6 +19,7 @@ pub(crate) struct FjallTxn {
     pub(crate) db: fjall::Database,
     pub(crate) keyspace: fjall::Keyspace,
     pub(crate) conflict_tracker: Arc<ConflictTracker>,
+    pub(crate) _conflict_snapshot: Option<ConflictSnapshot>,
     pub(crate) active_txn_count: Arc<AtomicUsize>,
     pub(crate) read_version: u64,
     pub(crate) snapshot: fjall::Snapshot,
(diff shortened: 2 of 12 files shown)


jackzampolin:
Consumer validation shows this patch does not fully close #1158 yet.

I pinned `source-inc/defra-agent` directly from current v0.16.2/main (`364db72`) to this PR head (`24d3b96fa0fcb1596e7bbbe74c49a177d03925ee`), rebuilt the exact focused CLI integration binary, and repeated its four serialized chat cases. The first 12 runs passed; repetition 13 failed in the tool-chat case:

```text
cannot advance progress for request_id=14dac5bf-4731-40d4-9137-c077bdba1ff7 doc_id=bae-0a0d606d-d4c2-5d85-85d4-847928029430: response is missing
```

The sequence remains: `create_AgentResponse` returned that DocID, the immediately following status-guarded `update_AgentResponse` matched zero rows, and lookup by DocID/request key found no response.

Before pinning this PR, I also tested consumer-side retries. A failed case remained missing across seven update/read attempts with exponential backoff (6.3 seconds total), confirming this is not ordinary visibility latency.

The #1162 active-snapshot retention fix is valid for the >1,000-commit overwrite mechanism, but this fresh run demonstrates a second path to the same acknowledged-create disappearance. I am continuing to narrow it in the consumer tracker: source-inc/defra-agent#853.

jackzampolin:
The acknowledged-create disappearance still reproduces with this branch alone. Raw downstream WAL evidence identified a second, independent RocksDB race: commit records the logical conflict version before the physical batch write, and new_txn reads that version before taking the physical snapshot. That permits a new logical version to be paired with stale physical state, so conflict detection can accept reuse of an already allocated short document ID.

I published the narrow follow-up as draft #1167, stacked directly on this PR. It adds a shared gate around logical/physical snapshot and commit pairing plus cancellation-safe active transaction accounting. Validation is green: three storage concurrency regressions, cargo check -p storage, the formerly failing downstream test, and 50/50 downstream stress passes (this branch alone failed by runs 13 and 18).

PR #1165, opened 2026-07-20
perf(query): pipeline ACP permission checks

### LINKED ISSUE #519: perf: Non-blocking ACP checks in query execution

iverc:
Closes #519.

## Problem

`PermissionFilterNode` awaited every ACP decision before advancing its source, serializing query execution against SourceHub or other remote-provider latency. The earlier parallelization attempt drained the complete source first, which traded latency for unbounded buffering and delayed the first streamed row.

## What changed

- Pipeline up to 16 ACP checks per permission-filter node, bounding both document buffering and provider pressure.
- Use an ordered future queue so checks run concurrently while authorized rows retain source order.
- Poll checks as scanning proceeds, preserving streaming latency without racing or cancelling the wrapped source.
- Preserve DAC bypass, task-local deferred ACP overlays, and fail-closed handling of provider errors.
- Cancel pending checks when the node is reinitialized or closed.
- Add deterministic coverage for bounded concurrency, out-of-order completion, ordered emission, and fail-closed continuation.

## Validation

- [x] `cargo test -p query-plan -p query` (262 passed)
- [x] `cargo clippy -p query-plan -p query --all-targets -- -D warnings`
- [x] `cargo check -p query-plan --target wasm32-unknown-unknown`
- [x] `git diff --check`

### DIFF
--- a/crates/query-plan/src/plan/permission_filter.rs
+++ b/crates/query-plan/src/plan/permission_filter.rs
@@ -6,13 +6,22 @@
 use std::sync::Arc;
 
 use acp::{DocumentACP, DocumentPermission, Identity};
+use async_lock::Mutex;
 use async_trait::async_trait;
+use defra_core::thread_bounds::MaybeBoxFuture;
+use futures::{stream::FuturesOrdered, FutureExt, StreamExt};
 use identity::Did;
 
 use crate::planner::{index_selection::CursorSeek, Doc, PlanNode};
 use crate::txn::check_doc_access_with_overlay;
 use query_types::document::DocumentMapping;
 use query_types::error::Result;
+
+// Keep enough remote checks moving to hide latency without letting one query
+// exhaust the ACP provider's connection pool.
+const MAX_IN_FLIGHT_PERMISSION_CHECKS: usize = 16;
+
+type PermissionCheck = MaybeBoxFuture<'static, (Doc, bool)>;
 
 /// PermissionFilterNode filters documents based on ACP permissions.
 ///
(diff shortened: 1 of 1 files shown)


jackzampolin:
Merged — verified fail-closed on every path (emit only on explicit allow, provider errors deny, init/close drop pending futures so cancellation can't emit unchecked rows), deterministic source ordering via FuturesOrdered matching Go, and both ACP paths intact (User queries via PermissionFilterNode; Commits/time-travel checks in runner/commits.rs and runner/version.rs untouched).

Two non-blocking follow-ups worth a test/issue:
1. Nothing pins the identity-context propagation contract: correctness currently depends on check futures being polled inline on the caller's task (thread-local `dac_bypass`/signing context + task-local deferred-ACP overlay stay in scope). A refactor moving checks onto `tokio::spawn` would silently break it with no failing test. A regression test asserting bypass/overlay visibility from inside a pipelined check would lock this in, plus the claimed-but-absent cancellation test.
2. `pending.len() < MAX_IN_FLIGHT_PERMISSION_CHECKS` counts completed-but-unyielded futures, so a slow head-of-line check throttles dispatch below 16 — a throughput nuance, though defensible as a memory bound on buffered docs.

PR #1166, opened 2026-07-20
fix(query): resolve cursor seeks through short IDs

iverc:
Follow-up to #1131.

## Problem

Go v1 DocID parity stores index tie-breaker suffixes as node-local short IDs. Cursor planning still appended public DocID bytes, producing seek keys that do not exist for non-unique indexes and unique indexes with null values. Indexed cursor pagination could therefore skip or repeat rows at the boundary.

## What changed

- Carry the public boundary DocID separately from the storage-encoded index prefix.
- Resolve that DocID through the collection short-ID map immediately before applying the seek.
- Apply the same resolution in regular, lensed, and lensed auto-commit index fetchers.
- Keep unique non-null index seeks suffix-free and reject stale cursor boundaries with no local mapping.
- Add focused coverage for the storage seek-key suffix.

## Validation

- [x] `cargo test -p query-plan -p db --lib` (122 query-plan tests and 144 db tests passed)
- [x] `cargo clippy -p query-plan -p db --all-targets -- -D warnings`

### DIFF
--- a/crates/db/src/doc_fetcher.rs
+++ b/crates/db/src/doc_fetcher.rs
@@ -212,7 +212,7 @@ impl<S: Store + 'static> DocFetcher for DbDocFetcher<S> {
         use std::collections::HashSet;
         use storage::index::IndexIterator;
 
-        let (_collection, datastore, systemstore, index_manager) =
+        let (collection, datastore, systemstore, index_manager) =
             get_collection_with_index_manager(&self.txn, collection_name).await?;
 
         // Get the index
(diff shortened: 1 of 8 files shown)


jackzampolin:
Merged after independent verification: reproduced the bug on main (`composite_index::rust_cursor_paginates_through_duplicate_index_keys` fails — the exclusive seek at `[values]+docID` ASCII sorts above every `[values][SEPARATOR][uvarint]` on-disk key, skipping the whole duplicate tier), and the full 25-test cursor integration suite passes on this branch. The constructed seek key is byte-identical to the write path, and the length-prefixed uvarint short-ID encoding is self-delimiting, so variable-length IDs can't prefix-collide. `validate_cursor_index` is correctly untouched (#961 latent issue was closed as Go-parity-by-design, tracked as #1005).

Two non-blocking notes:
1. `crates/db/src/index_seek.rs:41-47` — a boundary DocID with no local short-ID mapping now returns `cursor_invalid` (purged docs, or tokens minted on another node whose boundary doc isn't ingested locally). Go degrades gracefully there, but with short-ID keys erroring is the safest correct choice; worth a test in `cursor/error_paths.rs` for the stale-boundary case.
2. Worth re-running the Go #4617 FFI cursor suite against main as a parity check now that this is in.

PR #1171, opened 2026-07-21
fix(p2p): update noq to 1.1.0

### LINKED ISSUE #1090: noq-proto 1.0.0-rc.0: `.expect("known path")` panic at connection/mod.rs:4135 on datagram-vs-path-teardown race — kills production hub node
## Summary

Production hub node (amygdala fleet coordinator, macOS/studio-1, defra-agent embedding defradb.rs rev `1b78bd21`) crashes with a panic inside `noq-proto` when an incoming datagram references a `PathId` that has already been removed from the connection's path map. The panic is on a `tokio-rt-worker` thread and takes down the whole process.

```
thread 'tokio-rt-worker' (15943070) panicked at .../noq-proto-1.0.0-rc.0/src/connection/mod.rs:4135:14:
```

`connection/mod.rs:4128-4136` is the incoming-datagram path:

```rust
self.path_data_mut(path_id)
    .inc_total_recvd(data.len() as u64);
```

and `path_data_mut` (line 785-787) is:

```rust
fn path_data_mut(&mut self, path_id: PathId) -> &mut PathData {
    &mut self.paths.get_mut(&path_id).expect("known path").data
}
```

so a datagram racing path teardown (multipath state churn) is fatal.

## Production occurrences

5 crashes on the hub in 48h — 2026-07-05 14:39Z, 14:42Z, 14:48Z; 2026-07-06 06:47Z, 12:15Z — each followed by a full process relaunch. All occurred amid `noq_proto: PTO expired while unset` floods during the PushLog fan-in saturation storm (defra-agent#630 conditions, 16 pushers against one hub), i.e. exactly when path churn is highest. The hub restarted 15× total in that window; this panic is a major contributor.

## Why filed here

`noq`/`noq-proto` are n0-computer crates from crates.io (`noq` 1.0.1 / `noq-proto` 1.0.0-rc.0 are the latest published — no fixed release to bump to). Filing against defradb.rs as the integration owner: needs an upstream report to n0-computer/noq, and until that lands, consider a `[patch.crates-io]` pin with defensive handling (drop the datagram when the path is unknown instead of `.expect`), since any spurious/late datagram is currently a remote process-kill on every node.

Sibling production panic in the same crate family: `noq-1.0.1/src/endpoint.rs:648` counter underflow (separate issue).

### LINKED ISSUE #1091: noq 1.0.1: `active_connections` underflow panic at endpoint.rs:648 on draining event (defra-agent#634 in production)
## Summary

Production steward nodes crash with an integer-underflow panic in `noq` when a connection *draining* event arrives while `active_connections` is already 0:

```
thread 'tokio-rt-worker' (2278) panicked at .../noq-1.0.1/src/endpoint.rs:648:17:
attempt to subtract with overflow
```

`endpoint.rs:646-651`:

```rust
if event.is_draining() {
    self.recv_state.connections.active_connections -= 1;
    if self.recv_state.connections.active_connections == 0 {
        shared.all_draining.notify_waiters();
    }
```

A double-decrement / draining-vs-drained ordering race lets the counter hit 0 before the last draining event is processed. This is the same crash site as **defra-agent#634** (r5_cross_deployment conformance abort: subtract overflow, then poisoned-lock panic in Drop) — that flaky test is a real production defect.

## Production occurrences

spark-1 steward (Linux) crashed twice on 2026-07-06 — 18:21Z and 19:58Z, each followed by a restart banner — *after* the fleet p2p storm had subsided, so this races on ordinary connection teardown, not just under saturation.

## Note on release behavior

The fleet binaries are built with overflow checks enabled, so this dies loudly. With default release settings the subtraction would silently wrap to `u64::MAX`-ish garbage instead — `active_connections == 0` then never holds and `all_draining` waiters are never notified, which turns a crash into a silent hang. Worth keeping checks on until fixed.

## Why filed here

`noq` 1.0.1 is the latest crates.io release (n0-computer/noq upstream) — no fixed version to bump to. Needs an upstream report plus, if it keeps crash-looping nodes, a `[patch.crates-io]` interim (saturating_sub + debug_assert). Sibling production panic: `noq-proto connection/mod.rs:4135` `.expect("known path")` (separate issue).

iverc:
## Problem

DefraDB is locked to `noq` 1.0.1, which contains two production crash paths:

- A late datagram for a removed multipath path reaches `.expect("known path")` and panics.
- A duplicate `Draining` event decrements `active_connections` twice, causing an underflow panic with overflow checks or a shutdown hang without them.

Both defects were fixed upstream after 1.0.1, but no released upgrade was available when they were reported.

## What changed

- Update `noq`, `noq-proto`, and `noq-udp` to 1.1.0.
- Pick up the upstream stale abandoned-path packet fix from [n0-computer/noq#732](https://github.com/n0-computer/noq/pull/732).
- Pick up the upstream duplicate draining-event fix from [n0-computer/noq#748](https://github.com/n0-computer/noq/pull/748).
- Refresh the transitive lockfile entries required by the new release.

## Validation

- [x] `cargo test -p p2p --features iroh-transport`
- [x] `cargo test -p defra-p2p-adapter --features iroh`
- [x] `cargo clippy -p p2p -p defra-p2p-adapter --all-targets --features p2p/iroh-transport,defra-p2p-adapter/iroh -- -D warnings`
- [x] Confirmed the published `noq-proto` 1.1.0 source includes regression coverage for both failure modes.

Closes #1090.
Closes #1091.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -608,7 +608,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -619,7 +619,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3369,7 +3369,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -4381,7 +4381,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4958,7 +4958,7 @@ dependencies = [
  "libc",
  "log",
  "rustversion",
- "windows-link 0.1.3",
+ "windows-link 0.2.1",
  "windows-result 0.4.1",
 ]
 
@@ -5673,7 +5673,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "system-configuration 0.7.0",
  "tokio",
  "tower-service",
@@ -8158,7 +8158,7 @@ dependencies = [
  "pin-project-lite",
  "rustc-hash",
  "rustls 0.23.37",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "thiserror 2.0.18",
  "tokio",
  "tokio-stream",
@@ -8203,7 +8203,7 @@ checksum = "bde7a5d5102f1cff03d482240f0ed20551661f63663620f4b26112ed751165e9"
 dependencies = [
  "cfg_aliases",
  "libc",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "tracing",
  "windows-sys 0.61.2",
 ]
@@ -8223,7 +8223,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


jackzampolin:
The noq/noq-proto/noq-udp 1.0.1 → 1.1.0 bump itself looks right (picks up n0-computer/noq#732 and #748, matching what we were waiting on for #1090/#1091, and keeps overflow checks untouched).

One concern before merging: the Cargo.lock diff contains a number of downgrades on crates unrelated to noq — e.g. `windows-sys` 0.61.2 → 0.48.0/0.59.0/0.60.2 across several untouched packages, `socket2` 0.6.3 → 0.5.10 in quinn/hyper-util, and one dep entry moving from `syn 2.0.117` to `syn 1.0.109`. A targeted `cargo update -p noq -p noq-proto -p noq-udp` shouldn't touch those entries, so it looks like the lockfile was regenerated wholesale (or from a stale base). The downgrades are latent risk on paths CI doesn't exercise (Windows builds in particular).

Could you redo the lock change as a targeted update so the diff only carries the noq crates and their genuinely-required transitive bumps? Happy to merge right after.

jackzampolin:
Lockfile redone in 213ec7eb — the diff vs main is now exactly the three noq crates at 1.1.0 plus `cfg_aliases` 0.2.1→0.2.2 (a hard requirement of noq 1.1.0). No windows-sys/socket2/syn changes remain, and `overflow-checks = true` is untouched.

Root cause of the original churn, for the record: with the workspace's `rust-version = 1.91`, cargo ≥1.84's MSRV-aware resolver (`incompatible-rust-versions = fallback`) de-prioritizes MSRV-incompatible versions whenever it rewrites the lock, and full re-resolution flips ambiguous dep edges (e.g. windows-sys 0.61.2→0.60.2 where both exist in the lock). So even a targeted `cargo update -p noq` reproduces most of the downgrades locally — the minimal diff required restoring main's lock and applying only the required stanzas, verified with `cargo metadata --locked`.

Validated on the final lockfile: `cargo test -p p2p --features iroh-transport`, `cargo clippy --all -- -D warnings`, `cargo fmt --all -- --check` — all clean. Merging.

Issue #885, opened 2026-04-16
P2P replay / push noise under concurrent live desktop chats


jackzampolin:
This note is intended for an agent investigating P2P stability in
`defradb.rs`.

## Repro

Command:

```bash
DEFRA_AGENT_DESKTOP_LIVE_BACKEND_ENDPOINT=http://workstation-1:8000/v1 \
DEFRA_AGENT_DESKTOP_LIVE_BACKEND_MODEL=MiniMax-M2.7-NVFP4 \
DEFRA_AGENT_DESKTOP_LIVE_BACKEND_PROVIDER=openai-compatible \
cargo test -p defra-agent-desktop desktop_app_live_inference_smoke -- --ignored --nocapture
```

What the smoke does:

- starts a desktop embedded node
- starts a peer embedded node
- spawns a live agent backed by the workstation model
- configures file tools
- creates two conversations
- submits overlapping live requests
- verifies both conversations can complete tool-driven responses
- switches activities and returns to chat

Important outcome:

- the smoke now passes end-to-end
- the transport still emits a large amount of replay / push / shutdown noise during the run

## Observed Warning / Error Families

These show up repeatedly during a passing run:

- `Existing document replay PushLog failed ... error=dial error: closed`
- `Existing document replay PushLog failed ... error=codec error: failed to read length: connection lost`
- `Existing document replay PushLog failed ... error=channel send error`
- `Existing document replay PushLog failed ... error=channel receive error`
- `Failed to send two-stream response via token ... error=codec error: failed to write length: connection lost`
- `Failed to complete connection handshake: aborted by peer: the application or application protocol caused the connection to be closed during the handshake`
- `Failed to get replicators for push error=channel send error`
- `Failed to broadcast to both topics ... doc_error=channel send error collection_error=channel send error`
- `Fire-and-forget broadcast failed — document committed locally but NOT replicated`
- `Iroh endpoint task did not stop after graceful shutdown; aborting`
- `Endpoint dropped without calling Endpoint::close. Aborting ungracefully.`

## Why This Is Interesting

The desktop smoke is no longer failing on basic chat behavior:

- two concurrent live chats complete
- tool-driven file reads work
- follow-up messages work
- session switching works

But the underlying transport still looks unhealthy while that is happening.

That suggests at least one of these:

- replay repair is retrying against stale or already-closed peer/stream state
- replicator bookkeeping can outlive the connection it depends on
- transport/channel teardown races with replay push scheduling
[...]

PR #1176, opened 2026-07-22
fix(db-merge): bound parent replay stack usage

iverc:
## Problem

Deep replicated document and collection histories are replayed through boxed async recursion. The configured depth bound allows chains far deeper than a Tokio worker stack can hold, so the desktop-like follow-up workload from #885 aborts the process with a stack overflow before replication can recover.

## What changed

- Traverse composite and collection parent chains with explicit heap-backed enter/exit frames.
- Preserve parent-first merge order, depth limits, deduplication, retryable outcomes, and standalone versus batch transaction behavior.
- Keep each root block's exact payload while loading parent payloads from the blockstore.
- Add worker-stack regressions covering 256-level composite and collection chains.

This isolates the merge-stack failure found through #885. The separate full-DAG sender backlog observed under the same stress workload remains tracked by #1116.

## Validation

- [x] `cargo test -p db-merge` (100 passed)
- [x] `cargo test -p defra-node --features p2p live_replicator_same_session_followup_turn_converges -- --ignored --nocapture` (passed without stack exhaustion)
- [x] `cargo clippy -p defra-node --features p2p --all-targets -- -D warnings`
- [x] `git diff --check`

Advances #885.

### DIFF
--- a/crates/db-merge/src/merge_handler/collection.rs
+++ b/crates/db-merge/src/merge_handler/collection.rs
@@ -1,13 +1,79 @@
 use super::batch::{PendingFieldBlockFinalization, PendingMergeEvent, PendingPostCommitAction};
 use super::*;
 
+enum CollectionMergeFrame {
+    Enter {
+        cid: Cid,
+        block: Option<Block>,
+        payload: Option<defra_core::block::CollectionDeltaPayload>,
+        child_cid: Option<Cid>,
+        depth: usize,
+        is_root: bool,
+    },
+    Exit {
+        cid: Cid,
+        block: Block,
+        payload: defra_core::block::CollectionDeltaPayload,
+        depth: usize,
+        is_root: bool,
+    },
+}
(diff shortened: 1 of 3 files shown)


jackzampolin:
Merged — verified the iterative rewrite is order-faithful across all four recursive paths (heads pushed reversed so pop order preserves parent-first, oldest-to-newest post-order; dedup, depth-limit, and error-propagation semantics byte-for-byte equivalent), and empirically confirmed the new 256-level regression tests abort with a stack overflow when run against main's recursive implementation — they genuinely bite.

One note for future hook authors: on a non-root frame, a `CompositeMergePreparation::Complete(Merged)` from a `CompositeMergeHook` would now return early where the old code continued past the parent. Unreachable today (the only hook returns `None`/`retryable_skip`), but worth a comment on the trait if a hook ever gains a `Merged` return.

Issue #474, opened 2026-02-18
Implement browser-to-server document sync in WASM client


jackzampolin:
## Scope

This issue is the umbrella for the WASM browser-client track. It folds in what were previously separate sub-tracks (sync impl, passkey identity, browser-p2p research, docs demo). Browser-to-server document sync (the original body, below) is the highest-priority concrete work; the other concerns are listed under *Related sub-concerns folded in* and can be split back out if/when they materialize as concrete PRs.

## Browser-to-server document sync (original scope)

Once the sync protocol is designed and WebSocket subscriptions work, implement the actual sync in the WASM `DefraClient`.

### Work

- Add `sync(server_url: &str)` method to `DefraClient` (wasm-bindgen)
- Browser → server: push local mutations via HTTP POST to server's GraphQL API
- Server → browser: subscribe to changes via WebSocket (or SSE fallback)
- On receiving server updates, merge into local CRDT state
- Handle initial sync (first connect / reconnect after offline)
- Handle authentication (include JWT from passkey identity)
- Connection lifecycle: connect, reconnect on failure, backoff

### JS API

```javascript
const client = await DefraClient.create({ db_name: 'myapp' });
await client.add_schema(\`type User { name: String }\`);

// Start syncing with a server
await client.sync('https://defra.example.com');

// Mutations are automatically synced
await client.mutate(\`mutation { create_User(input: {name: "Alice"}) { _docID } }\`);

// Server changes appear locally via subscription
const result = await client.query(\`{ User { name } }\`);
```

### Depends on

- Sync protocol design (HTTP API)
- GraphQL WebSocket subscriptions
- Identity API in WASM crate (for authenticated sync)

## Related sub-concerns folded in (previously separate issues)

These were closed and consolidated here on 2026-05-07. If any grows concrete enough to need its own tracking issue, link the new issue as a child of this one.

- **Browser-to-server sync protocol spike** (was #472) — js-libp2p vs custom HTTP/WS
- **Passkey-to-DID identity research** (was #475) — architecture for binding WebAuthn-derived material to a DID/identity
- **JS WebAuthn ceremony wrapper + DID key management** (was #476)
- **Identity API in WASM crate** (was #477)
- **Browser P2P transport viability research** (was #479) — libp2p WebSocket vs Iroh
- **Docs playground update for current WASM API** (was #480)
- **Browser sync demo in docs playground** (was #481)

jackzampolin:
[...]

PR #1177, opened 2026-07-22
feat(wasm): add browser-to-server document sync

iverc:
Addresses the browser-to-server synchronization scope of #474.

## Problem

The WASM client persisted documents locally but had no supported path to exchange CRDT state with a DefraDB server. Browser applications could not perform an initial catch-up, push later mutations, receive remote updates, authenticate sync requests, or recover after a dropped connection.

## What changed

- Add authenticated `DefraClient.sync` support with initial push/pull, automatic local mutation sync, SSE-driven remote updates, reconnect recovery, and bounded exponential backoff.
- Add Go-versioned `POST /api/v0/sync` and `/api/v1/sync` endpoints backed by the database CRDT merge path.
- Enforce NAC and document ACP permissions before writes, register ownership only for new authenticated protected documents, and filter pull results per document.
- Validate payload limits, CIDs, DAG reachability and shape, canonical genesis DocIDs, field links, signatures, and collection ownership before storing any request block.
- Add bounded pagination and response handling, root-based echo suppression, serialized exchanges, event overflow recovery, and task cleanup when clients resync, close, or are dropped.
- Move merge contracts into `defra-core` and make the channel event bus runtime-neutral so the existing merge engine and update stream can run under WASM without changing native P2P behavior.
- Add CRDT convergence, authorization, pagination, atomic validation, route compatibility, malformed-DAG, and SSE framing coverage.

## Validation

- [x] `cargo test --workspace --exclude integration-test --quiet`
- [x] `cargo test -p db-merge -p cli -p defra-http browser_sync --lib` (14 passed)
- [x] `cargo clippy --workspace --exclude integration-test --all-targets -- -D warnings`
- [x] `cargo clippy -p db-merge --all-targets -- -D warnings`
- [x] `cargo check -p defra-wasm --target wasm32-unknown-unknown --tests`
- [x] `cargo test -p defra-wasm --target wasm32-unknown-unknown --no-run`
- [x] `cargo clippy -p defra-wasm --target wasm32-unknown-unknown --no-deps -- -D warnings -A clippy::arc-with-non-send-sync`
- [x] `cargo build -p cli`


jackzampolin:
Reviewed. Overall quality is high — the validation layer (CID/codec/size, DAG reachability/shape, canonical genesis docID, cross-collection and field-link checks), NAC/ACP gating, body limits, and test coverage are well above typical for a feature this size, and the runtime-neutral events/db-merge refactor is clean. The failing `P2P Integration (Linux)` check is the known pre-existing `manage_relay` 408 flake (`manage_collection_add_over_p2p_authorized`, relay wait timeout) — this branch doesn't touch the manage-relay path, so it's not attributable to this PR.

**One blocking issue:**

1. **Ownership squatting on protected documents** — `crates/cli/src/browser_sync_adapter.rs:176-199`: `apply_document` registers a new protected document to the *caller's* DID rather than the block's cryptographically verified creator. Scenario: authenticated Bob obtains the DAG of a document authored and signed by Alice (pulled from another node or shared out-of-band), pushes it to a server that doesn't have it yet, and becomes the registered ACP owner — when Alice later syncs, `can_access(Update)` forbids her own document. This diverges from both existing conventions in `crates/db-merge/src/acp_merge_handler.rs:169-198` (replicated docs register `effective_creator()` from the block signature; under Local ACP replicated docs aren't registered at all, matching Go). Fix: register the verified creator from the genesis block's signature, or refuse registration when verified creator ≠ caller DID.

**Non-blocking follow-ups:**
2. One oversized document (> `MAX_SYNC_PAYLOAD_BYTES`) permanently wedges sync — `load_document` hard-errors instead of skipping, so every unfiltered pull page 422s and `recover_full_sync` retries forever with no per-document quarantine (`browser_sync.rs:1186`, `session.rs:246-258`).
3. An empty push with no pull performs no NAC check (route is Dynamic, middleware doesn't gate it), letting unauthenticated callers probe whether sync is enabled; and the push path checks only `DocumentUpdate` even when creating brand-new documents (`handlers/browser_sync.rs:24-42`).
4. No fetch timeout/AbortController and no SSE staleness detection — a half-dead connection blocks `exchange` indefinitely while holding `exchange_lock` (`wasm/src/sync/http.rs:104-140`, `session.rs:224-233`).
5. Multi-document apply isn't atomic despite the framing — a mid-batch retryable failure leaves earlier docs applied while the batch reports failure, triggering a full re-sync. Convergence-safe, but worth per-document results (`browser_sync_adapter.rs:358-367`).
6. Unfiltered pull pages do a full datastore scan then apply the cursor in memory — O(N²/page) for a complete initial pull; fine for browser-scale datasets, worth a note for large stores.

Happy to merge once #1 is addressed.

jackzampolin:
Blocker addressed in 1c2ee705 (plus merge of main in 12fbe6ee, resolving the #1176 iterative merge-handler rewrite against this PR's type relocation).

Ownership registration now anchors to the **cryptographically verified genesis creator**, mirroring `acp_merge_handler.rs`'s replication convention: a new protected document registers only the DID verified from the genesis block's signature — never the transport caller; an unsigned genesis registers nobody (unregistered == public under Local ACP, matching Go); an unverifiable genesis signature rejects the push, consistent with merge-time `SignatureVerificationFailed` handling. Squat-by-signature-stripping is impossible since altering the genesis block changes its CID and therefore the derived docID.

Evidence: new tests `foreign_signed_document_cannot_be_squatted_by_pushing_caller` and `unsigned_document_stays_unregistered_for_authenticated_caller` failed pre-fix (Bob's DID registered as owner of Alice's signed doc) and pass post-fix, plus `validation_rejects_forged_genesis_signature`. Full runs green: `cargo test -p cli browser_sync` (8), `cargo test -p db-merge` (107), full ACP integration suite (198 passed incl. go_* parity against the pinned Go binary), workspace tests, clippy `-D warnings`, fmt.

One semantic note worth flagging: browser clients that create documents **without** a signing identity now produce docs that stay public server-side (no ownership registration) — intended Local-ACP semantics, but a behavior change from the original branch. Merging once checks are green.

PR #1186, opened 2026-07-23
fix(p2p): pending-DAG quota reporting + replicator retry redial

### LINKED ISSUE #1180: p2p: pending-DAG peer-quota follow-ups (fairness e2e, error field, dead registration API)
Follow-ups from the PR #1152 review (per-source-peer pending-DAG admission quotas):

1. **Multi-pusher fairness e2e.** The W2 fairness property (a noisy peer cannot starve a healthy peer) is only unit-tested. The integration binary `p2p_admission` runs with `DEFRA_P2P_MAX_PENDING_DAGS=1`, where per-peer quota = max(1/4,1) = 1 equals the global cap, so it exercises the quota code path but cannot distinguish quota from the global bound. Add a multi-pusher e2e with MAX >= 8 asserting a healthy pusher's docs land while a noisy pusher is saturated (fleet-conditions standard: concurrent traffic, anti-vacuity asserts).

2. **Misleading error field on quota rejection.** `pushlog.rs:371`: a per-peer-quota rejection returns `Error::PendingDagCapacity { max: global_max }`, so the typed error/HTTP surface reads "at capacity (1000)" when the peer's ~250-entry quota tripped. WARN logs carry `max_per_peer` and the wire reply is the fixed sentinel either way — polish only, but ops triage will see the wrong number.

3. **Dead public API.** `pending_dag.rs:364,409`: `register_docsync_dag`/`register_branchable_dag` have zero callers (real DocSync goes through the poll-based dag_fetcher) and silently drop on quota. Remove them.

iverc:
Closes #1180. Improves #1154 (kept open — see below).

### #1180 — per-source-peer pending-DAG quota follow-ups

- **Error field (#1180.2).** A per-peer-quota rejection returned `PendingDagCapacity { max: global_max }`, so the nack and its typed error read "at capacity (1000)" when the smaller per-peer quota (~250) actually tripped. `insert_pending_dag` now has a `try_insert_pending_dag` variant that returns which limit tripped, and the PushLog admission path reports that number.
- **Dead API (#1180.3).** Removed `register_docsync_dag` / `register_branchable_dag` — zero callers (real DocSync goes through the poll-based dag fetcher), and they silently dropped on quota.
- **Fairness e2e (#1180.1).** New `p2p_admission_fairness` binary: with `MAX_PENDING_DAGS=8` (per-peer quota 2), a frozen noisy pusher can hold at most its quota while a healthy pusher's writes must all land. Anti-vacuity is clean: with only the noisy peer active, the global cap can never fill from one peer, so any "at capacity" rejection is necessarily the per-peer quota. Passes in ~45–60s across runs.

### #1154 — replicator retry stalls after a target restart (partial fix, issue kept open)

Root cause: the libp2p retry pass skips any replicator peer not in `connected_peers()`, and **nothing redials a replicator target once it restarts** — so a pusher whose connection dropped keeps its entire persisted retry ledger parked behind the connectivity gate. The iroh path already dials on demand; the libp2p path did not. `redial_replicator` dials the peer's stored replicator addresses when it is absent, in both retry-loop copies (embedded + CLI).

This is a real fix — retries re-push the full DAG (immediate merge, no cap-1 bottleneck), so once the connection is re-established the ledger drains. The `hub_restart_recovers_success_acked_pending_dags` acceptance test went from **consistently losing ~744/3048 documents to full recovery (0 lost, 170s)** on a clean run.

**Why #1154 stays open:** recovery is still timing-racy. When reconnection lands, everything drains; when it doesn't, ~750 documents stall (same as before the fix). The residual is the reconnection reliability itself — a stale connection can linger after the hub dies (so the pusher pushes into a dead socket and times out instead of redialing) and the retry backoff ladder (30s→60s→120s…) races the test's 240s deadline. Fully closing #1154 needs the reconnection/expedite piece and warrants hub-log-level investigation; this PR removes the "ledger never fires at all" failure mode.

Validation: workspace `clippy --all -D warnings` + feature-gated `db`/`defra-node --features p2p` clippy clean, fmt clean, p2p unit tests green, pending-dag/pushlog tests green, fairness e2e green ×2.

### DIFF
--- a/crates/cli/src/commands/start/server_p2p.rs
+++ b/crates/cli/src/commands/start/server_p2p.rs
@@ -14,6 +14,35 @@ use p2p::P2PTransport;
 type WireDocumentAcp = Option<Box<dyn FnOnce(Arc<dyn acp::DocumentACP>)>>;
 type WireKms = Option<Box<dyn FnOnce(Arc<dyn kms::KmsService>) + Send>>;
 
+/// Redial a replicator target that dropped, using its stored addresses, so a
+/// stalled connection cannot strand the peer's persisted retry ledger after
+/// the target restarts (the connectivity gate below would otherwise skip it
+/// forever, while nothing else redials it).
+async fn redial_replicator<S: storage::corekv::Store>(
+    peerstore: &storage::stores::Peerstore<S>,
+    handle: &p2p::P2PHostHandle,
+    peer_id_str: &str,
+    peer_id: libp2p::PeerId,
+) {
+    let Ok(Some(bytes)) = peerstore.get_replicator(peer_id_str).await else {
+        return;
+    };
+    let Ok(info) = p2p::ReplicatorInfo::from_bytes(&bytes) else {
+        return;
+    };
+    let addrs: Vec<libp2p::Multiaddr> = info
+        .addresses
+        .iter()
+        .filter_map(|addr| addr.parse().ok())
+        .collect();
+    if addrs.is_empty() {
+        return;
+    }
+    if let Err(error) = handle.dial(peer_id, addrs).await {
+        tracing::debug!(peer_id = %peer_id, %error, "replicator retry redial failed");
+    }
+}
+
 async fn set_persisted_replicator_status<S: storage::corekv::Store>(
     peerstore: &storage::stores::Peerstore<S>,
     peer_id: &str,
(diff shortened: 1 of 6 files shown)


iverc:
Addressed both review comments in b3cbd850: the duplicated `redial_replicator` now lives once in `defra-p2p-adapter` (gated behind the `libp2p` feature) and both retry loops call it; the fairness test's status poll now uses a bounded 5s timeout and returns `Option<u64>`, so the global-cap assertion only runs on a successful read. Fairness e2e still green after the refactor.

PR #1189, opened 2026-07-24
fix: make the wasm32 build clippy-clean

iverc:
Nothing enforced clippy on wasm32. CI's WASM job runs `wasm-pack build`, so warnings on that target never failed anything and quietly accumulated across seven crates. `cargo clippy -p defra-wasm --target wasm32-unknown-unknown --all-targets -- -D warnings` fails on main with 26 errors; it passes here.

## Native-only code that was never gated

- **storage**: `ReadSet`, `ReadRange` and their helpers were left ungated even though `ConflictTracker` — the only thing that consumes them — is already `not(target_arch = "wasm32")`. Whoever gated the tracker missed the types it is built on. The `HashSet` and `IterOptions` imports move with them.
- **acp**: both persistent stores require `S: Store + Send + Sync`, which the wasm LevelDB store cannot satisfy, so their trait impls were excluded on wasm32 — leaving the modules exporting a struct with no impl behind it. Gate the modules instead. Each also carried a `#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]` written to support wasm that the blanket cfg above it made permanently dead; those are removed rather than left implying support that cannot exist.
- **db**: `process_downsample_update`'s only caller is the downsample task, already native-only; `view_ops`' imports are used only inside its native-gated block.
- **query**: the executor fetched `deferred_acp_mutations` and dropped it on wasm32. The task-local scoping it feeds is native-only, and the getter is a side-effect-free `Arc` clone, so the binding is now gated — no behaviour change.

## False positives

`clippy::arc_with_non_send_sync` fires in four places because wasm32 deliberately drops `Send + Sync` (see `MaybeSendSync` in defra-core `thread_bounds`) for the single-threaded browser runtime. Those `Arc`s have no threads to cross, and `Rc` is not available since the surrounding types and public signatures are `Arc`. Allowed per-site, scoped to wasm32 where the crate also builds natively, each with the reason recorded.

## Preventing recurrence

The WASM job now runs clippy with `-D warnings` before `wasm-pack build`, and the toolchain step gains the `clippy` component. Without this the warnings just re-accumulate.

## Verification

The originally-failing command now exits 0.

No native regression:

- `cargo clippy --all --all-targets -- -D warnings` exit 0, zero warnings
- `cargo fmt --all -- --check` clean
- all five CI feature-gated clippy passes exit 0
- `cargo build -p defra-wasm --target wasm32-unknown-unknown --release` exit 0 (what `wasm-pack` wraps)
- `cargo test -p storage -p acp -p datastore -p db -p query -p db-merge` — 50 binaries, zero failures

The SSI conflict tests (`detects_write_to_committed_read_prefix`, `detects_read_of_committed_write_key`, `ignores_document_collection_scan_prefixes`) and the `RecordGuard` tests from #1183 all pass, confirming write-skew detection is untouched.

Gating the `query` binding revealed a real functional gap that this PR does not attempt to close — filed separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
+          components: clippy
 
       - uses: Swatinem/rust-cache@v2
         with:
@@ -76,6 +77,9 @@ jobs:
 
       - name: Exclude tools with private deps from workspace
         run: sed -i '/"tools\/integration-test"/d; /"tools\/ffi-test"/d' Cargo.toml
+
+      - name: Clippy (wasm32)
+        run: cargo clippy -p defra-wasm --target wasm32-unknown-unknown --all-targets -- -D warnings
 
       - name: Install wasm-pack
         run: cargo install wasm-pack --locked
--- a/crates/acp/src/lib.rs
+++ b/crates/acp/src/lib.rs
@@ -32,6 +32,9 @@ mod identity;
 mod local;
 pub mod nac;
 mod permission;
+// Both persistent stores require `S: Store + Send + Sync`, which the wasm
+// LevelDB store cannot satisfy, so they are native-only.
+#[cfg(not(target_arch = "wasm32"))]
 mod persistent;
 pub mod policy_yaml;
 pub mod read_access;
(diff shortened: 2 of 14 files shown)


iverc:
The wasm32 transactional scoping gap noted above is now #1190.

PR #1192, opened 2026-07-26
bench(storage): compare backends through the DefraDB transaction wrapper

### LINKED ISSUE #1009: Expose Lark production tuning knobs in DefraDB storage options
## Severity
Medium

## Problem
The DefraDB Lark backend maps only a subset of `lark_kv::Options` into `LarkStoreOptions`. That makes it harder to run Lark with the same production/performance intent as the RocksDB backend, especially when comparing backends through DefraDB rather than raw microbenchmarks.

The benchmark retest showed that Lark performance is sensitive to options such as write-buffer sizing and block layout. The current DefraDB options expose some of these, but not the full production set that Lark already supports.

## Evidence
`crates/storage/src/backends/lark/config.rs` currently exposes:

- block cache size
- write buffer size
- block size
- bloom bits
- compression on/off
- L0 trigger
- level base/multiplier
- target file size
- close timeout
- durability

Important Lark options not surfaced include:

- `max_background_compactions`
- `max_write_buffer_number`
- `level0_slowdown_writes_trigger`
- `level0_stop_writes_trigger`
- pending-compaction byte limits
- `evict_compaction_data_from_page_cache`
- `partitioned_index` and `metadata_block_size`
- `block_cache_num_shard_bits` and `strict_capacity_limit`
- per-level compression configuration
- prefix extractor support where DefraDB key layout can supply a stable prefix

By comparison, `crates/storage/src/backends/rocksdb/config.rs` exposes background jobs, write-buffer count, L0 triggers, compaction style, block size, compression, and related production knobs.

## Expected Fix
Expand the DefraDB Lark configuration surface deliberately:

- Add builder methods and environment variables for the most important Lark production knobs.
- Map those fields into `lark_kv::Options` in `to_lark_options()`.
- Document recommended profiles for DefraDB workloads instead of relying on raw defaults.
- Update the DefraDB storage benchmark harness to honor those options when comparing Lark with RocksDB.
- Keep defaults conservative and compatible with existing behavior.

iverc:
Closes #1009.

The Lark config surface was already expanded (block/cache/level/compaction/write-buffer knobs, `LARK_*` env vars, `to_lark_options` mapping) in #894. The one open acceptance item was the benchmark: the storage bench only exercised `redb`, so there was no way to compare Lark and RocksDB against it *through DefraDB* — the comparison the "until Lark" migration decision needs.

This makes the fixture and workload generic over the backend and runs the same seven benches against every enabled store, grouped as `storage/<backend>`. Each store is built through its own options type (honoring its `*_` env vars), so a *configured* backend is measured, not just raw defaults:

```
cargo bench -p storage --features "redb,lark,rocksdb"
```

Sample point-read result (`get_from_tree`): lark 192ns, rocksdb 414ns, redb 464ns.

## On the prefix-extractor knob

I did not surface it, and recorded why in `to_lark_options`. Lark's only built-in extractor is fixed-length, but DefraDB keys are variable-length namespaced (`d/d/<collection>/<docID>`), so a fixed-length prefix bloom cannot align with the collection/index boundaries lookups and scans key on. The issue scoped this as "where DefraDB key layout can supply a stable prefix" — it does not.

## Recommended profiles

Served by the existing `LARK_*` env-var table in `config.rs`; a separate profiles doc would violate the repo's no-docs-files rule.

## Validation

- Bench compiles with `--features "redb,lark,rocksdb"` and with `--features redb` alone (backward-compatible).
- Smoke run confirms all three backends execute and report.
- `cargo fmt --all -- --check` clean; `cargo clippy -p storage --features "redb,lark,rocksdb" --all-targets -- -D warnings` exit 0; `cargo test -p storage --features lark` green.

### DIFF
--- a/crates/storage/benches/transaction.rs
+++ b/crates/storage/benches/transaction.rs
@@ -1,9 +1,9 @@
+use std::hint::black_box;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::{Arc, OnceLock};
+use std::sync::Arc;
+use std::sync::OnceLock;
 
 use criterion::{criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
-use std::hint::black_box;
-use storage::backends::RedbStore;
 use storage::corekv::{IterOptions, Reader, Store, Writer};
 use tempfile::TempDir;
 
@@ -12,9 +12,15 @@ fn runtime() -> &'static tokio::runtime::Runtime {
     RUNTIME.get_or_init(|| tokio::runtime::Runtime::new().unwrap())
 }
 
-struct BenchStore {
-    _temp_dir: TempDir,
-    store: Arc<RedbStore>,
+/// Fixture holding a seeded store plus the keys the benchmarks read against.
+///
+/// Generic over the backend so the same workload runs identically against
+/// every enabled store (`redb`, `lark`, `rocksdb`), giving a side-by-side
+/// comparison of the DefraDB transaction wrapper on each. Each backend
+/// honours its own `*_`-prefixed environment options at construction, so the
+/// harness measures a configured backend, not just its raw defaults (#1009).
+struct BenchStore<S: Store> {
+    store: Arc<S>,
     tree_key: Vec<u8>,
     pending_key: Vec<u8>,
     pending_value: Vec<u8>,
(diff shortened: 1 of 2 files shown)


iverc:
Addressed CodeRabbit's note in 5e487d31: the doc comment overclaimed. `RedbStoreOptions` has no `from_env` (only cache-size/close-timeout/durability builders), so redb has no environment-configuration path. Narrowed the comment to state that Lark and RocksDB are built through their `*_` env options while redb uses its defaults, rather than inventing a redb env path (out of scope for #1009).

PR #1196, opened 2026-07-27
fix(storage): don't conflict blind writes of identical content-addressed blocks

### LINKED ISSUE #1194: Transaction conflicts under concurrent write load — streaming response snapshots fail at ~8 concurrent writers
# Transaction conflicts under concurrent write load — streaming response snapshots fail at ~8 concurrent writers

## Summary

When multiple agent request executors concurrently flush streaming response snapshots to DefraDB via `update_AgentResponse` mutations, transaction conflicts occur frequently enough to exhaust client-side retry budgets (3 retries, 100ms exponential backoff). At 8 concurrent writers, ~75% of write attempts fail with `transaction conflict. Please retry`, causing the agent runtime to abort the response before it can persist results.

## Environment

- **defradb.rs**: pinned at rev `ef772a439ec84d3fb1a7c0ef740085666ea008b4` (v0.16.2 fork with active-snapshot conflict retention)
- **Embedding**: embedded node, single-process, HTTP GraphQL API on `127.0.0.1:9191`
- **Workload**: 8 concurrent agent request executors, each flushing streaming response snapshots (~1 write per ~2-4 seconds per executor) via `update_AgentResponse` mutations with `filter: { _docID: { _eq: "..." }, status: { _eq: "streaming" } }`
- **Client retry**: 3 retries with exponential backoff (100ms, 200ms, 400ms), then bail

## Reproduction

1. Start an embedded DefraDB node
2. Create 8 concurrent tasks that each repeatedly execute `update_AgentResponse` mutations (filtering by `_docID` and `status`) at ~1 mutation per 2-4 seconds
3. Observe transaction conflicts on ~75% of mutations after 3 retries

## Observed error

```
flush_streaming_response_snapshot failed: transaction conflict. Please retry
```

The error string comes from DefraDB's storage layer and propagates through the GraphQL response's `errors` array. The gents client-side retry logic (`execute_graphql_with_conflict_retry`) retries up to 3 times with 100ms exponential backoff, but this is insufficient under the contention level.

## Impact

At concurrency=8, ~75% of streaming response snapshot flushes fail after exhausting retries. This causes:
- Agent responses to abort mid-stream (the runtime bails on the flush failure)
- Downstream effects (tag writes, tool calls) never execute because the response snapshot couldn't be persisted
- The agent appears to "complete" from the CLI's perspective (exit 0) but produces zero useful output (0 tokens, 0 tags written)

At concurrency=2 with 30s stagger, conflicts are rare and the system works correctly. At concurrency=4, conflicts are intermittent. At concurrency=8, the system is effectively broken.

## Expected behavior

DefraDB should either:
1. **Handle concurrent writes more gracefully** — the `update_AgentResponse` mutations target different documents (different `_docID`s), so there is no logical conflict. The transaction conflict appears to be a storage-layer artifact of concurrent transactions touching the same collection/index, not a true data conflict.
2. **Provide better conflict resolution** — automatic retry at the storage layer (opaque to the client) would be ideal for cases where the transactions do not actually conflict on the same keys.
3. **Support higher concurrency** — a single embedded node should handle 8+ concurrent writers without chronic transaction conflicts, especially when writes target different documents.

## Additional context

- The gents client already has retry logic (see `execute_graphql_with_conflict_retry` in `crates/gents/src/retry.rs`), but 3 retries with 100ms base backoff is insufficient at this contention level.
- Increasing client-side retries to 8+ with longer backoff (500ms+) would help, but pushes the problem to the client rather than fixing the storage layer.
- The workload is write-heavy but the writes are to distinct documents — this should be parallelizable without conflicts.
- This was discovered while running a batch session classification job (8 concurrent agent requests classifying coding sessions), where each agent reads from MCP tools and writes streaming response snapshots + tag mutations.

jackzampolin:
Fixes #1194 — chronic `transaction conflict. Please retry` failures when concurrent writers update **different** documents through the HTTP GraphQL API.

## Root cause

Field-delta blocks are content-addressed (blockstore key = `b` + CID) and contain no document identity — a delta is `{field_name, priority, schema_version_id, data}` plus prev-head links. Documents created with the same field value have byte-identical delta chains for that field, so when two documents both rewrite `status: "streaming"` at the same CRDT chain depth, their update transactions write the **same** blockstore key. `ConflictTracker::check_and_record` treated that blind write-write overlap as a `TxnConflict` and aborted the second committer.

Demonstrated with instrumentation on the reported workload shape (redb, HTTP GraphQL, `update_X` with `filter: {_docID: {_eq}, status: {_eq: "streaming"}}`, input rewriting `status`): every conflict at concurrency 8 was `our write hit a committed WRITE key` on a `b`+CID key shared across workers — never a document, index, head, or metadata key. Variants that don't produce identical deltas (content-only updates, indexed or not) showed **zero** conflicts on current `main`, i.e. the earlier conflict-retention/serialization fixes (f5930d67, fa594d56, 1b479d1c, 2450e0d8) removed the amplifiers, and this shared-block collision is the remaining defect.

## Fix

Exempt content-addressed block data keys from the **blind write-write** check only, in the shared `ConflictTracker` used by all five backends (redb, memory, fjall, rocksdb, lark). The predicate requires the blockstore namespace prefix, a non-merge-tracking payload, and fully valid CID bytes (`cid::Cid::try_from`). Writers of the same content-addressed key write identical bytes by construction, so the overlap is not a serializability hazard — Go's badger likewise only conflicts reads against writes and never blind write-write.

Still enforced:
- read-vs-write conflicts on block keys (anti-dependencies),
- blockstore merge-tracking keys (`ToMergeIndexKey`, mutable presence state) and blockstore-namespace keys that are not valid CIDs,
- every other same-key conflict (documents, versions, heads, indexes, metadata),
- SSI write-skew rejection (`conformance ssi_write_skew_prevented` passes).

## Regression coverage

`tools/integration-test/tests/issue1194_repro.rs` (redb, full stack over HTTP GraphQL):
- **Deterministic** core: two overlapping interactive transactions updating two *distinct* docs with identical deltas — both commits must succeed. Before the fix this failed 100% of runs with `HTTP 409 … transaction conflict. Please retry` (verified by neutralizing the carve-out).
- Control: two overlapping transactions updating the *same* doc — second commit must still conflict.
- Workload shape: 8 docs, 8 barrier-synchronized writers rewriting `status` unchanged for 10 rounds; asserts zero conflicts/errors and correct final per-document state (no lost/misapplied writes).
- Unit controls in `crates/storage/src/backends/shared.rs`: identical block write-write commits; merge-index write-write conflicts; non-CID blockstore-namespace write-write conflicts; block write vs committed read conflicts.

## Before / after (concurrency-8 barrier workload, status-rewrite variant)

| | attempts | conflicts |
|---|---|---|
| before | 80 | 6–8 per run (and 1–2 at concurrency 2/4); txn-pair repro fails deterministically |
| after | 80 | 0; txn-pair passes |

## Validation (re-run on the branch merged with current main)

- `cargo test -p storage` — pass
- `cargo test -p integration-test --test issue1194_repro -- --test-threads=1` — 3/3 pass
- `cargo test -p conformance --test tla_conformance -- ssi::` — pass
- `cargo test --workspace --exclude integration-test --exclude conformance --exclude ffi-test` — all pass (pre-merge); `cargo test -p integration-test --test basic`: all 24 `rust_*` pass, the 10 `go_*` failures are environmental (`Go defradb binary not available` locally)
- `cargo fmt --all --check` clean; `cargo clippy --all -- -D warnings` clean; `cargo clippy -p defra-wasm --target wasm32-unknown-unknown --all-targets -- -D warnings` clean

### DIFF
--- a/crates/storage/src/backends/shared.rs
+++ b/crates/storage/src/backends/shared.rs
@@ -218,6 +218,24 @@ impl ReadSet {
     }
 }
 
+/// Content-addressed blockstore keys: namespace byte `b` followed by raw CID
+/// bytes (0x01 for CIDv1, 0x12 for a CIDv0 sha2-256 multihash).
+///
+/// The key is the hash of the value, so any two writers of the same key write
+/// identical bytes — blind write-write overlap on these keys is not a
+/// serializability hazard, and Go's badger likewise never conflicts blind
+/// writes (it only checks reads against committed writes). Without this
+/// carve-out, concurrent updates to DIFFERENT documents that produce
+/// byte-identical field deltas (e.g. both rewrite `status: "streaming"`)
+/// collide on the shared delta block and spuriously abort (#1194).
+///
+/// Merge-tracking keys under the blockstore namespace (`b` + `m` + CID) do
+/// not match: their first payload byte is `m` (0x6d), and their presence is
+/// mutable state that must stay conflict-checked.
+fn is_content_addressed_block_key(key: &[u8]) -> bool {
+    key.len() >= 2 && key[0] == b'b' && matches!(key[1], 0x01 | 0x12)
+}
+
 fn is_document_collection_scan_prefix(prefix: &[u8]) -> bool {
     // Namespaced datastore document scans use `d/d/...`; root datastore scans
     // use `/d/...`. Go's SSI conflict in the relation tests comes from FK index
@@ -369,7 +387,9 @@ impl ConflictTracker {
         // Check for conflicts against transactions committed after our snapshot.
         for (_, committed_writes, committed_reads) in state.committed_after(read_version) {
             for write_key in &write_keys {
-                if committed_writes.contains(*write_key) || committed_reads.conflicts_key(write_key)
+                if (committed_writes.contains(*write_key)
+                    && !is_content_addressed_block_key(write_key))
+                    || committed_reads.conflicts_key(write_key)
                 {
                     return Err(crate::corekv::Error::TxnConflict);
                 }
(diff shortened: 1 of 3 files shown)


jackzampolin:
Merged current `origin/main` into the branch (no history rewrite). The `shared.rs` conflict was main's new `cfg(not(wasm32))` gating landing where the block-key helper sits — resolved by gating the helper the same way. While resolving I hardened the predicate per review guidance: it now requires the blockstore namespace prefix, excludes `ToMergeIndexKey` merge-tracking payloads explicitly, and validates the payload as a real CID via `cid::Cid::try_from` instead of sniffing the first two bytes; a new unit control proves a blockstore-namespace key with a non-CID payload still conflicts. Also trimmed the integration suite from the diagnostic-era 4-scenario matrix down to the deterministic txn pair, the same-doc conflict control, and one concurrency-8 workload in the reported shape. Full validation re-run on the merged branch (storage, regression 3/3, SSI conformance, fmt, native + wasm32 clippy) — details in the updated PR body.

PR #1202, opened 2026-07-28
Remove dead query modules

jackzampolin:
## Summary

- remove an undeclared `query_parse/types.rs` module whose definitions duplicate the active parser implementation
- remove three comment-only placeholder modules from query and SDL parsing
- remove the corresponding module declarations and stale module-map entries
- remove `async-trait` from `query-parse`; the compiler confirms it is unused

## Why

These files describe planned extractions but contain no executable behavior. One is not part of the module graph at all, while the remaining placeholders compile as empty modules. Keeping them makes the architecture map misleading and conflicts with the repository's preference for documenting future work in issues.

## Impact

This removes 89 dead lines without changing the public API or runtime behavior.

## Validation

- `cargo fmt --all --check`
- `cargo test -p query-parse --all-targets` (195 passed)
- `cargo check -p query-parse --no-default-features`
- `cargo check -p query-parse --target wasm32-unknown-unknown`
- `cargo test -p query --lib` (69 passed)
- `cargo check -p query --all-targets`
- `git diff --check`


jackzampolin:
## Review

Verified each removal claim against `origin/main` — everything checks out. **Verdict: approve.**

**Confirmed:**
- `query_parse/types.rs` was genuinely outside the module graph: `query_parse/mod.rs` on main declares `aggregates, cursor, explain, filters, limits, mutations, ordering, parser, validation, values, variables` — no `types`. The live `ExplainType`/`ParsedOperation` are defined in `query_parse/parser.rs:35,59` and re-exported from there; the deleted file was a near-identical dead copy.
- `sdl_parse/preprocess.rs`, `sdl_parse/types.rs`, and `query/src/runner/helpers.rs` were comment-only placeholders (doc comments + "Future: extract..." notes, zero code items).
- Repo-wide grep on main for `sdl_parse::preprocess`, `sdl_parse::types`, `runner::helpers` → zero hits; no `include!` usage. No re-exports touched.
- `async-trait` appears nowhere in query-parse source at head; Cargo.lock diff is exactly the one dependency-edge removal, no version churn.

**Nits (optional / follow-up):**
1. `crates/query-parse/src/sdl_parse/mod.rs` doc list still advertises `fields`, `validation`, `builder` — and `builder` has no `mod builder;` declaration at all. Pre-existing doc drift, but since this PR is cleaning exactly this category, worth sweeping in the same pass or a follow-up.
2. PR description slightly understates the `runner/helpers.rs` change: it was a *declared* (empty) module that's fully deleted, not just a stale map entry. No action needed — just noting for the record.

PR #1203, opened 2026-07-28
Run runtime interoperability tests in CI

jackzampolin:
## Summary

- remove two lens unit tests that hardcoded a developer-local WASM path and returned success when it was absent
- retain source-built lens migration coverage in the query integration suite
- make three Rust/Go keyring interoperability tests fail explicitly when the Go binary is unavailable
- resolve the Rust CLI through `DEFRA_RUST_BINARY`, with the workspace debug binary as a local fallback
- add the identity test target to the existing Go-compat CI leg

## Why

The removed lens tests reported success on normal CI while executing no assertions. The keyring interoperability tests had the same silent-success behavior and were not selected by the Go-compat naming filter.

The Go-compat job already prepares both pinned Go v1.0.0 and Rust binaries. Adding the identity target schedules 22 existing compatibility tests, including the renamed cross-binary cases, without introducing new infrastructure.

## Impact

Runtime interoperability regressions now fail CI instead of being silently skipped. Unique lens behavior remains covered by source-reproducible integration tests.

## Validation

- `cargo fmt --all --check`
- `cargo check -p lens --tests`
- `cargo test -p lens --lib wasm::tests` (8 passed)
- exact Go-compat identity command against pinned Go v1.0.0 (22 passed)
- `cargo test -p integration-test --test query -- lens_reindex_secondary_index_979:: --test-threads=1` (3 passed)
- `git diff --check`


jackzampolin:
## Review

Verified all three claims against the diff, both branches' CI configs, and the passing CI run. **Verdict: approve.** The change is strictly coverage-positive.

**Confirmed:**
- The removed lens tests in `crates/lens/src/wasm.rs` hardcoded a developer-local absolute path and early-returned (green) when it was absent — i.e., they passed vacuously on every CI machine. Real retained coverage is `lens_reindex_secondary_index_979` (plus the iroh `schema_migration` and cursor reindex tests), which build the `set_default` lens from the committed source crate via `WasmLens::build()` and assert the transformed value end-to-end through indexed queries.
- The keyring hard-fail matches the file's own convention: the 13 pre-existing `go_*` tests in `keyring_lifecycle.rs` already `expect("Go defradb not in PATH")`; the 3 changed tests were the outliers with a silent `else { return }`. A dev without Go already got 13 failures on main — now 16, same class.
- CI routing is exact on both legs: renamed tests' full names contain `::go_`, so the rust/signed-docs legs' `--skip ::go_` correctly excludes them, and the go-compat leg's positional `go_` filter selects them. The count of 22 is exact (13 keyring `go_*` + 3 renamed + 6 macro-generated) and confirmed in this PR's go-compat job log: `22 passed; 0 failed; 35 filtered out`. These 22 previously ran in **no** CI leg.
- The `DEFRA_RUST_BINARY` env hookup was load-bearing, not cosmetic: in CI's cache-hit case `target/debug/defra` doesn't exist, so without the env lookup the now-hard-failing tests would have broken the leg. Wired correctly (job-level env already points at `target/ci/defra`).

**Minor / nits (no action required):**
1. `go_binary()` accepts any `defradb` on PATH via `--help` with no version check against `GO_COMPAT_TAG` — a stale local Go binary silently changes what's tested. Pre-existing pattern and CI is pinned; noting for a possible follow-up.
2. The `go_rust_*` prefix elsewhere (via `for_each_p2p_topology!`) means "mixed Go+Rust cluster"; here it labels CLI keyring interop. Filter-compatible, but slightly overloads the convention.
3. PR body's "lens migration coverage in the query integration suite" is loosely worded — `query/lens.rs` (`lens_migration`) doesn't run WASM; the real retained coverage is the reindex/schema-migration tests the validation section already cites.

PR #1232, opened 2026-07-29
fix(db): complete Go lens migration parity

### LINKED ISSUE #1230: fix(lens): port Go updateDataStore write-back and expose eager materialize
## Summary

Port Go DefraDB's lens **datastore-only write-back** into `defradb.rs`, and expose an eager `materialize_collection` (or equivalent) so callers can advance documents without waiting for organic reads.

This is **Rust ↔ Go parity**, not a new product surface. Tracked from Gents lens-first migration design (`source-inc/gents` PR #931, design §3 / §7.1).

Related umbrella: #717 (lens parity investigation).

## Go reference (behavior to match)

`defradb` `internal/lens/fetcher.go` — `updateDataStore`:

- On a successful lensed read, persist **migrated field values** and the **document version key** into the datastore.
- **No new CRDT commits**, nothing gossipped to peers.
- Subsequent reads hit cached datastore values instead of re-running the full lens pipeline every time.

Go reindex-after-migration effectively materializes whole collections by iterating the lensed fetcher in a write path.

## Rust pin today (broken / stubbed)

In `crates/db/src/lensed_fetcher/migration.rs` (current pin used by Gents: `8eba3d5`):

- Field write-back is stubbed: only `trace!("Would cache migrated field value")` — values are **not** persisted.
- Version key is written under a **malformed** key: `format!("/v/{}", doc_id)` instead of the real datastore layout (`/d/<collection_short_id>/<doc_short_id>/v` or the collection helper equivalent).
- Comments admit: "In a full implementation, we would use the collection's field keys."

Secondary effects called out in Gents review:

- Reindex can leave **index entries and datastore values disagreeing** after migration.
- Purely additive chains (`hasMigrations == false`) never advance stored version keys via the lens path; neither Go nor Rust does identity-only restamp today. Prefer fixing write-back first; identity restamp can be part of the same materialize API.

## Requested work

1. **Port `updateDataStore` semantics** into the Rust lensed fetcher(s) (`lensed_fetcher` and `lensed_auto_commit_fetcher` as applicable):
   - Write migrated fields with correct field keys / encoding (match write path, including DateTime → `Time` for indexes).
   - Write the document version key with the correct key layout.
   - Datastore-only; no new commits; no P2P traffic.
2. **Remove the stub** that writes under the fake `/v/{doc_id}` prefix (must not pollute the keyspace).
3. **Expose eager materialization** on the public/node surface, e.g. `materialize_collection(collection)`:
   - Iterate the collection through the lensed fetcher in a write transaction (same idea as Go reindex-after-migration).
   - Include **identity materialization**: re-stamp version keys for docs whose stored version ≠ active when the path has **no** transforms (additive migrations).
4. **Fix reindex** so index entries and datastore values agree after migration (same encoding path as fresh writes).
5. Tests mirroring Go coverage for write-back + eager materialize + restart (values durable, version advanced, no commit/gossip side effects).

## Why Gents cares

Gents is moving to a lens-first schema engine (`ensure_migrations`). Eager materialization after upgrade must not be a GraphQL-update sweep (that would emit ordinary commits and rewrite history fleet-wide). Without this port, upgrades rely only on lazy read-time transforms and never durably restamp docs.

## Acceptance

- [ ] Lensed read persists migrated values + version key with correct keys
- [ ] No new commits / no P2P from materialization write-back
- [ ] Public eager materialize API for a collection
- [ ] Additive (transform-less) paths can restamp version keys
- [ ] Reindex post-migration matches fresh-write index encoding
- [ ] Regression tests; ideally cross-check against Go behavior

### LINKED ISSUE #1231: fix(lens): pass through documents of unknown collection versions (Go parity)
## Summary

When a document’s stored collection version is **not** present in the local version history, Rust currently **fails the whole query**. Go passes those documents through **unchanged** so mixed-version P2P (newer peer → older node) does not brick reads.

This is **Rust ↔ Go parity**, required for rolling binary upgrades. Tracked from Gents lens-first migration design (`source-inc/gents` PR #931, design §7.2).

Related: #717 (lens parity umbrella), #1230 (write-back / materialize).

## Go reference

`defradb` `internal/lens/lens.go` (~139–149):

```go
// We may receive documents of unknown collection versions, they should
// still be fed through the pipe system in order to preserve order.
err = l.unknownVersionPipe.Put(doc.Doc)
```

Unknown versions are emitted without transform and without failing the query.

## Rust pin today

`crates/db/src/lensed_auto_commit_fetcher/migration.rs` (~287–292 on Gents pin `8eba3d5`):

```rust
if !history.contains_key(&doc_version) {
    return Err(query::error::QueryError::execution(format!(
        "no migration path found for document {} from version {} to {}",
        doc_id_str, doc_version, target_version_id
    )));
}
```

Same class of failure may exist on the non-auto-commit lensed fetcher — audit both.

## Why it matters

During a rolling upgrade:

1. Newer binary applies a schema version + lens edge; documents get stamped with the new version (or arrive via P2P already stamped).
2. Older binary has not yet applied that version to its local history.
3. Rust older node: **reads fail** for those docs.
4. Go older node: docs still readable (untransformed / pass-through).

Gents can promote nodes promptly as policy, but hard-failing the query is still a fleet footgun and diverges from Go.

## Requested work

1. Match Go: if `doc.schema_version_id` is absent from the targeted collection history, **emit the document unchanged** (preserve order if the pipeline cares about order).
2. Do **not** fail the entire query for a single unknown-version doc (unless there is a documented safety reason — call it out if so; Go chose pass-through).
3. Apply consistently across lensed fetch paths (auto-commit and non-auto-commit).
4. Tests:
   - Seed / inject a document stamped with a version id not in local history → query succeeds, doc returned.
   - Known-version docs still migrate via lenses.
   - Optional: multi-doc query with mix of known and unknown versions.

## Out of scope

- Inventing transforms for unknown versions (impossible without the module).
- Changing Gents schema-DAG rejection of **foreign schema versions** in the collection history (that is intentional at `ensure_migrations`). This issue is **document** version stamps only.

## Acceptance

- [ ] Unknown collection version on a document → pass-through, no query-level hard error
- [ ] Known versions still lens-migrate
- [ ] Both lensed fetcher paths covered
- [ ] Regression tests

jackzampolin:
## Summary

- persist lazily migrated documents and version keys without creating new commits
- rebuild migrated secondary indexes and preserve scalar-array representations
- add eager `materialize_collection` support through the DB, node, and C FFI layers
- pass documents with unknown/newer collection versions through unchanged on reads
- include update blocks in FFI subscriptions and await node background-task shutdown so persistent stores can immediately reopen

## Why

This completes the Rust parity work tracked by #1230 and #1231. Lazy migrations previously returned transformed values without fully writing the migrated datastore/index state back, while reads of versions outside the local collection history could fail instead of preserving replicated data from newer peers.

The FFI additions allow the Go `rust_ffi` harness to exercise eager materialization directly and run the existing Go migration suite against the Rust implementation.

## Compatibility note

Go's `updateDataStore` restamps the version key to the local target even when an unknown/newer document version is passed through unchanged. Rust intentionally preserves that foreign version marker so a document received from a newer peer is not mislabeled as the local version. Query results remain compatible with Go; the stored-marker difference is deliberate.

## Hardening included

- generation-keyed migration-context invalidation for positive and negative cache entries (#1233)
- conflict-safe, idempotent concurrent lazy write-back with guarded fresh re-reads (#1234)
- atomic secondary-index maintenance for late lazy migrations (#1235)
- read-only query snapshots with short, explicit write-back transactions only when documents were transformed

## Validation

- `cargo clippy -p db -p document -p ffi --all-targets -- -D warnings`
- `cargo test -p db migration::tests -- --nocapture`
- `cargo test -p db unknown_document_version_passes_through -- --nocapture`
- `cargo test -p ffi -- --nocapture` (152 passed)
- focused concurrent dual-read regression: `concurrent_lazy_reads_of_same_document_are_idempotent`
- direct Go FFI eager-materialization test passes against current Go `develop` (`13c46ec9`, `v1.0.0-19`)
- 10 targeted existing Go migration tests pass through `ffi-test`, including migrated indexes, arrays, sparse and multi-step histories, restart, and both newer-version P2P pass-through cases

Closes #1230
Closes #1231

### DIFF
--- a/crates/db/src/lens_utils.rs
+++ b/crates/db/src/lens_utils.rs
@@ -206,3 +206,50 @@ pub fn build_collection_history(
 
     result
 }
+
+/// Return whether the targeted path from `source_version_id` crosses a transform.
+///
+/// The walk mirrors the lens pipeline's forward-first navigation and is used by reindexing to
+/// avoid stamping transform-less paths before an application has explicitly requested eager
+/// identity materialization.
+pub fn migration_path_has_transform(
+    history: &HashMap<String, TargetedHistoryLink>,
+    source_version_id: &str,
+    target_version_id: &str,
+) -> bool {
+    if source_version_id == target_version_id {
+        return false;
+    }
+
+    let mut current = source_version_id;
+    let mut visited = std::collections::HashSet::new();
+    visited.insert(current.to_string());
+    let mut has_transform = false;
+
+    while current != target_version_id {
+        let Some(link) = history.get(current) else {
+            return false;
+        };
+
+        if let Some(next) = link.next.as_deref().filter(|next| !visited.contains(*next)) {
+            let Some(next_link) = history.get(next) else {
+                return false;
+            };
+            has_transform |= next_link.transform.is_some();
+            current = next;
+        } else if let Some(previous) = link
+            .previous
+            .as_deref()
+            .filter(|previous| !visited.contains(*previous))
+        {
+            has_transform |= link.transform.is_some();
+            current = previous;
+        } else {
+            return false;
+        }
+
+        visited.insert(current.to_string());
+    }
+
+    has_transform
+}
--- a/crates/db/src/lensed_auto_commit_fetcher/fetcher.rs
+++ b/crates/db/src/lensed_auto_commit_fetcher/fetcher.rs
@@ -11,7 +11,12 @@ use crate::commits_fetcher::{CommitsFetcher, CommitsQueryOptions as DbCommitsOpt
 use crate::txn::DbTxn;
 use crate::versioned_fetcher::VersionedFetcher;
 
+use super::migration::MigrationWriteBack;
 use super::LensedAutoCommitFetcher;
+
+type ProcessedDocs = (Vec<Document>, Vec<MigrationWriteBack>);
+type ProcessedDocsWithStatus = (Vec<(Document, bool)>, Vec<MigrationWriteBack>);
+type ProcessedDocsById = (Vec<Document>, Vec<String>, Vec<MigrationWriteBack>);
 
 impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
     pub(super) async fn get_all_impl(
@@ -24,7 +29,8 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
             .map_err(|e| query::error::QueryError::execution(format!("db error: {}", e)))?
             .ok_or_else(|| query::error::QueryError::collection_not_found(collection_name))?;
 
-        let (has_migrations, preloaded_history) = self.load_migration_context(&collection).await?;
+        let (migration_generation, has_migrations, preloaded_history) =
+            self.load_migration_context(&collection).await?;
 
         let target_version_id = &collection.schema().version_id;
 
@@ -36,7 +42,7 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
             );
         }
 
-        let txn = self.db.new_txn(!has_migrations).await.map_err(|e| {
+        let txn = self.db.new_txn(true).await.map_err(|e| {
             query::error::QueryError::execution(format!("failed to create txn: {}", e))
         })?;
 
@@ -47,57 +53,56 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
             query::error::QueryError::execution(format!("failed to get systemstore: {}", e))
         })?;
 
-        let docs = collection
-            .get_all_with_datastore(&datastore, &systemstore)
-            .await
-            .map_err(|e| query::error::QueryError::execution(format!("storage error: {}", e)))?;
+        let read_result: query::error::Result<ProcessedDocs> = async {
+            let docs = collection
+                .get_all_with_datastore(&datastore, &systemstore)
+                .await
+                .map_err(|e| {
+                    query::error::QueryError::execution(format!("storage error: {}", e))
+                })?;
 
-        let needs_migration_count = docs
-            .iter()
-            .filter(|doc| Self::doc_needs_migration(doc, target_version_id, has_migrations))
-            .count();
+            trace!(
+                collection = %collection_name,
+                doc_count = docs.len(),
+                "Fetched documents"
+            );
 
-        trace!(
-            collection = %collection_name,
-            doc_count = docs.len(),
-            needs_migration = needs_migration_count,
-            "Fetched documents"
-        );
-
-        let mut processed_docs = Vec::with_capacity(docs.len());
-        for doc in docs {
-            let processed = self
-                .process_document(
-                    doc,
-                    &collection,
-                    &datastore,
-                    &systemstore,
-                    has_migrations,
-                    &preloaded_history,
-                )
-                .await?;
-            processed_docs.push(processed);
+            let mut processed_docs = Vec::with_capacity(docs.len());
+            let mut write_backs = Vec::new();
+            for doc in docs {
+                let outcome = self
+                    .process_document(doc, &collection, has_migrations, &preloaded_history)
+                    .await?;
+                if let Some(source_document) = outcome.source_document {
+                    write_backs.push(MigrationWriteBack {
+                        source_document,
+                        migrated_document: outcome.document.clone(),
+                        migration_generation,
+                    });
+                }
+                processed_docs.push(outcome.document);
+            }
+            Ok((processed_docs, write_backs))
         }
+        .await;
 
         drop(datastore);
         drop(systemstore);
-        if has_migrations {
-            txn.commit().await
-        } else {
-            txn.discard()
-        }
-        .map_err(|e| {
+        txn.discard().map_err(|e| {
             query::error::QueryError::execution(format!(
-                "failed to finish lensed fetch transaction: {}",
+                "failed to discard lensed read transaction: {}",
                 e
             ))
         })?;
+        let (processed_docs, write_backs) = read_result?;
+        self.persist_migrated_documents(&collection, write_backs)
+            .await?;
 
-        if needs_migration_count > 0 {
+        if !processed_docs.is_empty() {
             debug!(
                 collection = %collection_name,
-                migrated = needs_migration_count,
-                "Documents migrated"
+                returned = processed_docs.len(),
+                "Lensed read completed"
             );
         }
 
@@ -115,9 +120,10 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
             .map_err(|e| query::error::QueryError::execution(format!("db error: {}", e)))?
             .ok_or_else(|| query::error::QueryError::collection_not_found(collection_name))?;
 
-        let (has_migrations, preloaded_history) = self.load_migration_context(&collection).await?;
+        let (migration_generation, has_migrations, preloaded_history) =
+            self.load_migration_context(&collection).await?;
 
-        let txn = self.db.new_txn(!has_migrations).await.map_err(|e| {
+        let txn = self.db.new_txn(true).await.map_err(|e| {
             query::error::QueryError::execution(format!("failed to create txn: {}", e))
         })?;
 
@@ -128,39 +134,44 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
             query::error::QueryError::execution(format!("failed to get systemstore: {}", e))
         })?;
 
-        let docs_with_status = collection
-            .get_all_with_datastore_include_deleted(&datastore, &systemstore, show_deleted)
-            .await
-            .map_err(|e| query::error::QueryError::execution(format!("storage error: {}", e)))?;
+        let read_result: query::error::Result<ProcessedDocsWithStatus> = async {
+            let docs_with_status = collection
+                .get_all_with_datastore_include_deleted(&datastore, &systemstore, show_deleted)
+                .await
+                .map_err(|e| {
+                    query::error::QueryError::execution(format!("storage error: {}", e))
+                })?;
 
-        let mut processed_docs = Vec::with_capacity(docs_with_status.len());
-        for (doc, is_deleted) in docs_with_status {
-            let processed = self
-                .process_document(
-                    doc,
-                    &collection,
-                    &datastore,
-                    &systemstore,
-                    has_migrations,
-                    &preloaded_history,
-                )
-                .await?;
-            processed_docs.push((processed, is_deleted));
+            let mut processed_docs = Vec::with_capacity(docs_with_status.len());
+            let mut write_backs = Vec::new();
+            for (doc, is_deleted) in docs_with_status {
+                let outcome = self
+                    .process_document(doc, &collection, has_migrations, &preloaded_history)
+                    .await?;
+                if let Some(source_document) = outcome.source_document {
+                    write_backs.push(MigrationWriteBack {
+                        source_document,
+                        migrated_document: outcome.document.clone(),
+                        migration_generation,
+                    });
+                }
+                processed_docs.push((outcome.document, is_deleted));
+            }
+            Ok((processed_docs, write_backs))
         }
+        .await;
 
         drop(datastore);
         drop(systemstore);
-        if has_migrations {
-            txn.commit().await
-        } else {
-            txn.discard()
-        }
-        .map_err(|e| {
+        txn.discard().map_err(|e| {
             query::error::QueryError::execution(format!(
-                "failed to finish lensed fetch transaction: {}",
+                "failed to discard lensed read transaction: {}",
                 e
             ))
         })?;
+        let (processed_docs, write_backs) = read_result?;
+        self.persist_migrated_documents(&collection, write_backs)
+            .await?;
 
         Ok(processed_docs)
     }
@@ -176,9 +187,10 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
             .map_err(|e| query::error::QueryError::execution(format!("db error: {}", e)))?
             .ok_or_else(|| query::error::QueryError::collection_not_found(collection_name))?;
 
-        let (has_migrations, preloaded_history) = self.load_migration_context(&collection).await?;
+        let (migration_generation, has_migrations, preloaded_history) =
+            self.load_migration_context(&collection).await?;
 
-        let txn = self.db.new_txn(!has_migrations).await.map_err(|e| {
+        let txn = self.db.new_txn(true).await.map_err(|e| {
             query::error::QueryError::execution(format!("failed to create txn: {}", e))
         })?;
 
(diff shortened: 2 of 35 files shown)


crates/db/src/lensed_auto_commit_fetcher/migration.rs
```diff
@@ -47,9 +48,12 @@ impl<S: Store> LensedAutoCommitFetcher<S> {
 
         let result = (has_migrations, if has_migrations { history } else { None });
 
-        // Store in cache using version-aware key
-        if let Ok(mut cache) = self.migration_cache.lock() {
-            cache.insert(cache_key, result.clone());
+        // A transform can be registered without changing the active version ID, so a negative
+        // result must not be cached. Positive entries remain version-scoped.
+        if has_migrations {
```
jackzampolin: **[bug]** Positive migration contexts (including the full targeted history graph) are cached for the life of `LensedAutoCommitFetcher` under `"{collection_id}:{active_version_id}"`. That fetcher is long-lived in the embedded query runner. Registering an additional transform without changing the active version ID (common when backfilling `v1→v2` after `v2→v3` is already active, or when adding a second hop on the same tip) leaves the cached history missing the new transform, so later reads can skip a required migration or apply an incomplete pipeline. The PR correctly stops caching negatives, but positives have the same invalidation problem (related to, but not solved by, follow-up #1233).

**Suggestion:** Invalidate or clear `migration_cache` from `set_migration` / collection-version mutations (preferred), or stop caching the history map and only cache a cheap “has any transform” probe with explicit invalidation. Add a regression test that registers two migrations without changing the active version and asserts the second transform is applied.

crates/db/src/lensed_auto_commit_fetcher/fetcher.rs
```diff
@@ -36,7 +36,7 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
             );
         }
 
-        let txn = self.db.new_txn(true).await.map_err(|e| {
+        let txn = self.db.new_txn(!has_migrations).await.map_err(|e| {
```
jackzampolin: **[bug]** When `has_migrations` is true, every auto-commit fetch path opens a **write** transaction (`new_txn(!has_migrations)`) and always `commit()`s it, even when zero documents need migration (all already at target, or all unknown/pass-through). On the default redb single-writer store this turns steady-state reads into writers and serializes concurrent queries for the entire lifetime after the first migration is registered. Unknown-version pass-through also never writes, yet still takes the write path. This is a major readiness risk for concurrent reads under anger and is introduced by the write-back design as implemented here.

**Suggestion:** Open a read txn first; only escalate to a write txn when at least one document actually needs transform write-back (or perform write-back in a short follow-up write txn after the read/migrate phase). Prefer discarding when no keys were mutated.

crates/db/src/migration/helpers.rs
```diff
@@ -70,15 +76,104 @@ pub fn json_to_native_value(
...
+    doc.set_schema_version_id(collection.version_id());
+    doc
+}
+
+/// Persist a lensed document directly to the datastore without creating CRDT commits.
+///
+/// Rust stores document fields in one CBOR blob, so this is the current-layout equivalent of
+/// Go's per-field `updateDataStore`: replace the blob and update the real version key in the same
+/// transaction.
+pub(crate) async fn cache_migrated_document(
```
jackzampolin, line 143: **[bug]** Lazy write-back via `cache_migrated_document` replaces the full CBOR document blob and version key but does **not** rebuild or patch secondary indexes. After `set_migration`, reindex materializes known docs once, but any later old-version document (P2P merge, delayed peers, restore) that is lazily migrated will have a migrated blob while index entries still reflect pre-migration values. Indexed filters/seeks can then miss or falsely hit. Go’s `updateDataStore` has a similar “cache fields only” shape and assumes indexes were reindexed at migration registration, so this is partly shared, but this PR’s title/claims of completing migration parity still leave a real post-reindex hole (follow-up #1235). Whole-blob overwrite also amplifies concurrent mutation loss versus Go’s per-field writes (follow-up #1234).

**Suggestion:** Until #1234/#1235 land, document this as an explicit unsupported window, or have lazy write-back update/remove+reinsert index entries for the affected doc in the same write txn. Prefer per-doc index maintenance over full collection rebuild on the hot path.

crates/db/src/lensed_auto_commit_fetcher/fetcher.rs
```diff
@@ -69,11 +67,32 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
...
-                .process_document(doc, &collection, has_migrations, &preloaded_history)
+                .process_document(
+                    doc,
+                    &collection,
+                    &datastore,
+                    &systemstore,
+                    has_migrations,
+                    &preloaded_history,
+                )
                 .await?;
```
jackzampolin: **[suggestion]** If `process_document` fails mid-batch (`?`), the function returns without explicitly `commit`/`discard`ing the write transaction. `DbTxn`/`BasicTxn` do not implement a clean RAII discard; storage backends rely on Drop as a safety net (redb warns and skips discard callbacks). Failed migrations can therefore leave messy txn teardown under load.

**Suggestion:** Structure the finish path with a local scope/guard that always discards on error and only commits on full success (same pattern as other implicit-txn DB methods).

crates/db/src/migration/helpers.rs
```diff
@@ -70,15 +76,104 @@ pub fn json_to_native_value(
...
+
+    if let Some(id) = original_doc.id() {
+        doc.set_id(id.clone());
+    }
+
+    for field in &collection.schema().fields {
+        if field.name == DOC_ID_FIELD {
+            continue;
+        }
+        if let Some(value) = lens_doc.remove(&field.name) {
```
jackzampolin, line 120: **[suggestion]** `lens_doc_to_document` only copies fields present in the lens output. Go’s `updateDataStore` treats fields present in the original but missing from the migrated map as explicit `nil` clears. For Rust CBOR storage, omitted fields are roughly equivalent to nulls on rewrite (nulls are skipped in `to_cbor`), but the in-memory `Document` returned to the current query has a missing field rather than `NormalValue::Null`, which can diverge for callers that distinguish absence vs null before re-read.

**Suggestion:** When a schema field existed on the original document (or always for active schema fields) and is absent from lens output, set `NormalValue::Null` (or explicitly remove and document the GraphQL null semantics). Align with Go’s clear-on-remove behavior.

crates/db/src/lensed_auto_commit_fetcher/migration.rs
```diff
@@ -285,10 +260,13 @@ impl<S: Store> LensedAutoCommitFetcher<S> {
         };
 
         if !history.contains_key(&doc_version) {
```
jackzampolin, line 277: **[suggestion]** Unknown/newer document versions pass through without restamping the stored version key. Go’s path still runs the doc through the lens pipe and `updateDataStore` always writes `targetVersionID`, so Go mislabels foreign versions as local. The Rust choice is intentional and safer for P2P rolling upgrades; query field values remain compatible. Residual behavior difference: Rust will re-enter the “needs migration” branch on every subsequent read of that doc (no restamp), which interacts badly with Issue 2’s always-write-txn policy.

**Suggestion:** Keep the intentional non-restamp, but ensure the fetch path uses a read-only txn for pure pass-through docs. Optionally memoize “known-foreign version” to avoid repeated history walks (with invalidation when local history gains that version).

crates/ffi/src/subscription/create.rs
```diff
@@ -98,6 +100,7 @@ pub(crate) fn message_to_json(message: &events::Message) -> String {
             "doc_id": update.doc_id,
             "cid": update.cid.to_string(),
             "collection_id": update.collection_id,
+            "block": base64::engine::general_purpose::STANDARD.encode(&update.block),
```
jackzampolin, line 107: **[nit]** FFI update events now include `block` as standard base64, while the HTTP events handler encodes `block` as hex. Consumers bridging FFI and HTTP will see incompatible encodings for the same payload. This is a useful completeness fix for FFI subscriptions, but the wire encoding is inconsistent across surfaces.

**Suggestion:** Document the FFI encoding explicitly in the manage.rs comment and/or align with one canonical encoding (or include an `encoding` field). Confirm Go harness expectations.

crates/db/src/migration/tests.rs
```diff
@@ -0,0 +1,260 @@
...
+        count(txn.headstore().unwrap()).await,
+        count(txn.peerstore().unwrap()).await,
+    );
+    txn.discard().unwrap();
+    counts
+}
+
+#[tokio::test]
+async fn lazy_lensed_read_writes_back_without_new_commits() {
+    let store = Arc::new(MemoryStore::new());
```
jackzampolin, line 392: **[suggestion]** New tests cover write-back without commit creation, restart persistence, identity materialize, array coercion, and unknown pass-through — good core coverage. Gaps for “ready under anger”: (a) second migration registered without active-version change must apply (Issue 1), (b) indexed query after lazy migration of a post-reindex old-version doc, (c) concurrent update during lazy write-back, (d) filter on a field whose value changes in the transform (pre- vs post-migration filter application), (e) FFI `materialize_collection` NAC identity path.

**Suggestion:** Add at least (a) and (b) before calling migration parity production-ready; track (c)/(d) with #1234/#1235.

jackzampolin (commented):
## Summary

This PR is a substantial and largely well-structured step toward Go lens-migration parity: real datastore write-back without commits, schema-aware scalar/array re-encoding, eager `materialize_collection` through DB/node/FFI, unknown-version pass-through, and FFI node reopen after background-task shutdown. The intentional foreign-version marker divergence vs Go’s `updateDataStore` restamp is real, documented, safer for rolling upgrades, and does not affect normal GraphQL field results (`_version` is commit history, not schema version). It is **not fully ready for production under concurrent load / P2P anger** yet: long-lived positive migration-history caching can go stale when additional transforms are registered without an active-version change; every lensed auto-commit read opens a write txn whenever any migration exists (including no-op paths), which serializes readers on single-writer backends like redb; and lazy write-back still neither updates secondary indexes nor is conflict-safe with concurrent mutations (acknowledged as #1234/#1235, but still correctness holes if this PR is treated as complete). Core write-back and materialize paths look correct for the single-threaded / post-`set_migration` reindex happy path covered by the new unit tests.

## Issue counts by severity

- bugs: 3
- suggestions: 5
- nits: 1

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[suggestion]** crates/db/src/lensed_fetcher/migration.rs:157 -- `LensedDocFetcher` caches history by `collection_id` only (no version key, no invalidation). It is usually per-transaction so the window is smaller than auto-commit, but long interactive txns that register a migration mid-flight (or reuse a fetcher) can still observe stale history.
  - **Suggestion:** Mirror whatever invalidation strategy is chosen for the auto-commit cache; at minimum key by active version and clear on migration registration visible inside the txn.

crates/db/src/lensed_auto_commit_fetcher/migration.rs
jackzampolin: Fixed in 62a40b53cc4bdb284ef52aaeb05bfe9f25b21ccc. migration_generation invalidates positive and negative migration contexts when the graph changes. Regression: migration_context_cache_invalidates_when_graph_changes_without_version_change.

crates/db/src/lensed_auto_commit_fetcher/fetcher.rs
jackzampolin: Fixed in 62a40b53cc4bdb284ef52aaeb05bfe9f25b21ccc. Fetches now use read-only transactions; only transformed documents enter a short separate persist path, and no-op/pass-through reads discard without writing. Regressions: implicit_query_flushes_lazy_migration_after_read_snapshot_closes and unknown_document_version_passes_through.

crates/db/src/migration/helpers.rs
jackzampolin: Fixed in 62a40b53cc4bdb284ef52aaeb05bfe9f25b21ccc, with direct dual-read coverage in 590a42188bab54be145df696ffa025c1f50c44da. Lazy write-back atomically maintains indexes and uses guarded fresh re-reads. Regressions: lazy_write_back_updates_secondary_indexes_for_late_document, concurrent_update_wins_over_stale_lazy_write_back, and concurrent_lazy_reads_of_same_document_are_idempotent. #1234 and #1235 are closed.

crates/db/src/lensed_auto_commit_fetcher/fetcher.rs
jackzampolin: Fixed in 62a40b53cc4bdb284ef52aaeb05bfe9f25b21ccc. The read transaction is explicitly discarded before persistence; persist_migrated_documents explicitly discards on errors and no-op batches and commits only after a complete successful batch. Regression path: implicit_query_flushes_lazy_migration_after_read_snapshot_closes.

crates/db/src/migration/helpers.rs
jackzampolin: Fixed in 62a40b53cc4bdb284ef52aaeb05bfe9f25b21ccc. Fields removed by a lens are NormalValue::Null in the current-query Document while remaining omitted from CBOR, preserving clear semantics across the query and re-read. Regression: lens_removed_field_is_returned_as_explicit_null.

crates/db/src/lensed_auto_commit_fetcher/migration.rs
jackzampolin: Intentional non-restamp retained. Foreign/newer versions pass through on the read-only path without write amplification, so their marker is not mislabeled as local. Regressions: unknown_document_version_passes_through in both lensed fetchers.

crates/ffi/src/subscription/create.rs
jackzampolin: Documented in 62a40b53cc4bdb284ef52aaeb05bfe9f25b21ccc next to message_to_json: FFI update blocks use standard padded base64 because that is the Go rust-FFI bridge contract; HTTP remains hex. Encoding unification is deferred as a non-blocking API change.

crates/db/src/migration/tests.rs
jackzampolin: Covered in 62a40b53cc4bdb284ef52aaeb05bfe9f25b21ccc and 590a42188bab54be145df696ffa025c1f50c44da. Regressions: migration_context_cache_invalidates_when_graph_changes_without_version_change, lazy_write_back_updates_secondary_indexes_for_late_document, concurrent_update_wins_over_stale_lazy_write_back, concurrent_lazy_reads_of_same_document_are_idempotent, and field_value_fetch_filters_after_lens_transform.

jackzampolin:
Review-body follow-up: the transaction-scoped LensedDocFetcher cache finding is also addressed in 62a40b53cc4bdb284ef52aaeb05bfe9f25b21ccc. DbTransactionRegistry::set_migration_in_txn explicitly clears that transaction context’s history cache, and a successful commit bumps the DB migration generation for long-lived auto-commit fetchers. The visibility/promotion paths remain covered by test_set_migration_in_txn_is_only_visible_inside_transaction and test_set_migration_in_txn_promotes_transform_on_commit.

crates/db/src/lens_utils.rs
```diff
@@ -206,3 +206,50 @@ pub fn build_collection_history(
...
+    target_version_id: &str,
+) -> bool {
+    if source_version_id == target_version_id {
+        return false;
+    }
+
+    let mut current = source_version_id;
+    let mut visited = std::collections::HashSet::new();
+    visited.insert(current.to_string());
+    let mut has_transform = false;
```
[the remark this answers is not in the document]
jackzampolin: No change. build_targeted_history intentionally implements the Go getTargetedCollectionHistory contract: TargetedHistoryLink is a target-rooted traversal tree with at most one previous/next edge, and Lens walks that same tree forward-first. Preserving every viable DAG edge would change the shared Go-parity API and pipeline semantics rather than fix this PR. Existing Go branching migration coverage passes.

crates/db/src/lens_utils.rs
```diff
@@ -206,3 +206,50 @@ pub fn build_collection_history(
...
+            .filter(|previous| !visited.contains(*previous))
+        {
+            has_transform |= link.transform.is_some();
+            current = previous;
+        } else {
+            return false;
+        }
+
+        visited.insert(current.to_string());
+    }
```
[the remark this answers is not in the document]
jackzampolin: No change. migration_path_has_transform only receives maps produced by build_collection_history, which roots traversal at the target; every included node is connected back to that target by its insertion edge. The caller also rejects source versions absent from the map before this helper. The fallback arms are unreachable for a valid targeted history, while unknown/foreign versions are deliberately passed through rather than restamped.

crates/db/src/lensed_auto_commit_fetcher/fetcher.rs
```diff
@@ -47,38 +53,56 @@ impl<S: Store + 'static> LensedAutoCommitFetcher<S> {
...
+                        source_document,
+                        migrated_document: outcome.document.clone(),
+                        migration_generation,
+                    });
+                }
+                processed_docs.push(outcome.document);
+            }
+            Ok((processed_docs, write_backs))
         }
+        .await;
```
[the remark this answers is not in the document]
jackzampolin: Valid scalability follow-up, not a parity correctness blocker. The read snapshot remains read-only and the guarded persistence path is correct, but candidate memory/write-transaction size should be bounded for very large first scans. Tracked with acceptance criteria in #1241.

crates/db/src/migration/reindex.rs
```diff
@@ -97,44 +119,16 @@ impl<S: Store> DB<S> {
...
-                None => return Ok(()),
-            }
-        };
+        let history = crate::lens_utils::build_collection_history(&versions, &target_version_id)
+            .ok_or_else(|| {
+                Error::Lens(format!(
+                    "failed to build migration history for collection '{}'",
+                    collection_name
+                ))
+            })?;
```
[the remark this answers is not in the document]
jackzampolin: No change; the strict error is intentional. The old silent Ok path could leave indexes stale while making a version switch appear successful. Reindex and eager materialization now surface an unbuildable local history, while unknown document versions inside a valid history are still skipped safely. This behavior was explicitly checked in the human review.

PR #1273, opened 2026-07-30
feat(cli)!: make postgres wire support opt-in

jackzampolin:
pg-compat (pgwire + sqlparser, ~1.3MiB of .text) is now behind a \`postgres\` cargo feature on cli, **off by default** per maintainer decision — no one uses the pg wire protocol from default builds; enable with \`--features postgres\`.

- Without the feature, \`--pg-address\` and \`api.pg_address\` simply do not exist (absent, not runtime-erroring). An existing config's \`pg_address\` entry is silently ignored — release-notes callout.
- All use sites cfg-gated; the 6-tuple \`init_store_and_server\` return became a \`ServerSetup\` struct (also dropped a \`clippy::type_complexity\` allow).
- cli was the only pg-compat consumer in the workspace.

Verified both ways: \`cargo test -p cli\` (default and \`--no-default-features --features redb\`) pass, clippy \`-D warnings\` clean on both, \`cargo tree -i pg-compat\` absent from the default graph.

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -41,7 +41,7 @@ keyring = { path = "../keyring" }
 orbis = { path = "../orbis" }
 sourcehub = { path = "../sourcehub" }
 defra-version = { path = "../defra-version" }
-pg-compat = { path = "../pg-compat" }
+pg-compat = { path = "../pg-compat", optional = true }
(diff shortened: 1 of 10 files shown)


jackzampolin:
Consolidated into #1275 (rollup carries these commits verbatim) to reduce CI amplification — branch retained.

PR #1261, opened 2026-07-30
fix(ffi): close major Go compatibility gaps

iverc:
Closes #1200, #1249, #1253, #1254, #1255, #1256, and #1257.

## Problem

The Rust FFI compatibility baseline had not been refreshed since the Go v1.0 pin. Go `develop` added and changed integration coverage across collection APIs, schema types, indexes, ACP, queries, and P2P behavior, while the paired-worktree harness also made complete sweeps difficult to reproduce outside one checkout layout.

## What changed

- Restore FFI parity for actions, collection deletion, transaction reads, request-scoped signing, identities, events, and subscriptions.
- Align scalar and array kinds, schema patches, view CIDs, introspection, collection validation, counters, encrypted historical reads, and document-ID index encoding with Go behavior.
- Correct ACP event ordering, anonymous bypass isolation, NAC re-enable authorization, and P2P disconnect authorization.
- Match Go explain scan prefixes and nested-limit execution metrics.
- Fix P2P peer reporting, one-way replication, document sync, replication events, task ownership, and shutdown so restart tests no longer cascade through a retained storage lock.
- Complete transactional collection operations and locking behavior.
- Restore index lifecycle, indexed-query, nullable datetime array, and mixed-ID deletion parity.
- Make `ffi-test` discover paired worktrees portably and fail on package-level harness errors.
- Add focused Rust regressions for the corrected behavior.

The companion Go adapter changes are in [sourcenetwork/defradb#5102](https://github.com/sourcenetwork/defradb/pull/5102) and [sourcenetwork/defradb#5107](https://github.com/sourcenetwork/defradb/pull/5107).

## Compatibility results

| Baseline | Leaf packages | Passed | Failed | Skipped | Total | Executed pass rate |
|---|---:|---:|---:|---:|---:|---:|
| Pinned `3de01484` | 89 | 2,723 | 48 | 174 | 2,945 | 98.27% |
| Go `develop` `13c46ec9` | 108 | 3,217 | 53 | 146 | 3,416 | 98.38% |

The recorded full sweep covers 471 more tests than the pin. The P2P shutdown fix removes the 11-test restart lock cascade, leaving the recorded networking result at 154 passed and 3 failed before the focused #1257 fixes.

This PR also resolves the implemented follow-ups:

- #1249: transactional collection-version operations and related locking cases
- #1253: NAC re-enable authorization
- #1254: explain output and execution metrics
- #1255: index lifecycle and indexed-query parity
[...]

PR #1281, opened 2026-08-02
fix(p2p): clear retry state when forgetting peers

### LINKED ISSUE #1269: fix(p2p): push-retry ledger has no removal API — deleted replicators keep being dialed forever
## Summary

Removing a replicator does not remove its retry state, and no API can. The result is that a dead peer is re-dialed every 2 seconds and re-joined to gossip on every boot, indefinitely.

Mechanics (v0.17.3):

- `DELETE /api/v0/p2p/replicators` deletes the persisted `/rep/id/{peer}` record only when the removal empties the collection set (`fully_deleted` gating `delete_persisted_replicator`, `crates/p2p-adapter/src/iroh.rs:640-648`), and touches **nothing** under `/rep/retry/*`.
- On boot, `restore_iroh_p2p_state` runs unconditionally (`crates/defra-node/src/p2p_runtime.rs:307`), and `spawn_iroh_retry_loop` (`p2p_runtime.rs:522-533`, the #1157 recovery) promotes every dormant watermark and then sweeps `get_all_retry_peers()` — keyed off `/rep/retry/id/*` **alone** — every `PERSISTED_RETRY_SWEEP_INTERVAL` = 2s (`crates/p2p/src/sync/mod.rs:67`), dialing the peer for each pending doc/commit **even if the `/rep/id/{peer}` replicator record is gone**.
- Each successful dial injects the connection, and gossip-heal joins the peer to every subscribed topic (`crates/p2p/src/iroh/endpoint.rs:331-345`), producing a permanent 60s-cadence `iroh_gossip::net: failed to send: connection task send loop terminated` WARN stream for a peer that is dead (`crates/p2p/src/iroh/gossip_heal.rs:70,316-333`).
- `Peerstore::clear_retry_peer` exists (`crates/storage/src/stores/peerstore.rs:701`) but deliberately preserves dormant watermarks, and nothing routes to it from the replicator-delete path. The only reliable cleanup today is stopping the node and deleting the `p/rep/retry/id/{peer}`, `p/rep/retry/doc/{peer}/`, `p/rep/retry/commit/{peer}/` prefixes from the raw store.

## Production evidence

An embedded-node deployment carried a replicator to a peer decommissioned months ago; its retry ledger survived a v0.14→v0.17 storage migration and still re-arms dials + gossip-heal warnings every boot (60+ WARN lines/hour).

## Ask

1. Replicator deletion (HTTP + `P2POperations::remove_replicator`) should also clear that peer's `/rep/retry/*` entries (doc, commit, and the retry-id watermark) when `fully_deleted`.
2. Provide an explicit "forget peer" operation (API + CLI) that clears retry state for a peer without needing a full replicator record — for exactly the orphaned-ledger case above.
3. The 2s retry sweep should skip (or age out) peers whose replicator record no longer exists, so an orphaned ledger cannot dial forever.

iverc:
Closes #1269.

## Problem

Removing the final collection from a replicator deleted its persisted configuration but left its document, collection-commit, and peer retry records behind. Retry sweeps were keyed from that orphaned ledger alone, so removed peers could be dialed again every two seconds and restored into gossip after each restart. There was also no supported API or CLI command for clearing retry state once the replicator record was already gone.

## What changed

- Delete a replicator configuration and all of its retry state atomically in one conflict-retried storage transaction.
- Add an explicit HTTP `Forget` operation and `p2p replicator forget <peerID>` CLI command for cleaning up orphaned retry ledgers.
- Restrict retry sweeps to peers that still have persisted replicator configurations.
- Apply full-removal semantics consistently with and without a sync coordinator for both P2P transports.
- Propagate persistence cleanup failures instead of reporting a successful removal while retry state remains.
- Add storage, HTTP routing, and CLI parsing regressions for deletion, orphan cleanup, sweep filtering, and the new command.

## Validation

- [x] `cargo +stable fmt --all -- --check`
- [x] `cargo clippy -p cli --features iroh --all-targets -- -D warnings`
- [x] `cargo check -p cli --features iroh`
- [x] `cargo test -p storage`
- [x] `cargo test -p cli`
- [x] `cargo test -p defra-p2p-adapter`
- [x] `cargo test -p embedded`
- [x] `git diff --check`


edjroz:
@coderabbitai review

edjroz (commented):
The fix seems alright to me some notes on the test coverage though:
- **No repro test for #1269.** Storage-level behavior is pinned, but the reported symptom (orphan re-dialed across restart) has no test. The four changed sweep call sites gain zero coverage.
- **Dormant records invisible to the assertions.** `get_retry_documents` filters to `pending == true`, so `is_empty()` would pass even if dormant watermarks survived deletion. The file's own stronger idiom is `activate_dormant_push_retries() == 0`.
- **Three new 400 branches untested.** `Forget` + `Collections`, missing `ID`, whitespace-only `ID`. Only the happy-path deserialization is covered.
- **Error propagation unpinned.** Swapping `tracing::warn!` for `?` on `delete_persisted_replicator` is a deliberate behavior change; reverting it would keep every test green.

jackzampolin (commented):
I reviewed the cleanup path and the overall direction looks right, but I do not think this is ready to merge on the current head.

This branch predates #1261 and overlaps the P2P paths changed there, so please merge/rebase current main and rerun the full matrix.

Before re-review, please also close out the unresolved retry/forget race thread (or document why a single already-in-flight send is the intended boundary), add a restart-level regression for the original redial behavior, and assert that dormant retries are actually gone (for example via `activate_dormant_push_retries() == 0`). The HTTP validation/error cases called out in the existing review should be pinned as well.

jackzampolin (approved):
The forget/retry race is now coordinated across all libp2p and iroh retry and failure-recording paths. Added storage race coverage, HTTP contract tests, and a restart E2E regression proving a forgotten peer is not redialed and queued data is not delivered. Focused storage/HTTP/E2E tests and embedded/defra-node compile checks pass locally.

jackzampolin:

PR #1282, opened 2026-08-03
perf(db): reserve document short IDs outside mutation transactions

### LINKED ISSUE #1264: perf(db): global /seq/doc short-ID allocation makes every pair of concurrent creates conflict
## Summary

Every document create performs a get-then-set on **one global key** — systemstore `s` + `/seq/doc` — to allocate the document short ID (`next_doc_short_id`, `crates/db/src/doc_id_map.rs:18-30`; key constant `crates/storage/src/keys/doc_id_index.rs:247-268`). Under the shared `ConflictTracker` (SSI: both write-write and read-write rules, `crates/storage/src/backends/shared.rs:416-460`), this means **any two concurrent transactions that each create at least one document conflict with each other with probability ~1**, regardless of collection, document, or field values — one of them aborts with `transaction conflict. Please retry`.

Call sites: `auto_commit_mutator/create.rs:80`, `batch.rs:207`, `create_many.rs:80` (per doc in a loop), `doc_mutator.rs:200`, and the P2P merge path `db-merge/src/merge_handler/composite.rs:421,817` — so a remote merge landing concurrently with a local create also conflicts on this key.

There is an existing regression test that (correctly, under current semantics) asserts this RMW conflicts: `crates/storage/src/backends/memory/tests.rs:149-180`.

## Production evidence

Coding-session store (embedded node, RocksDB, v0.17.3, ~3.2M docs): sustained ingest of ~200 batched creates/min from multiple concurrent writers produces a steady stream of client-visible `commit error: datastore error: storage error: transaction conflict. Please retry` failures (~0.2% of mutations, in bursts when writers overlap). Batches to *different* collections and *different* parent documents conflict identically — consistent with the global sequence key being the shared conflict point. We are about to add a second concurrent write workload (session labeling) and every labeling `create` will contend with every ingest `create`.

## Why this is the bottleneck to fix first

The v0.17.2 branchable-head commutation (#1211/#1226) and the content-addressed blockstore carve-out (#1194) removed other cross-document false-conflict sources, but `/seq/doc` still serializes *all* creates database-wide at the optimistic-concurrency layer. With N concurrent single-create writers, N-1 must retry; for batched writers the whole batch re-executes.

## Possible directions

1. **Allocate short IDs outside conflict tracking**: a dedicated in-process atomic allocator that persists a high-water mark (write-ahead reservation of a range, e.g. 1024 IDs at a time), so the per-create hot path never reads or writes a contended key. Crash losing a reserved range only wastes IDs — short IDs only need uniqueness, not density.
2. **Commutative carve-out**: treat `/seq/doc` like the branchable-head transition — both txns may bump it; on commit, replay the increment on top of the committed value instead of aborting (the value is opaque monotonic state, increments commute).
3. **Per-collection sequences** would only shrink the blast radius, not fix same-collection ingest; option 1 or 2 preferred.

Related: #658 (typed errors — the conflict that escapes is also stringified, see companion issue on the HTTP retry path).

iverc:
Closes #1264.

## Problem

Every document create and previously unknown merged document read and updated the same global `/seq/doc` key inside its mutation transaction. Under the shared conflict tracker, unrelated creates therefore conflicted across collections and workloads, forcing individual writes or entire batches to retry.

## What changed

- Add a database-owned allocator that reserves 1,024 document short IDs in a dedicated conflict-retried systemstore transaction and serves the reserved IDs from memory.
- Route auto-commit, batched, explicit-transaction, and P2P merge allocation through the shared allocator so mutation transactions no longer touch the global sequence key.
- Retain the transactional sequence helper only for startup migration work and preserve the existing persisted high-water format for compatibility.
- Allow crashes or restarts to skip unused reserved IDs while preserving uniqueness across database instances and restarts.
- Add regressions for multi-instance reservations, restart behavior, idempotent resolution, and concurrent unrelated caller transactions.

## Validation

- [x] `cargo test -p db -p db-merge`
- [x] `git diff --check`

### DIFF
--- a/crates/db/src/doc_id_map.rs
+++ b/crates/db/src/doc_id_map.rs
@@ -5,22 +5,115 @@
 //! holds the bidirectional mapping between short IDs and the public
 //! genesis-CID-derived DocIDs, plus a block-CID -> DocID ownership index.
 
+use async_lock::Mutex;
 use datastore::NamespaceView;
-use storage::corekv::Key;
+use std::sync::Arc;
+use storage::corekv::{Key, Store};
 use storage::keys::doc_id_index::{
     BlockCIDToDocIDKey, DocIDToDocRefKey, DocRef, DocShortIDSequenceKey, DocShortIDToDocIDAliasKey,
     DocShortIDToDocIDKey,
 };
+use storage::stores::Systemstore;
 
 use crate::error::{Error, Result};
 
-/// Allocate the next document short ID from the systemstore sequence.
-pub async fn next_doc_short_id(systemstore: &NamespaceView) -> Result<u64> {
-    let key = DocShortIDSequenceKey::new().bytes();
-    let current: u64 = match systemstore.get(&key).await.map_err(Error::Storage)? {
+const DOC_SHORT_ID_RESERVATION_SIZE: u64 = 1024;
+const MAX_RESERVATION_CONFLICT_RETRIES: u32 = 16;
+
+#[derive(Default)]
+struct ReservedRange {
+    next: u64,
+    remaining: u64,
+}
+
+/// Allocates document short IDs from persisted ranges.
+///
+/// The sequence stores the end of the latest reserved range. A crash can leave
+/// gaps, but every returned ID remains unique across restarts and DB instances.
+pub(crate) struct DocShortIdAllocator<S: Store> {
+    systemstore: Systemstore<S>,
+    range: Mutex<ReservedRange>,
+    reservation_size: u64,
+}
+
+impl<S: Store> DocShortIdAllocator<S> {
+    pub(crate) fn new(store: Arc<S>) -> Self {
+        Self::with_reservation_size(store, DOC_SHORT_ID_RESERVATION_SIZE)
+    }
+
+    fn with_reservation_size(store: Arc<S>, reservation_size: u64) -> Self {
+        assert!(reservation_size > 0, "reservation size must be non-zero");
+        Self {
+            systemstore: Systemstore::new(store),
+            range: Mutex::new(ReservedRange::default()),
+            reservation_size,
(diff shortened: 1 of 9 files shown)


crates/db/src/doc_id_map.rs
```diff
@@ -420,12 +505,65 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn resolve_or_allocate_is_idempotent() {
-        let store = systemstore().await;
-        let a = resolve_or_allocate_doc_short_id(&store, 3, "bae-y")
+    async fn allocator_reserves_ranges_across_instances_and_restarts() {
+        let store = Arc::new(MemoryStore::new());
+        let allocator_a = DocShortIdAllocator::with_reservation_size(store.clone(), 4);
+        let allocator_b = DocShortIdAllocator::with_reservation_size(store.clone(), 4);
```
edjroz, line 511: I think this is alright, however I think we should also have a test that spins up up to N nodes, to ensure the behavior doesn't persist.

Image

jackzampolin (commented):
The allocator design looks sound on review: reserving persisted ranges outside mutation transactions removes the global short-ID write conflict while preserving uniqueness, and gaps after abort/restart are an acceptable consequence of high-water allocation.

I am holding approval only because this branch predates #1261 and overlaps most of the same DB/runtime files. Please merge/rebase current main and rerun the full CI matrix. A multi-DB-instance stress case, as Ed suggested, would also make the cross-instance guarantee explicit.

jackzampolin (approved):
Updated onto current main. The added eight-instance concurrent allocation regression passes locally and makes the cross-instance uniqueness guarantee explicit. No remaining review blockers.

jackzampolin:

PR #1283, opened 2026-08-03
perf(index): shard full-text corpus statistics

### LINKED ISSUE #1265: perf(index): full-text _stats key is a collection-wide RMW on every create
## Summary

`FullTextIndex::save` performs `read_stats` → `write_stats` on a **single collection-wide key** — `d/{colShortID}/{ftIndexID}/_stats` — on every document create (`crates/storage/src/index/fulltext.rs:100-139, 314-329`). The postings entries themselves are per-document (`fulltext.rs:92-98`), but the stats key is shared, so under the SSI conflict tracker two concurrent creates into any collection with a `@fulltext` field conflict on both the write-write and read-write rules — same shape as the global `/seq/doc` problem (#1264), scoped to the collection.

In our deployment the two highest-write collections (`CodingMessage`: two `@fulltext` fields; `CodingSessionTag`: one) are exactly the ones written by concurrent ingest and labeling workloads, so this fires constantly on the hottest path.

## Impact

- Even if #1264 (global doc-ID sequence) is fixed, all concurrent creates *within* a fulltext collection still mutually conflict via `_stats` — the per-collection ceiling remains one create-transaction at a time.
- Each conflict aborts the entire (possibly large, multi-alias) batch after the full CRDT/block/index work has been done, then the client re-executes everything.

## Possible directions

1. **Commutative delta-merge**: stats are additive counters (doc count, token totals). Record a delta instead of read-modify-write, and fold deltas at commit/compaction — increments commute, so concurrent writers need not conflict (same philosophy as the #1226 branchable-head commutation).
2. **Move stats maintenance off the create transaction**: accumulate deltas via the commit-callback channel and fold them in a single background writer; BM25 scoring tolerates slightly stale corpus stats.
3. **Shard the stats key** (N sub-keys, sum on read) as a low-effort stopgap.

iverc:
Closes #1265.

## Problem

Every full-text document create read and rewrote the same collection-wide `_stats` key inside its mutation transaction. Concurrent writes to otherwise unrelated documents therefore conflicted after their posting and document work had already been performed, serializing writes to collections with full-text indexes.

## What changed

- Replace the single writable corpus statistics record with 256 deterministic shards selected from the document short ID.
- Store signed document-count and field-length deltas so creates, updates, and deletes remain exact without rewriting collection-wide state.
- Preserve the existing `_stats` value as an immutable compatibility base and combine it with shard deltas when calculating BM25 statistics.
- Reserve a binary metadata namespace that cannot collide with UTF-8 posting terms and keep storage bounded to at most 256 shard records per full-text index.
- Add regressions for concurrent creates, legacy statistics compatibility, and scoring after updates and deletes.

## Validation

- [x] `cargo test -p storage -p db-index -p db`
- [x] `cargo test -p storage index::tests`
- [x] `git diff --check`

### DIFF
--- a/crates/storage/src/index/fulltext.rs
+++ b/crates/storage/src/index/fulltext.rs
@@ -5,7 +5,8 @@
 //!
 //! Key layout:
 //!   Posting:  /[col_id]/[idx_id]/[term]/[doc_short_id] -> [term_freq, field_len]
-//!   Stats:    /[col_id]/[idx_id]/_stats                 -> [total_docs, total_field_len]
+//!   Legacy stats: /[col_id]/[idx_id]/_stats             -> [total_docs, total_field_len]
+//!   Stats shard:  /[col_id]/[idx_id]/0xff/s/[shard]     -> [doc_delta, field_len_delta]
 
 use async_trait::async_trait;
 use bm25::{DefaultTokenizer, Language, Tokenizer};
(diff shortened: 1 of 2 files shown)


jackzampolin (approved):
Reviewed the sharded stats keyspace, signed delta accounting, legacy compatibility, and mutation/scoring regressions on 36c84225. The implementation preserves existing stats while reducing the global write hotspot; the full CI matrix is green and I found no blocking issues.

PR #1286, opened 2026-08-04
feat(sourcehub): invalidate cached access from CometBFT events

### LINKED ISSUE #520: feat: Cosmos ACP cache invalidation via CometBFT WebSocket subscription
## Context

Phases 1 and 2 introduce caching of access decisions, but rely on TTL-based expiry. This creates a staleness window where revoked permissions may still be honored — a security concern, not just a performance issue.

This issue adds event-driven cache invalidation for the Cosmos/Go SourceHub path using CometBFT WebSocket subscriptions.

### Event Subscription

Subscribe to CometBFT WebSocket events on the Go SourceHub node:

```
ws://<comet_rpc_address>/websocket
Subscribe: tm.event='Tx' AND message.module='acp'
```

Filter for ACP-relevant transaction events:
- `MsgBearerPolicyCmd` (set_relationship, delete_relationship, archive_object)
- `MsgCreatePolicy`

### Filtering by Relevance

The node knows which policies are attached to its schemas. On receiving an ACP event:

1. Extract `policy_id` from event attributes
2. If `policy_id` is not in the node's active policy set → ignore
3. If relevant → invalidate affected cache entries

### What Gets Invalidated

| Event | Cache Invalidation |
|-------|-------------------|
| `set_relationship` | Invalidate access entries for `(policy_id, resource, object_id)` |
| `delete_relationship` | Same — this is the permission revocation case |
| `archive_object` | Remove all entries for `(policy_id, resource, object_id)` |
| `create_policy` | No invalidation needed (new policy, no cached entries) |

### Trade-offs (Cosmos Path)

- **Noisier**: CometBFT events include all ACP transactions network-wide, not just those relevant to this node. Filtering is client-side.
- **No proofs**: Events are trusted based on the WebSocket connection to a known validator. No cryptographic proof of event authenticity.
- **Available now**: This works with the existing Go SourceHub without modifications.

### Implementation

- New module: `crates/sourcehub/src/event_subscriber.rs`
- WebSocket client using `tokio-tungstenite` or similar
- Background task that receives events and calls into the cache invalidation API
- Configurable: `--sourcehub-events-ws` (WebSocket URL, optional — if not set, fall back to TTL-only)

## Depends on

- #517 (access cache — this invalidates it)
- #509 (configurable parameters)

## Related

- #521 (hub.rs native path — same cache, different invalidation mechanism)
- #522 (revocation SLA — the policy layer above both invalidation paths)

iverc:
Closes #520.

## Problem

SourceHub ACP access decisions remain cached until their TTL expires. Relationship changes submitted by another client bypass the local mutation path, so revoked permissions can remain effective during that staleness window.

## What changed

- Add an optional `--sourcehub-events-ws` / `DEFRA_SOURCE_HUB_EVENTS_WS` endpoint while preserving TTL-only behavior when unset.
- Subscribe to successful ACP transaction events over the CometBFT WebSocket and decode their transaction payloads.
- Invalidate exact object entries for relationship and object commands, whole-policy entries for policy edits, and the full cache for opaque or unknown ACP mutations.
- Clear cached decisions on malformed event payloads and reconnect with bounded exponential backoff after connection or subscription failures.
- Replace delimiter-based cache keys with typed components so invalidation matches policy, resource, and document identifiers exactly.
- Stop the background subscriber with its owning ACP instance and cover decoding, fail-safe handling, subscription requests, invalidation, and cancellation.

## Validation

- [x] `cargo test -p sourcehub`
- [x] `cargo test -p cli --no-default-features --features sourcehub,lark --all-targets`
- [x] `cargo clippy -p sourcehub --all-targets -- -D warnings`
- [x] `cargo clippy -p cli --no-default-features --features sourcehub,lark --all-targets -- -D warnings`
- [x] `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11422,15 +11422,18 @@ dependencies = [
  "crypto",
  "defra-core",
  "events",
+ "futures",
  "hex",
  "identity",
  "k256",
+ "prost 0.14.3",
  "reqwest 0.12.28",
  "serde",
  "serde_json",
  "sha2 0.10.9",
  "thiserror 2.0.18",
  "tokio",
+ "tokio-tungstenite 0.26.2",
  "tracing",
  "zanzibar",
 ]
@@ -12210,8 +12213,12 @@ checksum = "7a9daff607c6d2bf6c16fd681ccb7eecc83e4e2cdc1ca067ffaadfca5de7f084"
 dependencies = [
  "futures-util",
  "log",
+ "rustls 0.23.37",
+ "rustls-pki-types",
  "tokio",
+ "tokio-rustls 0.26.4",
  "tungstenite 0.26.2",
+ "webpki-roots 0.26.11",
 ]
 
 [[package]]
@@ -12627,6 +12634,8 @@ dependencies = [
  "httparse",
  "log",
  "rand 0.9.2",
+ "rustls 0.23.37",
+ "rustls-pki-types",
  "sha1 0.10.6",
  "thiserror 2.0.18",
  "utf-8",
@@ -13363,6 +13372,15 @@ name = "webpki-roots"
 version = "0.25.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f20c57d8d7db6d3b86154206ae5d8fba62dd39573114de97c2cb0578251f8e1"
+
+[[package]]
+name = "webpki-roots"
+version = "0.26.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "521bc38abb08001b01866da9f51eb7c5d647a19260e00054a8c7fd5f9e57f7a9"
+dependencies = [
+ "webpki-roots 1.0.6",
+]
 
 [[package]]
 name = "webpki-roots"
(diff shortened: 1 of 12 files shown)


jackzampolin (changes requested):
Reviewed the cache invalidation path on `fcbdb0fa`. Two correctness/security blockers remain:

1. Relationship mutations must invalidate the policy, not only the directly named object. Indirect Zanzibar-style relations can change access to other objects in the same policy, so object-only invalidation leaves stale grants alive until TTL expiry. Keep object-level invalidation for register/archive/unarchive, but use policy-level invalidation for set/delete relationship.
2. Clear the access cache after every successful WebSocket subscription and before consuming events. CometBFT does not replay events missed while disconnected; without a reconnect clear, a revocation during the gap remains stale until TTL expiry.

Please add focused regressions for both cases. These correspond to the two unresolved inline threads.

jackzampolin (approved):
Relationship mutations now invalidate the whole policy, subscription success clears cache state across event gaps, and the focused SourceHub Cosmos suite passes locally (24 tests). My requested changes are addressed.

jackzampolin:

Issue #523, opened 2026-02-24
security: Post-1.0 re-audit (target: late March 2026)


jackzampolin:
## Context

The pre-1.0 security audit (February 2026) is complete and archived in `past-audits/2026-02-pre-1.0-security-audit/`. All 1.0 blockers were resolved. 354 findings across 7 streams were triaged and remediated.

## Why Re-audit

Significant work is landing between the original audit and post-1.0 stabilization:

- **Iroh P2P transport** (#496, #500-505) — new transport layer with different trust model
- **SourceHub ACP caching** (#516 epic) — new caching layers with security-critical invalidation semantics
- **HubRsProvider** (#514) — new SourceHub backend with different wire protocol
- **Ongoing feature development** — query engine, schema, storage changes

A re-audit approximately one month after the original will catch regressions introduced by this work and validate the new subsystems.

## Scope

- All 7 original audit streams, focused on **changed code** since the original audit
- New subsystems added since February 2026 (iroh transport, ACP cache, HubRsProvider)
- Verification that original remediations are still intact (no regression from squash merges or refactors)

## Target Date

Late March 2026 (~4 weeks from original audit completion).

## Related

- Pre-1.0 audit archive: `past-audits/2026-02-pre-1.0-security-audit/`
- #516 — SourceHub ACP Performance epic (security-relevant caching work)

jackzampolin:
Closing per 2026-08-05 triage: the late-March target is long past and this tracking issue has gone stale. Security-relevant work has continued issue-by-issue (KMS peer-identity binding #1287, admission control #1088, ACP cache invalidation #520). If a formal re-run of the seven pre-1.0 security audit streams is scheduled, re-file with fresh scope against the current crate set.

PR #1287, opened 2026-08-04
fix(kms): bind key release to authenticated peer identities

iverc:
Targets #523.

## Problem

KMS pubsub requests carried a caller-controlled `Identity` field. The serving node used that field directly for DAC/NAC authorization, while the authenticated gossip source was retained only as metadata. A connected peer could therefore claim another DID and have an authorized DEK wrapped to its own ephemeral key.

Rust also sent a request-context user DID on cross-peer fetches, even though encrypted replication gates key access by the requesting node.

## What changed

- Resolve each inbound libp2p request through the existing signed peer-identity protocol and pass the authenticated DID across the KMS transport boundary.
- Authorize DEK release against the authenticated peer DID instead of the request payload, while retaining the payload field for Go wire compatibility and logging mismatches.
- Send the node DID on every outbound cross-peer fetch so Rust-to-Go requests preserve the node-level access gate.
- Keep transports without authenticated identity resolution anonymous, causing protected key release to fail closed.
- Add regressions for claimed-DID spoofing, outbound node identity, and resolver propagation into the KMS handler.

## Validation

- [x] `cargo test -p kms`
- [x] `cargo test -p p2p kms::pubsub_transport`
- [x] `cargo check -p cli -p embedded`
- [x] `cargo clippy -p kms -p p2p -p cli -p embedded --all-targets -- -D warnings`
- [x] `git diff --check`

### DIFF
--- a/crates/cli/src/commands/start/server.rs
+++ b/crates/cli/src/commands/start/server.rs
@@ -217,8 +217,8 @@ impl Node {
                 policy.set_node_acp(Arc::new(db::DbNodeAcpRead::new(adapter.nac_manager())));
             }
 
-            // Node identity for the wire `identity` fallback on gossip-initiated
-            // fetches. Anonymous nodes use a stable placeholder DID.
+            // Node identity used for cross-peer DEK requests. Anonymous nodes
+            // use a stable placeholder DID.
             let node_did = database.node_did().unwrap_or_else(|| {
                 identity::Did::new("did:key:z6MkhaXgBZDvotDkL5257faiztiGiC2QtKLGpbnnEGta2doK")
                     .expect("static anonymous DID parses")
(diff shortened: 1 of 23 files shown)


jackzampolin (commented):
The security model looks coherent on review: key release is bound to the authenticated transport peer, the wire identity is no longer trusted, and delegated replay is explicitly signed and constrained.

Because this branch predates #1261 and overlaps its merge/security wiring, please merge/rebase current main and rerun the full CI matrix before approval. This one is security-sensitive enough that I do not want to rely on the pre-#1261 green run.

jackzampolin (approved):
Updated cleanly onto current main. The authenticated-peer key-release design remains coherent after integration; no remaining review blockers.

jackzampolin:

Issue #699, opened 2026-04-10
Meta: Error handling and typed error propagation


jackzampolin:
## Summary

Track the migration from stringly-typed and ad hoc error handling toward typed error propagation across the Rust codebase.

This meta issue covers:
- replacing `String`-backed error plumbing with typed error variants where practical
- removing error-to-string conversions in `From` impls
- reducing repetitive `.map_err(|e| format!(...))` and `.to_string()` chains by using `From` and `?`
- preserving structured error context across crate boundaries
- leaving user-facing HTTP/FFI error messages stable where compatibility matters

This issue does **not** automatically cover every panic, unwrap, or shutdown-path bug. Concrete runtime/correctness issues should stay as separate issues when they have a clear behavioral impact.

## Primary work items

- Core error type redesign and typed wrapping
- Crate-boundary `From` conversions for `?` propagation
- Eliminate avoidable error stringification in internal APIs
- Keep compatibility at API boundaries while preserving internal structure

## Issues in this cluster

- #655 Redesign defra_core::Error to use typed variants instead of String
- #654 Eliminate 175+ repetitive .map_err() chains via proper error conversions
- #656 Fix error-to-string conversions in From impls
- #658 Replace string-matching for error types with structured variants
- #667 Replace Result<_, String> with typed errors in P2POperations trait

## Suggested decomposition

- Treat #655 as the foundational/core error model work
- Treat #658 and #667 as downstream consumers that should align with the core direction
- Fold purely derivative cleanup issues into this umbrella where they do not justify separate tracking

## Exit Criteria

- Internal error handling preserves structured context across major crate boundaries
- Repetitive string-conversion-based propagation is materially reduced
- String matching on error text is eliminated in core flows where typed variants are feasible
- Public API behavior remains compatible where required

jackzampolin:
Current state update:\n\nThis umbrella is still valid, but it is no longer purely aspirational. At least some downstream typed-error work has already landed, including the P2P typed-error refactor that completed issue #667. The remaining open work is now more clearly split between foundational core-error redesign (#655) and broader cleanup/reduction of string-based propagation across crate boundaries.\n\nRecommendation: keep this issue open as the umbrella, but update linked sub-issues/checklist items to distinguish completed downstream work from still-open foundational redesign and follow-on cleanup.

jackzampolin:
Current status update as of 2026-04-13:

This umbrella is still valid, but it is no longer untouched.

Progress already landed includes:
[...]

PR #1290, opened 2026-08-04
fix(db): preserve stale update conflict type

iverc:
Targets #699.

## Problem

Stale conditional updates detected a transaction conflict before commit but returned it as a generic query execution error. Although the message told clients to retry, the lost error type prevented the query boundary from emitting the stable `TXN_CONFLICT` code used by other transaction conflicts.

## What changed

- Return the existing typed `TransactionConflict` error for stale conditional updates.
- Strengthen the regression to verify the exact error variant and public message.

## Validation

- [x] `cargo test -p db`
- [x] `cargo test -p query transaction_conflict_response_has_stable_graphql_code`
- [x] `cargo clippy -p db -p query --all-targets -- -D warnings`
- [x] `git diff --check`

### DIFF
--- a/crates/db/src/auto_commit_mutator/update.rs
+++ b/crates/db/src/auto_commit_mutator/update.rs
@@ -116,7 +116,7 @@ impl<S: Store + 'static> AutoCommitMutator<S> {
                 || expected.is_deleted() != current_doc.is_deleted()
                 || !values_unchanged
             {
-                return Err(query::error::QueryError::execution(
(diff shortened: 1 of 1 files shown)


jackzampolin (approved):
Reviewed the typed conflict propagation and regression coverage on f69f2a08. The change preserves the existing public TXN_CONFLICT path, the full CI matrix is green, and I found no blocking issues.

PR #1291, opened 2026-08-04
chore(p2p): upgrade libp2p to 0.56

### LINKED ISSUE #1289: chore(deps): bump libp2p to 0.56 to clear two high gossipsub advisories
`Cargo.toml:93` pins `libp2p = { version = "0.53", ... }`, which resolves `libp2p-gossipsub 0.46.1`. That version is subject to two high-severity advisories:

| Advisory | Summary | Patched |
|---|---|---|
| [GHSA-xqmp-fxgv-xvq5](https://github.com/advisories/GHSA-xqmp-fxgv-xvq5) | Remote crash via unchecked `Instant` overflow in heartbeat backoff expiry handling | 0.49.3 |
| [GHSA-gc42-3jg7-rxr2](https://github.com/advisories/GHSA-gc42-3jg7-rxr2) | Gossipsub `PRUNE.backoff` `Duration` overflow | 0.49.4 |

Both are remote-crash paths in gossipsub's backoff handling, which is reachable on any node running the gossipsub behaviour — `crates/p2p/Cargo.toml:24` enables it.

I checked what a bump actually resolves rather than assuming: `libp2p 0.56` pulls `libp2p-gossipsub 0.49.5`, which is past both.

## What this does not fix, so you are not chasing it

`yamux 0.12.1` ([GHSA-vxx9-2994-q338](https://github.com/advisories/GHSA-vxx9-2994-q338), remote panic via a malformed Data frame, patched at 0.13.10) stays in the graph even at libp2p 0.56, and that is not something this repo can change. `libp2p-yamux 0.47.0` depends on **both** lines unconditionally:

```
libp2p-yamux 0.47.0
├── yamux 0.12.1
└── yamux 0.13.10
```

There is no feature to select between them, so the old line comes along regardless. That one belongs upstream in rust-libp2p; mentioning it only so it does not look like an omission from this report.

## Downstream context

We consume this repo as a git dependency. The chain is `embedded` → `native` (a default feature) → `defra-p2p-adapter/native` → `libp2p`, so gossipsub is in the default build for every embedding consumer, including those that never enable replication. Unlike `wasmtime-runtime` — which we could drop, and which removed `wasmtime` cleanly — `native` is the storage backend itself, so there is no way to opt out of this one downstream.

Version bump only as far as I can tell, no API surface involved on our side. Glad to open a PR if that is useful, though you will have a much better sense of the 0.53 → 0.56 migration across `p2p`, `p2p-adapter`, `db`, `ffi` and `cli` than I do.

iverc:
Closes #1289.

Depends on sourcenetwork/beetle#5.

## Problem

The workspace resolved `libp2p-gossipsub 0.46.1`, which is affected by two high-severity remote crash advisories in heartbeat backoff handling. The old Bitswap and `libp2p-stream` dependencies also kept P2P behaviours on the previous swarm release line, so upgrading the top-level libp2p dependency alone produced an incompatible graph.

## What changed

- Upgrade libp2p to 0.56, resolving `libp2p-gossipsub 0.49.5`, and move `libp2p-stream` to the matching swarm release line.
- Pin the companion Bitswap update that accepts the new outbound connection callback.
- Migrate Kademlia configuration, request-response events, slow-peer events, and infallible behaviour events to the current APIs.
- Preserve ephemeral source ports through per-dial options after removal of the TCP-level port reuse setting.
- Preserve the existing `InsufficientPeers` retry signal when gossipsub reports that no peers subscribe to a topic.
- Remove the obsolete `void` dependency and add a regression for publish retry compatibility.

## Validation

- [x] `cargo check -p defra-p2p-adapter -p embedded -p cli -p ffi --all-targets`
- [x] `cargo test -p p2p --features libp2p-transport`
- [x] `cargo tree -i libp2p-gossipsub --workspace` resolves only `libp2p-gossipsub 0.49.5`
- [x] `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -606,7 +606,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -617,7 +617,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -868,9 +868,9 @@ checksum = "71938f30533e4d95a6d17aa530939da3842c2ab6f4f84b9dae68447e4129f74a"
 
 [[package]]
 name = "asn1-rs"
-version = "0.6.2"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5493c3bedbacf7fd7382c6346bbd66687d12bbaad3a89a2d2c303ee6cf20b048"
+checksum = "b7f43a50ac4fdca5df8e885c21b835997f0a1cdee65494a6847694a98652d9d8"
 dependencies = [
  "asn1-rs-derive",
  "asn1-rs-impl",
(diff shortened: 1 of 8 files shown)


jackzampolin (commented):
This is currently blocked on three concrete items:

1. sourcenetwork/beetle#5 is still open, so the required companion dependency is not yet available through the intended merge path.
2. P2P Integration (Linux) fails in `rust_connection_manager_prunes_to_low_water_oldest_first`: the node did not prune to the low-water set, oldest-first, within 10 seconds. Because this PR changes libp2p connection behavior, that failure needs a focused reproduce/rerun and an explanation or fix before merge.
3. The branch predates #1261 and overlaps P2P code changed there; please merge/rebase current main and run the full matrix after the Beetle dependency is settled.

The other checks on the current head are green.

jackzampolin (approved):
Reviewed the libp2p 0.56 migration and validated the upstream Beetle dependency locally (52/52 iroh-bitswap tests passing); sourcenetwork/beetle#5 is now merged. The refreshed DefraDB CI run is exercising the updated stack.

jackzampolin:

Issue #1299, opened 2026-08-04
sync_documents returns Ok(()) whether it delivered every document, some, or none


edjroz:
## Summary

`sync_documents` returns `P2PResult<()>`. It counts merged documents to decide when to stop waiting, then discards that count and ends in an unconditional `Ok(())`. A caller cannot distinguish a complete sync from a partial one, or from one that delivered nothing at all.

## Context

**The site.** `crates/p2p-adapter/src/iroh.rs`, ending at `:934-935`. The function computes `total_expected = connected_peers.len() * doc_ids.len()` and accumulates `total_received` from `MergeComplete` events. That comparison is load-bearing: it is how the wait loop decides to exit early. It is then thrown away, and every path falls through to the same `Ok(())`.

**Three different outcomes, one return value.**

| Outcome | Returned |
|---|---|
| every requested document arrived | `Ok(())` |
| some arrived, the rest did not | `Ok(())` |
| nothing arrived | `Ok(())` |

There is no payload to inspect, so unlike a result-carrying API this cannot be worked around by a diligent caller. The information exists inside the function and is deliberately dropped.

**Observed.** While building a pull benchmark, a run returned `Ok` having delivered **zero of fifty** documents, with the caller given no indication. To be precise about the trigger: that particular run failed because the benchmark had not subscribed the receiver to the collection, which is a fault in the test setup and not in the transport. The return-value behaviour is the point, and it is independent of why the documents were missing.

**Also returns `Ok(())` with no peers.** An early return fires when `connected_peers.is_empty()`, so a sync against zero peers reports success. Go returns `ErrTimeoutDocSync` in the same situation. This overlaps with an existing draft from the #1261 review covering the zero-peer case specifically.

**Timing, for context rather than as the complaint.** The happy path is fast and correct: 605 ms for one document and 1.2 s for fifty, with every document arriving, measured on the memory backend. The failure path pins at about 9.2 s, which is three attempts at the 3 s `idle_timeout` under the 10 s overall cap (`iroh.rs:889-890`). Performance is not the issue here; the contract is.

**Suggested direction.** Return what the function already knows. A count of merged documents, or the set of document IDs that did not arrive, would let callers retry or surface the gap. If the signature must stay `()` for compatibility, an error on `total_received < total_expected` after the timeout would at least stop success being reported for a partial transfer.

[...]

PR #1301, opened 2026-08-04
bench: add block, CRDT, DAG merge, HTTP, and P2P benchmarks with a Go wire-format parity gate

### LINKED ISSUE #1300: No benchmark coverage for blocks, DAG merge, HTTP transport, or P2P
## Summary

The workspace has one benchmark: an LWW-only CRDT merge. There is no coverage for IPLD block encode/decode/CID, blockstore writes, DAG merge, HTTP transport, or P2P replication, so a regression on any of those paths is undetectable.

## Context

**What exists today.** `crates/crdt/benches/merge.rs` covers LWW only. Nothing else in the workspace has a `[[bench]]` target. There is no live tracking issue: #568 ("Phase 0") and its children #569 to #576 are all closed, and #568 explicitly deferred regression tracking.

**What is missing, and why each matters.**

| Area | Why it needs coverage |
|---|---|
| IPLD block encode / decode / CID | The wire format. A regression here changes document identity |
| Blockstore write | The write pipeline is per-field bound, so cost scales with schema width |
| DAG merge | Depth-sensitive; the merge path is where P2P work lands |
| HTTP transport | Measured at 98.5% of request overhead against a no-op executor |
| P2P replication and catch-up | No coverage at all, in either implementation |
| Document pull (`sync_documents`) | Never measured. Building it surfaced that the call returns `Ok(())` regardless of what arrived, filed as #1299 |
| Cross-runtime interop | Go to Rust replication has never been measured |

**Two things the existing benchmark got wrong**, both of which coverage would have caught earlier:

- `crates/crdt/benches/merge.rs` runs teardown inside the timed body, overstating LWW merge cost by roughly 2.4x. Every number it has produced is wrong in the same direction.
- Rust's signing and NAC `spawn_blocking` path adds 2.13x to per-request API overhead. It is inactive by default, so no existing measurement had ever exercised it.

**A wire-format gate is the highest-value piece.** Benchmarking block encode/decode makes it cheap to also assert byte-level parity with Go: decode Go-produced bytes, re-encode, compare bytes, compare the CID. That converts "we believe the formats match" into a test that fails when they stop matching.

**Prior work.** A branch implementing all of the above exists and is ready to propose: eight benchmark files plus Cargo wiring, rebased onto current `main`, compiling clean under `cargo check --benches --workspace --all-features`.

It includes **28 golden-vector assertions** across 7 block variants: Go-produced bytes decode in Rust, re-encode byte-identically, and match the Go-computed CID. That converts wire-format parity from something believed into something that fails a build when it stops being true. The gate runs before any timing, so the benchmarks refuse to report numbers if parity breaks.

Building these surfaced defects that no existing test covers, including #1299. Happy to open it as a PR against this issue.

edjroz:
Closes #1300.

Adds benchmark coverage for the layers that had none, and a wire-format parity gate that fails the build if Rust and Go stop agreeing on block encoding.

## What this adds

One commit per component, so this can be reviewed a piece at a time.

| Commit | Area | Lines |
|---|---|---:|
| `bench(defra-core)` | IPLD block encode / decode / CID, plus the Go golden vectors | 331 |
| `bench(crdt)` | CRDT merge, timed region corrected, coverage extended past LWW | 363 |
| `bench(db-blocks)` | Blockstore write pipeline | 168 |
| `bench(db-merge)` | DAG merge, parameterized by depth | 213 |
| `bench(http)` | Transport and middleware overhead | 262 |
| `bench(defra-node)` | P2P push and catch-up | 391 |
| `bench(integration-test)` | Cross-runtime Go/Rust interop | 253 |
| `bench(defra-node)` | Document pull (`sync_documents`) | 163 |

Before this, the workspace had one benchmark: an LWW-only CRDT merge. Nothing covered block encoding, blockstore writes, DAG merge, HTTP transport, or P2P, so a regression on any of those paths was undetectable.

## The parity gate is the part worth reviewing first

`crates/defra-core/benches/block.rs` carries **28 golden-vector assertions**: 7 block variants (`lww_simple`, `lww_high_priority`, `lww_deletion`, `counter`, `composite_active`, `composite_deleted`, `collection`) times 4 checks each.

For every variant, using bytes captured from the Go implementation:

1. Rust decodes Go's bytes
2. Rust re-encodes them **byte-identically**
3. Hashing Go's bytes yields Go's CID
4. Decode, re-encode and hash yields Go's CID

Check 4 is the one that matters, because CIDs are content addresses and therefore document identity. If the two implementations compute different CIDs for the same block, they disagree about what a document *is*, and replication between them cannot work.

`assert_go_parity()` runs **before any timing** and panics on mismatch, so the benchmark refuses to report numbers when parity is broken. A fast encoder that produces the wrong bytes is worthless, and this makes that state loud instead of silent.

## What building these found

- **#1299**, `sync_documents` returning `Ok(())` regardless of what was delivered. Found by writing the pull benchmark.
- The existing CRDT merge benchmark ran teardown **inside** the timed body, overstating LWW merge cost by ~2.4x. Every number it had produced was wrong in the same direction. Fixed here, and the contaminated version is deliberately retained alongside so the difference is provable: 654.8 ns against 266.7 ns clean, a 2.46x ratio.

## Verification

- `cargo check --benches --workspace` and `--all-features`: clean, no warnings
- `cargo check --tests -p defra-node --features p2p` and `-p integration-test`: clean. Two benchmarks sit behind `cfg(test)` or as `[[test]]` targets and are not covered by `--benches`, so they are checked separately
- `cargo bench --no-run -p defra-core --bench block`: links
- clippy `-D warnings` and `fmt --check`: clean
- All 28 golden vectors pass
- Rebased onto current `main`

## Known limitations

Stated up front rather than left to be discovered.

**No benchmark exercises `lark`.** `db-blocks/write`, `db-merge/merge_dag` and `crdt/merge` construct `MemoryStore` directly, and the P2P benchmarks build a node with no data path, which also falls to memory. Since `lark` is the default backend and `memory` is documented testing-only, none of the storage-touching numbers here should be used to characterise production. The block and transport benchmarks touch no storage at all and are unaffected. Worth fixing, and happy to do it in this PR if preferred.

**`cbor/generate_cid` is misnamed.** It calls `generate_cid()`, which encodes and *then* hashes, so it measures roughly encode plus hash (409 ns) rather than hashing alone (45 ns). Anyone comparing it against a hash-only figure would report a false regression of about 9x. Renaming it to `encode_and_cid` would be clearer.

**P2P benchmarks are `#[ignore]`d** and need real networked nodes; the interop one additionally needs a `defradb` binary on PATH and starts six node processes. They do not run in a normal `cargo test`.

**Sample counts are low** on the P2P and interop benchmarks. The harnesses are sound, the numbers are single-shot, and they should be repeated before anyone quotes them.


tools/integration-test/tests/p2p_interop_bench.rs
```diff
@@ -0,0 +1,252 @@
...
+        TestCluster::builder()
+            .go_nodes(2)
+            .with_p2p()
+            .build()
+            .await
+            .expect("build go/go cluster"),
+    )
+    .await;
+
+    let go_rust = measure_topology(
```
jackzampolin, line 322: The mixed-runtime direction is reversed here. The locked `TestClusterBuilder` appends Rust nodes before Go nodes, so with `.rust_nodes(1).go_nodes(1)`, node 0 is Rust and node 1 is Go. `establish_replication` and `measure` hardcode node 0 as the sender, but the report labels this result `go->rust`. That makes the reported result semantically wrong. Please either rename this case and its output to `rust_go` / `rust->go`, or parameterize sender and receiver and measure both cross-runtime directions.

crates/db-merge/benches/merge_dag.rs
```diff
@@ -0,0 +1,213 @@
...
+    group.sample_size(10);
+
+    for field_count in FIELD_COUNTS {
+        for depth in DEPTHS {
+            let dag = rt.block_on(build_dag(depth, field_count));
+            let label = format!("depth{depth}_fields{field_count}");
+            group.throughput(Throughput::Elements(depth as u64));
+
+            group.bench_function(BenchmarkId::from_parameter(&label), |b| {
+                b.iter_batched(
```
jackzampolin: Please use `iter_batched_ref` here. `handler` is moved into the async closure and dropped before the timed routine returns, so this measurement includes teardown of the handler, DB, and blockstore rather than isolating `handle_block`. This is the same teardown contamination already corrected in the other benchmark sites.

crates/defra-core/benches/block.rs
```diff
@@ -72,5 +226,162 @@ fn bench_block(c: &mut Criterion) {
...
+    for (name, go_bytes, go_cid) in GO_VECTORS {
+        let expected_cid = Cid::from_str(go_cid).unwrap();
+        group.throughput(Throughput::Bytes(go_bytes.len() as u64));
+
+        group.bench_function(BenchmarkId::new("decode_go", name), |b| {
+            b.iter(|| black_box(Block::from_dag_cbor(black_box(go_bytes)).unwrap()));
+        });
+
+        group.bench_function(BenchmarkId::new("reencode_go", name), |b| {
+            b.iter_batched(
```
jackzampolin: This has the same teardown issue: `iter_batched` passes the decoded `Block` by value, so the block and its owned allocations are dropped inside the timed routine after `to_dag_cbor()`. The emitted `Vec<u8>` is excluded, but the input destruction is not. Use `iter_batched_ref` so `reencode_go` measures re-encoding without block teardown.

crates/crdt/benches/merge.rs
```diff
@@ -117,5 +194,250 @@ fn bench_merge(c: &mut Criterion) {
...
+                );
+            },
+        );
+    }
+
+    group.finish();
+}
+
+/// Cost of the machinery the contaminated benches include but the clean ones do
+/// not. Subtracting these from `crdt/lww_merge_*` reconstructs the merge cost.
```
jackzampolin: These independent measurements cannot be subtracted to reconstruct the contaminated merge cost. For example, `txn_new_and_discard` includes `MemoryStore::new()`, transaction creation, and the custom executor even though that setup occurs before timing in `bench_merge_contaminated`; meanwhile, the contaminated path also times `lww.value()`, for which there is no matching baseline. Please remove the reconstruction claim and treat these as separate measurements, or build an exact A/B baseline with the same store lifetime, executor, and operation sequence.

jackzampolin (changes requested):
Thanks for putting this coverage together - the harnesses are thoughtful, particularly the explicit merge-outcome checks and P2P shutdown handling. I am requesting changes because the mixed-runtime direction is currently reported backwards and two benchmarks still include fixture destruction in their timed regions, which makes the resulting numbers misleading. The CRDT overhead suite also should not claim that its independent measurements can be subtracted to reconstruct merge cost.

One non-blocking framing note: the Go golden-vector checks already exist in `block_tests.rs`, while `assert_go_parity()` here only executes when the benchmark runs. It is useful as a benchmark preflight, but it is not itself a new normal-build gate. Sharing the vector corpus would also prevent the test and benchmark copies from drifting.

edjroz:
All three are fixed. Direction: sender and receiver are now explicit and all four topologies are measured,  so latency tracks the sender runtime rather than the crossing. For *Fixture destruction* both sites moved to `iter_batched_ref`. Swept the remaining `iter_batched` on the branch.
reconstruction claim is removed.

edjroz:
Deferring the corpus sharing: this PR is meant to be purely additive and that change modifies `block_tests.rs`. Filing it as a chore follow-up.

edjroz:
Addressed the convergence loop in `f318e749`: the deadline assertion sat after the success branch, so a receiver converging past `CONVERGENCE_TIMEOUT` broke out and recorded the elapsed time as a valid sample. It now checks the deadline first, so a late convergence fails rather than being measured.

Dismissing the other two:

`merge_dag.rs` `matches!` ownership: `matches!` expands to a match whose patterns bind nothing, so the scrutinee is not moved. If it were, both `"{outcome:?}"` and `black_box(outcome)` on the following lines would fail to compile, and `cargo check --benches --workspace --all-features` exits 0.

ast-grep SQL injection on `p2p_interop_bench.rs:150-154`: that is a GraphQL mutation string and `.query()` is the GraphQL client, not a SQL sink. No SQL is involved on this path.

crates/db-merge/benches/merge_dag.rs
```diff
@@ -0,0 +1,216 @@
...
+                                .handle_block(&dag.tip_cid, &dag.tip_bytes, metadata)
+                                .await
+                                .unwrap();
+                            // A skipped or rejected block does no walk, so timing it
+                            // would report a merge cost that was never paid.
+                            assert!(
+                                matches!(outcome, MergeOutcome::Merged),
+                                "tip was not merged: {outcome:?}"
+                            );
+                            black_box(outcome)
```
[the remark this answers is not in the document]
edjroz: > merge_dag.rs matches! ownership: matches! expands to a match whose patterns bind nothing, so the scrutinee is not moved. If it were, both "{outcome:?}" and black_box(outcome) on the following lines would fail to compile, and cargo check --benches --workspace --all-features exits 0.

As commented bellow I think this should be dismissed

Issue #1288, opened 2026-08-04
chore(embedded): make sourcehub optional, as the CLI already does


rainfall-rajesh:
`crates/embedded/Cargo.toml:29` declares SourceHub unconditionally:

```toml
sourcehub = { path = "../sourcehub" }
```

The CLI moved the same dependency behind a feature in 0423919 (`feat(cli)!: make sourcehub and orbis opt-in`), citing the cosmrs/tendermint/commonware/acp-light-client subtree. `embedded` did not follow, so any consumer that embeds DefraDB still compiles that whole subtree whether or not it uses on-chain ACP.

## Why it matters downstream

The subtree carries an advisory a downstream cannot reach:

```
embedded -> sourcehub -> acp-light-client (backbone.git)
         -> commonware-cryptography -> ark-ed-on-bls12-381-bandersnatch
         -> ark-r1cs-std -> ark-relations -> tracing-subscriber 0.2.25
```

`tracing-subscriber 0.2.25` is [GHSA-xwfj-jgwm-7wp5](https://github.com/advisories/GHSA-xwfj-jgwm-7wp5) / RUSTSEC-2025-0055 (low: ANSI escape sequences in logged user input), patched at 0.3.20. That is a major-version move for `ark-relations`, four levels above the consumer, so there is no override or feature toggle that clears it from outside this repo. It has to sit in an ignore file indefinitely.

We use `DocumentAcpConfig::Local` exclusively and never construct a SourceHub ACP, so we are compiling and shipping that subtree purely as a side effect of the default feature set.

## Why it looks straightforward

`sourcehub` appears in two files, 16 references:

- `src/node.rs` — the `sourcehub_acp: Option<Arc<sourcehub::SourceHubDocumentACP>>` field, `with_sourcehub()`, and passing it through the builder.
- `src/node_acp.rs` — one arm of the `match` on `DocumentAcpConfig`. The `Local` arm already returns `acp::ZanzibarDocumentACP` with `None` in the SourceHub slot, so the non-SourceHub path is fully formed today.

That suggests the CLI's pattern transfers directly:

```toml
sourcehub = { path = "../sourcehub", optional = true }

[features]
sourcehub = ["dep:sourcehub"]
```

with `#[cfg(feature = "sourcehub")]` on the field, the builder method and the `DocumentAcpConfig::SourceHub` arm.

[...]

PR #1303, opened 2026-08-04
chore(embedded): make SourceHub support opt-in

iverc:
Targets #1288.

## Problem

The `embedded` crate depended on SourceHub unconditionally, so consumers using only local document ACP still compiled the entire SourceHub dependency tree. This also prevented downstream embedders from excluding advisories in unreachable SourceHub dependencies.

SourceHub-enabled validation also exposed that embedded shutdown did not await its node-owned background task before closing the database, allowing that task to retain the final persistent-store handle briefly after shutdown returned.

## What changed

- Make the embedded `sourcehub` dependency optional and expose it through an opt-in `sourcehub` feature, matching the CLI feature policy.
- Gate the SourceHub ACP configuration, builder API, retained ACP handle, and construction path behind that feature.
- Keep local ACP behavior unchanged when SourceHub support is disabled.
- Await embedded background-task cancellation before closing the database so shutdown reliably releases persistent-store locks.

## Validation

- [x] `cargo test -p embedded`
- [x] `cargo test -p embedded --features sourcehub`
- [x] `cargo clippy -p embedded --all-targets -- -D warnings`
- [x] `cargo clippy -p embedded --features sourcehub --all-targets -- -D warnings`
- [x] `cargo clippy -p embedded --no-default-features --lib -- -D warnings`
- [x] `cargo tree -p embedded --features sourcehub -i sourcehub`
- [x] `git diff --check`

### DIFF
--- a/crates/embedded/Cargo.toml
+++ b/crates/embedded/Cargo.toml
@@ -26,7 +26,7 @@ p2p = { path = "../p2p", features = ["libp2p-transport"] }
 query = { path = "../query" }
 replication-filter = { path = "../replication-filter" }
 schema = { path = "../schema" }
-sourcehub = { path = "../sourcehub" }
+sourcehub = { path = "../sourcehub", optional = true }
 storage = { path = "../storage" }
 
 anyhow.workspace = true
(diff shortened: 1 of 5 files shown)


jackzampolin:
Pushed a633caee: the workspace lint/build failures were the ffi crate consuming embedded's SourceHub API (`SourceHubConfig`, `sourcehub_acp`, `DocumentAcpConfig::SourceHub`) after it was feature-gated. ffi already depends on `sourcehub` unconditionally, so it now opts into `embedded/sourcehub`. Verified locally: `cargo clippy --all -- -D warnings` and `cargo clippy -p ffi --all-targets -- -D warnings` both clean.

PR #1311, opened 2026-08-06
perf: Make `LIMIT n` cost `n`, not the collection

### LINKED ISSUE #1298: LIMIT is not pushed down for non-index scans
## Summary

#234 added limit pushdown for index scans. Limit-only queries against a non-indexed field still bypass the planner entirely and fetch the whole collection, so the cost of a `LIMIT 10` grows with collection size rather than with the limit.

## Context

**The path.** `crates/query/src/runner/query/select.rs:287` dispatches to `simple.rs:106`, which calls an unbounded `get_all`. The limit is applied after the fetch completes. Because the query never reaches the planner, none of the planner's pushdown machinery is consulted.

**Measurement.** On the default `lark` backend, `q_limit10` p50 at 16,000 documents is roughly 137x the Go reference implementation, and the ratio grows with collection size, 4.2x at 250 documents, 57.8x at 4,000, 137x at 16,000. A CPU profile confirms the mechanism directly: the limit query and a full scan produce the same sample distribution, because they are doing the same work.

**A note on an earlier number.** A figure of 248x circulated previously. That came from runs on the `memory` backend, which the project marks testing-only and which deep-clones the database per transaction. On the default backend the gap is 137x. The finding survived the correction; the magnitude did not, and 137x is the number to use.

**Sequencing note.** Open PR #1261 changes the `DocFetcher` trait signature and `ScanNode`'s contract. Work on this seam should land after it, and should also remove the post-fetch filter that PR introduces at `scan.rs:215-223`, which is O(docs x doc_ids).

---

## Measured scaling

Image 

Same query (`LIMIT 10`), same harness, against a collection that grows 64x.

| documents | Go | rust/lark | rust/redb | rust/memory |
|----------:|-----:|----------:|----------:|------------:|
| 250 | 0.395 | 0.920 | 1.090 | 1.632 |
| 1 000 | 0.375 | 2.911 | 3.567 | 6.883 |
| 4 000 | 0.417 | 12.241 | 15.325 | 26.999 |
| 16 000 | **0.391** | **53.437** | **63.778** | **121.093** |
| **growth, 250 -> 16k** | **0.99x** | **58x** | **58.5x** | **74x** |

p50 milliseconds.

A `LIMIT 10` returns ten documents whether the collection holds 250 or 16,000. Go's cost is flat, which is what pushdown looks like. Every Rust backend grows roughly in step with the data, which is what a full scan looks like.

**The backend is not the variable.** `lark`, `redb` and `memory` all grow 58-74x for 64x the data. The materialization happens above the storage layer, so no storage-level change can fix it. A CPU profile confirms this independently: the limit query and a full scan produce the same sample distribution.

### Notes on these numbers

- `lark` is the default backend and the headline. `memory` is documented testing-only and is included here to show the pattern is backend-independent, not to characterize performance.
- The Go column is Go's in-memory store, i.e. Go's most favourable configuration. A slower Go baseline narrows the ratio at 16,000 from ~137x to ~111x but does not change the shape, which is the finding.
- An earlier version of this work quoted 248x. That figure came from comparing against Rust's memory backend and is superseded. ~137x on `lark` is the number to use.

edjroz:
Fixes #1298

## What was wrong

`LIMIT 10` against a 4000-document collection read all 4000 documents. The plan
already stopped pulling at 10 (`LimitNode::next()`), but two layers below it had
already materialized everything:

- `ScanNode::init()` slurped the whole collection into a `Vec` before the first
  `next()` call.
- The redb, fjall, rocksdb and memory storage backends each built a full snapshot
  of the scanned range up front, so even a lazy scan paid for the entire prefix.

## Re-framing of the issue

The issue asks for limit *pushdown*. Go does not do that. `Fetcher` takes no
limit parameter, there is no capability negotiation and no routing gate anywhere
in the Go fetcher stack. Go is fast here because its pull chain is lazy end to
end: `limitNode` stops calling `Next()` and nothing below it reads further.

Rust's plan was already pull-based and already stopped at the limit. So this PR
makes the two materialization points lazy rather than introducing a pushdown
mechanism Go has no counterpart for. No routing gate, no capability negotiation.

## The change

**Query layer**

- New `DocStream` pull-source on `DocFetcher`. `DbDocFetcher`, both lensed
  fetchers, `AutoCommitFetcher` and `FetcherWrapper` all stream.
- `ScanNode::init()` opens a stream; `next()` pulls one document per call.
- `execute_simple_select` passes `ScanSource::Fetcher` to `build_plan` instead of
  a materialized `Vec`. The docID and index-scan paths still pass `Docs` — they
  are already bounded and gain nothing from streaming.

**Storage layer**

- Shared `ChunkedSnapshot` (`crates/storage/src/chunked.rs`): 256-pair windows,
  resuming from `last_key + 0x00`. Wired into redb, fjall, rocksdb and memory.
- The `MergingIterator` that merges a snapshot against a transaction's pending
  writes now lives once in `crates/storage/src/merging.rs`. It had been copied
  byte-identically into four backends; sharing it is a net −679 lines.

`lark` is deliberately untouched. It was already lazy, and its design is what
the other four backends now approximate.

## Results

`q_limit10`, p50 ms, median of 3 runs, sweeping 250 → 4000 documents:

| backend | before | after | growth |
|---|---|---|---|
| lark | 0.882 → 11.848 | 0.260 → 0.282 | **1.1×** |
| redb | 1.007 → 14.331 | 0.321 → 0.331 | **1.0×** |
| fjall | (no baseline) | 0.316 → 0.327 | **1.0×** |

Growth was 13–14× before and is flat now. Go's reference is ~0.4–0.48 ms flat,
so all three backends now sit below Go's absolute latency with equivalent
flatness.

Control queries on the same final build stayed linear — `q_all` 13.2–13.7×,
`q_filter` 12.5–13.2×. That is the evidence the win is targeted rather than a
measurement artifact: an unlimited query must still read everything, and a
filter with no limit must still examine every document.

Chunk size 256 is a stated default, not a tuned one.

## Things a reviewer should know

**The defaulted-trait-method trap.** `DocFetcher::stream_all_with_deleted`
originally had a default body that fell back to the eager path. Three
implementations silently inherited it — `FetcherWrapper`, both lensed fetchers
and `AutoCommitFetcher` — which made the entire optimization a no-op while
passing every correctness test. It happened three separate times. The method is
now **required**, with no default body.

**Reverse scans stay eager**, in all four backends. A chunked forward read cannot
be reversed after the fact. `b2cd18d1` does remove a double-hold of the result
set on that path — `ChunkedSnapshot::from_window` takes ownership of the window
instead of cloning it, which also drops a re-clone on every `reset()`.

**Lens write-backs.** The streaming path defers write-back per document; the
eager paths keep their full-scan write-back. Both coexist and are
order-independent. Note that Rust's `defer_full_scan_write_back` has no Go
counterpart — Go's write-back is per-document, inline and opportunistic
(`internal/lens/fetcher.go:296-300` describes it as a read-through cache).

## Testing

- New `tools/integration-test/tests/query/limit_pushdown.rs`.
- `e491876e` counts real storage reads for a limited query, so a regression to
  eager behaviour fails a test rather than only showing up in a benchmark.
- `b7b19346` covers pending writes landing exactly on a 256-key chunk boundary.
  It went into the shared backend suite rather than four per-backend copies, so
  it runs against all five backends including `lark`. Mutation-checked for
  non-vacuity: removing the `Equal`-branch snapshot skip fails exactly the four
  chunked backends and leaves `lark` green.
- Full `cargo test --all --all-features` run. Remaining failures are
  environmental, or verified failing identically on `origin/main` — including `p2p_backlog` and the
  two `p2p` restart tests, which were suspected of being caused by the redb
  chunked closure and then confirmed pre-existing.


crates/storage/src/backends/memory/transaction.rs
```diff
@@ -129,22 +135,74 @@ impl Reader for MemoryTxn {
...
+                let lower_bound = match &after {
+                    Some(k) => {
+                        let mut succ = k.clone();
+                        succ.push(0);
+                        Bound::Included(succ)
+                    }
+                    None => start_bound_for_chunk.clone(),
+                };
+                let items: Vec<(Vec<u8>, Vec<u8>)> = snapshot
+                    .range((lower_bound, end_bound_for_chunk.clone()))
```
jackzampolin, line 177: **[P2] Return an empty iterator for disjoint bounds instead of panicking.** `compute_range_bounds` can produce a lower bound greater than the upper bound (for example `start = b"z"`, `end = b"a"`, or a start beyond the requested prefix). `BTreeMap::range` panics for a non-empty map in that case, whereas the previous `MemoryIterator` predicate returned an empty scan. Please guard the range intersection before the snapshot and pending `range` calls and add forward/reverse regression coverage.

crates/db/src/limit_pushdown_tests.rs
```diff
@@ -0,0 +1,240 @@
...
+/// Counts documents as the query pulls them, one per `next`.
+struct CountingStream {
+    inner: Box<dyn DocStream>,
+    documents_read: Arc<AtomicUsize>,
+}
+
+#[async_trait]
+impl DocStream for CountingStream {
+    async fn next(&mut self) -> Result<Option<(Document, bool)>> {
+        let pulled = self.inner.next().await?;
```
jackzampolin: **[P2] This counter still observes yielded documents, not storage reads.** If `stream_all_with_deleted` eagerly materializes all 300 documents before returning the stream (or during its first `next`), this wrapper increments only for the 10 values pulled by `LimitNode`, so both tests still pass. Instrument the storage iterator/refill path instead. In particular, `MemoryStore::new_txn` itself clones the full `BTreeMap`, so this assertion cannot demonstrate that underlying work is bounded by the limit.

crates/db/src/auto_commit_fetcher.rs
```diff
@@ -352,3 +406,66 @@ impl<S: Store + 'static> DocFetcher for AutoCommitFetcher<S> {
...
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl<S: Store + 'static> DocStream for AutoCommitDocStream<S> {
+    async fn next(&mut self) -> query::error::Result<Option<(Document, bool)>> {
+        let pulled = match self.inner.as_mut() {
+            Some(inner) => inner.next().await?,
+            None => None,
+        };
+        if pulled.is_none() {
+            self.release_read_txn();
```
jackzampolin: **[P2] Close the inner stream before clearing it.** On the normal `None` path, `release_read_txn()` sets `inner = None`, so the later `ScanNode::close` cannot reach `CollectionDocStream::close`; `LensedAutoCommitDocStream::close` has the same release-without-close behavior. The storage iterator contract requires explicit closure. Please await `inner.close()` on exhaustion and explicit close before discarding the transaction, while still releasing the transaction if closing fails.

jackzampolin (changes requested):
Requesting changes for three issues: a memory-backend panic on disjoint iterator bounds, regression coverage that still does not observe actual storage reads, and auto-commit stream cleanup that drops the inner iterator before closing it.

edjroz:
All findings addressed:

**1. Disjoint bounds**

- The bounds helper now returns an `Option`; `None` means no key can match, so a call site can't forget the empty case. All five backends short-circuit to a shared empty iterator, restoring the silent empty scan the memory backend had before this PR.
- Four new tests in the shared backend suite — forward, reverse, start-beyond-prefix, and with pending writes — so they run against all five backends.

**2. The counter didn't observe storage reads**

- Added a counting `Store`/`Txn`/`Iterator` decorator, modelled on the existing encrypted-store decorator, and rebuilt both assertions on it.
- Fixture grown 300 → 2000 documents, since the chunk size is 256 and 256-against-300 isn't a discrimination.
- A `LIMIT 10` query pulls **11 keys** and **23 point lookups**. Making the stream eager takes it to **2001** and fails both tests — the discrimination the old counter structurally couldn't provide.
- Runs against memory, redb, rocksdb and fjall.

**3. Close before clearing**

- Both streams now close the inner stream before releasing the transaction. The lensed one was worse than reported: its explicit close never closed inner either, so both routes leaked.
- Close errors propagate, keeping parity with Go.
- The transaction is released whether or not the close succeeds, so a failed close can't leak it.
- The lensed close flushes write-backs even when the inner close failed, then surfaces both errors joined; returning early would have discarded already-computed migration work.

crates/db/src/lensed_auto_commit_fetcher/stream/tests.rs
```diff
@@ -0,0 +1,312 @@
...
+#[tokio::test]
+async fn close_joins_inner_close_error_with_persist_error() {
+    let transform_store = Arc::new(AlwaysFailingTransformStore::default());
+    let mut raw_db = DB::new(MemoryStore::new()).unwrap();
+    raw_db.lens_store = transform_store;
+    let db = Arc::new(raw_db);
+
+    let mut migratable_schema = test_schema();
+    migratable_schema.is_materialized = true;
+    db.create_collection(migratable_schema).await.unwrap();
```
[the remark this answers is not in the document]
edjroz: Declined — this follows the existing precedent at `migration/tests.rs:400-402`, there is no `DB` constructor that accepts a `lens_store`, and the shared-store hazard does not apply here because the test holds no second reference to the store.

edjroz:
Two nitpicks declined: the shared `RecordingStream` helper (~15 lines across two disjoint private test trees — worth promoting once a third stream needs it, not before), and extracting the seeding block (that test's coupling to db internals is deliberate, and hiding it in a helper makes it easier to break unknowingly). Everything else from the review is addressed in `b8941a1c`.

PR #1313, opened 2026-08-06
fix(crypto): derive block encryption from the DAG instead of a process-global store

### LINKED ISSUE #1292: Encryption config store evicts by clearing every entry
## Summary

When the document encryption store reaches its capacity limit, eviction is implemented as `store.clear()`. Crossing the threshold discards the encryption configuration of every document held, rather than evicting a single entry.

## Context

**The site.** `crates/defra-core/src/encryption.rs:192-199`. The capacity check is followed by a full clear rather than an eviction of the oldest or least-recently-used entry, so the structure behaves as a bounded cache that empties itself instead of one that ages entries out.

**Why it is easy to miss.** The threshold is high enough (~10,000 entries) that no test or short-lived run will reach it. It requires a long-running node handling a large number of distinct encrypted documents in a single process, which is exactly the shape of a production deployment and not the shape of a test.

**Suggested direction.** Replace the clear with a real eviction policy, or if entries are cheap to rebuild, make the miss path explicit and document the cache as best-effort.

edjroz:
Closes #1292.

## Problem

The document encryption store treated "at capacity" as "discard everything": at 10,000 entries
`store_doc_encryption` called `store.clear()`. Reproduced before touching anything, crossing the
threshold left 0 of 9,999 entries alive.

The stored `EncryptionConfig` was what told a later update to keep encrypting a document created
encrypted. A lost entry raised no error: the update path fell back to `None`, and `None` is
indistinguishable from "encryption was never requested", so the delta was written in plaintext. One
insert past the threshold silently turned encryption off for every encrypted document the process
had seen. Process restart had the same effect, since the store lived only in memory.

Go carries no such cache. `determineBlockEncryption` (`internal/core/block/store.go:144-198`) reads
the document's head blocks on every write and inherits the previous block's encryption when the
request supplies none. The Rust store stood in for a mechanism that was simply never ported, and
every input it needed was already available: head CIDs from `DocHeadsSnapshot`, and
`Block.encryption: Option<Cid>`. Bounding the eviction would have reduced the blast radius from
~10,000 documents to one per eviction while leaving the silent-plaintext hazard in place.

Note: the issue cites lines 192-199; the defect had drifted to 199-205.

## What changed

- Add `inherited_encryption_cid` / `inherited_encryption` in `crates/db-blocks/src/write.rs`, wired
  into both the per-field and composite encryption sites. When a write carries no explicit config,
  the field's heads are consulted and the previous block's key and `Encryption` link are reused
  as-is, with no new encryption block written, matching `store.go:192-195`.
- Delete the process-global store outright: the static, the accessor, `MAX_DOC_ENCRYPTION_ENTRIES`,
  `store_doc_encryption`, `get_doc_encryption` and `clear_doc_encryption_store`, plus all seven call
  sites in `crates/db`.
- Add `NamespaceView::sibling` so the encryption-block lookup checks the encstore before the
  blockstore, matching every other read path in the codebase
  (`db-merge/src/merge_handler/encryption.rs`, `versioned_fetcher.rs`, `kms_adapters.rs`).
- Derive the inherited link in `write_delete_block` too, so a delete tombstone on an encrypted
  document carries the same `Encryption` link as the composite it supersedes. Go's delete path runs
  through the same `addDelta` as every other write (`internal/db/document_delete.go:181-195`).
- Document the nonce-uniqueness invariant that the inherited-key model now depends on, at both the
  module and function level.
- Add eight unit tests in `crates/db-blocks/src/write_encryption_tests.rs`, the first coverage this
  behavior has had.

The comment in `update.rs` claiming to "match Go's behavior where encryption propagates through the
DAG" is now actually true.

## Behavior changes worth reviewing

**Key lifecycle.** Previously every encrypted write minted a fresh random key and wrote a new
`Encryption` block, so a field updated N times had N keys and N blocks. It now has one key and one
block referenced N times, and keys rotate only when a request explicitly supplies a config. This is
Go's model and the point of the port, not a side effect.

Reusing a key across a field's history is sound because every delta is sealed with a fresh random
96-bit nonce (`crates/crypto/src/encryption/aes.rs`, `nonce.rs`); the birthday bound of roughly 2^32
messages per key is unreachable for one field. That uniqueness is now load-bearing in a way it was
not under fresh-key-per-write, hence the added documentation: a deterministic or counter-based nonce
would repeat a (key, nonce) pair across updates.

**A missing head block is now an error, not a miss.** Guessing "not encrypted" is precisely the
silent downgrade this change exists to remove, and Go returns `ErrCouldNotFindBlock` in the same
spot.

**Known gap, deliberately matching Go.** A field first written by an update has no heads of its own,
so it inherits nothing and is stored in plaintext even on a document created with
`encrypt_doc: true`. Go behaves identically: `addDelta` builds its head set from the field's own
prefix (`store.go:82-86`), so `determineBlockEncryption` sees no heads. This chooses parity over a
unilateral divergence and pins the behavior with a characterization test,
`field_added_by_update_is_written_plaintext_matching_go`. It is being disclosed upstream to the Go
team rather than silently patched on one side.

## Validation

- [x] `cargo clippy --all --all-targets -- -D warnings`
- [x] `cargo test -p db-blocks -p defra-core -p db` (18 / 85 / 165)
- [x] `cargo test -p integration-test --test basic` (34)
- [x] `cargo test -p integration-test --test encryption` (12 passing, 4 known environmental)
- [x] `cargo test -p integration-test --test acp` (198, 16 ignored)
- [x] `cargo test -p integration-test --test p2p` (98, 11 ignored)
- [x] `git diff --check`

All boxes were run on this branch's tree and re-run after rebasing onto `e419da0f`.

Two suites cannot be verified locally. The 4 encryption cross-runtime tests and the Go-compat FFI
suite both need a Go binary at a hardcoded path that is absent on this machine; confirmed
pre-existing by reproducing the identical 4 failures on a clean tree. **CI must confirm the FFI
compat suite.** It is expected to pass unchanged: in deterministic mode the key is a pure function of
(docRef, field), so inheriting and regenerating produce identical bytes, which is also why this
divergence went unnoticed.

### DIFF
--- a/crates/db-blocks/src/write.rs
+++ b/crates/db-blocks/src/write.rs
@@ -1,5 +1,88 @@
 use super::*;
 
+/// The `Encryption` link a new block should carry forward, read off the
+/// previous block, when the write brings no config of its own.
+///
+/// Mirrors the fallback half of Go's `determineBlockEncryption`
+/// (internal/core/block/store.go): the DAG, not any process-local state, is
+/// what records that a document is encrypted, and the existing link is reused
+/// so no new encryption block is written.
+///
+/// A head whose block cannot be read is an error rather than a miss: guessing
+/// "not encrypted" is exactly the silent downgrade this derivation exists to
+/// prevent. Matches Go, which returns `ErrCouldNotFindBlock` here.
+///
+/// Heads with no encryption are skipped rather than ending the search, but an
+/// empty `heads` yields `None` — so a field first written by an update inherits
+/// nothing and is stored in plaintext. That is Go's behaviour too (its head set
+/// is per-field, `store.go:82-86`); see
+/// `field_added_by_update_is_written_plaintext_matching_go`.
+async fn inherited_encryption_cid(
+    blockstore: &NamespaceView,
+    heads: &[Cid],
+) -> Result<Option<Cid>, String> {
+    for head in heads {
+        let block_bytes = blockstore
+            .get(&head.to_bytes())
+            .await
+            .map_err(|e| format!("Failed to read previous block {}: {}", head, e))?
+            .ok_or_else(|| format!("Previous block {} not found", head))?;
+        let prev = Block::from_dag_cbor(&block_bytes)
+            .map_err(|e| format!("Failed to decode previous block {}: {}", head, e))?;
+
+        if prev.encryption.is_some() {
+            return Ok(prev.encryption);
+        }
+    }
+
+    Ok(None)
+}
+
+/// The inherited encryption link together with the key it points at, for the
+/// field deltas that actually need to encrypt with it.
+///
+/// Reusing one key across a field's whole history is safe only because every
+/// delta is sealed with a fresh random nonce (see `encrypt_delta`). Do not make
+/// that nonce deterministic or counter-based without revisiting this.
+async fn inherited_encryption(
+    blockstore: &NamespaceView,
+    heads: &[Cid],
+) -> Result<Option<([u8; 32], Cid)>, String> {
+    let Some(enc_cid) = inherited_encryption_cid(blockstore, heads).await? else {
+        return Ok(None);
+    };
+
+    // Encstore first, then blockstore: blocks arriving over P2P land in the
+    // encstore while locally-written ones go to the blockstore. Mirrors the
+    // merge decrypt path (`db-merge/src/merge_handler/encryption.rs`) and
+    // `versioned_fetcher`.
+    let enc_key = enc_cid.to_bytes();
+    let enc_bytes = match blockstore
+        .sibling(storage::namespace::Namespace::Encstore)
+        .get(&enc_key)
+        .await
+        .map_err(|e| format!("Failed to read encryption block {}: {}", enc_cid, e))?
+    {
+        Some(bytes) => bytes,
+        None => blockstore
+            .get(&enc_key)
+            .await
+            .map_err(|e| format!("Failed to read encryption block {}: {}", enc_cid, e))?
+            .ok_or_else(|| format!("Encryption block {} not found", enc_cid))?,
+    };
+    let enc_block = Encryption::from_dag_cbor(&enc_bytes)
+        .map_err(|e| format!("Failed to decode encryption block {}: {}", enc_cid, e))?;
+    let key: [u8; 32] = enc_block.key.as_slice().try_into().map_err(|_| {
+        format!(
+            "Encryption block {} has a {}-byte key, expected 32",
+            enc_cid,
+            enc_block.key.len()
+        )
+    })?;
+
+    Ok(Some((key, enc_cid)))
+}
+
 /// Write document blocks to blockstore and heads to headstore.
 ///
 /// This function creates CRDT blocks for a document and writes:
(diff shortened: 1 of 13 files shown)


edjroz:
@coderabbitai review

crates/db-blocks/src/write.rs
```diff
@@ -1,5 +1,88 @@
...
+    let Some(enc_cid) = inherited_encryption_cid(blockstore, heads).await? else {
+        return Ok(None);
+    };
+
+    // Encstore first, then blockstore: blocks arriving over P2P land in the
+    // encstore while locally-written ones go to the blockstore. Mirrors the
+    // merge decrypt path (`db-merge/src/merge_handler/encryption.rs`) and
+    // `versioned_fetcher`.
+    let enc_key = enc_cid.to_bytes();
+    let enc_bytes = match blockstore
```
jackzampolin, line 83: **[P2] Resolve inherited keys through the configured KMS.** This helper assumes every KMS-generated `Encryption` block is duplicated into this transaction’s encstore/blockstore and never consults the `kms` argument already supplied to `write_document_blocks`. `KmsService::generate_key` only promises local persistence, and supported stores such as `MemoryKeyStore` keep that state outside these namespaces. I reproduced a KMS-backed create where `get_keys` can resolve the returned CID, followed by an unconfigured update failing here with `Encryption block … not found`. Please thread KMS-aware key resolution into inheritance, or enforce the blockstore-backed storage requirement at the API boundary.

crates/db-blocks/src/write.rs
```diff
@@ -86,74 +169,79 @@ pub async fn write_document_blocks(
...
+
+                tracing::debug!(
+                    field = %field_name,
+                    ciphertext_len = encrypted.len(),
+                    "Encrypted field delta with the previous block's key"
+                );
+
+                (encrypted, Some(enc_cid))
             } else {
                 (value_bytes, None)
```
jackzampolin, line 332: **[P1] Preserve whole-document encryption when a field has no prior field head.** For a document created with `encrypt_doc`, a field first introduced by a later mutation has an empty `heads` slice. `inherited_encryption` therefore returns `None` and this branch writes the new value in plaintext. Before this PR, the stored document config encrypted that field, so removing the store introduces a confidentiality regression during normal operation—not only after eviction or restart. Please derive the document-level policy/key from the previous composite head (or otherwise persist that policy) before allowing the plaintext fallback. Matching the current Go behavior does not make this safe to ship.

jackzampolin (changes requested):
Requesting changes for two regressions in encryption inheritance: whole-document encryption can be lost for newly introduced fields, and inherited key resolution bypasses the configured KMS. Details are inline.

iverc (approved):

edjroz:
@jackzampolin both findings addressed, pushed as `4f0879b7`.

**P1 — whole-document encryption for newly introduced fields**

- Added a document-level policy fallback, read off the composite block's `Encryption` link. The chain is now explicit config → field-head inheritance → document policy → plaintext. `encryptFields` leaves that link `None`, so a field that was never in the list correctly stays in the clear.
- The fallback mints a fresh key through the same helper the explicit path uses rather than reusing the composite's key, since in production each `generate_encryption_key_for` call returns a distinct random key and the composite path ignores `kms` (pre-existing). This reproduces exactly what the store-backed path did before #1292.
- The trigger is "field inheritance found nothing", not "heads are empty" — a document created in the clear and later encrypted has heads carrying no encryption link, and gating on empty heads would have left that case regressed too.
- You were right that Go parity was the wrong standard here. Pre-#1292 this field was encrypted, so shipping as-is was a regression against our own behaviour, not an inherited gap.

**P2 — inherited key resolution and the KMS**

- `inherited_encryption` now takes the `kms` argument and falls back to `get_keys` when neither block namespace holds the `Encryption` block.
- Order is store first, KMS second: production behaviour is bit-identical to today and only the previously-hard-erroring path changes. KMS-first would have put `local_release_decision` on the local write path, where a `Deny` or `OwnershipUnknown` deferral would fail writes that currently succeed.
- Reproduced your scenario before fixing it — the existing `StubKms` already withholds its `Encryption` block the way `MemoryKeyStore` does, so the new test failed with the exact `Encryption block … not found` you hit.
- Also pinned the combination of the two findings — `encrypt_doc` plus a KMS plus a newly introduced field — in `document_policy_fallback_mints_through_the_kms`, since the P1 fallback has to mint through the KMS rather than inline.

PR #1314, opened 2026-08-06
fix(p2p): report failed and timed-out document syncs instead of returning Ok

### LINKED ISSUE #1299: sync_documents returns Ok(()) whether it delivered every document, some, or none

edjroz:
Closes #1299.

## Problem

`sync_documents` returned `Ok(())` in every situation where nothing was synced and nothing
could have been: no connected peers, no request reaching any peer, and a deadline expiring
having heard nothing back. A caller could not tell any of them from a successful sync.

Go reports all three. `syncDocuments` returns `ErrTimeoutDocSync` when it has no active peers
(`internal/db/p2p/sync_doc.go:102-104`), propagates the error when its topic publish fails
(`:118-121`), and returns `ErrTimeoutDocSync` again when the wait context expires having
collected no heads (`:155-158`).

## What changed

1. **Error on zero-peer sync**, in both adapters. `client.Host` documents `ActivePeers` as
   "peers that are currently connected to", so Go's `len(activePeers) == 0` (`sync_doc.go:102`)
   is exactly our `connected_peers.is_empty()`.

2. **Make the send path testable.** Both adapters held concrete transports, so nothing could
   stand in for a failing send. Adds `DocSyncDispatch`, implemented on `IrohTransport` and
   `P2PHostHandle` so each keeps its own signing and error mapping. No behaviour change.

3. **Error when no request reaches any peer.** One edit covering both transports, since they
   now share a send path.

4. **Guard the iroh no-reply path in the default suite.** iroh already errored here, but only
   incidentally: its send awaits a reply, so the send fails and the no-send branch catches it.
   The 30s that kept this `#[ignore]`d belonged to the test, which never drained the responding
   endpoint's event channel. It now costs **0.23s**, un-ignored and mutation-verified.

5. **Error when the deadline expires having heard nothing** (`sync_doc.go:155-158`). Go's rule
   is two-part: its loop succeeds once every peer answers, *even with an empty result*, so this
   errors only when a peer stayed silent **and** nothing was advertised. It checks the raw
   advertisement, which narrowing to unmerged heads first would turn into a false timeout, and
   counts `connected_peers` (Go's `activePeers`), not `topic_peers`.

6. **Count peers that answered with an error.** `parse_doc_sync_response` dropped error and
   undecodable replies entirely, so a peer refusing doc-sync authorization read as silence and
   would have tripped the check above. Go's `handleDocSyncResponse` clears such a peer from
   `pendingPeers`, so it is now recorded as having answered with no heads.

7. **Accept a caller-supplied timeout.** Go reads `timeout` from the request body, parses it
   with `time.ParseDuration`, and answers malformed input with **400**
   (`http/handler_p2p.go:272-280`). `SyncDocumentsRequest` gains `timeout: Option<String>`,
   threaded to both adapters as `Option<Duration>`; the parser rejects a bare `"5"`, as Go does.

8. **Use Go's 5s default** (`sync_doc.go:123-128`). We used 8s for the reply wait inside a 30s
   outer, and 10s on iroh, so accepting the field alone would have left the common no-timeout
   case non-parity. This is the last commit on the branch, so it can be dropped on its own.

The Rust `sync_documents` signature gains the timeout parameter, but the FFI entry point passes
`None`, so **the C ABI is unchanged**. Our FFI is not a drop-in for Go's cbindings regardless,
since Go's `SyncP2PDocuments` takes a uintptr identity and comma-separated docIDs where ours
takes a DID string and JSON, so adding it there is not a parity requirement and would break
existing consumers.

## Deliberately not addressed

- **Partial delivery.** `total_received < total_expected` is still not an error. The counters
  are an upper bound, not a target: a peer with nothing newer to send emits no `MergeComplete`,
  and with N peers and one document a fully successful sync merges once, not N times. Go does
  not detect this either. Its result map is populated from advertised heads *before* the fetch
  is attempted, and `syncDocumentAndMerge` failures are logged and discarded, so erroring here
  would put Rust ahead of Go rather than in step with it.

- **The two-stream fallback** (`doc_sync/sync.rs`) still returns `Ok(())` on silence. It has no
  reply channel, so it cannot express Go's condition at all, and the obvious substitute
  (erroring on zero merges) is not Go's condition either. Reachable on libp2p while pubsub
  services are still coming up, and masked on iroh, whose send awaits a reply.
  `libp2p::tests::doc_sync_with_unresponsive_peer_returns_ok` pins the current behaviour.
  A follow-up issue will track it.

- **`pubsub_sync_documents`'s early exit** counts responses rather than distinct peers, so
  duplicate replies from one peer can end the wait sooner than Go's map-keyed delete would. The
  decision function is not fooled, since it counts senders, but the wait is shorter. Left alone
  as it is a different concern in a different crate.

- **The sibling zero-peer gates** in `sync_branchable_collection` and `sync_collection_versions`,
  in both adapters. Go errors for the branchable case (`ErrTimeoutCollectionSync`,
  `sync_branchable_col.go:98-99`); for collection versions it has no zero-peer check at all and
  simply lets the Bitswap fetch fail, so that one needs a decision rather than a transcription.
  Left out to keep this reviewable, and a follow-up issue will track it.

`pubsub_sync_documents` in `crates/p2p/src/sync/coordinator/pubsub_client.rs` is a different
function with a confusingly similar name, and is untouched except for the error-reply fix noted
above.

## Validation

- [x] `cargo fmt --all`
- [x] `cargo clippy --all --all-targets -- -D warnings`, exit 0
- [x] `cargo test -p defra-p2p-adapter`, 42 passed, **0 ignored**
- [x] `cargo test -p defra-http`, 111 + 15 passed
- [x] `cargo test -p integration-test --test p2p`, 100 passed
- [x] `cargo test -p integration-test --test nac`, 26 passed
- [x] `cargo test -p integration-test --test acp`, 198 passed
- [x] `cargo test -p integration-test --test p2p_iroh`, 230 passed, 2 pre-existing failures

`go_p2p_sync_without_peers` passed *before* the first commit and still passes, so the Go
baseline is verified rather than assumed. The Go comparison was made against
`GO_COMPAT_COMMIT = 13c46ec9`, with `sync_doc.go` confirmed byte-identical to the checkout read.

The two `p2p_iroh` failures are `sync::version::with_view` and
`sync::version::with_view_activated_and_queried`, both missing a Lens WASM fixture that was
never built locally. They fail identically on `main`, and the failure set is unchanged by this
branch.

The 5s default cost nothing and made the suites faster: the adapter suite went from 9.26s to
5.53s, also losing its last ignored test, and `--test p2p` from 37.3s to 31.5s.


edjroz:
@coderabbitai review

crates/http/src/validation.rs
```diff
@@ -72,6 +72,78 @@ pub fn validate_doc_id(doc_id: &str) -> Result<(), HttpError> {
...
+
+        total_nanos += amount * nanos_per_unit;
+    }
+
+    // `Duration` is unsigned, and a negative timeout yields an already-expired
+    // deadline in Go. Zero reproduces that: the operation gives up immediately.
+    if negative {
+        return Some(std::time::Duration::ZERO);
+    }
+    Some(std::time::Duration::from_nanos(total_nanos as u64))
```
jackzampolin: **[P1] Reject overflowing durations instead of saturating them.** This parser accumulates through `f64` and then casts to `u64`; oversized values therefore become `Duration::MAX` instead of producing the HTTP 400 that Go's `time.ParseDuration` returns. In the libp2p pubsub path that value reaches `tokio::time::Instant::now() + wait`, which panics on an unrepresentable instant. Since this repository's release profile uses `panic = "abort"`, a crafted sync timeout can terminate the node. Please use checked arithmetic and enforce Go's signed 64-bit duration bounds, with an overflow regression test.

crates/p2p-adapter/src/doc_sync/sync.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    let mut sub = event_bus.subscribe(&[events::EventName::MergeComplete]);
+    let total_expected = connected_peers.len() * doc_ids.len();
+    let mut total_received = 0;
+    let idle_timeout = std::time::Duration::from_secs(3);
+    let start = std::time::Instant::now();
+    let doc_set: HashSet<String> = doc_ids.iter().cloned().collect();
+
+    for _attempt in 0..3 {
+        if total_received >= total_expected || start.elapsed() >= overall_timeout {
```
jackzampolin: **[P2] Return a timeout error when this condition is what exits the loop.** For connected peers and non-empty `doc_ids`, `timeout: "0"` is valid input, but `start.elapsed() >= overall_timeout` is immediately true, so no request is sent and the unconditional `Ok(())` below reports success. This is precisely the deadline-expired-with-no-response case this PR intends to map to `ErrTimeoutDocSync`. Please distinguish completion from deadline expiry and add a zero/already-expired timeout test.

crates/p2p-adapter/src/doc_sync/sync.rs
```diff
@@ -0,0 +1,182 @@
...
+            event_bus.unsubscribe(sub.id());
+            return Err(error);
+        }
+
+        // No request reached any peer, so nothing can arrive. Unlike a merge
+        // shortfall — where a peer legitimately has nothing newer to send —
+        // this has no benign reading, so it is an error rather than a silent
+        // success. Go reports its equivalent (a failed topic publish) the same
+        // way.
+        let any_sent = send_requests(&dispatch, &connected_peers, request, parallelism).await;
```
jackzampolin: **[P2] Bound dispatch by the caller's remaining timeout.** This await happens before the next deadline check, while iroh's `send_doc_sync_request` waits up to its hard-coded 30-second request/response timeout. Consequently, `timeout: "1s"` (and the new 5-second default) can still block for roughly 30 seconds. With multiple peers, one successful reply plus one 30-second timeout also leaves `any_sent == true`, after which the elapsed deadline can fall through to `Ok(())`, masking the silent peer. Please apply the remaining overall deadline around dispatch and retain enough per-peer response state to detect that case.

jackzampolin (changes requested):
Requesting changes for three timeout-contract issues: one release-build abort path and two cases where the caller's deadline is not enforced correctly.

edjroz:
requested changes made

**1. Overflowing durations**

- The parser is now a transcription of Go's `time.ParseDuration` — exact integer arithmetic with checked ops at each of Go's three overflow points, instead of an `f64` accumulator whose final cast saturated. Anything past `int64` nanoseconds is a 400.
- Differential-tested against go1.26.5 over ~11,200 inputs: systematic unit × digit × fraction coverage, randomized fuzz, and a boundary corpus walking each unit's exact overflow threshold. Zero value mismatches, one intentional acceptance difference.
- That difference: Go accumulates in a wrapping `uint64`, so two terms summing to exactly `1<<64` wrap and parse as `0s`. Reproducing it would turn a duration far past the maximum into "give up immediately", so we reject it. There's a test pinning it.
- Two defects the comment didn't name, fixed by the same rewrite: the `f64` accumulator lost exactness past ~104 days, and negative values collapsed to zero before any magnitude check, so `-9999999h` was accepted where Go rejects it.
- Added `checked_add` at the deadline arithmetic site.

On severity: I couldn't reproduce the abort. The saturation ceiling is ~584 years, and that adds cleanly to an `Instant` on darwin, with `checked_add` returning `Some`; the Linux path is `Timespec` with `i64` seconds, where a monotonic uptime plus 9.2e9 seconds can't overflow. So `panic = "abort"` isn't reached. The defect is real but is a hung request rather than a crash — the pubsub wait only exits early once every expected peer replies, so one silent peer holds the task and connection for the full deadline. Flagging in case it changes your prioritisation; the fix is in either way.

**2. `timeout: "0"` reported success**

- Split the loop's combined entry guard. Completion is checked first, then the deadline; an expiry that dispatched nothing now reports a timeout.
- Deliberately narrower than "distinguish completion from deadline expiry". Under the 5s default, a legitimate sync where peers had nothing newer exits on the idle timer, retries, and trips the deadline around 6s. Erroring there turns "peers had nothing to send" into a false timeout, and this path has no reply channel to tell that from silence — so only the never-dispatched exit is an error. Say the word if you want the broader rule and I'll take it.
- Checking completion first is load-bearing: an empty `docIDs` makes the expected count zero, and the handler doesn't reject empty lists. A rule phrased as "dispatched nothing means timeout" would have silently changed that case too.
- Three tests: the zero-timeout case, plus two regression guards for the empty-`docIDs` and peers-had-nothing cases, both of which passed before the change and must keep passing.

**3. Dispatch wasn't bounded by the caller's deadline**

- Each per-peer send is now bounded by the caller's remaining budget, so iroh's 30s request/response timeout can't outrun a shorter deadline.
- The dispatch step returns a per-peer outcome rather than a bool, so a peer that failed or was cut off is counted instead of discarded.
- For the masking case, the dispatch trait now declares whether a successful send proves the peer replied — true on iroh, whose send is request-response; false on libp2p, whose fire-and-forget send says nothing about whether that peer would have answered. Where it's true, a silent peer with zero merges is Go's `pendingPeers` condition and reports a timeout. Applying it uniformly would invent timeouts on libp2p, so it's iroh-only by design.
- Measured: the adapter suite went from 30.01s to 5.16s once dispatch was bounded — the overrun you described, showing up as wall clock and then disappearing.

**Worth knowing before merge:** bounding dispatch means the 5s default now caps the whole operation on iroh, where a send previously had up to 30s regardless of the caller's timeout. That's the Go behaviour, but on a high-latency link a sync that used to succeed at 12s will now fail.

PR #1322, opened 2026-08-07
chore: Update CLAUDE.md

edjroz:
Refresh `CLAUDE.md` so every factual claim is checkable against the tree.

- Crate tree: 24 to 40 entries, matching the workspace members.
- `tools/` block: 2 to 7 entries, noting which three are workspace members.
- Section 0 no longer names a Go release; it points at the `GO_COMPAT_*` constants in `crates/defra-version/src/lib.rs` that CI greps directly.
- Rust 1.82 to 1.91, per `Cargo.toml` `rust-version`.
- `protoc` attributed to `crates/orbis` (`proto/orbis.proto` via `tonic-prost-build`), not `opentelemetry-proto`.
- Integration-test tables now cover all 23 `[[test]]` binaries, split into area binaries and single-purpose ones.
- Storage table gains `lark` and names it the default; drops the `STORE` env var, which `--store` does not define.
- Naming table used `MAX_PRIORITY`, which does not exist; swapped for `STATUS_ACTIVE`.
- Go upstream section: hardcoded homedir path replaced with `$DEFRADB_GO_REPO`, plus a note that integration tests still resolve Go artifacts at compile time under `$HOME/go/src/github.com/sourcenetwork/defradb`.

Docs only, no code changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,7 +2,12 @@
 
 ## 0. The North Star: 1.0 Release
 
-**defradb.rs is at v1.0 RC1, tracking Go DefraDB v1.0.0-rc1.**
+**The Go compatibility baseline lives in `crates/defra-version/src/lib.rs`, not
+here.** Three constants define it: `GO_COMPAT_BRANCH`, `GO_COMPAT_COMMIT`, and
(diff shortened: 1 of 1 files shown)


jackzampolin (approved):
good bump! merging

Issue #1001, opened 2026-05-28
Telemetry: share one reqwest::Client between trace and metric OTLP exporters


jackzampolin:
Follow-up to #977 / PR #996.

## Problem

\`crates/telemetry/src/init.rs\` currently constructs the span exporter and the metric exporter independently:

\`\`\`rust
let span_exporter = SpanExporter::builder().with_http().build()?;
// ...
let metric_exporter = MetricExporter::builder().with_http().build()?;  // (under `metrics` feature)
\`\`\`

Each call instantiates its own \`reqwest::blocking::Client\`. That means per defra process:

- Two separate connection pools to the same OTLP endpoint
- Two TLS contexts
- Two DNS resolvers
- Two supporting thread pools

Wasteful when both signals target the same collector with the same auth — which is the normal case.

## Better

\`opentelemetry-otlp\`'s \`HttpExporterBuilder\` exposes \`.with_http_client(client)\` that takes a shared \`reqwest::Client\`. Build one client at \`init\` time, hand a clone to each exporter builder.

## Acceptance

- One \`reqwest::Client\` per \`telemetry::init\` call, shared between trace and metric exporters.
- Smoke tests still pass.
- \`netstat\` on a running defra shows one outbound connection (not two) to the collector.

Discovered by extra-high effort \`/code-review\` on PR #996. Lower-priority cleanup; the visible impact is FD count and thread count, not correctness.

jackzampolin:
**Triage: `Post-1.0`.** Procedure: #1391.

`Go Status: Has at v1.0.0`

Read crates/telemetry/src/init.rs pattern described in the issue (SpanExporter::builder().with_http().build() and MetricExporter::builder().with_http().build(), each spinning up its own reqwest::blocking::Client) and compared to Go v1.0.0's internal/telemetry/otel.go, which calls otlptracehttp.New(ctx) and otlpmetrichttp.New(ctx) independently with no shared *http.Client option passed to either -- Go has the identical two-client pattern, so Rust does not diverge from Go here; both implementations independently instantiate a client per exporter. The issue itself states impact is 'FD count and thread count, not correctness' -- no data loss, no silent failure, no security issue. Despite the 'bug' label this is a resource-hygiene cleanup, not a functional defect, and Go already 'has' the same shape so there's no parity work either.

[...]

PR #1323, opened 2026-08-07
Share one http_client for otel exporters

christopherbm:
Fix for [1001](https://github.com/sourcenetwork/defradb.rs/issues/1001)

reqwest clients can be cloned and all the underlying resources/pools are shared.

It was necessary to spawn the blocking clients on a separate thread in order to not crash the main async thread. 

**Testing** 
otel smoke-tests pass
netstat shows one connection


crates/telemetry/src/init.rs
```diff
@@ -72,8 +72,15 @@ pub fn init(config: TelemetryConfig) -> Result<(TelemetryHandle, SdkTracer), Ini
         .with_attribute(KeyValue::new("process.executable.name", executable_name))
         .build();
 
+    // let http_client = reqwest::blocking::Client::new();
```
edjroz: Commented-out code, `CLAUDE.md`. Please drop this line. The "necessary to build the blocking client off the async thread" comment below it is the useful WHY, keep that one.

crates/telemetry/src/init.rs
```diff
@@ -72,8 +72,15 @@ pub fn init(config: TelemetryConfig) -> Result<(TelemetryHandle, SdkTracer), Ini
...
+    // necessary to build the blocking client off the async thread, like opentelemetry-otlp does
+    let http_client = std::thread::spawn(reqwest::blocking::Client::new)
```
edjroz: Injecting the client skips `opentelemetry-otlp`'s timeout wiring: it only calls
`.timeout(...)` inside `if http_client.is_none()` (`exporter/http/mod.rs:232-258`),
and nothing downstream enforces a deadline (`periodic_reader.rs:100-101`,
`span_processor.rs:562-563`). So `OTEL_EXPORTER_OTLP_TIMEOUT` becomes a no-op and
every export deadline becomes reqwest's 30s default instead of 10s, which also
blows `shutdown()`'s 5s flush budget.

Can we build via `Client::builder().timeout(...)`? That closes CodeRabbit's
line-79 point too. Note `resolve_timeout` is private so the cascade needs
replicating, and one shared client means per-signal overrides are gone regardless.

edjroz (changes requested):
Sharing one blocking client is the right call, two clients means two inner
runtimes and two connection pools, and the clones do share one
`Arc<InnerClientHandle>`, so the intent works. One blocking issue inline that I
don't think the automated pass caught. Agreed with CodeRabbit on the
`ClientBuilder` / `InitError` point at line 79, no need to restate it.

crates/telemetry/src/init.rs
jackzampolin: **[bug]** Client::new() drops OTLP export timeout. Upstream applies OTEL_EXPORTER_OTLP_TIMEOUT (default 10s) only when building the default client; custom clients get no request timeout, so hung collectors can block export threads indefinitely.

**Suggestion:** Build shared client with .timeout(...) matching upstream (generic OTEL_EXPORTER_OTLP_TIMEOUT / 10s default).

jackzampolin: **[nit]** Commented-out code left in (project policy: no commented-out code).

**Suggestion:** Drop the dead Client::new() line; keep a single WHY comment on the spawn.

crates/telemetry/src/init.rs
jackzampolin: **[suggestion]** .expect panics instead of InitError on builder-thread join failure.

**Suggestion:** Map join/build failure into InitError for consistency with the public init API.

jackzampolin (dismissed):
## Summary

PR #1323 correctly addresses issue #1001: one `reqwest::blocking::Client` is built at `telemetry::init` time and handed (via `Clone`) to both the span and metric OTLP HTTP exporters through `WithHttpConfig::with_http_client`. That matches the intended API (`Client` is `Arc`-backed; clones share the connection pool) and covers the only two exporters this crate builds (no log exporter path). Feature-gating is clean: `reqwest` is `optional` and enabled only under `otlp` via `dep:reqwest`.

The thread-spawn around client construction is the right call — `opentelemetry-otlp` 0.32 does the same when auto-building a blocking client, because `reqwest::blocking::Client` must not be created on a Tokio worker. Lifecycle is fine: exporters retain the client (`Arc<dyn HttpClient>`), and the local handle drops at the end of `init` after both builders take clones.

Main concern is a real behavioral regression on export timeouts. When you supply a custom client, OTLP skips its default client builder that applies `OTEL_EXPORTER_OTLP_*_TIMEOUT` (default 10s). The SDK stores that timeout as `_timeout` and does **not** apply it per-request for the blocking-reqwest path, so `Client::new()` effectively means “no request timeout,” whereas pre-PR behavior had a 10s client timeout. Secondary nits: leftover commented-out code (against project policy) and a panic path via `.expect` instead of `InitError`.

go-compat CI failure noted as unrelated (CACHE_DIR bug / #1360). Conformance SKIPPED as expected.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 1

vertexclique (approved):

crates/telemetry/src/util.rs
```diff
@@ -0,0 +1,19 @@
...
+        .to_owned()
+}
+
+pub fn otel_timeout() -> Duration {
+    const DEFAULT: Duration = Duration::from_secs(10);
+    std::env::var("OTEL_EXPORTER_OTLP_TIMEOUT")
+        .ok()
+        .and_then(|ms| ms.parse::<u64>().ok())
+        .map(Duration::from_millis)
+        .unwrap_or(DEFAULT)
```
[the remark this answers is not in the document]
vertexclique: Not needed atm. This is better.

jackzampolin:
Non-blocking note on `otel_timeout()` — the approach here is right (it mirrors what
`opentelemetry-otlp` does internally, including the `std::thread::spawn(...).join()`, which is
required because `reqwest::blocking::Client` cannot be constructed inside a Tokio runtime — see
`exporter/http/mod.rs:254-263` — and this version propagates the error instead of `unwrap()`-ing).
The 10s default also matches upstream's `OTEL_EXPORTER_OTLP_TIMEOUT_DEFAULT`.

One behavior change worth recording: upstream resolves the timeout via `resolve_timeout`
(`exporter/mod.rs:296`), which checks the **signal-specific** variable first —
`OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` / `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` — before falling back
to `OTEL_EXPORTER_OTLP_TIMEOUT`. `otel_timeout()` only reads the generic one, so per-signal
timeouts silently stop having an effect.

That is inherent to sharing one client (the timeout is a client-level property, and the whole
point of the PR is that traces and metrics share it), so it is a tradeoff rather than a bug. A
line in the PR description or a `///` on `otel_timeout` noting that per-signal timeouts are
deliberately not honored would keep the next person from filing it as a regression.

The CI failure on this PR is unrelated — `downsample::rust_downsample_updates_stable_rollup_docs`
is a wall-clock race in the test that is currently failing on `main` too; fix in #1446.

PR #1325, opened 2026-08-07
Run signed embedded transactions on a node-owned runtime

jackzampolin:
## Summary

- expose EmbeddedNode::execute_request_in_txn
- apply the configured node signer and ambient identity context to explicit transaction requests
- run signed root queries on a node-owned Tokio runtime so caller cancellation cannot invalidate spawned query work
- close admission and drain signed queries before P2P shutdown
- bound native signed-node runtime threads to one async worker per node
- cover cancellation, spawned work, TLS context across awaits, deadlines, concurrent shutdown, node isolation, and async Drop

## Why

EmbeddedNode::execute installs the node signer for auto-commit work, but explicit transactions previously went through raw QueryExecutor::execute_in_txn and could silently create unsigned commits.

The initial transaction wrapper also exposed a lifecycle boundary: query futures and their spawned tasks could outlive the caller runtime or be cancelled while DefraDB still needed to finalize signed state. The node now owns the execution runtime and admission/drain protocol. Root signed work is either admitted and counted before shutdown closes, or rejected after close; admitted work has bounded drain semantics.

The caller-provided QueryRequest is preserved, including its ACP actor identity. Signing identity and query identity remain separate.

This is required by source-inc/gents#1065 for atomic request ingest, signed claim ancestry, and exact CID provenance.

## Shutdown contract

1. close signed-query admission;
2. wait up to 30 seconds for admitted root queries;
3. stop P2P while the signed runtime is still alive;
4. signal and join the runtime owner;
5. bound owner/runtime shutdown with five-second deadlines.

Drop uses the same close/signal path without attempting to drop a Tokio runtime from an async context.

## Validation

Green:

- cargo test -p defra-node
- cargo clippy -p defra-node --all-targets -- -D warnings
- cargo fmt --all -- --check
- GitHub: Build & Test
- GitHub: Integration (rust)
- GitHub: Integration (signed-docs)
- GitHub: P2P Integration (Linux)
- GitHub: Lint, WASM Build Check, Lean Conformance, Conformance
- independent Claude and Grok lifecycle reviews: no P0/P1 blockers

Two macOS integration jobs currently fail after successful CLI builds because their shared runner script copies into a host cache directory it never creates. That is CI infrastructure, not a signed-query test failure; no workflow change is included here.

### DIFF
--- a/crates/defra-node/src/db_impls.rs
+++ b/crates/defra-node/src/db_impls.rs
@@ -6,7 +6,60 @@ use identity::Did;
 use lens::{LensConfig, TransformId};
 use schema::CollectionVersion;
 
-use crate::SchemaOps;
+use crate::{BlockOps, SchemaOps};
+
+pub(crate) struct DbBlockOps<S: storage::corekv::Store + 'static> {
+    database: Arc<db::DB<S>>,
+    document_acp: Arc<dyn acp::DocumentACP>,
+    transaction_registry: Arc<db::DbTransactionRegistry<S>>,
+    caller_identity: acp::Identity,
+}
+
+impl<S: storage::corekv::Store + 'static> DbBlockOps<S> {
+    pub(crate) fn new(
+        database: Arc<db::DB<S>>,
+        document_acp: Arc<dyn acp::DocumentACP>,
+        transaction_registry: Arc<db::DbTransactionRegistry<S>>,
+    ) -> Self {
+        let caller_identity = database.node_did().into();
+        Self {
+            database,
+            document_acp,
+            transaction_registry,
+            caller_identity,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl<S: storage::corekv::Store + 'static> BlockOps for DbBlockOps<S> {
+    async fn verified_signer_did(&self, cid: &str) -> anyhow::Result<String> {
+        db::block_verify::verified_block_signer_did(
+            &self.database,
+            self.document_acp.as_ref(),
+            cid,
+            &self.caller_identity,
+        )
+        .await
+        .map_err(anyhow::Error::msg)
+    }
+
+    async fn verified_signer_did_in_txn(
+        &self,
+        cid: &str,
+        transaction: &query::TransactionHandle,
+    ) -> anyhow::Result<String> {
+        self.transaction_registry
+            .verified_block_signer_did_in_txn(
+                transaction.as_str(),
+                self.document_acp.as_ref(),
+                cid,
+                &self.caller_identity,
+            )
+            .await
+            .map_err(anyhow::Error::new)
+    }
+}
 
 pub(crate) struct DbSchemaOps<S: storage::corekv::Store + 'static> {
     database: Arc<db::DB<S>>,
--- a/crates/defra-node/src/lib.rs
+++ b/crates/defra-node/src/lib.rs
@@ -689,7 +689,7 @@ impl Drop for SignedQueryPermit {
         if self
             .state
             .active_queries
-            .fetch_sub(1, std::sync::atomic::Ordering::AcqRel)
+            .fetch_sub(1, std::sync::atomic::Ordering::SeqCst)
             == 1
         {
             self.state.active_queries_drained.notify_waiters();
@@ -708,7 +708,7 @@ impl Drop for SignedQueryRuntimeClosedGuard {
     fn drop(&mut self) {
         self.state
             .closed
-            .store(true, std::sync::atomic::Ordering::Release);
+            .store(true, std::sync::atomic::Ordering::SeqCst);
         self.state.closed_notify.notify_waiters();
     }
 }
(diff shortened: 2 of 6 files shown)


jackzampolin:
CI triage for 688f88a3:

- Integration (go-compat) and Integration (iroh) both successfully build the CLI.
- Both then fail in Restore or build binaries at cp .../$HASH/defra because $CACHE_DIR was not created.
- The script creates target/ci but does not mkdir -p "$CACHE_DIR" before populating the host-local cache.
- All code-relevant signed runtime, Rust integration, build/test, lint, WASM, Lean, conformance, signed-docs, and Linux P2P checks are green.

This appears to be shared runner-cache infrastructure and is intentionally not changed in this PR without separate CI-maintainer approval.

crates/defra-node/src/lib.rs
```diff
@@ -371,11 +473,30 @@ impl EmbeddedNode {
             let _ = tokio::time::timeout(Duration::from_secs(1), task).await;
```
jackzampolin, line 473: **[suggestion]** runner().execute_in_txn remains an unsigned escape hatch when a node identity is configured — callers can recreate the original unsigned-commit bug.

**Suggestion:** Expose begin_txn/commit_txn/rollback_txn on EmbeddedNode and document runner().execute_in_txn as unsupported for signed nodes, or wrap executor to always apply signing context.

crates/defra-node/src/lib.rs
```diff
@@ -371,11 +473,30 @@ impl EmbeddedNode {
...
         }
 
+        #[cfg(not(target_arch = "wasm32"))]
+        if let Some(runtime) = &self.signed_query_runtime {
+            if !runtime
+                .close_admission_and_wait_for(SIGNED_QUERY_DRAIN_TIMEOUT)
+                .await
```
jackzampolin, line 480: **[suggestion]** Force-shutdown after drain timeout can interrupt in-flight signed mutations (Drop is worse: ~5s then detach).

**Suggestion:** On drain timeout refuse further teardown or surface structured shutdown error; align Drop policy with explicit shutdown.

crates/defra-node/src/db_impls.rs
```diff
@@ -6,7 +6,60 @@ use identity::Did;
...
+    caller_identity: acp::Identity,
+}
+
+impl<S: storage::corekv::Store + 'static> DbBlockOps<S> {
+    pub(crate) fn new(
+        database: Arc<db::DB<S>>,
+        document_acp: Arc<dyn acp::DocumentACP>,
+        transaction_registry: Arc<db::DbTransactionRegistry<S>>,
+    ) -> Self {
+        let caller_identity = database.node_did().into();
```
jackzampolin, line 24: **[suggestion]** DbBlockOps ACP identity uses database.node_did() only; remote-signer-only nodes verify as Anonymous under DAC.

**Suggestion:** Construct caller_identity from EmbeddedNode::node_identity_did / registered DID string.

crates/defra-node/src/lib.rs
```diff
@@ -441,30 +562,356 @@ where
...
+#[cfg(not(target_arch = "wasm32"))]
+const SIGNED_QUERY_RUNTIME_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5);
+
+#[cfg(not(target_arch = "wasm32"))]
+struct SignedQueryPermit {
+    state: Arc<SignedQueryRuntimeState>,
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl Drop for SignedQueryPermit {
```
jackzampolin, line 682: **[suggestion]** ~300+ lines of SignedQueryRuntime land in an already large lib.rs against project file-size guidance.

**Suggestion:** Extract to signed_query_runtime.rs before undrafting.

jackzampolin (commented):
## Summary

DRAFT PR #1325 wires signed embedded GraphQL work (auto-commit and explicit transactions) through a node-owned Tokio runtime with admission/drain, installs the node signer + ambient identity for both `execute` and new `execute_request_in_txn`, fails closed when the configured signer is missing, and adds trust-on-verify block signer DID APIs (including in-txn). Shutdown order (close admission → drain up to 30s → P2P → join runtime) is coherent, TLS signing cleanup is improved via `ThreadSigningContextGuard`, and coverage for cancel/spawn/context/drain/isolation/async Drop plus an end-to-end txn signing test is strong for a draft.

No clear P0. Main residual risks before merge: public `runner().execute_in_txn` still bypasses signing (easy footgun for the gents use case), Drop/timeout paths can force-kill in-flight signed work after short drains, `DbBlockOps` ACP identity is tied to local-key `database.node_did()` (remote-signer-only nodes become anonymous for verify), and `lib.rs` absorbs ~900 lines of runtime machinery that should be split out. go-compat/iroh failures are noted as known CI cache issues outside this change.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 0

jackzampolin:
Superseded by #1373, which rebases the signed execution work onto current main, consolidates the stack into one commit, and addresses the prior review findings.

PR #1352, opened 2026-08-08
Release persistent store after P2P shutdown

jackzampolin:
## Summary

- retain the pending-DAG resync and retry-clock task handles in `P2PLifecycleInner`
- abort and join both tasks during `EmbeddedNode::shutdown`
- cover persistent P2P shutdown, drop, and immediate reopen with a focused regression test

## Root cause

`setup_p2p` spawned the pending-DAG resync and retry-clock loops without retaining their `JoinHandle`s. Both futures own a cloned sync coordinator, which in turn retains the persistent store. `EmbeddedNode::shutdown` therefore completed while those detached tasks still kept the redb database alive, and immediately reopening the same path failed with `Database is locked by another process`.

The Iroh lifecycle now owns these tasks alongside its other background work and aborts/joins them before shutting down the coordinator.

## Regression evidence

The new `p2p_shutdown_releases_persistent_store_for_immediate_reopen` test performs this sequence:

1. start a persistent P2P node and wait until it is listening
2. shut down and drop the node
3. immediately reopen the same data path and P2P identity
4. shut down and drop the reopened node

With the lifecycle fix removed, the test fails at step 3 with redb's database-lock error. With this change applied, it passes. The existing `p2p_document_subscriptions_survive_restart_and_delete` three-restart test also failed on its first reopen before the fix and now passes all cycles.

## Validation

- `cargo test -p defra-node --features p2p p2p_shutdown_releases_persistent_store_for_immediate_reopen`
- `cargo test -p defra-node --features p2p p2p_document_subscriptions_survive_restart_and_delete`
- `cargo check -p defra-node --features p2p --all-targets`
- `cargo clippy -p defra-node --features p2p --lib --tests -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

This unblocks the persistent P2P shutdown/reopen path required by [Gents PR #1065](https://github.com/source-inc/gents/pull/1065).

PR #1353, opened 2026-08-08
Attach node identity and expose signed block material

jackzampolin:
## What changed

Embedded node queries without an explicit document actor now default to the configured node DID. Prepared requests with an explicit identity keep that actor, retry attempts preserve it, and transactional execution follows the same rule. Unsigned nodes remain anonymous, and HTTP behavior is unchanged.

The block API now also exposes an ACP-authorized `/api/v0/block/signed?cid=...` endpoint that returns the canonical DAG-CBOR bytes for a commit and its detached signature. Clients can verify the requested CID, signature CID, cryptographic signature, and signer DID locally instead of trusting a server-side verification verdict.

## Why

A signed embedded node previously signed commits as the node while submitting document-ACP queries without an actor. Runtime clients therefore could not rely on node identity being attached to every embedded read and write. Remote provenance clients also lacked the signed source material needed to verify authorship independently.

## Impact

Embedded applications can use the node DID as their default ACP principal without rebuilding every query as a prepared request. Explicit actor delegation remains possible inside the trusted embedded application boundary. Authorized remote readers can independently verify commit provenance without receiving block material they cannot read.

This PR is stacked on #1352 because it uses that series per-node signing runtime and configured node identity.

## Validation

- focused embedded identity seam tests
- signed-block HTTP route and permission tests
- block signature verification tests
- cargo test -p defra-node
- cargo test -p query
- cargo check --workspace --all-targets
- cargo check -p defra-http -p cli
- cargo fmt --all --check
- git diff --check

### DIFF
--- a/crates/db/src/block_verify.rs
+++ b/crates/db/src/block_verify.rs
@@ -39,6 +39,43 @@ pub async fn verified_block_signer_did<S: Store>(
     .await
 }
 
+/// Load one authorized signed block and its detached signature block as
+/// canonical DAG-CBOR bytes. Remote clients use this to verify CID integrity
+/// and authorship locally instead of trusting a server-side yes/no verdict.
+pub async fn authorized_signed_block_bytes<S: Store>(
+    database: &Arc<DB<S>>,
+    document_acp: &dyn acp::DocumentACP,
+    cid_str: &str,
+    caller_identity: &acp::Identity,
+) -> Result<(Vec<u8>, Vec<u8>), String> {
+    let txn = database
+        .new_txn(true)
+        .await
+        .map_err(|e| format!("failed to create transaction: {}", e))?;
+    let blockstore = txn
+        .blockstore()
+        .map_err(|e| format!("failed to get blockstore: {}", e))?;
+    let systemstore = txn
+        .systemstore()
+        .map_err(|e| format!("failed to get systemstore: {}", e))?;
+    let (block, signature) = load_authorized_block_signature(
+        database,
+        document_acp,
+        blockstore,
+        systemstore,
+        cid_str,
+        caller_identity,
+    )
+    .await?;
+    let block_bytes = block
+        .to_dag_cbor()
+        .map_err(|error| format!("failed to encode authorized block: {error}"))?;
+    let signature_bytes = signature
+        .to_dag_cbor()
+        .map_err(|error| format!("failed to encode authorized signature block: {error}"))?;
+    Ok((block_bytes, signature_bytes))
+}
+
 /// Verify a block visible inside an active transaction and return its signer DID.
 pub async fn verified_block_signer_did_in_txn<S: Store>(
     database: &Arc<DB<S>>,
--- a/crates/defra-node/src/embedded_query_identity_tests.rs
+++ b/crates/defra-node/src/embedded_query_identity_tests.rs
@@ -106,6 +106,32 @@ fn no_retry() -> ExecuteRetryPolicy {
 }
 
 #[tokio::test]
+async fn builder_rejects_invalid_node_identity_did() {
+    let _serial = SIGNING_STORE_GUARD.lock().await;
+    defra_core::signing::clear_identity_store();
+    let invalid_did = "invalid:key:zNode";
+    register_test_remote_node_identity(invalid_did);
+
+    let result = EmbeddedNode::builder()
+        .with_node_identity_did(invalid_did)
+        .build()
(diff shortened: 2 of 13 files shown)


crates/db/src/block_verify.rs
```diff
@@ -39,6 +39,43 @@ pub async fn verified_block_signer_did<S: Store>(
...
+    let (block, signature) = load_authorized_block_signature(
+        database,
+        document_acp,
+        blockstore,
+        systemstore,
+        cid_str,
+        caller_identity,
+    )
+    .await?;
+    let block_bytes = block
```
jackzampolin, line 70: **[bug]** authorized_signed_block_bytes re-encodes decoded blocks instead of returning CID-verified blockstore bytes. Re-encoding can break client CID checks on Go-originated blocks or non-identical CBOR round-trips.

**Suggestion:** Return the original blockstore slices verified by load_authorized_block_signature, not to_dag_cbor() re-encodings.

crates/http/src/handlers/block_tests.rs
```diff
@@ -0,0 +1,82 @@
...
+        _public_key: &str,
+        _key_type: Option<&str>,
+        _caller_did: Option<&str>,
+    ) -> Result<(), String> {
+        Ok(())
+    }
+}
+
+#[tokio::test]
+async fn signed_block_route_returns_canonical_material() {
```
jackzampolin, line 44: **[suggestion]** Coverage for the new route is mock-only happy path. Missing: ACP denial, SignatureVerify NAC denial, Authorization identity forwarding, real blockstore CID integrity.

**Suggestion:** Add end-to-end tests against the real loader for DAC denial and cid(block_bytes)==requested cid.

crates/http/src/handlers/block.rs
```diff
@@ -17,6 +19,40 @@ pub struct VerifySignatureParams {
     pub key_type: Option<String>,
 }
 
+#[derive(Debug, Deserialize)]
+pub struct SignedBlockParams {
+    pub cid: String,
+}
+
+#[derive(Debug, Serialize)]
```
jackzampolin, line 27: **[suggestion]** SignedBlockResponse returns cid/block/signature but not the signature block CID, though the PR narrative lists signature CID as a first-class local check.

**Suggestion:** Add signature_cid to the response, or document that clients must parse the commit to obtain it.

crates/defra-node/src/embedded_query_identity_tests.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    let defaulted = node.execute("query { defaulted }").await;
+    assert!(!defaulted.has_errors(), "defaulted query failed");
+    let explicit = node
+        .execute_request_with_retry(
+            QueryRequest::new("query { explicit }").with_identity(Some(explicit_did.clone())),
+            no_retry(),
+        )
+        .await;
+    assert!(!explicit.has_errors(), "explicit-actor query failed");
```
jackzampolin, line 151: **[suggestion]** Identity tests only use a recording executor; they do not show the defaulted node DID drives document ACP under a real policy.

**Suggestion:** Add a small end-to-end ACP case for bare execute() on signed nodes.

jackzampolin (commented):
## Summary

Draft PR attaches the configured node DID as the default GraphQL/document-ACP actor on embedded `execute` / retry / in-txn paths, and adds an ACP-gated HTTP surface that returns base64 DAG-CBOR for a signed commit plus its detached signature block. Identity attachment is implemented cleanly at the embedded seam (`with_default_query_identity`), preserves explicit `QueryRequest` identities, keeps unsigned nodes anonymous, and does not change HTTP GraphQL actor resolution. Secret material is not leaked: signature blocks carry public key + signature value only; private keys stay out of the response. Stacking on `agent/p2p-shutdown` is real at the code level (uses `load_authorized_block_signature`, signed-query runtime, and related base APIs). Main correctness concern is that `authorized_signed_block_bytes` re-encodes decoded structures instead of returning the already-CID-verified blockstore bytes, which undercuts the endpoint’s “verify CID locally” contract. Tests cover the identity seam and a mock happy-path HTTP route, but not end-to-end CID integrity, DAC denial, or authenticated caller forwarding.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 0

jackzampolin:
Superseded by #1373, which consolidates the node identity and signed-block work onto current main in one reviewable commit.

PR #1355, opened 2026-08-10
ci: report release artifact size changes

### LINKED ISSUE #1328: CI: track binary size per release variant
## Context

There is no size tracking in CI today: zero hits for size/bloat/twiggy/machete anywhere in `.github/workflows/`. Without it, every size PR's claim is unverifiable and regressions are invisible. This is the foundation issue for the whole epic, since every other issue's acceptance criteria says "measured delta recorded".

## Scope

Record, per release variant:

- `libffi.dylib` bytes for each entry in the ffi matrix (`release.yml:231-267`)
- `defra` bytes for each cli variant
- wasm raw + gzip + **brotli** (wasm job at `release.yml:198`)

Brotli is what CDNs actually serve and is currently measured nowhere.

Publish in the job summary, and/or as a comparison against the previous release.

## Acceptance criteria

- [ ] Every release variant reports its artifact size in CI output.
- [ ] wasm reports raw, gzip, and brotli.
- [ ] A size change against the prior release is visible without downloading artifacts.

---

Part of #1327. Effort: S. Blocked by: nothing.

iverc:
Closes #1328.

## Problem

Release builds publish multiple CLI, FFI, and WASM variants without recording their payload sizes. Size regressions are therefore invisible in CI, and validating optimization claims requires downloading and inspecting artifacts manually.

## What changed

- Report the exact shipped CLI binary and desktop FFI library size for every release matrix entry.
- Report deterministic WASM raw, gzip-9, and brotli-11 sizes.
- Compare each artifact with the matching asset from the preceding reachable release tag.
- Show absolute and percentage changes together in one release job summary.
- Warn when a historical variant is unavailable while still reporting the current size.
- Fail the reporting step when any expected CLI, FFI, or WASM artifact family is missing.

## Validation

- [x] `cargo test --workspace --lib`
- [x] Bash syntax validation
- [x] Release workflow YAML parsing
- [x] Synthetic current/prior archive comparison
- [x] Measurement against published v0.17.4 CLI, FFI, and WASM archives

### DIFF
--- a/.github/scripts/report-release-sizes.sh
+++ b/.github/scripts/report-release-sizes.sh
@@ -1,0 +1,129 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+ARTIFACT_DIR=${1:?artifact directory is required}
+PREVIOUS_DIR=${2:?previous artifact directory is required}
+CURRENT_TAG=${3:?current tag is required}
+PREVIOUS_TAG=${4:-}
+SUMMARY=${GITHUB_STEP_SUMMARY:-/dev/stdout}
+
+tmp=$(mktemp -d)
+trap 'rm -rf "$tmp"' EXIT
+
+format_size() {
+  awk -v bytes="$1" 'BEGIN {
+    if (bytes >= 1048576) printf "%.2f MiB", bytes / 1048576
+    else printf "%.2f KiB", bytes / 1024
+  }'
+}
+
+format_delta() {
+  awk -v current="$1" -v previous="$2" 'BEGIN {
+    delta = current - previous
+    printf "%+.2f MiB (%+.1f%%)", delta / 1048576, delta * 100 / previous
+  }'
+}
+
+member_size() {
+  tar -xOzf "$1" "$2" | wc -c | tr -d ' '
+}
+
+report_row() {
+  local label=$1 current=$2 previous=${3:-}
+  if [ -n "$previous" ]; then
+    printf '| `%s` | %s | %s | %s |\n' \
+      "$label" "$(format_size "$current")" "$(format_size "$previous")" \
+      "$(format_delta "$current" "$previous")" >> "$SUMMARY"
+  else
+    printf '| `%s` | %s | - | - |\n' "$label" "$(format_size "$current")" >> "$SUMMARY"
+  fi
+}
+
+previous_archive() {
+  local name=$1
+  if [ -n "$PREVIOUS_TAG" ]; then
+    local previous_name=${name/"$CURRENT_TAG"/"$PREVIOUS_TAG"}
+    if [ -f "$PREVIOUS_DIR/$previous_name" ]; then
+      printf '%s\n' "$PREVIOUS_DIR/$previous_name"
+    fi
+  fi
+}
+
+{
+  echo '## Release artifact sizes'
+  echo
+  if [ -n "$PREVIOUS_TAG" ]; then
+    echo "Compared with \`$PREVIOUS_TAG\`."
+  else
+    echo 'No previous release tag was found.'
+  fi
+  echo
+  echo '| Artifact | Current | Previous | Change |'
+  echo '|---|---:|---:|---:|'
+} >> "$SUMMARY"
+
+cli_count=0
+ffi_count=0
+wasm_count=0
+
+for archive in "$ARTIFACT_DIR"/*.tar.gz; do
+  name=${archive##*/}
+  label=${name/"_$CURRENT_TAG"/}
+  label=${label%.tar.gz}
+  previous=$(previous_archive "$name")
+
+  case "$name" in
+    defra-wasm_*)
+      current_wasm="$tmp/current.wasm"
+      tar -xOzf "$archive" defra_wasm_bg.wasm > "$current_wasm"
+      current_raw=$(wc -c < "$current_wasm" | tr -d ' ')
+      current_gzip=$(gzip -9 -n -c "$current_wasm" | wc -c | tr -d ' ')
+      current_brotli=$(brotli --quality=11 --stdout "$current_wasm" | wc -c | tr -d ' ')
+
+      previous_raw=
+      previous_gzip=
+      previous_brotli=
+      if [ -n "$previous" ]; then
+        previous_wasm="$tmp/previous.wasm"
+        tar -xOzf "$previous" defra_wasm_bg.wasm > "$previous_wasm"
+        previous_raw=$(wc -c < "$previous_wasm" | tr -d ' ')
+        previous_gzip=$(gzip -9 -n -c "$previous_wasm" | wc -c | tr -d ' ')
+        previous_brotli=$(brotli --quality=11 --stdout "$previous_wasm" | wc -c | tr -d ' ')
+      fi
+
+      report_row "$label raw" "$current_raw" "$previous_raw"
+      report_row "$label gzip-9" "$current_gzip" "$previous_gzip"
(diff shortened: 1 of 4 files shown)


.github/scripts/report-release-sizes.sh
```diff
@@ -0,0 +1,129 @@
...
+      [ -n "$member" ] || { echo "No FFI library found in $name" >&2; exit 1; }
+      current=$(member_size "$archive" "$member")
+      previous_size=
+      if [ -n "$previous" ]; then
+        previous_member=$(tar -tzf "$previous" | awk '/^libdefra_ffi\.(so|dylib)$/ { member = $0 } END { print member }')
+        [ -n "$previous_member" ] || { echo "No FFI library found in ${previous##*/}" >&2; exit 1; }
+        previous_size=$(member_size "$previous" "$previous_member")
+      fi
+      report_row "$label" "$current" "$previous_size"
+      ffi_count=$((ffi_count + 1))
```
jackzampolin, line 127: **[bug]** Missing previous assets are handled softly (warn + `-` columns), but a *present* previous archive that fails listing/extraction hard-exits the script under `set -e` / explicit `exit 1`. Because the report step runs before `softprops/action-gh-release`, a corrupt previous asset, layout rename (e.g. future FFI member name change), or unexpected archive content blocks publishing the new release even though current artifacts are fine. Same failure class applies to CLI `member_size "$previous" defra` (line 124) and WASM previous extraction (lines 93–98).

**Suggestion:** Treat previous-archive parse/extract failures like missing assets: log `::warning::`, leave previous empty, and continue reporting current sizes so comparison never gates release publication.

.github/workflows/release.yml
```diff
@@ -507,6 +511,37 @@ jobs:
...
+          mkdir -p previous-artifacts
+          if [ -n "$PREVIOUS_TAG" ]; then
+            for archive in artifacts/*.tar.gz; do
+              name=${archive##*/}
+              case "$name" in
+                defra-ffi-*ios_xcframework*) continue ;;
+              esac
+              previous_name=${name/"$GITHUB_REF_NAME"/"$PREVIOUS_TAG"}
+              url="https://github.com/${GITHUB_REPOSITORY}/releases/download/${PREVIOUS_TAG}/${previous_name}"
+              if ! curl --fail --location --silent --show-error --retry 3 \
```
jackzampolin, line 532: **[bug]** Previous-release downloads use `curl --fail --location --retry 3` with no `--connect-timeout` / `--max-time`. A hung TCP connection to GitHub or a stuck redirect target can block the publish job indefinitely. This step is on the critical path before the GitHub Release is created.

**Suggestion:** Add bounded timeouts (e.g. `--connect-timeout 20 --max-time 300`) and keep the existing failure path that `rm -f`s the partial file and emits a warning. Optionally prefer `gh release download "$PREVIOUS_TAG" -p "$previous_name" -D previous-artifacts` for first-party asset resolution.

.github/workflows/release.yml
```diff
@@ -477,6 +477,10 @@ jobs:
       needs.docker.result == 'success'
     runs-on: [self-hosted, Linux, X64]
     steps:
```
jackzampolin, line 479: **[suggestion]** New `actions/checkout@v4` does not set `persist-credentials: false`. This job only needs the tree (for the report script) and git history/tags (for `git describe`); it does not push. Leaving credentials persisted on a self-hosted runner is unnecessary exposure (zizmor artipacked class).

**Suggestion:** Add `persist-credentials: false` alongside `fetch-depth: 0`.

.github/scripts/report-release-sizes.sh
```diff
@@ -0,0 +1,129 @@
...
+}
+
+format_delta() {
+  awk -v current="$1" -v previous="$2" 'BEGIN {
+    delta = current - previous
+    printf "%+.2f MiB (%+.1f%%)", delta / 1048576, delta * 100 / previous
+  }'
+}
+
+member_size() {
```
jackzampolin, line 31: **[nit]** `format_delta` always prints the absolute delta in MiB and computes `delta * 100 / previous` with no zero guard. A zero-byte previous member would yield awk division-by-zero / `inf` percentage. Also, sub-MiB deltas render as `+0.00 MiB` and lose useful precision versus `format_size`'s KiB path.

**Suggestion:** If `previous == 0`, print delta without a percentage (or `n/a`). Reuse unit selection similar to `format_size` for the absolute delta while keeping signed percent when `previous > 0`.

.github/scripts/report-release-sizes.sh
```diff
@@ -0,0 +1,129 @@
...
+      report_row "$label" "$current" "$previous_size"
+      cli_count=$((cli_count + 1))
+      ;;
+  esac
+done
+
+if [ "$cli_count" -eq 0 ] || [ "$ffi_count" -eq 0 ] || [ "$wasm_count" -eq 0 ]; then
+  echo "Incomplete release artifact set: cli=$cli_count ffi=$ffi_count wasm=$wasm_count" >&2
+  exit 1
+fi
```
jackzampolin, line 147: **[suggestion]** Completeness is family-level (`cli_count`, `ffi_count`, `wasm_count` each ≥ 1), not matrix-complete. If a single platform archive were absent while another of the same family remained, the step would still pass and silently omit that variant from the summary. Today this is largely mitigated by `needs.*.result == 'success'` on every build matrix job, so risk is low.

**Suggestion:** Optionally assert expected counts (e.g. 12 CLI, 8 desktop FFI, 1 WASM) or fail on any current archive whose case arm is empty/unmatched, if stricter matrix coverage is desired.

jackzampolin (changes requested):
## Summary

PR #1355 correctly wires release-time size reporting for CLI binaries, desktop FFI libraries, and WASM (raw/gzip-9/brotli-11), comparing against the previous reachable `v*` tag and failing only when current CLI/FFI/WASM artifact families are absent. Bash hygiene is solid (`set -euo pipefail`, quoting, controlled member extraction via `tar -xO`, soft-missing previous downloads with `::warning::`), and YAML ordering (full-history checkout → download artifacts → report → publish) is right. Remaining issues are robustness: historical extract failures can hard-fail the release gate, curl has no timeouts, and a few small hardening/UX nits.

## Issue counts by severity

- bugs: 2
- suggestions: 2
- nits: 1

iverc:
Addressed the review feedback:\n\n- previous archive listing or extraction failures now emit a warning and leave the comparison unavailable instead of blocking the release\n- release downloads now have bounded connection and total timeouts\n- checkout no longer persists credentials, and public release downloads no longer carry an authorization header\n- zero baselines and sub-MiB deltas are formatted safely and precisely\n\nI kept completeness tied to successful matrix jobs and uploads rather than hard-coding artifact counts, so adding a release variant does not require synchronizing a second fixed-count list.

jackzampolin (approved):
Follow-ups from my earlier review are addressed in `cbf5f31e`:

- Previous-archive listing/extraction failures now warn + fall back to `-` columns instead of hard-failing the release gate (CLI, FFI, and WASM paths).
- Previous-asset downloads bounded with `--connect-timeout 20 --max-time 300`; dropping the Authorization header also removes the cross-host redirect leak CodeRabbit flagged.
- `persist-credentials: false` on the new checkout.
- `format_delta` guards `previous == 0` and picks KiB/MiB by magnitude.

Family-level completeness (vs. matrix-complete counts) stays as-is — fine given every build matrix job is a `needs.*.result == 'success'` gate.

Green across the board. Approving.

Issue #1310, opened 2026-08-05
Fix TLA run-all RED verdict classification under pinned TLC


jackzampolin:
## Problem

`proofs/tla/run-all.sh` does not classify all expected RED models correctly with the TLA+ 1.8.0 command-line jar. A clean archive of commit `63c0be62340bc72a08cdb19e13aa038f20404e42`, using the official v1.8.0 `tla2tools.jar` (SHA-256 `e22f8ffb4bacdea0a871f444dd94fe5fb0d8013b3388ae39e82e26f852c735d5`), finishes with `5/84 MISMATCHED`.

Expected RED rows reported as `ERROR`:

- `M1Naive.cfg`
- `MC_S4_Naive.cfg`
- `MC_Conv_NoHeadRediscovery.cfg`
- `MC_Claim_Split_Eventual.cfg`
- `MC_Replicator_Naive_Red.cfg`

The `MC_PushLogAdmission_Green` and `MC_PushLogAdmission_Red_SuccessOnFull` rows both classify correctly. Direct #1089 checks also behave as intended: RED exits 12 with `Invariant INV_SuccessImpliesRegisteredOrMerged is violated`, and GREEN exits 0 with 93 distinct states.

## Desired outcome

Make the RED/ERROR parser distinguish expected counterexamples from TLC/tool errors under the pinned command-line toolchain, and add a regression check for the five rows above. Also reconcile `proofs/README.md` with `proofs/tla/tools/tlc`: the README says the git-ignored jar is downloaded automatically, but the wrapper currently only references a local jar.

vertexclique:
The cause was one missing phrase in the verdict parser. TLC reports liveness failures as "Temporal property X **was** violated", but `run-all.sh` only grepped for "is violated". All five models listed here fail on a temporal property, so they fell through to ERROR instead of RED. The pattern on `main` now includes `was violated`.

This can be closed right now.

jackzampolin:
Follow-up from #1502: the stage-3 ownership proof pins TLC 1.7.4 because the 1.8.0 prerelease asset was mutable under the same version label/checksum. TLC 1.7.4 has now checked all 131 registered configurations, including LIVE_EventualCurrency and LIVE_EventualReceiverQuiescence for document, collection, and Iroh-relay fairness; all green/red verdicts matched. A future 1.8.x update should use an immutable released asset and explicit checksum update.

PR #1358, opened 2026-08-10
build: add a documented justfile with a no-sudo setup target

vertexclique:
Adds a `justfile` covering the full development surface, and makes
`proofs/tla/tools/tlc` fetch its own jar as `proofs/README.md:78` already claims
it does.

## Onboarding

`just setup` installs everything needed to build, test and verify the database,
including the formal-methods toolchain: Rust with CI's components and the wasm32
target, cbindgen, wasm-pack, a lockfile-matched wasm-bindgen-cli, protoc, Go, a
JDK, elan/lake, and the TLC jar.

It uses no package manager and needs no root. Every system tool is fetched from
its official release into a git-ignored `.tooling/` inside the repo, which the
justfile puts on `PATH`. That makes it identical on Arch, Debian, Fedora, Alpine
and macOS, with x86_64 and aarch64 both handled, and removes the per-distro
package-name matrix that would otherwise need maintaining.

Versions are pinned to what CI already uses: Go 1.25.12 (`ci.yml:449` pins
1.25), Lean v4.18.0 (`proofs/lean/lean-toolchain`), TLC 1.8.0 by checksum
(#1310), protoc 35.1, Temurin 21.

`just doctor` reports what resolved and what is missing, and never fails the
build.

## Targets

Roughly 50, grouped, all with one-line docs in `just --list`:

| Group | Targets |
|---|---|
| setup | `setup`, one per tool, `doctor` |
| build | `build`, `build-release`, `build-fast`, `build-wasm`, `build-ffi-header`, `build-apple` |
| test | `test`, `test-crate`, `test-features`, `integration`, `integration-suite <area>`, `integration-p2p`, `integration-go`, `integration-all`, `test-ffi` |
| proofs | `proofs`, `tla`, `tla-model`, `lean`, `conformance`, `conformance-behavioral` |
| check | `fmt`, `fmt-check`, `lint`, `lint-all-targets`, `lint-wasm`, `doc`, `gate`, `ci` |
| run | `start`, `start-backend`, `cli`, `lens-host`, `otel-smoke`, `pg-compat`, `embedding-server` |
| misc | `bench`, `clean`, `clean-tooling`, `audit`, `machete`, `go-baseline` |

`just lint` reproduces all nine workspace clippy invocations from `ci.yml:66-86`
in the same order, and `just ci` reproduces the pipeline, so a green `just ci`
should predict a green PR.

## TLC jar

`proofs/README.md:78` says the git-ignored jar is re-downloaded when missing.
The wrapper never did that. It now fetches it, pinned by version and SHA-256,
and re-verifies an existing jar so a truncated or swapped file fails loudly
instead of producing a misleading verdict. `--fetch-only` downloads without
running a model, which is what `just setup` calls, so the version and checksum
live in exactly one place.

Two smaller fixes in the same area:

- The java resolver rejected `/usr/bin/java` on every platform, though its
  comment says it is guarding the macOS stub. On Linux that is the normal distro
  path, so a Linux user with only a system JDK got "no usable Java found". The
  rejection is now scoped to Darwin.
- `proofs/tla/.gitignore` covered `states/`, `*.out` and `*.toolbox/` but not
  TLC's error-trace exports, so a full `run-all.sh` left 122 untracked files
  behind. Added `*_TTrace_*.tla` and `*_TTrace_*.bin`.

## Verification

Run against a clean checkout on Linux x86_64:

| Command | Result |
|---|---|
| `just setup-tla` | jar fetched, checksum matches the #1310 pin |
| `just setup-protoc` | `libprotoc 35.1` |
| `just setup-go` | `go1.25.12 linux/amd64` |
| `just setup-jdk` | `openjdk 21.0.12 LTS` |
| `just setup-lean` | elan + Lake 5.0.0 / Lean 4.18.0 |
| `just setup-cargo-tools` | cbindgen, wasm-pack, wasm-bindgen installed |
| `just tla-model MC_DeferredAcp_Green.cfg MC_DeferredAcp_Green.tla` | 377521 states, finished in 3s |
| `just lean` | `Build completed successfully` |
| `just doctor` | every tool resolved |

Every download URL was checked for a 200 before being written into the file, and
the jar's SHA-256 was verified against the value #1310 pins. The checksum guard
was tested by deliberately corrupting the jar: it detected the mismatch,
re-fetched, and restored the pinned hash.

Not verified: macOS and aarch64 paths are written but have only been exercised
on Linux x86_64. The Temurin `Contents/Home` handling in particular is untested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,25 @@ Compatible with Go DefraDB v1.0.0-rc1. Full feature parity across CLI, HTTP API,
 Install [`just`](https://github.com/casey/just), then let it install everything else:
 
 ```bash
-cargo install just    # one time
+cargo install just    # one time, if you already have a Rust toolchain
 just setup            # Rust, protoc, Go, a JDK, Lean/lake, the TLC jar, wasm tooling
 ```
 
+Without Rust yet, install a prebuilt `just` first, since `just setup` is what
+installs the toolchain:
+
+```bash
+curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh \
+  | bash -s -- --to ~/.local/bin      # or grab a binary from the just releases page
+```
+
 `just setup` needs no root and no package manager: every tool is fetched from its
-official release into a git-ignored `.tooling/` inside the repo and put on `PATH`,
-so it behaves identically on any Linux distribution and on macOS, x86_64 or arm64.
-Versions are pinned to the ones CI uses. `just doctor` reports what resolved and
-what is missing.
+official release into a git-ignored `.tooling/` inside the repo and put on `PATH`.
+It is written for Linux and macOS on x86_64 and arm64, and is currently verified
+on Linux x86_64. Downloads are pinned by version and checked against a SHA-256
+before use. The host needs `bash`, `curl`, `tar`, `unzip` and `git`.
+
+`just doctor` reports what resolved and what is missing.
 
 ```bash
 just build             # Build all crates
--- a/README.md
+++ b/README.md
@@ -24,13 +24,30 @@ Compatible with Go DefraDB v1.0.0-rc1. Full feature parity across CLI, HTTP API,
 
 ## Building
 
+Install [`just`](https://github.com/casey/just), then let it install everything else:
+
 ```bash
-cargo build                            # Build all crates
-cargo build --release -p cli           # Release binary
-cargo test                             # Unit tests
-cargo clippy --all -- -D warnings      # Lint
-cargo fmt --all                        # Format
+cargo install just    # one time
+just setup            # Rust, protoc, Go, a JDK, Lean/lake, the TLC jar, wasm tooling
 ```
+
+`just setup` needs no root and no package manager: every tool is fetched from its
+official release into a git-ignored `.tooling/` inside the repo and put on `PATH`,
+so it behaves identically on any Linux distribution and on macOS, x86_64 or arm64.
+Versions are pinned to the ones CI uses. `just doctor` reports what resolved and
+what is missing.
+
+```bash
+just build             # Build all crates
+just build-release     # Release binary
+just test              # Unit tests
+just lint              # Lint (every clippy invocation CI runs)
+just fmt               # Format
+just gate              # fmt + lint + docs + tests, before asking for a review
+just ci                # Reproduce the CI pipeline locally
+```
+
+Run `just` on its own to list every target, grouped.
 
 ## Configuration
 
--- a/justfile
+++ b/justfile
@@ -24,16 +24,29 @@ jdk_root    := tooling / "jdk"
 # TLC 1.8.0 by checksum.
 protoc_version := "35.1"
 go_version     := "1.25.12"
-jdk_version    := "21"
+jdk_release    := "jdk-21.0.12+8"
+jq_version     := "1.8.1"
 tla_version    := "1.8.0"
 tla_sha256     := "e22f8ffb4bacdea0a871f444dd94fe5fb0d8013b3388ae39e82e26f852c735d5"
+
+# protoc publishes no per-asset checksum file, so its hashes are embedded.
+# Go and Temurin publish theirs, and are verified against upstream at install
+# time rather than duplicated here.
+protoc_sha256_linux_x86_64  := "6930ebf62bd4ea607b98fff052596c6ee564b9835b4ce172c75a3f53ae9d91b7"
+protoc_sha256_linux_aarch64 := "01bf9d08808c7f96678b63f4bd8efa559bb4f83d5a7a270d5edaf507f9d5d9cf"
+protoc_sha256_osx_x86_64    := "537d73604a344ded6fc94e98e07e529d4fe3e4a0b09e59905353950fafc2a1f7"
+protoc_sha256_osx_aarch64   := "193289af0470c6a1aada357d4fba0bbf8d78bfaac8b5e42ca30af2ef75583de2"
 
 # .tooling/bin wins over the system copies, so a repo-local protoc/java/go is
 # used even when the distro ships an older one. elan installs lake into its own
 # home rather than .tooling, so that bin directory has to be on PATH too or
 # `just lean` cannot find lake after `just setup`.
-elan_bin := env('ELAN_HOME', home_directory() / ".elan") / "bin"
-export PATH := tooling_bin + ":" + go_root + "/bin" + ":" + elan_bin + ":" + env('PATH')
+# rustup is installed with --no-modify-path, so cargo's bin directory is only
+# on PATH if the user already had it. Without this, every recipe after
+# setup-rust fails to find cargo on a fresh machine.
+cargo_bin := env('CARGO_HOME', home_directory() / ".cargo") / "bin"
+elan_bin  := env('ELAN_HOME', home_directory() / ".elan") / "bin"
+export PATH := tooling_bin + ":" + go_root + "/bin" + ":" + cargo_bin + ":" + elan_bin + ":" + env('PATH')
 
 # Integration areas, each a [[test]] binary in tools/integration-test.
 integration_suites := "basic query acp nac p2p encryption identity backup sourcehub hubrs fts p2p_iroh cursor"
@@ -41,17 +54,30 @@ integration_suites := "basic query acp nac p2p encryption identity backup source
 _default:
     @just --list --unsorted
 
+# sha256 of a file, portable across Linux (sha256sum) and macOS (shasum).
+[private]
+_sha256 file:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if command -v sha256sum >/dev/null 2>&1; then sha256sum "{{ file }}" | awk '{print $1}'
+    elif command -v shasum >/dev/null 2>&1; then shasum -a 256 "{{ file }}" | awk '{print $1}'
+    else echo "error: no sha256 tool found" >&2; exit 1; fi
+
 # ---------------------------------------------------------------------------
 # Setup
 # ---------------------------------------------------------------------------
 
 # Install every dependency needed to develop, test and verify the database.
 [group('setup')]
-setup: setup-rust setup-cargo-tools setup-protoc setup-go setup-jdk setup-lean setup-tla
+setup: setup-rust setup-jq setup-cargo-tools setup-protoc setup-go setup-jdk setup-lean setup-tla
     @echo
     @just doctor
 
 # Rust toolchain, the components CI needs, and the wasm target.
+#
+# Trust boundary: this and setup-lean pipe an upstream install script into sh
+# (rustup.rs, elan-init.sh), which is how both ecosystems distribute. Every
+# other tool here is fetched as an archive and checksum-verified.
 [doc("Rust toolchain, CI's components, and the wasm32 target.")]
 [group('setup')]
 setup-rust:
@@ -69,6 +95,31 @@ setup-rust:
     rustup target add --toolchain stable wasm32-unknown-unknown
     echo "rust: $(rustc --version)"
 
+# jq is required by setup-cargo-tools (lockfile-matched wasm-bindgen) and by
+# setup-jdk (Temurin checksum lookup). Installed rather than assumed, so a host
+# without it does not silently skip those steps.
+[doc("jq, used by the wasm-bindgen and JDK checksum lookups.")]
+[group('setup')]
+setup-jq:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if command -v jq >/dev/null 2>&1; then echo "jq: $(jq --version)"; exit 0; fi
+    case "$(uname -s)" in
+        Linux)  os=linux ;;
+        Darwin) os=macos ;;
+        *) echo "unsupported OS: $(uname -s)" >&2; exit 1 ;;
+    esac
+    case "$(uname -m)" in
+        x86_64|amd64)  arch=amd64 ;;
+        aarch64|arm64) arch=arm64 ;;
+        *) echo "unsupported arch: $(uname -m)" >&2; exit 1 ;;
+    esac
+    mkdir -p "{{ tooling_bin }}"
+    curl -fsSL -o "{{ tooling_bin }}/jq" \
+        "https://github.com/jqlang/jq/releases/download/jq-{{ jq_version }}/jq-${os}-${arch}"
+    chmod 0755 "{{ tooling_bin }}/jq"
+    echo "jq: $({{ tooling_bin }}/jq --version)"
+
 # cbindgen generates the FFI C header; wasm-pack and wasm-bindgen-cli build and
 # test the browser client. wasm-bindgen-cli must match the wasm-bindgen version
 # in Cargo.lock exactly or the test runner refuses the module, so it is derived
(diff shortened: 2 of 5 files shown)


justfile
```diff
@@ -0,0 +1,507 @@
...
+jdk_version    := "21"
+tla_version    := "1.8.0"
+tla_sha256     := "e22f8ffb4bacdea0a871f444dd94fe5fb0d8013b3388ae39e82e26f852c735d5"
+
+# .tooling/bin wins over the system copies, so a repo-local protoc/java/go is
+# used even when the distro ships an older one. elan installs lake into its own
+# home rather than .tooling, so that bin directory has to be on PATH too or
+# `just lean` cannot find lake after `just setup`.
+elan_bin := env('ELAN_HOME', home_directory() / ".elan") / "bin"
+export PATH := tooling_bin + ":" + go_root + "/bin" + ":" + elan_bin + ":" + env('PATH')
```
jackzampolin: **[bug]** `export PATH` includes `.tooling/bin`, Go, and elan, but not `~/.cargo/bin`. `setup-rust` installs rustup with `--no-modify-path` and only `source`s cargo’s env inside that recipe’s subshell. The next recipe (`setup-cargo-tools`) and every later `just` target therefore cannot find `cargo`/`rustc`/`rustup` if the user did not already have them on PATH.

**Suggestion:** Add `cargo_bin := env('CARGO_HOME', home_directory() / ".cargo") / "bin"` to `export PATH` the same way `elan_bin` is handled.

justfile
```diff
@@ -0,0 +1,507 @@
...
+    failed=()
+    for suite in {{ integration_suites }}; do
+        echo "== $suite =="
+        cargo test -p integration-test --test "$suite" || failed+=("$suite")
+    done
+    if [ ${#failed[@]} -gt 0 ]; then
+        echo "FAILED: ${failed[*]}" >&2
+        exit 1
+    fi
+    echo "all ${#failed[@]} listed areas passed"
```
jackzampolin: **[bug]** On full success, `integration-all` prints `all ${#failed[@]} listed areas passed`. When nothing failed, `failed` is empty so `${#failed[@]}` is `0`, producing the false message `all 0 listed areas passed`.

**Suggestion:** Track the suite count separately (e.g. count loop iterations, or use the word count of `integration_suites`) and print that; only use `${#failed[@]}` in the failure branch.

justfile
```diff
@@ -0,0 +1,507 @@
...
+    RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps
+
+# What to make green before asking for a review.
+[group('check')]
+gate: fmt-check lint doc test
+    @echo "gate: green"
+
+# Reproduce the CI pipeline locally, in CI's order.
+[group('check')]
+ci: fmt-check lint lint-wasm test test-features build-wasm integration proofs
```
jackzampolin: **[bug]** `just ci` is documented as reproducing the CI pipeline, but it ends with `proofs` → `proofs/verify-all.sh`, which **skips** the TLA behavioral conformance axis unless `target/release/defra` (or `DEFRA_CONFORMANCE_BINARY`) already exists. CI always builds a binary and runs `tla_conformance`. A local `just ci` can therefore report green while CI’s Conformance job would fail.

**Suggestion:** Make `ci` (or `proofs`) depend on a release/release-fast binary build before behavioral conformance, or have `ci` invoke `conformance-behavioral` explicitly after `build-fast`/`build-release`. Align the profile with CI if the goal is true local repro.

justfile
```diff
@@ -0,0 +1,507 @@
...
+        Linux)  os=linux ;;
+        Darwin) os=mac ;;
+        *) echo "unsupported OS: $(uname -s)" >&2; exit 1 ;;
+    esac
+    case "$(uname -m)" in
+        x86_64|amd64)  arch=x64 ;;
+        aarch64|arm64) arch=aarch64 ;;
+        *) echo "unsupported arch: $(uname -m)" >&2; exit 1 ;;
+    esac
+    url="https://api.adoptium.net/v3/binary/latest/{{ jdk_version }}/ga/${os}/${arch}/jdk/hotspot/normal/eclipse"
```
jackzampolin: **[suggestion]** `setup-jdk` downloads from Adoptium’s `.../binary/latest/21/ga/...` endpoint. That is a floating “latest 21.x” stream with no SHA-256 pin, unlike TLC. README/`just setup` claim versions are pinned like CI; Temurin patch releases can change under you day-to-day with no verification of the tarball contents.

**Suggestion:** Pin a concrete Temurin build (version + full version string or checksum from the Adoptium API package endpoint) and verify the archive before extract, matching the TLC approach.

justfile
```diff
@@ -0,0 +1,507 @@
...
+    tmp="$(mktemp -d)"; trap 'rm -rf "$tmp"' EXIT
+    echo "fetching Temurin JDK {{ jdk_version }} for ${os}-${arch}"
+    curl -fsSL -o "$tmp/jdk.tar.gz" "$url"
+    mkdir -p "$tmp/x" "{{ tooling_bin }}"
+    tar -C "$tmp/x" -xzf "$tmp/jdk.tar.gz"
```
jackzampolin, line 264: **[suggestion]** `setup-protoc` and `setup-go` curl official archives with no checksum verification. Acceptable for many bootstraps, but inconsistent with the careful TLC path and weaker against truncated/CDN mishaps.

**Suggestion:** Embed expected SHA-256 for the four OS/arch combos (or fetch and check go.dev’s checksum file for the exact Go tarball name) and verify before extract.

justfile
```diff
@@ -0,0 +1,507 @@
...
+go_root     := tooling / "go"
+jdk_root    := tooling / "jdk"
+
+# Pinned to what CI uses. ci.yml pins Go 1.25 (:449); Cargo.toml sets
+# rust-version 1.91; proofs/lean/lean-toolchain pins Lean v4.18.0; #1310 pins
+# TLC 1.8.0 by checksum.
+protoc_version := "35.1"
+go_version     := "1.25.12"
+jdk_version    := "21"
+tla_version    := "1.8.0"
```
jackzampolin, line 29: **[suggestion]** `tla_version` and `tla_sha256` are declared in the justfile but never used. `setup-tla` only runs `proofs/tla/tools/tlc --fetch-only`, which uses its own hardcoded defaults. The two pins can drift silently.

**Suggestion:** Either pass them through (`TLA_VERSION=... TLA_SHA256=... proofs/tla/tools/tlc --fetch-only`) so justfile is the single pin, or drop the unused justfile constants and document that the wrapper owns the pin.

justfile
```diff
@@ -0,0 +1,507 @@
...
+# an artifact from this profile.
+[doc("Thin-LTO build for fast iteration (never ship this profile).")]
+[group('build')]
+build-fast:
+    cargo build --profile release-fast -p cli
+
+# Browser client, via wasm-pack, into pkg/wasm (mirrors ci.yml:95).
+[group('build')]
+build-wasm:
+    wasm-pack build crates/wasm --release --target web --out-dir ../../pkg/wasm
```
jackzampolin: **[suggestion]** `build-wasm` runs `wasm-pack build ...` without `CARGO_PROFILE_RELEASE_OPT_LEVEL=z`. CI’s wasm job sets that env. Local `just build-wasm` / `just ci` therefore does not match CI’s wasm artifact profile.

**Suggestion:** Prefix the recipe with `CARGO_PROFILE_RELEASE_OPT_LEVEL=z` to match CI.

justfile
```diff
@@ -0,0 +1,507 @@
...
+# rather than pinned by hand.
+[doc("cbindgen, wasm-pack, and a lockfile-matched wasm-bindgen-cli.")]
+[group('setup')]
+setup-cargo-tools:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    command -v cbindgen  >/dev/null 2>&1 || cargo install cbindgen --locked
+    command -v wasm-pack >/dev/null 2>&1 || cargo install wasm-pack --locked
+    wb_version="$(cargo metadata --format-version 1 --filter-platform wasm32-unknown-unknown 2>/dev/null \
+        | jq -r '.packages[] | select(.name=="wasm-bindgen") | .version' | head -1 || true)"
```
jackzampolin: **[suggestion]** `setup-cargo-tools` shells out to `jq` to read the lockfile-matched `wasm-bindgen` version. `jq` is not installed by `setup`, only reported by `doctor`. Missing `jq` is soft-failed via `|| true`, which silently skips installing `wasm-bindgen-cli` and can leave wasm tests broken with a vague later error.

**Suggestion:** Install a static `jq` into `.tooling/bin`, parse with pure bash/`cargo metadata` + a tiny python one-liner, or fail loudly if `jq` is absent when wasm-bindgen is in the graph.

README.md
```diff
@@ -24,14 +24,31 @@ Compatible with Go DefraDB v1.0.0-rc1. Full feature parity across CLI, HTTP API,
 
 ## Building
 
+Install [`just`](https://github.com/casey/just), then let it install everything else:
+
```
jackzampolin, line 28: **[suggestion]** README states setup “behaves identically on any Linux distribution and on macOS, x86_64 or arm64” and that versions are pinned to CI. Author notes only Linux x86_64 was tested; JDK is not fully pinned; host still needs bash, curl, tar, unzip, etc. Overclaims make onboarding failures look like user error.

**Suggestion:** Soften to “intended to work on Linux/macOS amd64/arm64; verified on …”, list required host utilities, and avoid “pinned to CI” for tools that float or differ from runners.

justfile
```diff
@@ -0,0 +1,507 @@
...
+
+# Rust toolchain, the components CI needs, and the wasm target.
+[doc("Rust toolchain, CI's components, and the wasm32 target.")]
+[group('setup')]
+setup-rust:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if ! command -v rustup >/dev/null 2>&1; then
+        echo "installing rustup (user-space, no sudo)"
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path
```
jackzampolin, line 88: **[suggestion]** `setup-rust` and `setup-lean` pipe remote install scripts into `sh` (`rustup.rs`, `elan-init.sh`) with no script checksum pin. Common and intentional for those ecosystems, but it is the highest-trust step in an otherwise no-sudo local toolchain story.

**Suggestion:** Optionally pin install-script hashes or document the trust boundary next to `setup`.

justfile
```diff
@@ -0,0 +1,507 @@
...
+        aarch64|arm64) arch=aarch_64 ;;
+        *) echo "unsupported arch: $(uname -m)" >&2; exit 1 ;;
+    esac
+    url="https://github.com/protocolbuffers/protobuf/releases/download/v{{ protoc_version }}/protoc-{{ protoc_version }}-${os}-${arch}.zip"
+    tmp="$(mktemp -d)"; trap 'rm -rf "$tmp"' EXIT
+    echo "fetching protoc {{ protoc_version }} for ${os}-${arch}"
+    curl -fsSL -o "$tmp/protoc.zip" "$url"
+    unzip -q "$tmp/protoc.zip" -d "$tmp/out"
+    mkdir -p "{{ tooling_bin }}" "{{ tooling }}/protoc"
+    cp -R "$tmp/out/include" "{{ tooling }}/protoc/"
```
jackzampolin: **[nit]** protoc’s well-known includes are copied to `.tooling/protoc/include`, while the binary is installed at `.tooling/bin/protoc`. Upstream layout expects `prefix/bin/protoc` + `prefix/include/...` so protoc resolves `../include` next to the binary. Current layout would break if a proto ever imported `google/protobuf/*.proto`. Today `orbis.proto` does not, so this is latent.

**Suggestion:** Install as `.tooling/protoc/bin/protoc` + `.tooling/protoc/include` and symlink `bin/protoc` → that binary, or copy includes to `.tooling/include`.

justfile
```diff
@@ -0,0 +1,507 @@
...
+[group('setup')]
+setup-jdk:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if [ -x "{{ jdk_root }}/bin/java" ]; then echo "jdk: already installed"; exit 0; fi
+    case "$(uname -s)" in
+        Linux)  os=linux ;;
+        Darwin) os=mac ;;
+        *) echo "unsupported OS: $(uname -s)" >&2; exit 1 ;;
+    esac
```
jackzampolin, line 245: **[nit]** `setup-protoc` / `setup-go` / `setup-jdk` short-circuit on “already installed” without checking the pinned version. Bumping `protoc_version` / `go_version` in the justfile will not refresh an existing `.tooling` tree until `just clean-tooling`.

**Suggestion:** Compare `protoc --version` / `go version` / `java -version` to the pin and re-fetch on mismatch (or record a version stamp file under `.tooling`).

proofs/tla/tools/tlc
```diff
@@ -1,10 +1,69 @@
...
+#
+# The jar is git-ignored and fetched on first use, pinned by version and
+# SHA-256 so every run checks the same tool CI and #1310 refer to. Use
+# `--fetch-only` to download it without running a model (what `just setup` does).
 set -euo pipefail
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
+TLA_VERSION="${TLA_VERSION:-1.8.0}"
+TLA_SHA256="${TLA_SHA256:-e22f8ffb4bacdea0a871f444dd94fe5fb0d8013b3388ae39e82e26f852c735d5}"
+JAR="$DIR/tla2tools.jar"
```
jackzampolin, line 15: **[nit]** TLC 1.8.0 is a long-lived GitHub **pre-release** that upstream updates in place. Pinning SHA-256 is the right mitigation; worth a one-line comment that the asset under `v1.8.0` is mutable and the checksum—not the tag—is authoritative.

**Suggestion:** Comment that `TLA_SHA256` is the pin because the `v1.8.0` asset can move; when upgrading, update version + checksum together.

proofs/tla/tools/tlc
```diff
@@ -13,12 +72,13 @@ elif [ -x /opt/homebrew/opt/java/bin/java ]; then
   JAVA=/opt/homebrew/opt/java/bin/java
 elif [ -x /usr/local/opt/java/bin/java ]; then
   JAVA=/usr/local/opt/java/bin/java
-elif command -v java >/dev/null 2>&1 && [ "$(command -v java)" != /usr/bin/java ]; then
-  JAVA=java   # a real java on PATH that is not the macOS stub
+elif command -v java >/dev/null 2>&1 &&
```
jackzampolin, line 78: **[nit]** Darwin-only rejection of `/usr/bin/java` is correct and fixes Linux distros that legitimately ship java only as `/usr/bin/java`. Error text still always mentions the macOS stub, which is slightly confusing on Linux when no java is found at all.

**Suggestion:** Branch the error hint: on Darwin mention the stub; elsewhere just say install a JDK / `just setup-jdk`.

jackzampolin (changes requested):
## Summary

Solid onboarding PR: a well-structured justfile that largely mirrors CI’s clippy/test surface, a real TLC jar fetch+SHA-256 path (replacing docs that claimed re-download already worked), and a correct Darwin-only `/usr/bin/java` stub fix. Main gaps are PATH handling after a no-modify-path rustup install (breaks the advertised one-command setup for fresh machines), overstated “pinned like CI / works on all platforms” claims versus floating JDK downloads and untested macOS paths, and a few small correctness nits (`integration-all` success message, unused dual TLC pins, `just ci` skipping behavioral conformance without a release binary).

## Issue counts by severity

- bugs: 3
- suggestions: 7
- nits: 4

**Verdict:** Approve with changes — land after fixing PATH (Issue 1) and the `integration-all` success message (Issue 2); strongly consider making `just ci` actually run behavioral conformance (Issue 3).

PR #1362, opened 2026-08-10
fix(wasm): use web-time so opening a transaction does not panic in the browser

### LINKED ISSUE #1348: WASM: opening a transaction panics on wasm32-unknown-unknown (Instant::now), so every query aborts in the browser
## Summary

Opening a transaction panics on `wasm32-unknown-unknown`, the only wasm target this repo ships. Every GraphQL query opens an implicit transaction, so the browser client aborts on the first query. CI is green because a `panic!` body is valid Rust and nothing executes the wasm tests.

## Reproduction

Confirmed by running under node with `wasm-bindgen-test-runner`. Stack trace, bottom-up:

```
DbTransactionContext<LevelDbStore>::new_with_broadcaster
  <std::time::Instant>::now
    core::panicking::panic_fmt
      rust_begin_unwind -> panic_handler -> rust_panic -> __rust_abort
RuntimeError: unreachable
```

Minimal case (`registry.begin(false)` alone is enough, no query planning or storage I/O involved):

```rust
#![cfg(target_arch = "wasm32")]
use std::sync::Arc;
use db::{DbTransactionRegistry, DB};
use query::txn::TransactionRegistry;
use storage::LevelDbStore;
use wasm_bindgen_test::*;

#[wasm_bindgen_test]
async fn opening_a_transaction_does_not_panic() {
    let store = LevelDbStore::open("p0_instant_repro").unwrap();
    let db = Arc::new(DB::new(store).unwrap());
    let registry = Arc::new(DbTransactionRegistry::new(Arc::clone(&db)));
    let handle = registry.begin(false).await.unwrap();   // panics here
    registry.rollback(&handle).await.unwrap();
}
```

To run it: install `wasm-bindgen-cli` at the version the lockfile pins (currently `0.2.114` -- the CLI and crate schema versions must match exactly), add a runner to `.cargo/config.toml`:

```toml
[target.wasm32-unknown-unknown]
runner = "wasm-bindgen-test-runner"
```

then `cargo test -p defra-wasm --target wasm32-unknown-unknown`.

## Root cause

`crates/db/src/txn_context.rs:47`:

```rust
let now = Instant::now();
```

The file contains no `cfg` attributes; the module is ungated at `crates/db/src/lib.rs:98`; `db` is a direct dependency of the wasm crate (`crates/wasm/Cargo.toml:56`).

`std` dispatch: `sys/time/mod.rs`'s `cfg_select!` has no `wasm32` arm, so `wasm32-unknown-unknown` falls through to `sys/time/unsupported.rs`, where `Instant::now()` is `panic!("time not implemented on this platform")`. Note `target_os = "wasi"` IS handled (it maps to the unix impl), so `wasm32-wasip1` does not exhibit this -- but this repo does not build wasip1.

## Reachability

Every hop confirmed ungated:

```
crates/query/src/runner/executor.rs:134   registry.begin_implicit_read()   <- every ParsedOperation::Query
  -> crates/db/src/txn_registry.rs:1385   begin_implicit_read
  -> crates/db/src/txn_registry.rs:1310   begin_with_options
  -> new_with_broadcaster
  -> crates/db/src/txn_context.rs:47      Instant::now()
```

The only `cfg`s in `executor.rs` (`:20`, `:39`) cover `await_with_timeout`, not the transaction path.

## Other ungated sites of the same class

Confirmed ungated and reachable:
- `crates/db/src/txn_context.rs:70,74` (`touch()`, called per request on an existing transaction)
- `crates/db/src/txn_registry.rs:270,292,312,321` (idle sweep)

Ungated, reachability not traced end to end:
- `crates/query/src/runner/mutation.rs:655,735`
- `crates/query/src/runner/query/nested.rs:58,64,85,88,95,104,110,134,140,149,160,167`
- `crates/query/src/runner/query/nested_fulltext.rs:439,618,636`
- `crates/storage/src/stores/retry_info.rs:135,151,168` (`SystemTime::now()`)

Not reachable (module-gated, listed so they are not re-investigated): `crates/db/src/view_ops.rs:184`, and all `storage/src/backends/{memory,lark,fjall,redb,rocksdb}` sites.

`crates/storage/src/backends/leveldb/*` and `opfs_env.rs` are clean; `Env::micros()` correctly uses `js_sys::Date::now()`.

## Acceptance criteria

- [ ] Opening a transaction and running a query succeed on `wasm32-unknown-unknown`.
- [ ] Native codegen unchanged, proven by a machine-code diff rather than by inspection.
- [ ] A wasm test covering the query path runs in CI (see the companion CI issue).
- [ ] The remaining ungated `Instant`/`SystemTime` sites above are triaged, each either fixed or recorded as unreachable with the reason.

vertexclique:
Fixes #1348.

Opening a transaction in the browser panics:

```
panicked at library/std/src/sys/time/unsupported.rs:13:9: time not implemented on this platform
  <std::time::Instant>::now
  DbTransactionContext::new_with_broadcaster
  DbTransactionRegistry::begin
  DefraClient::mutate
```

`std::time` is unimplemented on `wasm32-unknown-unknown`, so every mutation and
every query that opens a transaction aborts.

## Fix

Swap `std::time::Instant`/`SystemTime` for `web_time` equivalents in the crates
that compile to wasm. **`web_time` re-exports `std::time` verbatim on every
native target** — I type-checked that a `web_time::Instant` assigns directly to
a `std::time::Instant` on native — so server and FFI builds are unchanged.

8 files, not the 49 call sites a naive grep suggests: every non-leveldb storage
backend is `#[cfg(not(target_arch = "wasm32"))]` and never compiles for wasm.

Also widens the CI `pull_request` base filter from `[main]` to `['**']`. The
filter matches the PR's *base*, so a stacked PR runs no jobs at all and shows an
empty, green-looking check list. This is the bottom of a 3-PR stack, so the
layers above it need this to be tested at all.

## Verification

Run against a real headless Firefox: the wasm suite goes from **46 passed / 5
failed** to **50 passed / 1 failed**. All four `Instant::now()` failures are
gone; the one remaining failure is a different bug, fixed in #1363.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,10 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    # Every base branch, not just main. The filter matches the PR's *base*, so
+    # limiting it to main means a stacked PR (one targeting the branch below it)
+    # runs no jobs at all and shows an empty, green-looking check list.
+    branches: ['**']
 
 # Only the tip of each ref keeps a run. Superseded PR pushes and main merge
 # trains cancel prior work so intermediate SHAs cannot stack the self-hosted
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3435,6 +3435,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tracing",
+ "web-time",
  "zeroize",
 ]
 
@@ -9626,6 +9627,7 @@ dependencies = [
  "storage",
  "tokio",
  "tracing",
+ "web-time",
 ]
 
 [[package]]
@@ -9665,6 +9667,7 @@ dependencies = [
  "storage",
  "tokio",
  "tracing",
+ "web-time",
 ]
 
 [[package]]
@@ -11660,6 +11663,7 @@ dependencies = [
  "uuid",
  "wasm-bindgen",
  "wasm-bindgen-futures",
+ "web-time",
  "zeroize",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,6 +87,7 @@ serde_json = "1.0"
 serde_cbor = "0.11"
 serde_bytes = "0.11"
 ciborium = "0.2"
+web-time = "1.1"
 
 # P2P networking
 # Keep these on the same libp2p-swarm release line.
--- a/crates/db/Cargo.toml
+++ b/crates/db/Cargo.toml
@@ -18,6 +18,7 @@ fjall = ["storage/fjall"]
 wasmtime-runtime = ["lens/wasmtime-runtime"]
 
 [dependencies]
+web-time.workspace = true
 # Async runtime
 async-trait.workspace = true
 async-lock.workspace = true
--- a/crates/db/src/txn_context.rs
+++ b/crates/db/src/txn_context.rs
@@ -6,8 +6,9 @@ use query::runner::DocFetcher;
 use query::txn::{DeferredAcpMutations, TransactionContext};
 use std::collections::BTreeMap;
 use std::sync::{Arc, Mutex};
-use std::time::{Duration, Instant};
+use std::time::Duration;
 use storage::corekv::Store;
+use web_time::Instant;
 
 use crate::collection_provider::TxnCollectionProvider;
 use crate::database::DB;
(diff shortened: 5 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -4,7 +4,10 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    # Every base branch, not just main. The filter matches the PR's *base*, so
+    # limiting it to main means a stacked PR (one targeting the branch below it)
+    # runs no jobs at all and shows an empty, green-looking check list.
+    branches: ['**']
```
vertexclique, line 10: this is for stacked prs

PR #1366, opened 2026-08-10
fix(document): unify NormalValue::Bytes JSON as lowercase hex

### LINKED ISSUE #1294: normal_value_to_json encodes the same value three different ways
## Summary

Three separate copies of `normal_value_to_json` disagree about how binary values are rendered: base64 in one, lowercase hex in another. Which representation a client receives depends on which layer produced the response, not on anything the client can control or predict.

## Context

**The three copies.**

| copy | encoding | location |
|---|---|---|
| 1 | base64 | `crates/document/src/encoding.rs:337` (`base64_encode`) |
| 2 | lowercase hex | `crates/query-types/src/json_convert.rs:268` (`write!("{:02x}")`) |
| 3 | see PR #1261 | `crates/query-plan/src/plan/mutation/create_conversions.rs` |

Copy 3 is currently being modified by open PR #1261.

**Why this is a correctness problem and not a style one.** These are wire-format outputs. A client that reads a binary field through one code path and writes it back through another is not round-tripping the same value. Any consumer that persists, compares or hashes the rendered form will see two different values for identical data, and there is no field in the response that says which encoding was used.

**The divergence is actively moving.** Open PR #1261 separately flips signature encoding from base64 to hex at `db/src/commits_fetcher/conversion.rs:194`. Independent decisions about representation are being made per site, which is the expected outcome when three implementations exist and none is canonical.

**Suggested direction.** Collapse to one shared implementation, then make an explicit decision about which encoding is the wire contract and document it. The choice matters less than having exactly one. If both encodings must be supported for compatibility, the format should be explicit in the response rather than implied by the code path.

**Open question for triage.** Whether either encoding is already load-bearing for an external consumer, which would determine whether unification is a fix or a breaking change.

edjroz:
## Summary

Closes #1294.

Three independent `normal_value_to_json` paths disagreed on how `NormalValue::Bytes` is rendered (base64 / lowercase hex / JSON array of numbers). This PR collapses them to **one shared helper** that emits **lowercase hex**, matching the query path and Go’s GraphQL Blob scalar when values are raw bytes.

### Context (not a “Blob create vs query is broken” story)

Production GraphQL Blob fields are usually stored as hex **strings**, so create and query already agreed on the happy path (L2 spike). The bug is still real for **`NormalValue::Bytes`**: mutation create returned a number array while query returned hex when Bytes was present.

### What changed

1. **Red (test):** e2e with harness env `DEFRA_TEST_BLOB_AS_BYTES=1` forces Blob inputs into `NormalValue::Bytes`, then asserts create + query both return `"00ff"`.
2. **Green (fix):** `document::encoding::bytes_to_json` is the canonical encoder; query-types and mutation create_conversions call it. Document’s old base64 and mutation’s array path are gone.

### Out of scope (documented for later)

- Commit `delta` / `signature.value` remain separate contracts (base64 / other).
- CBOR→JSON byte encoding in `encoding_cbor` unchanged.
- Long-term suite widening: natural Bytes product paths (lens, P2P `to_map`) without a test hook — noted in the e2e test comments.

## Test plan

- [x] `cargo test -p integration-test --test basic -- issue1294_bytes_json` (failed red, passes green)
- [x] `cargo test -p document -p query-types -p query-plan --lib`
- [x] CI full suite

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -200,6 +200,7 @@ Single-purpose binaries, each its own `[[test]]` with no submodules:
 | `--test issue1154_repro` | at-scale merge of every success-acked document after hub restart |
 | `--test issue1194_repro` | concurrent updates to distinct documents must not conflict |
 | `--test issue1211_repro` | same, through a branchable collection's head set |
+| `--test issue1294_bytes_json` | Blob-as-Bytes create+query must return lowercase hex |
 
 ### Rust Commands
 
--- a/crates/document/src/encoding.rs
+++ b/crates/document/src/encoding.rs
@@ -409,6 +409,46 @@ pub fn normal_value_to_json(value: &NormalValue) -> Result<serde_json::Value> {
     }
 }
 
+/// Env var for the #1294 e2e harness: store Blob inputs as hex-decoded
+/// [`NormalValue::Bytes`] so create/query response converters are exercised.
+///
+/// Production paths leave this unset; Blob fields remain hex **strings**.
+pub const TEST_BLOB_AS_BYTES_ENV: &str = "DEFRA_TEST_BLOB_AS_BYTES";
+
+/// True when [`TEST_BLOB_AS_BYTES_ENV`] is `1` or `true`.
+pub fn test_blob_as_bytes_enabled() -> bool {
+    matches!(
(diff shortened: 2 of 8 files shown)


vertexclique:
This is good. Let me merge cbor and then let's merge this and watch master.

vertexclique (approved):

PR #1371, opened 2026-08-10
fix(storage): HeadstoreDocKey::parse — harden binary short-id decoding

### LINKED ISSUE #1293: Key parsers split binary keys on '/' after a lossy UTF-8 conversion
## Summary

Around eleven call sites parse storage keys by running `String::from_utf8_lossy` and then splitting on `/`. Storage keys embed uvarint and CID bytes, and the uvarint encoding of 47 is `0x2F`, the ASCII `/`, so a key can be silently split in the wrong place.

## Context

**The counter-example, rather than a list of sites.** The list is the least useful part of this report because it is growing faster than it can be fixed, so the failure mode is stated first:

A key component whose uvarint value is 47 serialises to the single byte `0x2F`. After `from_utf8_lossy` that byte is an ordinary `/` character, indistinguishable from a real separator, and `split('/')` produces one extra segment. Downstream code then indexes the wrong field. Separately, `from_utf8_lossy` replaces any invalid UTF-8 sequence with U+FFFD, so once the conversion has happened the original bytes cannot be recovered even if the caller notices something is wrong. CID bytes make invalid sequences common.

**Nothing errors.** Both the false separator and the replacement character are silent, so a mis-parse surfaces later as a missing or wrong record rather than as a parse failure. That makes it hard to attribute when it does happen and is why we suspect it has not been reported despite being long-standing.

**Known sites** include `db-merge/src/head_provider.rs:252`, `db-blocks/src/collection.rs:45`, and `db/src/commits_fetcher/mod.rs:357`, among others.

**The class is actively growing.** Open PR #1261 adds another instance at `db/src/txn_registry.rs`: `String::from_utf8_lossy(&pair.key).rsplit('/').next()` over a `CollectionVersionKey` prefix scan. This is not a criticism of that PR. The pattern reads as idiomatic in a codebase where it already appears a dozen times. It is the argument for fixing the pattern rather than the instances, since new sites are being added in good faith faster than a site-by-site cleanup would close them.

**Suggested direction.** Parse over `&[u8]` using each key type's own decoder rather than converting to a string and splitting. #675 ("typed key structs instead of string concatenation") is the closest structural precedent in the tree and points the same way. A lint or a deliberately awkward helper name would stop new sites appearing while existing ones are migrated.

**Severity.** Latent correctness, not performance. We have not observed a production mis-parse; we are reporting reachability from the encoding, not from an incident.

edjroz:
## Summary

Hardening slice for #1293 (not a live consumer bugfix).

- Add `HeadstoreDocKey::parse(&[u8])` that decodes the binary `doc_short_id` uvarint via `decode_doc_short_id_prefix` (no `from_utf8_lossy` + `/` split on that segment), modelled on `ActionStatusKey::parse`.
- Switch `commit_priority_index` backfill to the typed parser.
- Encoding-level regression test: `doc_short_id = 47` serialises raw `0x2F` (`/`), so naive forward `split('/')` after lossy UTF-8 sees a different segment count than 46/48 — and typed parse still recovers 47.

Discovery on current `main` found **8** lossy-split product sites; **none mis-parse today** (rsplit/last or decimal-text keys). This PR converts the highest-reachability `HeadstoreDocKey` path into a guarded decoder and documents the hazard so a future fixed-index parse cannot land silently. Remaining sites + a pattern lint stay as follow-ups.

## Test plan

- [x] `cargo test -p storage --lib keys::headstore` (incl. roundtrip, reject, 47 false-separator)
- [x] `cargo test -p db --lib commit_priority_index` (4 tests)
- [x] `cargo clippy -p storage -p db --all-targets -- -D warnings`

Fixes #1293

### DIFF
--- a/crates/storage/src/keys/headstore.rs
+++ b/crates/storage/src/keys/headstore.rs
@@ -56,7 +56,10 @@ impl HeadstoreDocKey {
         let rest = bytes.strip_prefix(b"/d/")?;
         let (rest, doc_short_id) = decode_doc_short_id_prefix(rest).ok()?;
         let rest = rest.strip_prefix(b"/")?;
-        let text = std::str::from_utf8(rest).ok()?;
+        // SAFETY: `Key::bytes` always writes `field_id` and the CID as UTF-8
+        // (ASCII field names + base32 CID). After the binary short-id segment,
+        // the remainder is only those text fields.
+        let text = unsafe { std::str::from_utf8_unchecked(rest) };
         let (field_id, cid_str) = text.rsplit_once('/')?;
         if field_id.is_empty() || cid_str.is_empty() {
             return None;
--- a/crates/storage/src/keys/headstore.rs
+++ b/crates/storage/src/keys/headstore.rs
@@ -1,6 +1,7 @@
-use super::doc_id_index::encode_doc_short_id;
+use super::doc_id_index::{decode_doc_short_id_prefix, encode_doc_short_id};
 use crate::corekv::Key;
 use cid::Cid;
+use std::str::FromStr;
 
 const PRIORITY_HEX_WIDTH: usize = 16;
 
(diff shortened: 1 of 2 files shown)


crates/storage/src/keys/headstore.rs
```diff
@@ -44,6 +45,29 @@ impl HeadstoreDocKey {
...
+    ///
+    /// Decodes the binary `doc_short_id` uvarint without a lossy UTF-8 conversion,
+    /// so a short ID whose encoding is `0x2F` (`/`) cannot create a false separator.
+    /// Modelled on [`super::systemstore::ActionStatusKey::parse`]: fail-closed,
+    /// no trailing garbage.
+    pub fn parse(bytes: &[u8]) -> Option<Self> {
+        let rest = bytes.strip_prefix(b"/d/")?;
+        let (rest, doc_short_id) = decode_doc_short_id_prefix(rest).ok()?;
+        let rest = rest.strip_prefix(b"/")?;
+        let text = std::str::from_utf8(rest).ok()?;
```
vertexclique: TIOLI: this can be `from_utf8_unchecked` for us to be faster.

vertexclique (approved):

crates/storage/src/keys/headstore.rs
edjroz: I'll take it, faster is better in this context.

crates/storage/src/keys/headstore.rs
```diff
@@ -44,6 +45,32 @@ impl HeadstoreDocKey {
...
+    /// Modelled on [`super::systemstore::ActionStatusKey::parse`]: fail-closed,
+    /// no trailing garbage.
+    pub fn parse(bytes: &[u8]) -> Option<Self> {
+        let rest = bytes.strip_prefix(b"/d/")?;
+        let (rest, doc_short_id) = decode_doc_short_id_prefix(rest).ok()?;
+        let rest = rest.strip_prefix(b"/")?;
+        // SAFETY: `Key::bytes` always writes `field_id` and the CID as UTF-8
+        // (ASCII field names + base32 CID). After the binary short-id segment,
+        // the remainder is only those text fields.
+        let text = unsafe { std::str::from_utf8_unchecked(rest) };
```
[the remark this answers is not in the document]
vertexclique: Sure, but what is pushed down is already inside `String` with utf8 absorbed. So this is not correct in normal circumstances.

PR #1375, opened 2026-08-11
perf(wasm): remove Tokio from dependency graph

### LINKED ISSUE #1334: Remove tokio from the wasm graph
## Context

tokio is declared unconditionally in three crates that end up in the browser blob:

- `crates/query/Cargo.toml:25`
- `crates/query-plan/Cargo.toml:27`
- `crates/kms/Cargo.toml:46`

`storage` and `datastore` already cfg-gate this correctly and can be used as the reference. This is not just a cfg gate: the call sites need async-lock substitutes.

Est. 150-400 KiB wasm raw `[hypothesis]`.

## Acceptance criteria

- [ ] tokio absent from the wasm dependency graph.
- [ ] Native artifacts unchanged.
- [ ] wasm integration/browser smoke passes; size delta recorded.

---

Part of #1327. Effort: M. Blocked by: nothing.

iverc:
Closes #1334.

## Problem

Tokio was declared unconditionally by query, query-plan, and KMS, pulling it into the browser WASM dependency graph. Removing the dependency also required runtime-neutral replacements for task-local query context and KMS reply channels.

## What changed

- Restrict Tokio dependencies in query, query-plan, and KMS to native targets.
- Add poll-scoped WASM task-local context that preserves the active value across nested and interleaved futures.
- Use async-channel for KMS reply streams on WASM while preserving the existing Tokio channel behavior on native targets.
- Route query, ACP, and P2P KMS call sites through the target-specific abstractions.
- Add regression coverage for interleaved WASM task-local scopes.

## Validation

- [x] `cargo test -p defra-core -p kms -p query-plan -p query` (482 passed)
- [x] `cargo test -p p2p kms` (8 passed)
- [x] `cargo tree -p defra-wasm --target wasm32-unknown-unknown -i tokio` (no Tokio dependency)
- [x] Optimized WASM size: 7,096,189 -> 7,091,332 bytes (-4,857 bytes)
- [x] `git diff --check`

### DIFF
--- a/crates/defra-core/src/wasm_task_local.rs
+++ b/crates/defra-core/src/wasm_task_local.rs
@@ -1,0 +1,93 @@
+//! Single-threaded task-local storage for WASM futures.
+
+use std::cell::RefCell;
+use std::future::{poll_fn, Future};
+use std::thread::LocalKey;
+
+fn with_value<T: 'static, R>(
+    local: &'static LocalKey<RefCell<Option<T>>>,
+    value: &mut Option<T>,
+    f: impl FnOnce() -> R,
+) -> R {
+    struct Reset<'a, T: 'static> {
+        local: &'static LocalKey<RefCell<Option<T>>>,
+        value: &'a mut Option<T>,
+    }
+
+    impl<T: 'static> Drop for Reset<'_, T> {
+        fn drop(&mut self) {
+            self.local
+                .with(|local| std::mem::swap(self.value, &mut *local.borrow_mut()));
+        }
+    }
+
+    local.with(|local| std::mem::swap(value, &mut *local.borrow_mut()));
+    let reset = Reset { local, value };
+    let result = f();
+    drop(reset);
+    result
+}
+
+/// Poll a future with `value` installed in its task-local slot.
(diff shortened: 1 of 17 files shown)


edjroz (approved):
Approving. The 7,096,189 figure looks like a pre-`wasm-opt` stage. The shipped delta is slightly larger than the -4,857 reported. this yields about 0.11%

crates/defra-core/src/wasm_task_local.rs
```diff
@@ -0,0 +1,93 @@
...
+    poll_fn(move |cx| with_value(local, &mut value, || future.as_mut().poll(cx))).await
+}
+
+/// Access the current task-local value when a scope is active.
+pub fn try_with<T: 'static, R>(
+    local: &'static LocalKey<RefCell<Option<T>>>,
+    f: impl FnOnce(&T) -> R,
+) -> Option<R> {
+    local.with(|local| local.borrow().as_ref().map(f))
+}
```
vertexclique, line 51: This will have reentrancy problems with nested `scope` calls above. I guess the best prevention will be rewriting it like how `with_value` works.

Issue #1332, opened 2026-08-07
Un-inherit tokio "full"


edjroz:
## Context

`crates/storage/Cargo.toml:73` requests tokio with the `full` feature. Cargo feature unification then propagates it across the workspace, defeating the narrow tokio feature set that `crates/ffi` asks for.

Mostly a build-time and hygiene fix. Expected size impact is under 1% `[hypothesis]`.

## Acceptance criteria

- [ ] No crate inherits tokio `full`; each names the feature set it actually uses.
- [ ] Build and tests green.
- [ ] `cargo tree -e features` shows the narrow set in the ffi graph.

---

Part of #1327. Effort: S. Blocked by: nothing.

Issue #1327, opened 2026-08-07
epic: binary size reduction


edjroz:
## Context

A size analysis (baseline, measure, decide) was run against `30fb6525` on macOS arm64, all numbers from the true `release` profile with fat LTO. This epic tracks the work that came out of it.

The headline: crate counts and rlib sizes wildly overstate impact. Fat LTO already dead-strips the axum server and most of the sourcehub/commonware block from the FFI dylib, so `release-full`'s ~190 extra crates cost only +3.5 MiB. Only binary-level deltas count as evidence here.

### Measured baselines

| Artifact | Size |
|---|---|
| `defra` (default features) | 34.8 MiB |
| `defra` (release-full) | 38.3 MiB |
| `libffi.dylib` (shipped: redb + iroh) | 32.2 MiB |
| `libffi.dylib` (same, without iroh) | 25.6 MiB |
| wasm, shipped pipeline (opt-z + wasm-bindgen + wasm-opt -Oz) | 5.03 MiB raw / 1.92 MiB gzip |

Isolated deltas: iroh is +6.9 MiB, 20.4% of the shipped dylib and the single largest measured lever. The rocksdb block in FFI is ~5.8 MiB. The full sourcehub + orbis + pg-compat block is only +3.5 MiB.

The 434-497 MB `libffi.a` is **not** a target. The Go harness links the dylib (`tools/ffi-test/src/builder.rs:112`) and ld64 dead-strips the XCFramework `.a` at app link. See #1340 for measuring what a Swift consumer actually pays.

### Honest ceilings

Stated up front so no issue here oversells itself: wasm 8-15% without capability loss (20-30% with cuts), FFI 3-6% without capability loss (20% from iroh, ~35% for a mobile-lean variant), CLI 2-5% (10-20% only via the opt-level experiment). 50% is not achievable anywhere without giving up features.

## Child issues

Tier order is dependency order. Tier 0 makes every later claim verifiable, Tier 1 needs no decisions, Tier 2 is the decision issues, Tier 3 is gated on them, Tier 4 is measure-first experiments.

### Tier 0: foundation

| # | Issue |
|---|---|
| #1328 | CI: track binary size per release variant |
| #1329 | CI: add cargo-machete and delete the two known dead deps |

### Tier 1: no-decision wins

| # | Issue |
|---|---|
| #1330 | Replace `multicodec 0.1` with inline constants |
| #1331 | Make telemetry optional so wasm drops tracing-subscriber |
[...]

PR #1380, opened 2026-08-11
build(deps): declare the tokio features each crate uses

edjroz:
## Summary

Cargo workspace dependency inheritance is **additive** — a member's
`features = [...]` are unioned with the workspace entry's, never subtracted. The
workspace entry requests `full` (`Cargo.toml:79`), so every member writing
`tokio.workspace = true` silently receives `full`, including the ~20 that name a
narrow set and look correct. `cargo tree -p defra-core -e features -i tokio`
attributes `tokio feature "full"` to `defra-core`, which declares only `["rt"]`.

The result is that the manifests lie: crates compile only because a sibling
turns on a feature they use but never declare.

- Declare the real tokio feature set on all 21 inheriting lines across 19
  manifests; **the workspace entry is deliberately left at `full`**.
- Add the four missing tokio dev-dependencies: `defra-core`, `orbis`,
  `sourcehub` (which had no `[dev-dependencies]` section at all), and `storage`.
- 14 crates had a genuine declared-vs-used gap, each verified against source —
  e.g. `db` uses `tokio::fs` (`migration/mod.rs:66`) while declaring only
  `sync`; `p2p` uses `tokio::time::Instant`
  (`sync/push_fanout_coalescer.rs:70`) without declaring `time`;
  `defra-core`'s `#[tokio::test]` (`current_identity.rs:139`) compiled with no
  tokio dev-dependency at all.

**This is 1 of 3.** 1332-b drops `features = ["full"]` from `Cargo.toml:79`;
1332-c removes the four dev/test-only explicit `full`s. Splitting this way is
what makes the diff reviewable — see below.

### Why this is provably behaviour-neutral

Because inheritance is additive and the workspace entry still says `full`,
adding features to a member **cannot** change the resolved graph. The resolved
tokio feature set is byte-identical before and after for **all 45 workspace
packages**, under both default features and `--all-features`. Normalising the
full `cargo tree -e features` output shows no edge ever leaves any graph; every
difference is an added feature edge.

Reviewers can accept 19 files on that mechanical evidence rather than by reading
each line. In 1332-b, by contrast, any remaining gap surfaces as a compile error
against a baseline where the manifests are already believed correct.

### Out of scope

- `Cargo.toml:79` (1332-b) and the four explicit dev/test `full`s (1332-c).
- The already-narrow non-inherited lines owned by #1334 / #1375
  (`kms`, `query-plan`, `query` normal deps). This PR touches only their **dev**
  lines, which #1375 left inheriting.

## Test plan

- [x] `cargo tree -e features` tokio feature set identical before/after for all
      45 packages, default and `--all-features` (empty diff)
- [x] normalised full `cargo tree -e features`: no edge removed from any graph
- [x] per-crate `cargo check -p <pkg> --all-targets` for all 45 packages (44 ok;
      `defra-wasm` fails identically on the base commit — see below)
- [x] `cargo clippy -p cli --no-default-features --features release-full --all-targets -- -D warnings`
- [x] `cargo hack check --each-feature -p db -p lens` (fails identically on the
      base commit — see below)

### Pre-existing failures, confirmed not introduced here

`cargo check -p defra-wasm` and `cargo hack --each-feature` on `db` / `lens`
fail on `--no-default-features`, **identically on the base commit**.

Root cause, unrelated to tokio *features*: `crates/db` and `crates/lens` declare
tokio `optional = true` while their source uses `tokio::` unconditionally, so
any build without `db/native` or `lens/wasmtime-runtime` fails with
`E0433: cannot find module or crate tokio`. Worth its own issue; it is a
dependency-optionality bug, not a feature bug.

Part of #1327. Refs #1332.

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -51,7 +51,7 @@ telemetry = { path = "../telemetry", default-features = false }
 clap = { version = "4.5", features = ["derive", "env"] }
 
 # Async runtime
-tokio = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "fs", "signal", "sync", "time"] }
 futures = { workspace = true }
 async-trait = { workspace = true }
 
--- a/crates/crdt/Cargo.toml
+++ b/crates/crdt/Cargo.toml
@@ -28,7 +28,7 @@ tracing.workspace = true
 [dev-dependencies]
 criterion.workspace = true
 proptest.workspace = true
-tokio = { workspace = true, features = ["test-util"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "test-util"] }
 
 [[bench]]
 name = "merge"
--- a/crates/datastore/Cargo.toml
+++ b/crates/datastore/Cargo.toml
@@ -34,7 +34,7 @@ tokio = { workspace = true, features = ["rt"] }
 [dev-dependencies]
 # Testing
 futures.workspace = true
-tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "sync", "time"] }
 
 [lints]
 workspace = true
--- a/crates/db-merge/Cargo.toml
+++ b/crates/db-merge/Cargo.toml
@@ -66,7 +66,7 @@ events = { path = "../events", default-features = false }
 kms = { path = "../kms" }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio = { workspace = true, features = ["sync"] }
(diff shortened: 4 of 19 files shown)


jackzampolin (approved):
This one is provably a no-op for the resolved build graph: the workspace entry still requests `full`, and Cargo unions member features with it, so every feature added here was already present. `test-util` (the one feature outside `full`) was already declared where it's used. That makes the risk here zero and the value entirely in the manifests telling the truth — which is what makes #1387 safe to land on top.

The `Integration (go-compat)` failure was infrastructure, not code: the annotation is `The self-hosted runner lost communication with the server`, every step has a null conclusion, and adding feature declarations cannot affect go-compat. Re-queued; the rest of the pipeline is green.

Approving.

Issue #1336, opened 2026-08-07
Remove openssl from cli builds


edjroz:
## Context

openssl reaches cli builds through josekit in keyring (`crates/keyring/Cargo.toml:17`) plus reqwest defaults. CI vendors it statically only on darwin_amd64 (`release.yml:34,55,76`), so that variant pays the most.

Est. 2-3 MiB on darwin_amd64 `[hypothesis]`, minor elsewhere where it links the system dylib. FFI has no openssl at all.

## Path

- Pure-Rust JWE in keyring, replacing josekit.
- `default-features = false, features = ["rustls-tls"]` on the five reqwest declarations.

## Risk

JWE format round-trip against existing keyrings, and TLS behavior differences. Both need explicit coverage.

## Acceptance criteria

- [ ] A keyring created with the old code opens with the new code (round-trip test).
- [ ] openssl absent from all cli graphs.
- [ ] vendored-openssl steps deleted from `release.yml`.
- [ ] Size delta recorded per cli variant.

---

Part of #1327. Effort: M. Blocked by: nothing.

jackzampolin:
**Triage: `Post-1.0`.** Procedure: #1391.

`Go Status: Not in Go`

Confirmed on main: crates/keyring/Cargo.toml:17 still declares `josekit = "0.8"`, so the issue is unresolved on the working tree. PR #1381 (open, replaces josekit with a pure-Rust JWE) and PR #1382 (open, stacked, switches reqwest to rustls) together close this. Explicitly a binary-size/dependency-surface reduction with its own round-trip test requirement to guard correctness; no defect exists today.

<sub>Go reference: Go's TLS/crypto stack uses Go stdlib crypto plus its own keyring/ package, not josekit/openssl; no comparable openssl-dependency-removal concern exists in the Go tree.</sub>

PR #1381, opened 2026-08-11
perf(keyring): replace josekit with a pure-Rust JWE, dropping OpenSSL from macOS cli builds

edjroz:
Closes part of #1336. Part of #1327.

## Problem

`josekit` was in `crates/keyring` for exactly two operations — PBES2-HS512+A256KW key wrapping and A256GCM content encryption — and dragged OpenSSL into every cli graph to do it. On `x86_64-apple-darwin`, CI compensates with the `vendored-openssl` feature, which statically links OpenSSL into the shipped binary.

## What changed

`crates/keyring/src/jwe.rs` implements the format directly against RFC 7518 §4.8 (PBES2 key encryption) and §5.3 (AES-GCM content encryption), over RustCrypto crates. `file.rs` drops from ~30 lines of josekit plumbing to two delegating calls.

New dependencies are chosen to avoid forking versions already in the lock:

| Crate | Requires | Locked | Note |
|---|---|---|---|
| `pbkdf2 0.12` | `hmac ^0.12`, `sha2 ^0.10` | 0.12.1 / 0.10.9 | not 0.13, which wants `sha2 ^0.11` |
| `aes-kw 0.2` | `aes ^0.8.1` | 0.8.4 | not 0.3, which wants `aes ^0.9` and would split `aes` |

Net: −josekit −openssl −openssl-sys −flate2, +pbkdf2 +aes-kw.

## Header validation

josekit enforced several header rules that a direct reimplementation drops silently. `decrypt` rejects the following before deriving any key material:

| Rule | Why |
|---|---|
| `zip` present | RFC 7516 §4.1.3. Go jwx and josekit both write and inflate it. Ignoring it means the AEAD tag verifies and the caller receives a **DEFLATE stream as key material** rather than an error |
| `crit` present | RFC 7516 §4.1.13 requires rejecting extensions the recipient does not implement. We implement none |
| `p2s` under 8 octets | RFC 7518 §4.8.1.1. An empty salt makes the derived key a function of the password alone, precomputable across every such token |
| `p2c` outside `1..=10_000_000` | A hostile token could otherwise request billions of PBKDF2 iterations and hang the process |

Wrapped-key, IV, and tag lengths are checked too.

## On-disk compatibility

The format is unchanged, and decryption is self-describing — `p2s` and `p2c` come from each token's own header — so keyrings written by any prior release open unmodified.

**There was no golden fixture before this PR.** `test_jwe_format_go_compatible` only inspects the shape of a token it generates itself, so it would pass even if key derivation or the AAD changed. `tests/fixtures/golden-keyring-josekit.jwe` is an opaque ciphertext captured from the josekit implementation **before any code changed**, and must never be regenerated — regenerating it re-encrypts under whatever code is current and destroys the property being tested.

`tests/jwe_compat_tests.rs` pins both directions:

| Test | Proves |
|---|---|
| `golden_keyring_from_previous_release_still_opens` | pure Rust decrypts a josekit-written keyring |
| `tokens_we_write_are_readable_by_josekit` | josekit decrypts a pure-Rust-written token |
| `golden_keyring_rejects_wrong_password` | decryption is real, not a pass-through |
| `compressed_token_is_rejected_rather_than_returned_compressed` | a `zip` token errors instead of yielding DEFLATE bytes |
| `token_with_unsupported_critical_extension_is_rejected` | `crit` is honoured |
| `token_with_undersized_salt_is_rejected` | the RFC salt floor is enforced |

The second row is why `josekit` remains a **dev-dependency**: a self round-trip would pass even with a self-consistently wrong AAD, so the only real check is cross-implementation. It is dev-only, so OpenSSL stays out of every shipped graph. This also matters for Go parity — `FileKeyring` is documented as compatible with Go DefraDB's file keyring (`github.com/lestrrat-go/jwx/v2/jwe`).

### Known limit of key zeroization

`derive_kek`'s output and the CEK are wrapped in `Zeroizing`, but `Kek` and `Aes256Gcm` copy them into expanded key schedules that neither `aes-kw` (no zeroize support) nor `aes` (only on the armv8 path) clears. Documented at the call site rather than left implied.

## Scope

`openssl-sys` is now absent from **all macOS cli graphs** (`lark`, `rocksdb,lark`, `release-full`), verified with `cargo tree -i openssl-sys` and a passing control on the base commit. The `vendored-openssl` feature and its three release-matrix uses are therefore removed.

On **Linux**, openssl still arrives via `native-tls → hyper-tls → reqwest`, so the `libssl-dev` and `pkg-config` steps in `release.yml` intentionally stay. That half is the follow-up PR.

## Sizes

| Variant | before | after | delta |
|---|---:|---:|---:|
| `x86_64-apple-darwin` `lark` | 39,214,676 | 35,496,876 | **−3,717,800 B = −3.55 MiB (−9.48%)** |

Both sides are cold builds measured at base `d3a01ad7`: before is `--no-default-features --features lark,vendored-openssl` (what CI ships today), after is this branch with `--features lark`. The branch has since been rebased onto `5e5c4990`; the delta is not expected to move, but the figures were taken at the earlier base.

The cross-build is itself evidence: at base it fails with `rc=101` because `openssl-sys` demands an x86_64 OpenSSL that is not installed. After this change it builds, because nothing needs OpenSSL.

## Validation

- [x] `cargo test -p keyring` — 51 passed, 0 failed
- [x] `cargo test -p db-search -p sourcehub -p defra-node` — all passed
- [x] `cargo clippy --all -- -D warnings` (CI's gate)
- [x] `cargo tree -i openssl-sys` empty for all three macOS cli variants, control passes on base
- [x] `josekit` and `flate2` absent from release graphs


jackzampolin (approved):
Approving now that the gate I was holding this for has passed.

**Integration (go-compat) is green, and the keyring lifecycle tests are the reason it mattered.** All 16 ran and passed, including the three cross-binary round-trips that are the actual proof for a JWE rewrite:

```
keyring_lifecycle::go_rust_import_rust_export_go ... ok
keyring_lifecycle::go_rust_import_go_export_rust ... ok
keyring_lifecycle::go_rust_generate_go_list_rust ... ok
```

Those exercise the real binaries in both directions. This PR replaces josekit with a hand-rolled PBES2-HS512+A256KW / A256GCM implementation and changes how keyrings are wrapped on disk, so unit-level round-trips prove self-consistency but not interop; these do.

On the code, from the earlier read: salt construction (`UTF8(alg) || 0x00 || p2s`), the AAD (decrypt uses the token's original base64 header rather than re-encoding, which avoids a canonicalisation mismatch), and the length checks all match RFC 7518. The header validation added beyond josekit — rejecting `zip`, `crit`, undersized `p2s`, and out-of-range `p2c` — is the right call, particularly `zip`: without it a compressed token would authenticate and hand the caller a DEFLATE stream as key material.

The test design is what makes this reviewable: the golden fixture captured from josekit **before** any code changed, plus josekit retained as a dev-dependency so a token we write is proven readable by the library Go's format follows. A self round-trip would have passed with a self-consistently wrong AAD.

Green across Build & Test, Lint, Conformance, WASM, P2P Linux, Lean Conformance and go-compat. The remaining three Integration suites are queued and do not exercise keyring encryption.

PR #1382, opened 2026-08-11
build(deps): select rustls for reqwest, dropping OpenSSL from Linux lark and rocksdb builds

edjroz:
Closes the rest of #1336. Part of #1327.

**Stacked on #1381 — targets that branch, not `main`. Retarget after #1381 merges.**

## Framing, up front: this PR costs bytes

Measured on `x86_64-apple-darwin` `lark`, cold builds, base `d3a01ad7`:

| | bytes | vs base |
|---|---:|---:|
| PR1 only | 35,496,876 | −3.55 MiB |
| PR1 + this PR | 35,663,420 | −3.39 MiB |
| **this PR alone** | | **+166,544 B (+0.16 MiB)** |

macOS `native-tls` uses the system Security.framework, which costs essentially nothing to link. `rustls-tls-native-roots` replaces it with a bundled Rust TLS stack. On macOS this is a small net **regression**.

To be precise about the benefit, since it is easy to overstate: this does **not** fix a trust-store bug. `native-tls` already uses the OS trust store correctly today. What `rustls-tls-native-roots` does is **avoid introducing** the bug that #1336's literal `rustls-tls` would have caused.

The real trade is **+166 KB on macOS in exchange for removing OpenSSL — a C library with a long CVE history — from two Linux cli variants.** That may well be worth it, but it is a supply-chain judgement, not a size win.

## What changed

All five `reqwest` declarations move to `default-features = false` with an explicit feature set: `crates/cli`, `crates/db-search`, `crates/sourcehub`, `crates/defra-node`, `tools/integration-test`.

## ⚠️ Nothing in CI exercises the code path this ships

`defra-harness` and `hub-harness` depend on reqwest with default features, so `default-tls` unifies back on for the whole `integration-test` binary and every test HTTP client uses native-tls. reqwest prefers `default-tls` over `__rustls` whenever both are enabled.

So the rustls path ships in the `lark`/`rocksdb` release binaries and is never executed by any test. rustls is stricter than OpenSSL and Security.framework — it rejects TLS 1.0/1.1, SHA-1-signed certificates, and CN-only certificates without a SAN. A regression of that shape, for example `db-search` calling a user-configured embedding endpoint over legacy TLS, would reach users with no CI signal at all.

This argues for merging early in a release cycle rather than late, and for the manual check below.

## Scope, and what this cannot fix

Verified with `cargo tree --target x86_64-unknown-linux-gnu -i openssl-sys`:

| Linux variant | openssl-sys |
|---|---|
| `lark` | **absent** |
| `rocksdb,lark` | **absent** |
| `release-full` | **still present** |

`release-full` pulls `acp-light-client` (`sourcenetwork/backbone`), which declares `reqwest` with defaults left on. Cargo unifies features per crate-version across the whole graph, so that one out-of-repo declaration re-enables `default-tls` for every consumer in that graph. **It cannot be overridden from this repo** — Cargo offers no way to subtract a feature another crate enabled. The `libssl-dev` and `pkg-config` steps in `release.yml` therefore stay.

⚠️ #1336's acceptance criterion "openssl absent from all cli graphs" is **not achievable from this repository** and should be amended to "absent from all macOS graphs, and from Linux `lark`/`rocksdb`".

## The Linux gain is unmeasured, and is probably not bytes

Cross-compiling to Linux from the development machine needs a cross-linker that is not configured, so the Linux delta needs CI or a Linux host.

Reviewers should not assume it is large. Unlike `x86_64-apple-darwin`, Linux never used `vendored-openssl` — it links system libssl dynamically, which contributes almost nothing to binary size. The likely outcome is the same small increase seen on macOS. **The deliverable on Linux is the removed runtime dependency, not bytes.**

## Validation

- [x] `cargo build -p cli --no-default-features --features lark`
- [x] `cargo test -p db-search -p sourcehub -p defra-node` — all passed
- [x] `cargo clippy --all -- -D warnings` (CI's gate)
- [x] Linux `cargo tree -i openssl-sys` per the table above
- [ ] **Manual check wanted:** `ldd` on a Linux `lark` binary shows no `libssl`/`libcrypto`
- [ ] **Manual check wanted:** an HTTPS request against a host using a private/corporate CA succeeds, confirming `rustls-tls-native-roots` reads the OS trust store. Not unit-testable; the failure mode is runtime-only


vertexclique:
Ran the Linux side on my x86_64 box since you don't have one. Native `x86_64-unknown-linux-gnu`, exact `release.yml` commands, three build points (`main`, this branch's base #1381, and this branch) so each PR's own delta is isolated. Both of your unchecked boxes are ticked below.

## Linux x86_64 numbers

Measured on a native `x86_64-unknown-linux-gnu` host (36 cores, rustc 1.97.1, glibc,
OpenSSL 3.6.3), using the exact `release.yml` build commands, `[profile.release]`
as configured in the workspace (fat LTO, `codegen-units = 1`, `strip = true`,
`panic = "abort"`).

Three build points, so each PR's own contribution is isolated. The stack is
linear with no divergence: `main 5e5c4990` -> `#1381 8aa89e47` -> `#1382 667d469b`.

### Binary size

| variant | main | #1381 | #1382 | #1381 vs main | #1382 vs #1381 | #1382 vs main |
|---|---:|---:|---:|---:|---:|---:|
| `lark` | 41,934,656 | 41,663,416 | 41,827,440 | -271,240 B (-0.26 MiB) | **+164,024 B (+0.16 MiB)** | -107,216 B (-0.10 MiB) |
| `rocksdb,lark` | 51,884,944 | 51,594,384 | 51,758,032 | -290,560 B (-0.28 MiB) | **+163,648 B (+0.16 MiB)** | -126,912 B (-0.12 MiB) |
| `release-full` | 57,292,048 | 57,025,520 | 57,183,248 | -266,528 B (-0.25 MiB) | **+157,728 B (+0.15 MiB)** | -108,800 B (-0.10 MiB) |

Sizes only. A shared target dir was used, so no honest cold-build timings are
available from this run and none are quoted.

### The prediction in the PR body was right

> The likely outcome is the same small increase seen on macOS.

macOS was +166,544 B, Linux `lark` is +164,024 B. Within 2.5 KB. Confirmed.

### `ldd`: the unchecked box, now ticked

| binary | libssl / libcrypto |
|---|---|
| main `lark` | `libssl.so.3` `libcrypto.so.3` |
| #1381 `lark` | `libssl.so.3` `libcrypto.so.3` |
| **#1382 `lark`** | **none** |
| main `rocksdb,lark` | `libssl.so.3` `libcrypto.so.3` |
| #1381 `rocksdb,lark` | `libssl.so.3` `libcrypto.so.3` |
| **#1382 `rocksdb,lark`** | **none** |
| main / #1381 / #1382 `release-full` | `libssl.so.3` `libcrypto.so.3` |

`cargo tree -i openssl-sys` on a real Linux host reproduces the PR's table exactly,
and adds one detail: **neither PR alone drops OpenSSL from Linux.** On `main` the
edges are `josekit` *and* `native-tls`; #1381 removes only the first, #1382 only
the second. Linux `lark` needs the whole stack merged before `openssl-sys` leaves
the graph.

| | main | #1381 | #1382 |
|---|---|---|---|
| `lark` | present (`josekit` + `native-tls`) | present (`native-tls`) | **absent** |
| `rocksdb,lark` | present (`josekit` + `native-tls`) | present (`native-tls`) | **absent** |
| `release-full` | present | present | present (`native-tls`, via `acp-light-client`) |

### Trust store: `rustls-tls-native-roots` does read the OS store on Linux

The second unchecked box. Driven through the real shipped binary and the real CLI
reqwest client (`defra client action list` over https), against a local server
holding a cert signed by a throwaway private CA. Verdicts are taken from the
server's view of the handshake, because the CLI prints only reqwest's top-level
error and never names a cert failure.

| probe | trust configuration | main (native-tls) | #1381 (native-tls) | #1382 (rustls) |
|---|---|---|---|---|
| T1 | unrelated CA only | rejected, `unknown_ca` alert | rejected, `unknown_ca` alert | rejected, `unknown_ca` alert |
| T2 | private CA via `SSL_CERT_FILE` | handshake OK, `200` | handshake OK, `200` | handshake OK, `200` |
| T3 baseline | system store, CA not installed | rejected | rejected | rejected |
| T3 | system store, CA installed | handshake OK, `200` | handshake OK, `200` | **handshake OK, `200`** |

T3 is the one that matters: with `SSL_CERT_FILE` and `SSL_CERT_DIR` unset and the
test CA temporarily added to the OS trust store (`/etc/ca-certificates/trust-source/anchors`
+ `update-ca-trust`, since removed), the rustls binary completes the handshake.
It is reading the platform store, not a bundled Mozilla root set. The baseline row
is the control: the same probe before installing the CA is rejected by all three.

Behaviour is identical to native-tls on every probe.

## Two things the PR body does not say

### 1. On Linux, rustls was already in the binary. This consolidates, it does not add.

`rustls v0.23.37` is already in `main`'s Linux `lark` graph, via
`libp2p -> libp2p-quic -> libp2p-tls -> futures-rustls`, on the `ring` provider.
So before this PR the Linux `lark` binary shipped **both** TLS stacks: rustls for
QUIC and OpenSSL for reqwest. After it, one.

What #1382 actually adds to that graph is only the glue, against the same already
present `rustls` crate and the same `ring` provider:

- `hyper-rustls v0.27.7`
- `tokio-rustls v0.26.4`
- `rustls-native-certs` (via the `rustls-tls-native-roots` feature)
- the `tls12` feature on `rustls`, previously off because libp2p's QUIC usage is TLS 1.3 only

That is what the +164 KB buys, and it reframes the trade on Linux. The PR body
describes rustls as replacing native-tls with "a bundled Rust TLS stack", which
reads as introducing one. On Linux the stack was already compiled in and this
change removes the second one.

The `tls12` point is worth a look on its own: it is newly enabled, so the rustls
in the binary now carries TLS 1.2 code that the libp2p path never needed.

### 2. `release-full` pays the bytes and gets nothing back

It takes the full +157,728 B and still links `libssl`/`libcrypto` (`ldd` above,
consistent with the `acp-light-client` finding). That variant now carries rustls
compiled in *and* OpenSSL linked. It is the shipped default artifact
(`defra_<tag>_<platform>.tar.gz`). Not a blocker, but it belongs in the framing:
the cost lands on all three variants, the benefit on two.

## What was not checked

- `aarch64-unknown-linux-gnu`. x86_64 only.
- Nothing was run against the actual rustls code path in tests. The PR's own
  warning about that stands unchanged: `defra-harness` and `hub-harness` pull
  reqwest with default features, so `default-tls` unifies back on for the whole
  `integration-test` binary and every test HTTP client still uses native-tls.
- No cold-build timings, by design.

vertexclique (approved):

jackzampolin:
Heads-up @edjroz — this went `CONFLICTING` as a side effect of me updating #1381's branch from main, so flagging rather than leaving you to find it.

Both sides now touch the same five files: `Cargo.lock`, `crates/cli/Cargo.toml`, `crates/defra-node/Cargo.toml`, `crates/sourcehub/Cargo.toml`, `tools/integration-test/Cargo.toml`. #1380's per-crate tokio feature declarations landed in the same manifests where this PR edits reqwest.

**I'd suggest not resolving it yet.** This PR is based on `edjroz/1336a-keyring-pure-rust-jwe` rather than main, so once #1381 merges GitHub retargets this to main automatically (same as it did for #1387), and the resolution is then against main rather than against a stack base that is about to disappear. Resolving now means doing it twice.

Sequence: #1381 goes green and merges → this retargets → `gh pr update-branch` or a local merge of main → resolve the five manifests once.

The review still stands — `rustls-tls-native-roots` preserving the OS trust store, and `charset`/`http2`/`system-proxy` preserved from reqwest's defaults, are both right. This is purely mechanical now.

For context on the delay: all four sibling PRs were 16-18 commits behind main and missing `8ab32e5e` (the embedding-fixture semaphore from #1355), so their `Build & Test` runs were failing on flaky fixtures rather than anything real. I updated #1381/#1383/#1385/#1386 from main; their runs are queued behind a fairly deep studio-runner backlog.

Issue #1333, opened 2026-08-07
Dedup duplicate dependency versions


edjroz:
## Context

The graph carries duplicate major versions `[measured via cargo tree]`:

- `prost` x3
- `sysinfo` x3
- `sha2` x3
- a second ed25519/curve25519 generation, pulled by iroh's prerelease dalek line

Each is under 1% of binary size `[hypothesis]`. The real win is build time and supply-chain surface, so this should not be sold as a size fix.

## Acceptance criteria

- [ ] Duplicate counts reduced wherever semver allows pinning.
- [ ] Remaining duplicates documented here with the reason they cannot be unified.

---

Part of #1327. Effort: S-M. Blocked by: nothing.

edjroz:
## Measured correction to the counts above

Three of the four named duplicates can't be acted on in this repo; one can. The x3 figures are the `cargo tree -d --workspace` view, which unions every member's features including `tools/integration-test` and `crates/sourcehub`. Per shipped artifact, at `5e5c4990`:

| Crate | `-p cli` (default `defra`) | `-p ffi --features iroh` | `--workspace` |
|---|---|---|---|
| `prost` | **x1** (0.11.9 only) | x3 | x3 |
| `sysinfo` | **x2** | x3 | x3 |
| `sha2` | **x2** | x3 | x3 |
| ed25519 / curve25519 | **none** | dup | dup |

- **`prost`** — 0.13 pinned by the cosmrs/tendermint stack, 0.11 by our own beetle fork (cross-repo fix). We're already on the newest line.
- **`sha2`** — floor is 2 and the default CLI has reached it: `libp2p-identity 0.2.13` needs `0.10.8`, `multihash-codetable 0.2.2` needs `0.11` inside the CID path. Bumping ours changes docID/CID hashing and still leaves 0.10.
- **ed25519 / curve25519** — `iroh-base 1.0.1` exact-pins `= 3.0.0-rc.0`/`= 5.0.0-rc.0` while `libp2p-identity` floor-pins 2.x. Impossible in both directions; dies with #1341.
- **`sysinfo`** — ours. `crates/p2p` was the sole dependent of 0.29.11; moved to 0.33 to unify with `libp2p-memory-connection-limits`. `total_memory()` returns bytes in both, so the Go-parity memory divisor is unchanged.

[...]

Issue #1330, opened 2026-08-07
Replace multicodec 0.1 with inline constants


edjroz:
## Context

`multicodec 0.1` has a single import in the whole workspace, `crates/defra-core/src/block.rs:7`, and it drags `failure 0.1.8 -> backtrace -> gimli/object` into the graph. That chain defeats `panic=abort` stripping.

`crates/crypto` already hand-rolls the same constants at `crates/crypto/src/did/mod.rs:27-31`. Follow that pattern.

Est. 200-500 KiB native `[hypothesis]`, near-zero on wasm.

## Red line

The constants must be byte-identical. docID/CID parity with Go depends on them.

## Acceptance criteria

- [ ] `multicodec`, `failure`, `backtrace`, `gimli`, `object` gone from the shipped graphs (`cargo tree` proof in the PR).
- [ ] Integration tests pass unchanged; CIDs unaffected.
- [ ] Measured before/after delta recorded using the CI size tooling.

---

Part of #1327. Effort: S. Needs #1328 for measurement.

PR #1383, opened 2026-08-11
chore(deps): unify sysinfo on 0.33 and document the duplicates we cannot unify

edjroz:
Refs #1333.

**Of the four duplicates the issue names, one is ours to fix.** `prost` 0.13 is pinned by the
cosmrs/tendermint stack and 0.11 by our beetle fork; `sha2` is at its floor of 2, held by
`libp2p-identity` and by `multihash-codetable` in the CID path; the dalek pair is exact-pinned
by `iroh-base` in one direction and floor-pinned by `libp2p-identity` in the other. Measured
detail: [comment on #1333](https://github.com/sourcenetwork/defradb.rs/issues/1333#issuecomment-5258604793).

### Changes

- **`sysinfo` 0.29 → 0.33 in `crates/p2p`**, the sole dependent of `0.29.11`. 0.33 unifies with
  `libp2p-memory-connection-limits`, present in every Go-interop variant; 0.37 would unify with
  `commonware-runtime` and leave the default CLI on two versions. `total_memory()` returns
  bytes in both versions, so the Go-parity memory divisor is unchanged; a unit test asserts the
  magnitude (it returns 0 and fails if the refresh kind is dropped — verified deliberately).
- **`thiserror` 1 → 2 in `crates/cursor` and `tools/ffi-test`**, matching the workspace.
  **Zero duplicate-count change** — `graphql-parser`, `josekit`, `lark-kv`, `iroh-bitswap` and
  others still pull v1. First-party consistency only.
- **`deny.toml [bans].skip`**: 17 accepted duplicates, each with the path that pulls it and the
  condition under which it goes away. This is the issue's AC 2, put where `cargo deny` reads it
  instead of in a comment that rots. `multiple-versions` stays `"warn"`.

### Effect, measured at `5e5c4990`

| Variant | duplicated names | duplicate nodes |
|---|---|---|
| `-p cli` (default `defra`) | 39 → **38** | 82 → **80** |
| `-p cli --features release-full` | 54 → 54 | 128 → **127** |
| `-p ffi` | 53 → 53 | 124 → **123** |
| `-p ffi --features iroh` | 75 → 75 | 170 → **169** |
| wasm32 | 15 → 15 | 33 → 33 |
| `--workspace` | 92 → 92 | 207 → **206** |

**Net: one duplicated name removed, default CLI only** — elsewhere `sysinfo` goes 3 versions →
2 via `sourcehub → commonware-runtime` and stays duplicated. Not a size fix, and not a
build-time fix either: `sysinfo 0.29.11` costs 0.35 CPU-seconds to compile and ~1.7 MB of
target artifacts. The value is supply-chain surface — one fewer crate version to audit — plus
the inventory.

### Verification

`cargo clippy --all -- -D warnings` ✅ · `cargo clippy -p p2p --features test-utils
--all-targets` ✅ (the pass that lints the changed code, since `crates/p2p` is `default = []`) ·
`cargo test --workspace` ✅ 194 suites · `cargo test -p integration-test --test p2p` ✅ 100
passed · `cargo deny check bans` ✅ 17 skips matched, none unused · `cargo fmt --all` ✅

Two pre-existing failures, unchanged by this PR: `cargo deny check` advisories (21 errors
before and after), and `cargo clippy --all --all-targets` (two `useless_borrows_in_formatting`
errors in `tools/integration-test`, which neither CI nor `just lint` runs).

⚠️ **Known expiry:** the `unsigned-varint@0.2.3` skip exists only for `multicodec 0.1`, which
#1330 removes. Delete that line when #1330 lands or `cargo deny` will warn about it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3165,7 +3165,7 @@ dependencies = [
  "base64 0.22.1",
  "serde",
  "serde_json",
- "thiserror 1.0.69",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -4553,7 +4553,7 @@ dependencies = [
  "dirs",
  "serde",
  "serde_json",
- "thiserror 1.0.69",
+ "thiserror 2.0.18",
  "tokio",
 ]
 
@@ -8632,7 +8632,7 @@ dependencies = [
  "serde_json",
  "sha2 0.10.9",
  "storage",
- "sysinfo 0.29.11",
+ "sysinfo 0.33.1",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -11801,21 +11801,6 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.117",
-]
-
-[[package]]
-name = "sysinfo"
-version = "0.29.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cd727fc423c2060f6c92d9534cef765c65a6ed3f428a03d7def74a8c4348e666"
-dependencies = [
- "cfg-if",
- "core-foundation-sys",
- "libc",
- "ntapi",
- "once_cell",
- "rayon",
- "winapi",
 ]
 
 [[package]]
--- a/crates/cursor/Cargo.toml
+++ b/crates/cursor/Cargo.toml
@@ -9,4 +9,4 @@ license = "Apache-2.0 OR MIT"
 base64 = "0.22"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-thiserror = "1"
+thiserror = "2"
--- a/crates/p2p/Cargo.toml
+++ b/crates/p2p/Cargo.toml
@@ -56,7 +56,7 @@ thiserror = { workspace = true }
 anyhow = { workspace = true }
 tracing = { workspace = true }
 rlimit = { version = "0.9", optional = true }
-sysinfo = { version = "0.29", optional = true }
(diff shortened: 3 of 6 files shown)


jackzampolin (approved):
Merging with the Lint check red, deliberately, and recording why so the history is readable later.

**The red Lint is an sccache infrastructure failure, not a code defect.** The log shows `sccache: warning: The server looks like it shut down unexpectedly` followed by `sccache: error: Can't assign requested address (os error 49)`, and every `clippy-driver` invocation routed through it then failed. Inspecting studio-1 while it was live: the sccache server was running (1 LISTEN, 26 ESTABLISHED on 4226) with only 80 sockets in TIME_WAIT, so not port exhaustion — the host was simply at load 50-65 on 32 cores with a non-CI tenant taking ~12. Full analysis in #1429.

**Verified the code independently:** ran the exact failing invocation locally on this branch — `cargo clippy -p defra-node --features p2p --all-targets -- -D warnings` — **exit 0, zero warnings**.

**Everything else is green**, including Integration (go-compat), which is the suite that matters here: the sysinfo 0.29→0.33 migration changes `total_system_memory_bytes()`, which feeds the p2p resource-manager limits.

On the substance, reviewed earlier: I checked the one thing that could have silently broken — sysinfo 0.29.11 already documented `total_memory()` as returning bytes, so 0.29→0.33 does **not** rescale the Go-parity memory divisor by 1024×. The added unit test pins that magnitude. The `deny.toml` skip list with per-entry justifications is the right place for that inventory, versus a comment that rots.

Issue #1329, opened 2026-08-07
CI: add cargo-machete and delete the two known dead deps


edjroz:
## Context

No dead-dependency gate exists. Two dead manifest deps were found by hand, which means the gate lands with proof of value rather than as pure ceremony:

- `reqwest` in `crates/db/Cargo.toml:56` (zero src usages)
- `multicodec` in `crates/document/Cargo.toml:40`

## Scope

Add cargo-machete to CI and delete both deps. Note that `multicodec` also appears in `crates/defra-core` where it is genuinely imported; that one is handled separately in the epic (inline constants), not here.

## Acceptance criteria

- [ ] cargo-machete runs in CI and passes.
- [ ] Both dead deps removed; `cargo test` and the integration suite unchanged.

---

Part of #1327. Effort: S. Blocked by: nothing.

PR #1384, opened 2026-08-11
refactor: replace multicodec crate with inline codec constants

### LINKED ISSUE #1330: Replace multicodec 0.1 with inline constants

edjroz:
Closes #1330. Child of #1327 (Tier 1).

## Problem

`multicodec 0.1` had exactly one import in the workspace and existed only to produce two integer constants.
It was the sole path pulling the unmaintained `failure 0.1.8` (last released 2020, deprecated upstream) and its `backtrace` chain into the graph.

## What Changed

`DAG_CBOR_CODEC` and `SHA2_256_CODE` become inline `const` values, mirroring `crates/crypto/src/did/mod.rs:26-32`, which already hand-rolls the same multicodec table.

```rust
// before
pub static DAG_CBOR_CODEC: LazyLock<u64> = LazyLock::new(|| Codec::Dag_Cbor.code() as u64);
pub static SHA2_256_CODE:  LazyLock<u64> = LazyLock::new(|| Codec::Sha2_256.code() as u64);

// after
// Multicodec table: https://github.com/multiformats/multicodec/blob/master/table.csv
pub const DAG_CBOR_CODEC: u64 = 0x71;
pub const SHA2_256_CODE:  u64 = 0x12;
```

`LazyLock<u64>` forced `*CONST` at every use, so this removes that deref at 14 sites across 8 files in 5 crates (`defra-core`, `crypto`, `schema`, `db-merge`, `document`). Entirely mechanical and compiler-caught: a stale `*` on a `u64` does not compile.

It also finishes #1329's unfinished half. `crates/document` still declared `multicodec` as a dev-dependency for one `Codec::Bin.code()` call in a test, replaced with an inline `0x55`. The root workspace declaration could not be deleted while that stood.

**Removed outright** — zero entries remain in the lock: `multicodec`, `failure`, `failure_derive`, `backtrace`.

**Duplicate versions dropped**, one copy of each survives under another dependent: `gimli 0.32.3`, `synstructure 0.12.6`, `addr2line 0.25.1`, `unsigned-varint 0.2.3`.

These crates **remain** under different dependents, and this PR does not claim them:

| Crate | Actually pulled by |
|---|---|
| `gimli 0.33.1` | `cranelift-codegen` ← `wasmtime` ← `lens` |
| `object 0.39.1`, `addr2line 0.26.1` | `wasmtime` ← `lens` |
| `rustc-demangle 0.1.27` | `wasmtime-environ` ← `lens` |
| `object 0.37.3` | `ar_archive_writer` ← `psm` ← `stacker` ← `recursive` ← `sqlparser` ← `pg-compat` |
| `miniz_oxide 0.8.9` | `flate2` ← `josekit` ← `keyring` |
| `synstructure 0.13.2` | `multihash-derive-impl`, `asn1-rs-derive`, `yoke-derive`, `zerofrom-derive` |
| `unsigned-varint 0.7.2` | `multistream-select` ← `libp2p-core` ← `libp2p` |
| `unsigned-varint 0.8.0` | `cid`, `multihash`, `multiaddr`, and `crypto`/`p2p` directly |

**Merge-order note for #1383:** it adds a `deny.toml [bans].skip` for `unsigned-varint@0.2.3` that exists solely because of `multicodec 0.1`. Whichever lands second must drop that line, or `cargo deny` warns about a skip matching nothing. Same applies to any skip covering `gimli 0.32.3` or `synstructure 0.12.6`.

Not touched: `serde_ipld_dagcbor` and the dag-cbor encoding path (epic red line) — only the *source* of two integer literals changed. Nor `crates/crypto`'s constants (the reference pattern), nor `audit/past/**`.

## Validation

```
cargo test --workspace --exclude integration-test --exclude ffi-test --exclude conformance
    194 binaries, 4701 passed, 0 failed

cargo test -p integration-test --test basic
    34 passed, 0 failed  (includes all 10 go_* cross-runtime tests)

cargo clippy --all -- -D warnings    clean
cargo fmt --all --check              clean
```

The `--test basic` run used a Go `defradb` binary built from the pinned `GO_COMPAT_COMMIT` (`13c46ec9`), confirming the inline literals produce CIDs the Go implementation agrees with. That run predates the final rebase; CI covers it on the current head.

**Graph assertions:**

```
$ awk '/^name = /{n=$3} /^ "failure"/{print n}' Cargo.lock     # (no output)
$ grep -c '^name = "multicodec"$' Cargo.lock
0
```

**Guarding the constants.** These values feed docID/CID generation and Go parity — a wrong hex literal would silently rewrite every document ID. `test_codec_constants_match_go_cid_prefix` decodes both values out of the multicodec prefix of an existing Go-generated CID vector rather than restating the literals:

```rust
let go_cid = Cid::from_str(GO_LWW_SIMPLE_CID).unwrap();
assert_eq!(go_cid.codec(), DAG_CBOR_CODEC);
assert_eq!(go_cid.hash().code(), SHA2_256_CODE);
```

Asserting `DAG_CBOR_CODEC == 0x71` would be a tautology: the reference value would live inside the same change surface the test polices. A Go vector cannot be edited consistently without regenerating Go's output. Mutation-checked — setting `DAG_CBOR_CODEC = 0x72` fails it with `left: 113, right: 114`. All three values were read from `multicodec-0.1.0/src/lib.rs` (`:202`, `:72`, `:52`), not recalled. The 12 existing `test_go_wire_compat_*` golden-CID assertions pass untouched.

**This is not a size fix.** The issue estimated 200-500 KiB of native savings; the measured release binary delta is **0 bytes**.

| | bytes |
|---|---|
| `target/release/defra` before | 35,890,192 |
| `target/release/defra` after | 35,890,192 |
| **delta** | **0** |

Both are fat-LTO builds of the same worktree on arm64 macOS differing only by reverse-applying this commit; the baseline recompiled 29 crates, so it is a genuine relink rather than a cache hit. The chain was already dead-stripped: under `lto = true`, `codegen-units = 1`, `strip = true`, `panic = "abort"`, and with the only call being `Codec::code()` (an integer match), `failure`'s error types were never constructed. #1328's CI size tooling is unmerged (PR #1355), so these are local numbers.

Where it does pay off is build artifacts — **≈16.35 MiB** of compile units no longer produced per fresh release build, attributed via `cargo build --message-format=json` so duplicate-version crates split correctly between the copy that survives and the one that does not:

| Package | Artifacts no longer produced |
|---|---|
| `gimli 0.32.3` (dup) | 9.99 MiB |
| `failure_derive` | 2.19 MiB |
| `synstructure 0.12.6` (dup) | 1.37 MiB |
| `backtrace` | 1.30 MiB |
| `addr2line 0.25.1` (dup) | 0.85 MiB |
| `failure` | 0.36 MiB |
| `multicodec` | 0.19 MiB |
| `unsigned-varint 0.2.3` (dup) | 0.10 MiB |
| **Total** | **≈16.35 MiB** |

In proportion: a full `target/` is ~35 GiB, so this is ~0.8% of the release directory. A real saving on every fresh build and in CI cache size, not a fix for local disk pressure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,20 +56,11 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.25.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
-dependencies = [
- "gimli 0.32.3",
-]
-
-[[package]]
-name = "addr2line"
 version = "0.26.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "59317f77929f0e679d39364702289274de2f0f0b22cbf50b2b8cff2169a0b27a"
 dependencies = [
- "gimli 0.33.1",
+ "gimli",
 ]
 
 [[package]]
@@ -965,7 +956,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.117",
- "synstructure 0.13.2",
+ "synstructure",
 ]
 
 [[package]]
@@ -1385,21 +1376,6 @@ dependencies = [
  "fastrand",
  "gloo-timers",
  "tokio",
-]
-
-[[package]]
-name = "backtrace"
-version = "0.3.76"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb531853791a215d7c62a30daf0dde835f381ab5de4589cfe7c649d2cbe92bd6"
-dependencies = [
- "addr2line 0.25.1",
- "cfg-if",
- "libc",
- "miniz_oxide",
- "object 0.37.3",
- "rustc-demangle",
- "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -2776,7 +2752,7 @@ dependencies = [
  "cranelift-control",
  "cranelift-entity",
  "cranelift-isle",
- "gimli 0.33.1",
+ "gimli",
  "hashbrown 0.17.1",
  "libm",
  "log",
@@ -3615,7 +3591,6 @@ dependencies = [
  "futures",
  "ipld-core",
  "multibase",
- "multicodec",
  "multihash",
  "multihash-codetable",
  "rand 0.8.5",
@@ -4035,7 +4010,6 @@ dependencies = [
  "cid",
  "defra-core",
  "multibase",
- "multicodec",
  "multihash",
  "schema",
  "serde",
@@ -4397,28 +4371,6 @@ checksum = "7cd915d99f24784cdc19fd37ef22b97e3ff0ae756c7e492e9fbfe897d61e2aec"
 dependencies = [
  "indenter",
  "once_cell",
-]
-
-[[package]]
-name = "failure"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d32e9bd16cc02eae7db7ef620b392808b89f6a5e16bb3497d159c6b92a0f4f86"
-dependencies = [
- "backtrace",
- "failure_derive",
-]
-
-[[package]]
-name = "failure_derive"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aa4da3c766cd7a0db8242e326e9e4e081edd567072893ed320008189715366a4"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 1.0.109",
- "synstructure 0.12.6",
 ]
 
 [[package]]
@@ -4946,12 +4898,6 @@ dependencies = [
  "opaque-debug",
  "polyval",
 ]
-
-[[package]]
-name = "gimli"
-version = "0.32.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e629b9b98ef3dd8afe6ca2bd0f89306cec16d43d907889945bc5d6687f2f13c7"
 
 [[package]]
 name = "gimli"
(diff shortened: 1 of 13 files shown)


jackzampolin (approved):
Verified the constants against the multicodec table: DAG-CBOR `0x71`, sha2-256 `0x12`, raw `0x55` — all correct, and `test_codec_constants_match_go_cid_prefix` re-derives the first two from a Go CID vector so a wrong literal can't land silently. That guard is the right shape for this change.

Dropping `multicodec 0.1` also takes `failure` → `backtrace`, `syn 1.0`, `synstructure 0.12`, `unsigned-varint 0.2` and a second `gimli`/`addr2line` out of the graph — an unmaintained `failure` edge gone is worth more than the bytes.

`LazyLock<u64>` → `const u64` changes the deref at every call site; Build & Test, go-compat, and iroh integration are all green on it. Integration (rust)/(signed-docs) are still queued behind the studio-runner backlog, not failing.

Approving.

PR #1387, opened 2026-08-11
build(deps): drop tokio `full` from the workspace entry

### LINKED ISSUE #1332: Un-inherit tokio "full"

edjroz:
Closes #1332

## Problem

Cargo's workspace dependency inheritance is additive: a member's `features`
are unioned with the workspace entry's, never subtracted. The workspace entry
requested `full`, so every member received the whole tokio runtime no matter
what it declared.

The effect is that manifests lie. Crates use `tokio::fs` or `tokio::time`
without naming the feature and still compile, because a sibling turned it on,
leaving `cargo check -p <crate>` one unrelated change away from breaking.

## What changed

- `Cargo.toml`: workspace `tokio` drops `features = ["full"]`, becoming version-only.
- `crates/lens`: new tokio dev-dependency (`rt`, `macros`) for its `#[tokio::test]`s.
- `crates/p2p` dev: `["full", "test-util"]` → `["test-util"]`.
- `crates/embedded` dev: `["full"]` → `["rt-multi-thread"]`.
- `tools/integration-test`: own `version = "1.35"` + `full` → workspace inherit, narrow features.
- `proofs`: same, `["rt", "macros", "sync", "time"]`.

tokio's `default = []`, so no `default-features = false` is needed.

## Validation

- **Isolation gate — 44/45 packages pass `cargo check -p <pkg> --all-targets`.** This is the check a `--workspace` build cannot substitute for, since unification still masks omitted features inside it.
- `cargo clippy --all -- -D warnings` clean; `cargo fmt --all -- --check` clean.
- `cargo test --workspace` (excl. `integration-test`, `conformance`) green.
- `cargo test -p integration-test`: **794 passed, 21 failed** — 20 are missing external binaries (`sourcehubd`, `hubd`) or a hardcoded absolute path in test source; the 21st (`p2p_backlog`) is deterministic and reproduces identically at the base commit.
- `cargo hack check --each-feature --all-targets -p lens` went **FAIL → PASS**; the new dev-dependency also frees the test target from `wasmtime-runtime`.
- The one remaining isolated-check failure, `defra-wasm`, is **pre-existing and proven**: re-running at the base commit with `full` restored fails identically. Root cause is `crates/db` declaring tokio `optional = true` while its source uses `tokio::` unconditionally — a dependency-optionality bug, not a feature bug, and untouched here.

### On size

Measured on the default-feature `defra`: **35,890,192 → 35,890,176 bytes, −16.**

The two shipped artifacts are unchanged, and provably so: `acp-light-client`
and `commonware-runtime` both request tokio `full` directly, so `full` cannot
leave the `ffi` or `release-full` graphs. Under `lto = true` + `strip = true`
the unused runtime code was already dead-stripped, so this is a build-correctness
change, not a size change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ await_holding_refcell_ref = "warn"
 defra-p2p-adapter = { path = "crates/p2p-adapter", default-features = false }
 
 # Async runtime
-tokio = { version = "1.35", features = ["full"] }
+tokio = { version = "1.35" }
 async-trait = "0.1"
 async-lock = "3"
 futures = "0.3"
--- a/crates/embedded/Cargo.toml
+++ b/crates/embedded/Cargo.toml
@@ -64,7 +64,7 @@ iroh = ["p2p/iroh-transport", "defra-p2p-adapter/iroh"]
 [dev-dependencies]
 base64 = { workspace = true }
 tempfile = "3.17"
-tokio = { workspace = true, features = ["full"] }
+tokio = { workspace = true, features = ["rt-multi-thread"] }
 zeroize = "1"
 zstd = "0.13"
 
--- a/crates/lens/Cargo.toml
+++ b/crates/lens/Cargo.toml
@@ -43,5 +43,8 @@ wasmtime = { version = "46.0.2", optional = true, default-features = false, feat
     "std",
 ] }
 
+[dev-dependencies]
+tokio = { workspace = true, features = ["rt", "macros"] }
+
 [lints]
 workspace = true
--- a/crates/p2p/Cargo.toml
+++ b/crates/p2p/Cargo.toml
@@ -91,7 +91,7 @@ libp2p-transport = [
 iroh-transport = ["dep:iroh", "dep:iroh-gossip", "dep:iroh-tickets", "dep:noq", "dep:rand", "dep:blake3"]
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["full", "test-util"] }
+tokio = { workspace = true, features = ["test-util"] }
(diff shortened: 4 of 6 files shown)


jackzampolin (approved):
The isolation gate (`cargo check -p <pkg> --all-targets` on 44/45 packages) is the right evidence for this change, and the right thing to have run — a `--workspace` build cannot substitute for it, since unification is exactly what was masking the omitted features. Two jobs that compile `tools/integration-test` (go-compat, P2P Integration) are green, which covers the narrowed features on the biggest consumer.

Spot-checked the two narrowings that looked thinnest: `embedded` dev `["rt-multi-thread"]` and `p2p` dev `["test-util"]` both rely on `macros`/`rt` from their normal dep, which #1380 now declares — so they resolve. Everything else in this class fails at compile time, not runtime, so CI is a sufficient oracle here.

The `defra-wasm` exception is documented and proven pre-existing (`crates/db` declares tokio `optional = true` while its source uses `tokio::` unconditionally) — worth its own issue, not a blocker here.

Integration (rust)/(signed-docs) are queued behind the studio-runner backlog, not failing. Approving.

PR #1386, opened 2026-08-11
chore(telemetry): gate tracing-subscriber behind the otlp feature

edjroz:
closes #1331

## Problem

`crates/telemetry` declared `tracing-subscriber` as a plain, non-optional
dependency, so every consumer paid for it, including the browser client that
never installs a subscriber. Only `dedup` and `init` need it, and both are
otel-specific.

This is dependency hygiene, not a size win. #1331 estimated 100-300 KiB of wasm;
the artifact measured 4,263 bytes *larger*, because fat LTO plus `wasm-opt -Oz`
had already stripped the crate entirely.

## What changed

- `tracing-subscriber` is now `optional = true`, pulled in by the existing `otlp` feature.
- `mod dedup` and the `OtelDedupFilter` re-export are `#[cfg(feature = "otlp")]`.
- Module doc corrected: the filter is no longer "always available".
- No call sites touched, and no `crates/cli` manifest change.

## Validation

- Wasm graph: `tracing-subscriber`, `nu-ansi-term`, `sharded-slab`, `thread_local`,
  `lazy_static`, `tracing-log` all absent; 299 -> 293 crates.
- Measured wasm size, 4 builds with a determinism control (both arms reproduce
  byte-identically, so the noise floor is 0):

  | Metric | Before | After | Delta |
  |---|---:|---:|---:|
  | raw | 5,246,360 | 5,250,623 | +4,263 B |
  | gzip-9 | 2,017,300 | 2,019,173 | +1,873 B |
  | brotli-11 | 1,485,107 | 1,486,488 | +1,381 B |

- `cli`, `ffi`, `defra-node` resolve an identical feature set, so no shipped binary
  changes. `embedded` and `defra-http` shed the 6 crates; neither references
  `tracing_subscriber`.
- `cargo test -p telemetry` and `--features otlp` (6 dedup tests) green.
- `cargo test -p cli --features otel --test telemetry_dedup` green.
- The `telemetry::metrics` conflict counters still work, unmodified:
  `query_context::tests::auto_commit_execution_retries_typed_conflicts` (defra-http,
  115 lib tests green) and `tests::execute_retry_loop_retries_conflicts_until_success`
  (defra-node, 39 lib tests green) both pass.
- `cargo clippy -p defra-wasm --target wasm32-unknown-unknown --all-targets -- -D warnings`,
  `cargo clippy -p telemetry --all-targets` (both feature states), and
  `cargo fmt --all --check` clean.

[...]

PR #1397, opened 2026-08-12
build: add a super-dev profile selectable via the justfile profile variable (#1396)

### LINKED ISSUE #1396: build: add a super-dev profile for lean worktree targets, selectable via a justfile profile variable
## Problem

Dev builds on macOS retain one `.o` file per codegen unit forever:
`[profile.dev] debug = true` makes every profile-with-debug-info build use
`split-debuginfo = "unpacked"` (the macOS default, also set explicitly in
`.cargo/config.toml`), so debug info lives in `target/debug/deps/*.o` and cargo
must keep all of it alive for the debugger — and cargo never garbage-collects
stale fingerprints. The 22 integration-test `[[test]]` binaries each statically
link the full stack and retain ~5.7 GB of objects apiece.

Measured impact: on 2026-08-06 one worktree's `target/debug/deps` held 482,901
`.o` files at 123 GiB, and ten worktrees totaled 513 GiB (43 GiB free on a
926 GiB volume). All targets were deleted; by 2026-08-12 they had regrown to
~330 GiB across 25 worktrees. Concurrent worktree development (humans + agents)
multiplies whatever one target costs.

A measured mitigation exists — `debug = "line-tables-only"` for workspace
crates plus `debug = false` for dependencies cuts object volume ~7.5× with no
build-time regression — but it has only ever lived as uncommitted local edits,
which is why the problem recurred within a week of the first cleanup.

## Proposal

Formalize the lean configuration as a **sibling profile**, opt-in and explicit.
Stock `cargo build` / `cargo test` / rust-analyzer / lldb behavior does not
change.

**1. A `[profile.super-dev]` sibling profile in the workspace `Cargo.toml`.**
It inherits `dev`, keeps `line-tables-only` debug info for workspace crates so
panic backtraces retain real file:line frames, and compiles dependencies with
no debug info at the same `opt-level = 2` dev already gives them. Builds land
in their own `target/super-dev/` directory, fully independent of
`target/debug/` — a worktree that only ever builds this way never materializes
a fat debug dir, and nothing is shared across worktrees.

**2. A `profile` variable in the justfile, default `dev`.**
The variable is threaded through every cargo-invoking dev recipe (`build`,
`test`, `test-crate`, `test-integration`, `test-suite`, and siblings), so plain
`just test` behaves exactly as today while `just test profile=super-dev` is the
explicit opt-in to the lean build. Release recipes and CI invocations are
untouched.

**3. Profile hygiene: manifest is the single source of truth.**

Move the `[profile.dev.package."*"] opt-level = 2` override from
`.cargo/config.toml` into the workspace `Cargo.toml`, and delete the
`[profile.dev]` block from `.cargo/config.toml` entirely. Config-file profiles
silently override manifest ones — a foot-gun that already caused divergence.
No behavior change: `split-debuginfo = "unpacked"` is the macOS default for
debug-enabled profiles, so dropping the explicit line is a no-op, and dev
keeps `debug = true` exactly as today.

**4. Docs.** README dev-workflow section: when to use `profile=super-dev`
(day-to-day worktree development, agent worktrees, disk-constrained machines)
and when to stay on `dev` (interactive lldb with full variable inspection).
CLAUDE.md Common Commands: note the variable so agents in worktrees use it.

## Acceptance criteria

- [ ] `[profile.super-dev]` exists in the workspace `Cargo.toml` as above; no
      other profile's semantics change (`cargo build` output and `dev`
      behavior byte-identical intent).
- [ ] `.cargo/config.toml` contains no `[profile.*]` blocks; the dev package
      override lives in the manifest.
- [ ] Justfile dev recipes accept `profile`, default `dev`; `just test
      profile=super-dev` runs the workspace tests under the new profile.
- [ ] Measured evidence in the PR: on macOS arm64, target-dir size after
      `cargo test -p integration-test --no-run` under `dev` vs `super-dev`
      (expectation: ~7.5× less object volume in the super-dev dir).
- [ ] Backtrace check: a deliberately panicking test under `super-dev` with
      `RUST_BACKTRACE=1` shows workspace file:line frames.
- [ ] README + CLAUDE.md updated as in §4.
- [ ] CI workflow behavior unchanged (no `profile=` passed anywhere in CI).

## Non-goals

- Changing any default: bare cargo, `just` without `profile=`, rust-analyzer,
  and CI all behave exactly as today.
- Shared target dirs, sccache, or any cross-worktree artifact reuse
  (corruption risk; each worktree stays independent).
- Automated target cleanup (`cargo sweep` etc.) — separate concern.
- Reclaiming the existing ~330 GiB of accumulated targets — one-off chore,
  not this issue.

## Evidence appendix

- 2026-08-06: 513 GiB of `target/` across 10 worktrees; single
  `target/debug/deps` = 482,901 `.o` files / 123 GiB; per-binary retention
  ~5.7 GB × 22 integration-test binaries; deletion restored 549 GiB free.
- Mitigation measured the same day: `line-tables-only` + deps `debug=false` →
  ~7.5× less `.o` volume, no build-time change.
- 2026-08-12: regrowth to ~330 GiB across 25 worktrees (76% volume used) —
  confirming that an uncommitted/opt-in-by-hand-edit fix does not stick.

edjroz:
Closes #1396

## Problem

Dev builds on macOS retain one `.o` file per codegen unit forever: `[profile.dev] debug = true` puts full DWARF into `target/debug/deps/*.o` under `split-debuginfo = "unpacked"` (the macOS default), and cargo never garbage-collects stale artifacts. Bloat is therefore **accumulation, not single builds**: a fresh full build is ~5 G, but a sibling worktree was measured at 73 GiB holding 44 coexisting copies of `libstorage-*.rlib`, 53k `.o` files (~6 build generations), and a 36 GiB incremental cache. Across 25 concurrent worktrees this reached ~330 GiB (513 GiB at the 2026-08-06 peak). A measured mitigation (`line-tables-only` + no dependency debug info, ~7.5× less object volume) existed only as uncommitted local edits, which is why the problem recurred.

## What changes

- **`[profile.super-dev]`** in the workspace manifest: inherits `dev`, `debug = "line-tables-only"` for workspace crates (backtraces keep file:line), dependencies at `opt-level = 2` with no debug info. Builds land in `target/super-dev/`, independent of `target/debug/`.
- **A `profile` variable in the justfile** (default `dev`), threaded through the dev build/test recipes: `just profile=super-dev test`. The override precedes the recipe name. Nothing changes by default — bare cargo, plain `just`, rust-analyzer, and CI (which never invokes `just`) behave exactly as before.
- **Profile hygiene:** the `[profile.dev.package."*"] opt-level = 2` override moves from `.cargo/config.toml` into the manifest, and the config file's `[profile.*]` blocks are deleted — config profiles silently override manifest ones. `dev` semantics are unchanged on macOS; on Linux, dropping the explicit `split-debuginfo = "unpacked"` reverts to rustc's default (`off`), a layout-only change.
- **`just sweep [days=7]`:** deletes target artifacts untouched for N days in this worktree only; cargo-sweep installs on first use, same pattern as `audit`/`machete`. **Deliberate scope addition** relative to #1396's non-goals, per team decision: a lean profile shrinks each generation ~7×, but only sweeping bounds how many generations survive.
- Docs: README dev-workflow note + CLAUDE.md commands note.

## Verification

**Cold build, macOS arm64 (14 cores), fresh `rm -rf target` each, all 25 integration `[[test]]` binaries, exit 0:**

| Profile | Wall | Target | `deps/` | `.o` count | `.o` bytes |
|---|---:|---:|---:|---:|---:|
| `dev` | 109.1 s | 5.0 G | 4.1 G | 9,021 | **1.45 GiB** |
| `super-dev` | 92.7 s | 2.8 G | 2.0 G | 7,300 | **0.20 GiB** |

7.2× less object volume (the part that accumulates), no build-time regression.

**Warm rebuild cost of incremental compilation (`cargo build -p cli`, the binary integration tests execute) — why super-dev keeps `incremental = true`:**

| Edit → rebuild `defra` | incremental on | incremental off |
|---|---:|---:|
| `query-plan` (mid-stack) | 4.2 s | 20.2 s |
| `defra-core` (deepest) | 6.0 s | 25.8 s |

Disabling incremental would cost ~15–20 s on every edit-rebuild cycle to save cache disk; instead the cache is left on for the hot loop and bounded by sweep cadence. (Known limitation: cargo-sweep does not clean `target/*/incremental` — the same twelve measurement builds grew it 1.5 G → 4.0 G.)

**Option on the table: removing incremental builds entirely.** Not done in this PR, recorded so the trade-off is explicit:

| Option | Warm rebuild of `defra` | Incremental cache on disk | Suits |
|---|---:|---|---|
| `incremental = true` (this PR) | 4–6 s | grows until purged (cargo-sweep can't touch it) | humans in edit-compile loops |
| `incremental = false` in `[profile.super-dev]` | 20–26 s | none, by construction | disk-first worktrees; one-line follow-up if the cache proves worse than the rebuild tax |
| `CARGO_INCREMENTAL=0` in the caller's environment | 20–26 s | none for those runs | agents/batch runs, no repo change needed |

**Backtrace:** a temporarily-failing `crdt` test (not committed) under `RUST_BACKTRACE=1 --profile super-dev` shows `crates/crdt/tests/lww_tests.rs:16:5` in the frames — workspace file:line survives.

**Sweep:** `just sweep` on a same-day target deleted nothing (correct: warm working set spared); `just sweep 0` deleted 2.36 GiB (3.3 G → 1.2 G, zero rlibs left).

**Hygiene:** `cargo fmt --all -- --check` and `cargo clippy --all -- -D warnings` clean; `just test-crate crdt` without the variable still builds into `target/debug`. Known follow-up: the integration harness hardcodes `target/debug/defra` (`tools/integration-test/src/lib.rs:74`), so super-dev integration runs need `DEFRA_RUST_BINARY` until that path is parameterized.


vertexclique:
Le'ts keep incremental, otherwise it will be cumbersome with more code through time

vertexclique (approved):
Nice, I liked the sweep too.

jackzampolin (approved):

jackzampolin:
Updated this branch from main — the `Build & Test` failure is not caused by the profile change.

**The failure:** `fixture-hot-session: query execution timed out after 30 seconds`, in the `benchmark_support` fixture family.

**Why it is not this PR:** the branch was **32 commits behind main** and missing `8ab32e5e` ("Stabilize embedding fixture tests", merged in #1355 this morning), which serializes those fixtures behind an `EMBEDDING_FIXTURE_TEST` semaphore. `git show <branch>:crates/defra-node/src/benchmark_support_tests.rs | grep -c EMBEDDING_FIXTURE_TEST` returned **0** here and 2 on main. The same failure hit #1381, #1383, #1385 and #1386 earlier today for the same reason, and cleared once each was updated.

**Why the profile change is runtime-neutral for CI**, having checked rather than assumed:

- `opt-level = 2` for dependencies is **preserved** — it moved from `.cargo/config.toml` to `Cargo.toml`, so the crypto libraries the old comment warns about are still optimized. That was the one thing that could plausibly have caused a query timeout.
- `super-dev` is opt-in through the justfile, so CI still builds `dev`.
- The only genuinely dropped setting is `split-debuginfo = "unpacked"`, which affects macOS build time, not runtime.

Worth noting the host was also heavily oversubscribed at the time (see #1429 — studio-1 at load 50-65 on 32 cores), which is why a 30s query budget is fragile there even with the semaphore.

One thing worth confirming while you are here: dropping `split-debuginfo = "unpacked"` from `[profile.dev]` looks unintentional rather than part of the super-dev design — it was a macOS incremental-build optimisation. If it was deliberate, a line in the PR description would help; if not, it probably wants restoring alongside the moved `opt-level`.

PR #1426, opened 2026-08-12
ci: stop cancelling main's post-merge validation runs

jackzampolin:
## Problem

`a1cc7991` introduced the concurrency group and its message says:

> concurrency group cancels superseded PR runs (**main is never cancelled**)

The config does not implement that:

```yaml
concurrency:
  group: ci-${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true      # unconditional — main included
```

So every push to main cancels the previous push's run.

## Impact, measured

**11 of the last 12 `main` runs are `cancelled`.** That includes today's post-merge validation for **#1355, #1384, #1380, and #1387** — four PRs merged today, none of which ever completed CI on main.

This matters more than it sounds. A PR run tests the *merge preview*; the push-to-main run is the only thing that tests the *result*, and nothing runs after it. During a merge train — exactly when main changes fastest and integration risk is highest — each merge cancels the prior merge's validation, so effectively none of them are ever tested.

It also explains a symptom visible all day: main runs perpetually in the queue that never reach a conclusion.

## Fix

One line, implementing what the original commit message already claims:

```yaml
cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

PR runs keep being cancelled when superseded, which is the behaviour worth having. Main runs now complete.

## Cost

Main runs will no longer free their slot early, so a burst of merges queues rather than collapsing to one run. That is the point — the alternative is not testing them. Combined with #1424 doubling `Build & Test` capacity, the added load should be absorbable; if it is not, the answer is more capacity, not less validation.

Found while investigating why #1424's own validation vanished — merging that PR cancelled its in-progress run, which led me to look at what else was being cancelled.


jackzampolin:
## Correction: my framing was wrong — this reverts a deliberate decision

I filed this as "stated intent was never implemented." That is not what happened, and I want it on the record before anyone merges it.

`a1cc7991` (Aug 7) did make main exempt, and its comment said so:

> Superseded PR pushes release the studio fleet immediately; main pushes are **never cancelled** so every main commit gets full validation and artifacts.

`84a7e16c` ("ci: cancel superseded main runs and PR CI on close") then **deliberately reversed it**:

> Only the tip of each ref keeps a run. Superseded PR pushes and **main merge trains cancel prior work** so intermediate SHAs cannot stack the self-hosted fleet.

The comment currently in the file accurately describes current behaviour. I read the older commit message, saw it disagree with the code, and concluded bug — without checking whether the intent had changed in between. It had.

## The underlying concern still stands, as a trade-off rather than a defect

The cost is real and measurable: **11 of the last 12 main runs cancelled**, including post-merge validation for #1355, #1384, #1380 and #1387 — four merges today, none validated on main. PR runs test the merge preview, not the result.

The benefit is also real: on a two-host fleet, a merge train would otherwise stack N full runs for intermediate SHAs nobody will ever ship.

**What changed today is the capacity side of that trade.** #1424 lets `Build & Test` use both studio hosts and stops studio-2 duplicating every compile. The fleet argument for cancelling main was made against the old capacity.

## Options, none of which is "merge this as filed"

1. **Close this.** The trade was made deliberately and the capacity change may not be enough to revisit it.
2. **Validate the tip only, on a delay.** Keep cancelling intermediates, but ensure the *final* main SHA after a train always completes — closest to getting both properties.
3. **Merge queue** (what @jackzampolin raised separately). A queue tests each PR against the result of merging those ahead of it, which subsumes this problem entirely and batches runs rather than multiplying them. If that lands, this PR is moot.

I would not merge this as-is. Converting to draft pending that decision.

PR #1430, opened 2026-08-12
fix(db): preserve concurrent first counter increments

### LINKED ISSUE #1401: Counter nonce predicate diverges from Go: first increment of a counter field on an existing doc gets nonce=0, so concurrent equal increments collapse to one block
Split out of #1043, where a secondary symptom appeared that is unrelated to the Go double-apply bug (now upstream as sourcenetwork/defradb#5135): runs in which **every node agreed on a value one short of the issued sum, with the DAG missing an increment block entirely**. Present both with and without the Go-side fix applied, so it is ours.

## The divergence

Both implementations use `nonce = 0` to keep a document's genesis counter block reproducible, and a random nonce afterwards to keep sibling increments content-distinct. They disagree on **which predicate selects it**.

**Go** — `internal/core/crdt/counter.go`, `Counter.Delta`, keyed on *document* existence:

```go
// To ensure that the dag block is unique, we add a random number to the delta.
// This is done only on update (if the doc doesn't already exist) to ensure that the
// initial dag block of a document can be reproducible.
exists, err := c.store.Has(ctx, c.key.ToPrimaryDataStoreKey())
...
var nonce int64
if exists {
    nonce = <crypto/rand int64>
}
```

**Rust** — `crates/db-blocks/src/write.rs:342`, keyed on the *field's* DAG heads:

```rust
let nonce: i64 = if is_counter && !heads.is_empty() {
    rand::random()
} else {
    0
};
```

`heads` here is per-field — it comes from `.field_heads(field_name)` at `write.rs:271-275`, not from the document's composite heads.

## Why that matters

The two predicates agree in the common case and diverge in one window: **the first-ever write of a counter field on a document that already exists.**

| Situation | Go | Rust |
|---|---|---|
| Create, counter field in the create payload | nonce = 0 | nonce = 0 |
| Later increment, field already has heads | random | random |
| **First increment of a counter field on an existing doc** | **random** | **0** |

In that third row Rust emits `nonce = 0`. Two nodes independently performing that same first increment with the same value, at the same (empty) field-head position, therefore produce **byte-identical `CounterDeltaPayload`s and hence identical CIDs**. Content addressing dedupes them, and one increment is silently lost. Go is immune because both writers would have carried distinct random nonces.

This is silent data loss, not a merge error: no conflict is raised, and every node converges on the same wrong total, which is what makes it hard to notice.

## Why this is likely the observed symptom

`parity_counter_storm_mixed` creates accounts with `{handle, seed}` and no `hits` field, then increments `hits`. So every document in that test hits the divergent window on its first `hits` increment — matching the observation that the missing block was an increment and that the shortfall was exactly one.

Stated as a strong hypothesis rather than a proven cause: I have established the mechanism and its reachability, but have not yet run a test that isolates it from the double-apply bug that was present in the same runs.

## Discriminating test

Two nodes, one already-created document with no counter field written, each issuing the same first increment value concurrently, with no other traffic. Assert the DAG contains two distinct counter blocks and the total equals the issued sum. That isolates this from #1043 entirely, since no diamond and no re-walk is involved.

## Fix direction

Match Go's predicate: select the nonce on whether the *document* already exists, not on whether the field has heads. Worth checking `crates/db-blocks/src/compute.rs:97-98` at the same time — its `// For creates, heads are always empty -> nonce is always 0` is correct for the create path, but it encodes the same field-heads reasoning.

Note `crates/db-blocks/src/build.rs:52` already carries the comment `// Go uses nonce=0 for initial creation (deterministic CID)`, so matching Go was the intent here; the predicate chosen just is not the one Go uses.

Counter merge ignores the nonce in both implementations (#847), so this only affects block identity — which is precisely the problem, since block identity is what dedup keys on.

iverc:
Closes #1401.

## Problem

Rust selected a random counter nonce only when the counter field already had DAG heads. The first counter update on an existing document therefore used nonce zero, allowing equal concurrent increments from separate replicas to produce identical CIDs and collapse into one block. Go instead selects the nonce based on whether the document already exists.

## What changed

- Generate random counter nonces for every document update, including the first write of a counter field.
- Preserve nonce zero for document creation so genesis blocks remain deterministic.
- Add a regression that applies the same first counter increment on two replicas and verifies their blocks have distinct CIDs.

## Validation

- [x] `cargo test -p db-blocks` (24 passed)
- [x] `cargo clippy -p db-blocks --all-targets -- -D warnings`

### DIFF
--- a/crates/db-blocks/src/write.rs
+++ b/crates/db-blocks/src/write.rs
@@ -339,7 +339,7 @@ pub async fn write_document_blocks(
                 .unwrap_or(false)
                 || doc.get_counter_delta(field_name).is_some();
 
-            let nonce: i64 = if is_counter && !heads.is_empty() {
+            let nonce: i64 = if is_counter && !is_create {
                 rand::random()
             } else {
                 0
@@ -645,3 +645,7 @@ mod encryption_derivation_tests;
 #[cfg(test)]
 #[path = "write_kms_tests.rs"]
 mod kms_write_tests;
+
+#[cfg(test)]
+#[path = "write_counter_tests.rs"]
+mod counter_write_tests;
--- a/crates/db-blocks/src/write_counter_tests.rs
+++ b/crates/db-blocks/src/write_counter_tests.rs
@@ -1,0 +1,60 @@
+use super::*;
+use datastore::{NamespaceView, SharedTxn};
+use document::CType;
+use std::collections::HashSet;
+use storage::backends::MemoryStore;
+use storage::corekv::Store;
+use storage::namespace::Namespace;
+
+async fn first_counter_update() -> Cid {
+    let store = MemoryStore::new();
+    let txn = store.new_txn(false).await.unwrap();
+    let shared = SharedTxn::new(txn);
+    let blockstore = NamespaceView::new(shared.clone(), Namespace::Blockstore);
+    let headstore = NamespaceView::new(shared, Namespace::Headstore);
+    let identity = DocStorageIdentity::new(1, 1);
+
+    let mut doc = Document::new();
+    doc.set("name", NormalValue::String("Alice".to_string()));
+    let created = write_document_blocks(
+        &blockstore,
+        &headstore,
+        &doc,
+        "schema-v1",
+        identity,
+        None,
+        None,
+        None,
+        None,
+    )
+    .await
+    .unwrap();
+
+    doc.set_id(document::DocID::from_string(&created.doc_id).unwrap());
+    doc.set_with_crdt("count", CType::PnCounter, NormalValue::Int(1))
+        .unwrap();
+    doc.set_counter_delta("count".to_string(), NormalValue::Int(1));
+    let modified = HashSet::from(["count".to_string()]);
+
+    write_document_blocks(
+        &blockstore,
+        &headstore,
+        &doc,
+        "schema-v1",
+        identity,
+        Some(&modified),
+        None,
+        None,
+        None,
+    )
+    .await
+    .unwrap()
+    .field_cids[0]
+}
+
+#[tokio::test]
+async fn first_counter_updates_on_existing_document_have_distinct_cids() {
+    let (left, right) = tokio::join!(first_counter_update(), first_counter_update());
+
+    assert_ne!(left, right);
+}


jackzampolin (approved):
Verified against Go v1.0.0 (`3de01484`). The fix is right, and the predicate chosen is arguably better than Go's own.

## The change

```rust
- let nonce: i64 = if is_counter && !heads.is_empty() {
+ let nonce: i64 = if is_counter && !is_create {
```

Go's implementation infers "is this an update?" from a storage read — `c.store.Has(ctx, c.key.ToPrimaryDataStoreKey())` — but its comment states the actual intent plainly:

> To ensure that the dag block is unique, we add a random number to the delta. **This is done only on update** (if the doc doesn't already exist) to ensure that the initial dag block of a document can be reproducible.

`!is_create` implements that intent **directly**, rather than deriving it from document existence. The two agree on every normal path: a create takes `nonce = 0` on both sides (preserving the reproducible genesis block Go wants), and any update takes a random nonce on both sides.

It is also the more robust of the two formulations, since it depends on the caller's declared operation rather than on a storage read that could observe a partially-written document.

## Why the old predicate was wrong

`!heads.is_empty()` asked whether the *field* had DAG heads, not whether the *document* existed. Those diverge in exactly one window — the first write of a counter field to a document that already exists — where Rust emitted `nonce = 0`, so two nodes issuing the same first increment produced byte-identical `CounterDeltaPayload`s, identical CIDs, and content-addressed dedup silently dropped one. Go was immune because both writers carried distinct random nonces.

## The test targets the right window

`first_counter_updates_on_existing_document_have_distinct_cids` creates a document with only `name`, then updates it with a `count` PN-counter — precisely the divergent case, not a create-with-counter which would have passed before the fix too. Running both halves through `tokio::join!` and asserting `assert_ne!` on the resulting field CIDs is the correct assertion: the defect was *identical* CIDs, so CID inequality is the property, not the value.

One optional addition, if cheap: a companion asserting a **create** with a counter field still yields a deterministic CID across two runs. That pins the other half of the contract — the reproducible genesis block Go's comment is protecting — which this change could regress in the opposite direction without anything noticing.

Approving.

PR #1437, opened 2026-08-12
fix(db): scope field priorities to their own history

### LINKED ISSUE #1405: Field delta priority is derived from a document-wide max instead of the field's own head history, causing CID and LWW-arbitration divergence from Go on multi-field documents
Found by a systematic Go-parity audit of `crdt-blocks`. Triage procedure: #1391.

## Divergence

Go computes each field's new delta priority as max(that field's own prior head priorities) + 1, independent of how many times sibling fields or the composite were updated. Rust computes a single priority for the whole write call as max(priority over every field's AND the composite's heads in the document) + 1 and stamps that same number onto every field block written in that call. Concrete trace on a 2-field doc (A, B): create -> A=1,B=1,C=1 (both schemes agree). Update A only -> Go: A=2 (own last=1, +1), C=2; Rust: doc-wide max=1 -> priority=2 applied to A and C (still agrees, coincidentally, because the composite always mirrors the doc-wide max). Update B only (third write, B never touched before) -> Go: B's own last head priority is still 1, so B=2, C=3; Rust: doc-wide max is now 2 (from A/C) so priority=3 is stamped on B and C. Rust's field-B block ends up with priority=3 in its LWWDeltaPayload/CounterDeltaPayload; Go's equivalent field-B block has priority=2, for the exact same logical operation ('set B') applied to the exact same prior document state.

## Consequence

Priority is part of the delta payload that gets DAG-CBOR encoded into the block and hashed into the block's CID, so an identical sequence of document operations produces a DIFFERENT CID for the field-B block on Rust vs. Go -- a direct violation of 'Same CRDT operations -> Same merged state' / 'Same document content -> Same document ID'. It also affects real merge correctness: LWW conflict arbitration in both implementations compares raw priority numbers (if priority < curPrio: skip; if priority == curPrio: byte-compare values -- internal/core/crdt/lww.go:141-150 / crates/crdt/src/lww.rs). Because Rust inflates a field's priority based on unrelated sibling-field/composite activity rather than that field's own causal history, a concurrent multi-writer merge (two branches of the same document diverging then reconciling) can pick a different LWW winner on Rust than Go would for the identical operation set -- a genuine convergence/data divergence, not just a cosmetic CID difference. The bug reproduces on any ordinary partial-field update sequence that alternates which field is touched (a common real-world pattern), and is reachable through the normal Update API path (crates/db/src/auto_commit_mutator/update.rs passes a genuine caller-supplied subset as modified_fields).

## References

- **Rust:** crates/db-blocks/src/write.rs:217-230 (priority = snapshot.max_priority()+1, one value applied to every modified field's delta AND the composite delta) together with crates/db-blocks/src/lib.rs:297-358 (DocHeadsSnapshot::load computes `max_priority` by scanning the ENTIRE document's headstore prefix `/d/{doc_short_id}/`, i.e. across all fields plus the composite, not scoped per field)
- **Go v1.0.0 (`3de01484`):** internal/core/block/store.go:82-93 (addDelta: `headset := NewHeadSet(txn.Headstore(), crdtData.HeadstorePrefix())`; `height = headset.List() max + 1`; `delta.SetPriority(height)` -- scoped to THAT crdtData's own head-set only) + internal/core/crdt/lww.go:90 (`LWW.HeadstorePrefix()` returns `l.key.ToHeadStoreKey()`, a field-specific key) + internal/keys/datastore_doc.go:105-110 (`ToHeadStoreKey` keys by `DocShortID`+`FieldID`, i.e. one namespace per field, not per document) + internal/core/block/heads.go:78-118 (`heads.List` iterates only `hh.namespace` prefix) + internal/db/document.go:520-568 (`save()` loops over dirty fields and calls `AddDeltaWithOptions` once per field, each call independently computing that field's own height from its own head-set, before the composite delta gets its own separately-scoped priority)

## Verification done

Searched crates/db-blocks/src/{lib,write,compute,build,collection}.rs end-to-end for any per-field priority computation that might override or correct the doc-wide value before it reaches the delta payload -- none exists; priority is a single local variable threaded unchanged into every field's CrdtDelta and the composite's CrdtDelta in the same write call. Checked crates/db-blocks/src/tests.rs and grepped tools/integration-test for any test that updates two different fields on the same document in separate calls and asserts on field-specific commit heights/priorities -- found none, consistent with this being unpinned and unnoticed. Verified modified_fields genuinely reaches write_document_blocks as a real subset (not silently expanded to all fields) by reading crates/db/src/auto_commit_mutator/update.rs. Cross-checked that this is not simply a labeling difference: Go's per-field head-set scoping is structural (separate keyspace per field via DataStoreKey.FieldID in ToHeadStoreKey), confirmed by reading internal/keys/datastore_doc.go directly, not inferred from a comment. Confirmed the composite/delete path priority computation (crates/db-blocks/src/write.rs:562, write_delete_block) is not independently broken by the same defect in practice, since the composite's own scoped priority in Go always equals the document-wide max in Rust's scheme (composite is written on every document mutation), so only individual FIELD deltas are mis-primed.

iverc:
Closes #1405.

## Problem

Rust derived one priority from every head in a document and applied it to each modified field. Updates to unrelated sibling fields could therefore inflate a field's priority, producing block CIDs and LWW conflict ordering that diverged from Go.

## What changed

- Retain the maximum priority for each field in the existing document-head snapshot.
- Derive field delta priorities from that field's own heads while preserving the independent composite priority.
- Apply the field-scoped priority consistently to delta payloads, head entries, priority indexes, signing, and batch CID collection.
- Add regression coverage for alternating updates across two fields.

## Validation

- [x] `cargo test -p db-blocks -p db --lib` (215 passed)
- [x] `cargo clippy -p db-blocks -p db --all-targets -- -D warnings`


jackzampolin (approved):
Verified against Go v1.0.0 (`3de01484`). The shape is right, and it fixes more than #1405 reported.

## The core fix matches Go structurally

Splitting `priority` into `composite_priority` and a per-field `field_priority` from `field_max_priority(field_name) + 1` is exactly Go's model: `addDelta` builds `NewHeadSet(txn.Headstore(), crdtData.HeadstorePrefix())` and takes `height` from *that* CRDT's own heads (`internal/core/block/store.go:85-93`), with `HeadstoreDocKey` keyed by `DocShortID`+`FieldID` (`internal/keys/datastore_doc.go`).

**Keeping the composite on the document-wide max is correct**, and worth stating since it looks asymmetric. Go's composite gets its own scoped priority, and its head-set is the composite chain — which is document-scoped. Since the composite is written on *every* document mutation, its own count is always ≥ any individual field's, so `max_priority()` and the composite's own head priority coincide. The two schemes agree there.

## It also fixes a signing-parity bug that #1405 did not mention

`internal/core/block/signing.go:74-77`:

```go
// We sign only the first field blocks just to add entropy and prevent any collisions.
// The integrity of the field data is guaranteed by signatures of the parent composite blocks.
if block.Delta.IsField() && block.Delta.GetPriority() > 1 {
    return nil
}
```

Go gates field-block signing on **per-field** priority, so a field's first block is signed however old the document is. Under the doc-wide scheme Rust gave that block a priority > 1 and skipped signing it entirely — a silent divergence in which blocks carry signatures.

`initial_field_cids` collecting `field_priority == 1` CIDs for `batch_collect_cid` is the right mechanism for that, and it means the fix restores signing parity as well as CID parity. Worth calling out in the PR description, because a reviewer looking only at #1405 would read the signing changes as incidental refactoring rather than a second bug fixed.

## Remaining

The one claim from #1405 still unproven either way is the LWW arbitration consequence — that a peer with unrelated sibling-field activity could win an arbitration it should have lost on the tie-break. This fix removes the mechanism, so it becomes moot in practice, but if it is cheap to add a case to `write_priority_tests.rs` covering two writers converging on a never-before-touched field, that would close it explicitly.

Approving.

PR #1440, opened 2026-08-13
fix(document): preserve Blob values through lens migrations

### LINKED ISSUE #1406: Lens-migrated documents corrupt Blob field values (double hex-encoded) because the migration path's schema-typed JSON-to-NormalValue coercion diverges from Go's Blob=String treatment
Found by a systematic Go-parity audit of `document-encoding`. Triage procedure: #1391.

## Divergence

Go treats a Blob field's value as an ordinary NormalValue String in every code path (one switch statement, client/document.go's NewNormalValue, groups FieldKind_BLOB with FieldKind_STRING). Rust's mutation-create path correctly mirrors this (query-plan/src/plan/mutation/create_conversions.rs:193-214 special-cases Blob to stay a String unless a test-only env var is set). But Rust's lens-migration/reindex path calls the generic json_to_normal_value_for_kind helper instead, whose Blob branch is `value.as_str().map(|s| NormalValue::Bytes(s.as_bytes().to_vec()))` -- it converts the hex-text field value into NormalValue::Bytes holding the literal ASCII bytes of the hex string itself (not hex-decoded).

## Consequence

Permanent, silent corruption of the Blob field's externally visible value for every subsequent read after a lens schema migration touches the document: users querying the field get a garbled double-hex-encoded string instead of their original data, with no error surfaced. Diverges from Go, which returns the original hex text unchanged through migrations.

## References

- **Rust:** crates/document/src/encoding.rs:171-173 (json_to_normal_value_for_kind, ScalarKind::Blob branch), consumed by crates/db/src/migration/helpers.rs:65-99 (json_to_native_value) and crates/db/src/migration/helpers.rs:105-136 (lens_doc_to_document), reached from crates/db/src/lensed_fetcher/migration.rs:307, crates/db/src/lensed_auto_commit_fetcher/migration.rs:332, and crates/db/src/migration/reindex.rs:218; corrupted value surfaces via crates/query-types/src/json_convert.rs:19,267-269 (bytes_to_json) used throughout crates/query/src/runner/*
- **Go v1.0.0 (`3de01484`):** client/document.go:272 (case FieldKind_NILLABLE_STRING, FieldKind_NILLABLE_BLOB, FieldKind_STRING, FieldKind_BLOB: return NewNormalString(v)) - Go's single JSON-to-NormalValue conversion switch, used uniformly for every write path including lens-transformed field updates

## Verification done

Confirmed the create-path (query-plan/src/plan/mutation/create_conversions.rs:193-214) deliberately routes Blob through json_to_normal_value (plain String), bypassing the buggy shared helper, so fresh document creation is correct and matches Go. Confirmed the cursor/index-selection call site (query-plan/src/planner/index_selection/values.rs:48) does not corrupt index keys because encode_string_ascending/encode_bytes_ascending in crates/storage/src/encoding/bytes.rs:14-52 produce identical key bytes for identical byte content regardless of NormalValue variant -- so the divergence is confined to CBOR document storage and JSON query output, not index seeking. Grepped tools/integration-test and crates/db/src/migration for any Blob+lens test coverage or documented pinning of this behavior -- none exists.

Classification: **Release blocker**. Confidence: high.

iverc:
Closes #1406.

## Problem

Lens migrations converted Blob hex text into `NormalValue::Bytes` containing the literal ASCII characters. After CBOR persistence and JSON conversion, values were hex-encoded a second time, so a value such as `00ff` became `30306666`. This silently changed stored document behavior and diverged from Go, which keeps Blob values as strings.

## What changed

- Treat Blob values as strings in the shared schema-aware JSON conversion path, matching fresh writes and Go behavior.
- Add regression coverage for lens materialization and the subsequent CBOR round trip.

## Validation

- [x] `cargo clippy -p document -p query-plan -p db --all-targets -- -D warnings`
- [x] `cargo test -p db blob_field_remains_hex_text_after_lens_materialization`
- [x] `cargo test -p document -p query-plan -p db`

### DIFF
--- a/crates/db/src/migration/helpers.rs
+++ b/crates/db/src/migration/helpers.rs
@@ -318,6 +318,30 @@ mod tests {
     }
 
     #[test]
+    fn blob_field_remains_hex_text_after_lens_materialization() {
+        let collection = Collection::new(CollectionVersion::new(
+            "Files",
+            "v2",
+            "files",
+            vec![FieldDescription::new("1", "data", FieldKind::blob())],
+        ));
+        let mut lens_doc = LensDoc::new();
+        lens_doc.insert("data".to_string(), serde_json::json!("00ff"));
+
+        let migrated = lens_doc_to_document(lens_doc, &Document::new(), &collection);
+        assert_eq!(
+            migrated.get("data"),
+            Some(&document::NormalValue::String("00ff".to_string()))
+        );
+
+        let persisted = Document::from_cbor(&migrated.to_cbor().unwrap()).unwrap();
+        assert_eq!(
+            persisted.get("data"),
+            Some(&document::NormalValue::String("00ff".to_string()))
+        );
+    }
+
+    #[test]
     fn lens_removed_field_is_returned_as_explicit_null() {
         let collection = Collection::new(CollectionVersion::new(
             "Users",
--- a/crates/document/src/encoding.rs
+++ b/crates/document/src/encoding.rs
@@ -165,12 +165,9 @@ pub fn json_to_normal_value_for_kind(
         ScalarKind::Float64 => value.as_f64().map(NormalValue::Float64),
         ScalarKind::Float32 => value.as_f64().map(|f| NormalValue::Float32(f as f32)),
         ScalarKind::Bool => value.as_bool().map(NormalValue::Bool),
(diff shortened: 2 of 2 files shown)


vertexclique (approved):

jackzampolin:

PR #1441, opened 2026-08-13
fix(index): deduplicate typed array index values

### LINKED ISSUE #1409: Rust rejects legitimate documents with a duplicate array element on a unique array index that Go accepts (missing dedup before uniqueness check)
Found by a systematic Go-parity audit of `index-encoding`. Triage procedure: #1391.

## Divergence

For a scalar-array field (e.g. `tags: [String!] @index(unique: true)`), Go deduplicates the array's elements before generating index keys, so a document whose array contains a repeated value (e.g. ["a","a","b"]) produces only distinct keys and is indexed without conflict. Rust's expand_array! macro maps every array element into the expansion set without deduplication, so the same repeated value produces the identical (field-value) combo twice in `value_sets`; `validate_unique_value_sets` then sees `value_sets[..position].contains(values)` == true for the second occurrence and returns `UniqueConstraintViolation` before any document is written. This applies to every typed array kind (String/Int/Bool/Float32/Float64/Bytes/Time array and their Nillable variants); Go's JSON-array path (JSONFieldGenerator, no RemoveDuplicates) is deliberately NOT deduplicated on either side and both correctly reject self-duplicate JSON-array unique index values (confirmed by Go's own tests/integration/index/json_unique_array_test.go:72-80, `"numbers": []int{5, 8, 5}` expects the same 'can not index a doc's field(s) that violates unique index' error) — so this finding is specifically about the non-JSON typed-array path, where Go silently succeeds and Rust errors.

## Consequence

Creating a document with a unique-indexed scalar-array field that happens to contain a repeated value succeeds in Go but is rejected outright in Rust with a spurious UniqueConstraintViolation error, even though there is no real cross-document conflict — a correctness regression that turns valid writes into hard failures.

## References

- **Rust:** crates/db-index/src/index_manager/value_extraction.rs:77-88 (expand_array! macro, no deduplication) and :151-157 (invoked for BoolArray/IntArray/Float64Array/Float32Array/StringArray/BytesArray/TimeArray); crates/db-index/src/index_manager/mod.rs:115-134 (validate_unique_value_sets rejects any duplicate non-nil combo in the expanded value_sets) called from on_document_create at mod.rs:622
- **Go v1.0.0 (`3de01484`):** internal/db/index.go:127-143 (ArrayFieldGenerator.Generate calls slice.RemoveDuplicates(normVals) before generating index entries, 'Remove duplicates to avoid duplicate index entries') and :176-184 (getFieldGenerator routes every kind.IsArray() field, i.e. non-JSON typed arrays, to ArrayFieldGenerator); internal/utils/slice/slice.go:18-29 (RemoveDuplicates)

## Verification done

Searched crates/db-index and crates/document for any dedup/remove_duplicates logic (none found); confirmed Rust's array-unique-index integration coverage (tools/integration-test) has no equivalent test for this path, and cross-checked Go's array_unique_test.go (typed-array cases) and json_unique_array_test.go (JSON-array self-duplicate case, which both implementations correctly reject) to isolate that the divergence is confined to non-JSON typed arrays.

iverc:
Closes #1409.

## Problem

Typed array fields generated one index entry for every element, including repeated values within the same document. A unique index therefore treated a self-duplicate such as `["a", "a", "b"]` as a uniqueness violation, while Go deduplicates typed-array values before indexing. JSON arrays intentionally use different semantics and must continue rejecting duplicate indexed values.

## What changed

- Deduplicate typed-array values before generating index Cartesian products, including nillable array forms.
- Keep JSON-array expansion unchanged so positional entries and duplicate validation retain their existing behavior.
- Add regression coverage proving repeated typed-array values no longer reject an otherwise valid document.

## Validation

- [x] `cargo clippy -p db-index -p db --all-targets -- -D warnings`
- [x] `cargo test -p db --test index_manager_tests unique_`
- [x] `cargo test -p db-index -p db`

### DIFF
--- a/crates/db-index/src/index_manager/value_extraction.rs
+++ b/crates/db-index/src/index_manager/value_extraction.rs
@@ -80,8 +80,9 @@ impl IndexManager {
                 if $arr.is_empty() {
                     vec![NormalValue::Null]
                 } else {
-                    $arr.iter()
-                        .map(|v| NormalValue::$variant(v.clone()))
+                    Self::deduplicate_array_values($arr.iter().cloned())
+                        .into_iter()
+                        .map(NormalValue::$variant)
                         .collect()
                 }
             };
@@ -94,8 +95,9 @@ impl IndexManager {
                         if arr.is_empty() {
                             vec![NormalValue::Null]
                         } else {
-                            arr.iter()
-                                .map(|v| NormalValue::$variant(v.clone()))
+                            Self::deduplicate_array_values(arr.iter().cloned())
+                                .into_iter()
+                                .map(NormalValue::$variant)
                                 .collect()
                         }
                     }
@@ -109,9 +111,10 @@ impl IndexManager {
                 if $arr.is_empty() {
                     vec![NormalValue::Null]
                 } else {
-                    $arr.iter()
+                    Self::deduplicate_array_values($arr.iter().cloned())
+                        .into_iter()
                         .map(|v| match v {
-                            Some(val) => NormalValue::$variant(val.clone()),
+                            Some(val) => NormalValue::$variant(val),
                             None => NormalValue::Null,
                         })
                         .collect()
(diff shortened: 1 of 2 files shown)


jackzampolin:
Heads up @iverc — I updated this branch onto main as part of a CI-throughput sweep, and that surfaced a semantic merge conflict: main gained `IndexDescription::kind` (from the vector index work), and the `unique_typed_array_index_deduplicates_repeated_values` test added here constructs that struct without it. Git merged both sides cleanly because they touch different lines, so it only failed at compile time — `Build & Test` and `Lint` went red.

Since I triggered it, I pushed the one-line fix (`604256f3`): `kind: None` in the initializer, matching the placement of the existing ones in that file. `cargo test -p db --test index_manager_tests` → 37 passed, 0 failed, including your new test. Revert it freely if you'd rather handle it differently.

The update also brings this branch the parallel integration suites (#1454), the gossipsub fix (#1473) and the build-cache change (#1455), so CI here should be substantially faster than it was — the rust integration leg went from ~103 min to ~13 min on main.

PR #1453, opened 2026-08-14
fix(p2p): replay only composite document heads

### LINKED ISSUE #1450: Rust→Go PushLog replay rejected: field blocks pushed as standalone heads
`crates/db-merge/src/push_docs.rs:257-292` pushes **every** block of a document's DAG as its own
PushLog head, including field blocks. Go rejects a field block that arrives without its composite
parent — `internal/db/merge.go:501-508` errors when `mp.currentCompositeDocRef == nil`:

```
WARN db_merge::push_docs: Existing document replay PushLog was rejected ...
  completed_blocks=2 total_blocks=4
  error=failed to init CRDT for merge: failed to parse doc ID during merge:
        malformed document ID ... RawDocID: bafyrei...
```

So Rust→Go replicator replay of any LWW document is partially rejected and retried indefinitely.

Found while diagnosing #1449. It is **not** the cause of that failure — this rejection occurs on
passing runs too — but it is why gossipsub is the *only* heal path between a Rust and a Go node
rather than one of two redundant routes. Fixing it would make Rust↔Go replication resilient to a
pubsub outage instead of wholly dependent on it.

## Also worth noting

Two upstream Go observations recorded while investigating, neither causal:

- `internal/db/p2p/sync_doc.go:97-109` keys `pendingPeers` by multiaddr but deletes by peer ID at
  `:150`, so every Go doc-sync burns its full 5s timeout even when it succeeds.
- `quick_protobuf_codec`'s `Failed to encode/decode message` Display hides the underlying io
  error (a stream reset, in #1449's case), which made that diagnosis considerably slower.

iverc:
Closes #1450.

## Problem

Rust document backfill and retry expanded each document DAG and sent linked field blocks as standalone PushLog heads. Go requires document fields to arrive through their composite parent, so it rejected Rust LWW field blocks, left replay partially failed, and made cross-runtime healing depend entirely on gossipsub.

## What changed

- Load validated highest-priority composite document heads together with their block data.
- Replay only composite heads over libp2p and transport-agnostic paths; receivers fetch linked field blocks through DAG sync.
- Keep full-DAG transfer for collection commits and browser sync, where standalone document-head semantics do not apply.
- Add regression coverage for pre-existing Rust-to-Go LWW replication and encrypted filtered backfill.
- Update the encrypted replay model anchors to reflect composite-head delivery.

## Validation

- [x] `cargo clippy -p db-merge -p integration-test --test p2p -- -D warnings`
- [x] `cargo test -p db-merge` (121 passed)
- [x] `cargo test -p integration-test --test p2p --no-run`


jackzampolin:
Heads up @iverc — the red legs here weren't (only) the sweep flakiness. The new `rust_to_go_replicator_backfills_existing_lww_document` test escaped CI's test routing: the `rust`/`signed-docs` macOS legs and the Linux P2P leg all run with `--skip ::go_` and have no Go `defradb` on PATH, so the test ran there and died on `Go defradb binary not available` — deterministically, on every run. Meanwhile the `go-compat` leg (which has the binary) selects tests by the `go_` name filter.

Since the convention for mixed rust+go cluster tests is the `go_rust_` prefix (what `for_each_p2p_topology!` generates, and what both filters key on), I pushed `04c3577c` renaming it to `go_rust_replicator_backfills_existing_lww_document` — one-line change, no behavior touched. Revert freely if you'd rather route it differently.

The other failures on the Aug 14 runs (`filtered_replication::*`, `failed to start rust-1: did not become ready`) hit several unrelated PRs at once when the whole sweep re-ran CI simultaneously and look like runner contention, so this push's fresh run should tell us cleanly.

tools/integration-test/tests/p2p/filtered_replication.rs
```diff
@@ -509,6 +507,9 @@ async fn rust_filtered_replication_encrypted_respects_filter() {
         .expect("create second matching encrypted doc");
     let matching2_id = extract_doc_id(&matching2, "add_SecretDoc");
 
+    // Add the replicator after all writes so delivery can only come from backfill.
```
vertexclique, line 602: Isn't this also requires changes in TLA files for write of LLWs?

jackzampolin:

edjroz (changes requested):
The Rust→Go fix works. `go_rust_replicator_backfills_existing_lww_document` passes here against Go `13c46ec9`, `cargo test -p db-merge` is 121 green, and the `filtered_replication` module (40 tests) passes. One blocking regression, though.

**Filtered replicators under ACP never receive backfilled documents.** Backfill now sends the composite head only, so the receiver has to pull the linked field blocks, and under `AccessMode::Controlled` the sender denies exactly that to a filtered replicator (`bitswap/filter.rs:107`, *"filtered replicator data-block access uses direct push"*; same at `car.rs:202`). The live path compensates with `expand_dag = true` (`broadcast.rs:303`) plus a `selective_car_access` grant (`push_worker.rs:249`). The db-merge backfill/retry path does neither, so the composite lands and its fields never follow.

I wrote a test to confirm it: the Controlled-mode twin of `rust_filtered_replication_backfill_respects_filter`, with the replicator added after the writes so backfill is the only delivery path. On the merge-base the documents arrive in 3.5s; on this branch they never arrive, still nothing at 120s. Nothing in the suite covers that quadrant today, since both Controlled-mode filtered tests add the replicator before the writes and so exercise the live path, while every other filtered test runs in Open mode where `filter.rs:78` short-circuits before the gate is reached.

Two shapes that would fix it: keep DAG expansion for filtered-replicator backfill, or give the backfill path the same `selective_car_access` grant the live path takes out. Happy to hand over the test either way.

tools/integration-test/tests/p2p/filtered_replication.rs
```diff
@@ -509,6 +599,9 @@ async fn rust_filtered_replication_encrypted_respects_filter() {
...
+    add_filtered_replicator(&cluster, 0, &["SecretDoc"], &addr1, "agent_did", ALICE);
```
jackzampolin: **[suggestion]** `rust_filtered_replication_encrypted_respects_filter` now adds the replicator only after every write, so delivery is unordered backfill. The other backfill tests in this file (`rust_filtered_replication_backfill_respects_filter`, and the new Controlled-mode test) sleep `ABSENCE_GRACE` after the matching document arrives before asserting the non-matching document is absent. This test still treats the second matching document as an ordering anchor and then immediately asserts `count == 2`. That anchor only works on the live path, where create order is send order. If a regression started pushing the non-matching encrypted doc, this assertion can pass while that doc is still in flight.

**Suggestion:** After the poll succeeds, sleep `ABSENCE_GRACE` (or poll until the set is stable) before asserting the peer holds exactly the two matching documents. Drop or rewrite the “second matching doc just ensures delivery happened” comment; it is now describing live-push ordering that this test no longer has.

tools/integration-test/tests/p2p/filtered_replication.rs
```diff
@@ -459,10 +551,10 @@ async fn rust_filtered_replicator_cli_requires_filter_field() {
     );
 }
 
-/// #8: the filter must gate the encrypted (SE-artifact) push path too. A
-/// non-matching encrypted document must not reach the filtered peer; matching
-/// documents must arrive with decrypted secrets, proving their field blocks
-/// triggered DEK resolution.
+/// #8: the filter must gate encrypted (SE-artifact) backfill too. A non-matching
```
jackzampolin: **[suggestion]** This was the only filtered-replication test that both enables encryption (`.with_encryption()`, so nodes are not started with `--no-encryption`) and exercised the live push path. Rewriting its doc comment and moving `add_filtered_replicator` after the writes makes it a backfill test. Live filtered push is a different sender (`push_worker` / gossip) than replicator replay (`push_docs.rs`). `rust_filtered_replication_encrypted_rich_filter` still creates after the replicator is added, but that cluster is built without `.with_encryption()`, so it does not replace this coverage.

**Suggestion:** Keep a live-path encrypted filtered test (the previous body of this function) and add the backfill variant separately, the same way `#5` (`rust_filtered_replication_backfill_respects_filter`) sits beside the live `#1/#2` tests.

proofs/tla/EncryptedLwwReplay_DESIGN.md
```diff
@@ -10,7 +10,7 @@ including across a receiver restart.
 
 | Fact | Source | Model consequence |
 |---|---|---|
-| Filtered replication pushes every non-counter field block, including encrypted LWW blocks, as a head so receipt triggers DEK resolution. | `crates/db/src/merge/push_docs.rs`, `crates/db/src/merge/push_docs_transport.rs` | `FilterMode="PreserveEncrypted"` stores the ciphertext; the red mode drops it and violates `INV_NoFilteredLoss`. |
+| Filtered replay pushes composite heads, and pending-DAG completion fetches their linked encrypted LWW blocks so merge triggers DEK resolution. | `crates/db/src/merge/push_docs.rs`, `crates/p2p/src/sync/coordinator/selective_car_access.rs`, `crates/db/src/merge/merge_handler/composite_fields.rs` | `DeliverComposite` records the link and durable retry before acknowledgement; `FetchCiphertext` models the later child fetch. The red mode omits the link and violates `INV_NoFilteredLoss`. |
```
jackzampolin: **[suggestion]** The updated fact row and the conformance fence (lines 54–56) treat `selective_car_access.rs` plus `rust_filtered_replication_encrypted_respects_filter` as the bind for “pending-DAG completion fetches linked encrypted LWW blocks.” That encrypted test still runs in Open mode, where `bitswap/filter.rs` short-circuits before the filtered-replicator deny, so children can arrive without a selective-CAR grant. The new `rust_filtered_replication_backfill_controlled_mode` test is the one that actually needs the grant (Controlled mode denies Bitswap to filtered replicators), but it is not encrypted and does not trigger DEK resolution. Neither test alone binds both the TLA link obligation and the CAR authorization the fact table now cites. `INV_NoFilteredLoss` is also only about retaining `ciphertextLinked`; `FetchCiphertext` has no grant/deny state, so a missing `car_authority.register` on replay would still leave the green spec live.

**Suggestion:** Point the `INV_NoFilteredLoss` fence at an encrypted Controlled-mode backfill test (or say clearly that the Open-mode encrypted test binds the link/DEK obligation and the new Controlled-mode test binds CAR-granted child fetch). If the model is meant to include the grant, give `FetchCiphertext` an authorization flag and a red config that acks the composite without installing the grant.

tools/integration-test/tests/p2p/filtered_replication.rs
```diff
@@ -364,6 +364,98 @@ async fn rust_filtered_replication_backfill_respects_filter() {
...
+        P2P_POLL_INTERVAL,
+        "matching document did not backfill to controlled filtered peer",
+    )
+    .await;
+
+    tokio::time::sleep(ABSENCE_GRACE).await;
+    let result = node1
+        .query("query { User { agent_did } }")
+        .expect("query node1");
+    let dids: Vec<&str> = result["User"]
```
jackzampolin: **[nit]** The absence check collects `agent_did` with `filter_map` and then `assert_eq!(dids, vec![ALICE])`. Rows with a missing `agent_did` are dropped, so a leaked non-matching document whose field blocks never arrived would not fail this assertion. The live Controlled-mode twin (`rust_filtered_replication_excludes_nonmatching_controlled_mode`) asserts `rows.len()` first, which would catch that. This is the test that exists to catch “composite lands, fields never follow,” so counting rows is the tighter check.

**Suggestion:** Mirror the live test: assert the `User` row count is 1, then that every row’s `agent_did` is `ALICE`. Optionally also query `name` so a matching composite whose non-filter field never fetched cannot pass.

vertexclique (approved):

PR #1456, opened 2026-08-14
fix(db-blocks): let a remote signer sign blocks with a secp256r1 key

### LINKED ISSUE #1319: compute_signature rejects Secp256r1 before the remote-signer branch — Secure Enclave (P-256) identities cannot sign blocks
## Summary

`compute_signature` rejects `Secp256r1` before the remote-signer branch is reached, so a TEE-backed identity using a P-256 key cannot sign blocks — even though the code's own comment says remote signers exist so that "mobile/TEE and Orbis-backed identities use the same path".

macOS Secure Enclave keys are P-256. That is the one key type a Secure Enclave can produce.

## Detail

`crates/db-blocks/src/lib.rs:79-105`:

```rust
// Only secp256k1, Ed25519, and BLS are supported for block signing.
// ...
// BLS is a Rust-specific extension wired through a remote signer (Orbis ring)
match signer.key_type {
    SigningKeyType::Secp256k1 | SigningKeyType::Ed25519 | SigningKeyType::Bls => {}
    other => {
        return Err(format!("unsupported key type for signing. KeyType: {}", other));
    }
}

// Determine signature type and sign. Remote signers can back any supported
// key type so mobile/TEE and Orbis-backed identities use the same path.
let sig_type: SignatureType = signer.key_type.into();
let sig_bytes = if let Some(remote) = signer.remote_signer.as_ref() {
    remote.sign_sync(&block_bytes, signer.signing_authorization.as_ref())?
} else { ... };
```

The guard runs first, so a `SigningConfig` with `key_type: Secp256r1` and a perfectly functional `remote_signer` never reaches `remote.sign_sync`.

This is not a type that was never meant to sign — `Secp256r1` is a declared `SigningKeyType` (`crates/defra-core/src/signing.rs:56`) with a defined signature mapping to `SignatureType::ES256` (`signing.rs:79`). The mapping exists; the guard blocks it.

## How this is reached

In [gents](https://github.com/source-inc/gents), the macOS Secure Enclave identity backend registers exactly this shape (`crates/gents/src/identity.rs:271-283`):

```rust
let signer = MacosSecureEnclaveSigner::load(label, create_if_missing)?;
defra_core::signing::store_identity(&did, SigningConfig {
    key_type: SigningKeyType::Secp256r1,
    private_key_bytes: Vec::new(),          // key is non-extractable, as intended
    public_key_bytes: public_key_bytes.clone(),
    remote_signer: Some(Arc::new(signer)),  // TEE-backed
    signing_authorization: None,
});
```

Because the Secure Enclave backend also configures a node identity DID, block signing is active for that deployment. Any block write that goes through `compute_signature` would then fail with `unsupported key type for signing. KeyType: secp256r1` rather than signing via the Secure Enclave.

## Expected

`Secp256r1` should be permitted when a `remote_signer` is present, since the signing is delegated and the local-key path is not involved. The local-key `match` further down can continue to reject it with a clear "non-extractable key requires a remote signer" style error.

Suggested shape: allow the key type through the guard when `signer.remote_signer.is_some()`, keeping the existing rejection for local signing.

## Notes

- Also worth confirming the intended verification story: `SignatureType::ES256` must be verifiable by peers for these blocks to be useful across replication.
- Related but separate: `Signature.identity` carries the hex-encoded public key rather than a DID, and field blocks with `priority > 1` are unsigned by design (`lib.rs:85-89`). Both are fine, but they surprised us — worth a line in the signing docs.

vertexclique:
Closes #1319.

`compute_signature` had two lists of accepted key types: an early guard and the local-signing match. They disagreed, so a Secure Enclave identity with a working `remote_signer` was rejected before reaching it. Removed the guard; the match decides alone.

- `Secp256r1` + remote signer: signs, as `ES256`.
- `Secp256r1` without one: refused by name, since the key cannot be exported.

The existing test asserting the old behaviour is inverted rather than deleted, and now checks the signature actually verifies against the block bytes.

Btw Go verifies these via `VerifyBlockSignatureWithKey`, but its P2P path derives the key from the signature header, which has no `ES256` arm at v1.0.0. Replication to a Go peer needs that arm upstream. @jackzampolin

### DIFF
--- a/crates/db-blocks/src/lib.rs
+++ b/crates/db-blocks/src/lib.rs
@@ -88,30 +88,11 @@ pub(crate) fn compute_signature(
         return Ok(None);
     }
 
-    // Only secp256k1, Ed25519, and BLS are supported for block signing.
-    // secp256k1 and Ed25519 match Go's internal/core/block/signing.go:74-79
-    // and signature.go:186-193. BLS is a Rust-specific extension wired
-    // through a remote signer (Orbis ring); the local-signing match below
-    // still rejects local BLS with a clearer error.
-    match signer.key_type {
-        defra_core::signing::SigningKeyType::Secp256k1
-        | defra_core::signing::SigningKeyType::Ed25519
-        | defra_core::signing::SigningKeyType::Bls => {}
-        other => {
-            return Err(format!(
-                "unsupported key type for signing. KeyType: {}",
-                other
-            ));
-        }
-    }
-
     // Serialize the block (without signature) to get the bytes to sign
     let block_bytes = block
         .to_dag_cbor()
         .map_err(|e| format!("Failed to encode block for signing: {}", e))?;
 
-    // Determine signature type and sign. Remote signers can back any supported
-    // key type so mobile/TEE and Orbis-backed identities use the same path.
     let sig_type: defra_core::block::SignatureType = signer.key_type.into();
     let sig_bytes = if let Some(remote) = signer.remote_signer.as_ref() {
         remote.sign_sync(&block_bytes, signer.signing_authorization.as_ref())?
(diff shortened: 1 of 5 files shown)


iverc (changes requested):
Go `v1.0.0` supports secp256r1 keys and explicit-key verification, but its protocol-facing block signing and automatic P2P verifier recognize only `ES256K` and `EdDSA`. P2P sync calls `VerifyBlockSignature`, so a Rust node using this new `ES256` path can create signed blocks that a Go peer rejects during replication.

Before enabling this in the protocol-facing block path, we need companion Go verification support and a mixed-runtime regression test. Otherwise the Rust path should remain rejected or explicitly gated until peers can consume the resulting blocks.

iverc (approved):

PR #1461, opened 2026-08-14
fix(storage): encode DateTime index keys as separate seconds and nanos

### LINKED ISSUE #1410: DateTime index encoding overflows and errors for years outside ~1677-2262 that Go's index encoding handles natively
Found by a systematic Go-parity audit of `index-encoding`. Triage procedure: #1391.

## Divergence

Go's time index encoding is a composite (seconds, sub-second-nanos) key built from two independent CockroachDB-style varints, which is exact and lossless for any Go time.Time value (bounded only by int64 seconds, i.e. hundreds of billions of years). Rust's encoder first collapses the timestamp into a single i64 count of nanoseconds since the Unix epoch via chrono's `timestamp_nanos_opt()`, which is documented to return `None` outside roughly the year 1677-2262 range. A DateTime value parsed from a perfectly valid RFC3339 string (e.g. "9999-12-31T23:59:59Z", which `chrono::DateTime::parse_from_rfc3339` at crates/document/src/encoding.rs:175/296/304/334 accepts, and which Go's `time.Parse(time.RFC3339, s)` at client/document.go:511 also accepts and can index without issue) fails to encode in Rust's field-value indexer.

## Consequence

Creating or updating a document whose indexed DateTime field holds a legal RFC3339 timestamp outside ~1677-2262 (e.g. far-future/far-past event dates, which the GraphQL schema and RFC3339 parser both permit) fails the document mutation in Rust with an index-encoding error, whereas the identical document and identical index definition succeed in Go.

## References

- **Rust:** crates/storage/src/field_value.rs:103-115 (NormalValue::Time) and :116-128 (NormalValue::NillableTime(Some(v))) — both call `v.timestamp_nanos_opt().ok_or_else(...)`, returning `Error::Other("timestamp ... cannot be encoded as nanoseconds (overflow)")` when the value doesn't fit in a single i64 nanosecond count; crates/storage/src/encoding/time.rs:11-20 (encode_time_ascending/descending take a single i64 unix_nanos)
- **Go v1.0.0 (`3de01484`):** internal/encoding/time.go:19-41 (EncodeTimeAscending/EncodeTimeDescending -> encodeTime encodes `t.Unix()` (int64 seconds, full int64 range) and `t.Nanosecond()` (0-999999999) as two SEPARATE order-preserving varints; no conversion to a single nanosecond-since-epoch scalar, so no overflow for any time.Time representable value)

## Verification done

Checked crates/document/src/encoding.rs and query-parse for any prior validation that would reject out-of-nanosecond-range DateTime values before they reach the index encoder (none found — parse_from_rfc3339 accepts any 4-digit year and is not range-checked against the nanosecond limit before storage/indexing).

vertexclique:
Closes #1410.

DateTime index keys collapsed a timestamp into a single i64 nanosecond count, which only spans ~1677-2262, so `encode_field_value` returned an overflow error outside it and the mutation failed. Go encodes `(t.Unix(), t.Nanosecond())` as two order-preserving varints and covers the full second range.

`encode_time_ascending`/`descending` now take `(seconds, nanos)` and write two varints, matching Go's `encodeTime`. The fallible nanosecond path and the decoder's negative-remainder renormalization are both gone. `encode_varint_descending` is already byte-identical to Go's complement-then-encode-ascending, so descending needs no separate handling.

Verified against the pre-fix encoder: writing `1000-06-15T12:30:45Z` into an indexed DateTime field failed with `cannot be encoded as nanoseconds (overflow)`, and passes after.

Tests: `crates/storage/tests/datetime_index_range_tests.rs` (range, ordering, sub-second precision, pre-epoch), the time property tests rewritten for the pair plus a descending-order property, and `tools/integration-test/tests/query/datetime_index_range.rs` running mutations, ASC/DESC ordering and range filters against both runtimes.

The `go_*` variants were not run locally (no Go binary on PATH).


iverc (changes requested):
There are two correctness blockers here:

- This changes the persisted DateTime index-key format from one encoded nanosecond value to separate seconds and nanos. Existing secondary-index entries are not rebuilt on startup, and this PR adds no store migration or reindex. After an upgrade, existing keys remain in the old format while seeks, range bounds, updates, and deletes generate the new format. That can hide old rows and leave stale entries behind. Please add an upgrade migration or reindex path and a regression test that opens a store containing an index written with the old encoding.
- `encode_time_value` clamps `timestamp_subsec_nanos()` to `999_999_999`. Chrono accepts RFC3339 leap seconds, so this silently aliases values such as `:60` to `:59.999999999`, including collisions in unique indexes. Please reject unsupported leap-second values or encode them without losing information, and pin the behavior with a test.

PR #1480, opened 2026-08-14
Enforce `--max-schema-size` and `--max-backup-size` at the request boundary

### LINKED ISSUE #1296: --max-schema-size and --max-backup-size are accepted but never enforced
## Summary

Both flags are parsed, validated and stored in the server configuration, but no code reads them at any enforcement point. An operator who sets them receives no protection and no indication that the setting is inert.

## Context

**The site.** `crates/http/src/server.rs:41,43`. The config fields are populated and never subsequently read. Searching the tree for the field names returns the definition, the CLI wiring, and a test, but no consumer.

**Why the test suite does not catch it.** `cli/tests/start_tests.rs:202` asserts that the configured value *arrives* in the server config. It does not assert that a request exceeding the limit is rejected. The test is green and tests the wrong half of the contract, so the gap is invisible from CI.

**Why this is worth filing rather than quietly wiring up.** These are denial-of-service controls. A flag that silently does nothing is worse than an absent flag: an absent flag prompts the operator to find another control, whereas a present one ends the search. Anyone who has set `--max-schema-size` in a deployment currently believes they have a bound on schema size and does not.

**Suggested direction.** Either enforce both at the request boundary and extend the test to assert rejection, or remove the flags and document the absence. Enforcing is preferable, but removing is better than the current state. If enforcement lands, the accompanying test should exercise the reject path rather than the plumbing.

edjroz:
Closes #1296

## Problem

`--max-schema-size` and `--max-backup-size` were parsed, validated, stored on `ServerConfig`, and
never read by anything. The existing test asserted the value arrived in the config struct, never that
an oversized request was rejected, so CI stayed green while operators had a DoS bound that did nothing.

## What changed

- Apply both as per-route `DefaultBodyLimit` caps returning `413`, matching the browser-sync precedent
- `max_schema_size` binds `POST /schema`, `/collections`, `/tx/{id}/schema` and `/views`
- `max_backup_size` binds `POST /backup/import`
- `/views` is included because Go parses view SDL with the same `ParseSDL` as a schema add (`view.go:47` vs `collection.go:276`)
- `PATCH /collections` stays uncapped: Go patches the description with `jsonpatch`, never `ParseSDL` (`collection_define.go:154`)
- Delete the import handler's hardcoded 100 MiB check; that number becomes `max_backup_size`'s default
- `Server::route_body_limit` clamps every route cap to `max_body_size`
- The `/collections` cap is merged onto the POST alone, so sibling patch/delete stay uncapped

## Verified

- Reverting `capped()` fails five tests; removing the `.min()` clamp fails exactly the clamp test; chaining instead of merging fails exactly the PATCH test
- `cargo test -p defra-http`, `-p cli`, `cargo clippy --all -- -D warnings`, `cargo fmt --all -- --check` all clean

## Notes

- The clamp is load-bearing, not defensive: a route-level `DefaultBodyLimit` overrides the global one rather than intersecting with it, so an unclamped looser route cap would raise the effective cap above the operator's global limit.
- `max_backup_size` defaulting to 100 MiB instead of 0 keeps the backstop a default node already had while making the flag able to raise the cap, not only lower it. Go cannot guide this — its import reads a server-side filepath and has no request body to bound (`http/handler_store.go:38-51`), whereas ours uploads the backup inline.


crates/http/src/server_tests.rs
```diff
@@ -0,0 +1,222 @@
+use axum::body::Body;
```
vertexclique, line 1: do you want to move the tests into tests dir? Or we want to do it in separate pass?

edjroz: I think it's best to move them as a follow up, since we have to review the rest of the refactors and determine merge sequence

christopherbm (approved):
Non-blocking comment: I would argue for a value enum instead of using 0 to mean unlimited.

iverc (approved):

Issue #1422, opened 2026-08-12
Build a config-enforcement test class: every flag must demonstrably do something


jackzampolin:
Sub-issue of #1393.

## The gap that parity testing is structurally blind to

Three filed defects share one shape: **an option a user can set that silently does nothing.**

- **#1296** — `--max-schema-size` / `--max-backup-size` accepted, never enforced
- **#1413** — `--no-signing` parsed and stored, never consulted; commits are always signed
- **#1414** — `--valuelogfilesize` parsed and stored, never passed to any backend

No amount of go-compat testing catches this class in general, because Go is not the oracle. The oracle is *"does setting this change observable behavior?"* — a question about the option's relationship to itself, not to Go.

That makes this a test class the suite does not currently have, and worth building deliberately rather than case by case.

## Two layers

**1. Dual-binary, for options Go also has.** `--no-signing` and `--valuelogfilesize` both exist in Go (`cli/start.go:293-295, 347-348`). For these the dual-binary suite works directly: start both with the flag set, assert both change behavior the same way. `--no-signing` is cleanly observable — query `_commits` and check for signature presence.

**2. Single-binary enforcement assertions, for the rest.** For options Go lacks (`--max-schema-size`, `--max-backup-size`), assert the flag has an effect at all: set it to a value that must change behavior, and assert the behavior changes. Set `--max-schema-size` below a schema's size and assert rejection.

## Preventing recurrence

The generalisable step, and the reason this is worth doing properly: **enumerate every CLI flag and config key, and require each to have either an enforcement test or an explicit annotation saying why it needs none.** Three instances of one bug shape found without anyone looking systematically suggests there are more; a checklist converts an open-ended hunt into a finite one.

`crates/cli/src/commands/start/mod.rs` and the config structs in `crates/defra-node/` and `crates/embedded/` are the surfaces to enumerate.

## Where

Dual-binary cases in `tools/integration-test/tests/basic/` (`go_*`-named); enforcement cases wherever they fit, since they do not need Go.

PR #1482, opened 2026-08-14
Map `--valuelogfilesize` onto the lark and rocksdb target file size

### LINKED ISSUE #1414: --valuelogfilesize is parsed and stored on DatastoreConfig but never passed to any storage backend
Found by a systematic Go-parity audit of `unenforced-cli`. Triage procedure: #1391.

## Divergence

Go wires the flag straight into the Badger datastore options (ValueLogFileSize), which controls value-log segment sizing/rotation and therefore write amplification and disk layout. Rust exposes the identical flag name and help text, stores it on DatastoreConfig, but no storage backend (lark, redb, fjall, rocksdb) or any other consumer in the workspace ever reads DatastoreConfig.valuelogfilesize.

## Consequence

A user tuning `--valuelogfilesize` to control datastore segment sizing (e.g. for disk-constrained or high-throughput deployments) gets no effect whatsoever in Rust; the flag is accepted and silently ignored.

## References

- **Rust:** crates/cli/src/commands/start/mod.rs:51-53 (clap arg 'Specify the datastore value log file size (in bytes)'), 427-428 (config.datastore.valuelogfilesize = size); crates/cli/src/config/sections.rs:208,243 (DatastoreConfig.valuelogfilesize field/default, never read by any consumer)
- **Go v1.0.0 (`3de01484`):** cli/start.go:293-295 (--valuelogfilesize flag, help text 'Specify the datastore value log file size (in bytes)...'), cli/start.go:97 (SetBadgerFileSize(int64(cfg.GetInt("datastore.badger.valuelogfilesize")))); node/store_badger.go:34 (badgerOpts.ValueLogFileSize = opts.BadgerFileSize) — the value directly configures the Badger datastore's on-disk value-log segment size.

## Verification done

Grepped crates/storage/src/backends/{lark,redb,fjall,rocksdb}/*.rs and crates/defra-node for value_log_file_size / ValueLogFileSize / value_log -- zero hits anywhere outside the CLI parsing/test files.

edjroz:
Closes #1414

## Problem

`--valuelogfilesize` was parsed into `config.datastore.valuelogfilesize` and never passed to any
storage backend, so tuning it had no effect whatsoever.

## What changed

- Map onto `target_file_size` (lark) and `target_file_size_base` (rocksdb); fjall, redb and memory no-op with a startup warning
- Apply only when explicitly set: the config field becomes `Option<u64>` and is marked `serde(skip)`
- Split env acquisition from the override (`lark_options_from` / `rocksdb_options_from`) so precedence is structural rather than a comment
- Precedence is CLI flag > `LARK_TARGET_FILE_MB` / `ROCKS_TARGET_FILE_MB` > backend default
- Add a `target_file_size` getter to the lark options wrapper, which rocksdb already had

## Verified

- Removing the apply in `lark_options_from` fails the wiring test with `left: 67108864, right: 4194304`
- Tests cover an explicitly set value reaching the backend, an unset flag leaving the resolved default alone, a legacy config file not resurrecting 1 GiB, and the flag beating the env var
- Warning observed on a live node: `--valuelogfilesize=1048576 has no effect on the Redb backend and is being ignored`
- Standard gates clean, plus `cargo test -p storage`

## Notes

**🔑 Why `serde(skip)` is load-bearing — do not simplify this away.** The old default was 1 GiB,
mirroring Go, but that is *Badger's* own default so passing it through is a no-op there. lark and
rocksdb default to 64 MiB, so applying it moves every operator who never set the flag by 16x. That is
a compaction-granularity change, not a disk-space one: `target_file_size` is the threshold at which a
compaction output file is cut (`lark` `src/engine/compaction.rs:1486-1488`), and at 1 GiB a single
output file would exceed the entire 256 MiB level-1 budget (`:586-588`). Keeping the backend's own
default is therefore a deliberate divergence from Go — a value log and an SST compaction target are
different mechanisms, and a shared default is meaningless across them.

Dropping the default is not enough on its own. `create_if_missing` serializes the whole default config
to `$rootdir/config.yaml` on every start, so every node ever started already has
`valuelogfilesize: 1073741824` written to disk; deserializing that back would deliver the 16x change
to every upgraded deployment while fresh installs got 64 MiB. `serde(skip)` closes it by making the
key neither read nor written, so `Some(_)` means the operator passed the flag and nothing else.

- No bounds validation: Go's Badger rejects outside `[1 MiB, 2 GiB)`; we accept `--valuelogfilesize 0`.
- Operator footgun: the flag is in bytes, so `--valuelogfilesize 64` is 64 bytes, not 64 MiB.
- That `Node::start` calls the option builders is unpinned — needs a node-level on-disk test, tracked under #1422.


iverc (approved):
LGTM

Issue #1404, opened 2026-08-12
Audit: Rust registers 25 of Go v1.0.0's 39 schema definition validators


jackzampolin:
Go v1.0.0 registers **39** schema definition validators; Rust registers **25**. Each unmatched Go validator is a schema Rust may accept and Go rejects — i.e. a candidate parity gap with data-integrity consequences, since schema validation is what prevents documents from entering illegal states.

One has already been confirmed and split out: #1403 (`validateNonNillableFieldNotAdded`). This issue tracks confirming or dismissing the rest.

Sources: Go `internal/db/definition_validation.go` (registration lists around `:135-187`); Rust `crates/schema/src/definition_validation/mod.rs`.

## Candidates

Name-variant matches are already accounted for (`validate_source_belongs_to_host` ≈ `validateCollectionSourceFromSameCollection`, `validate_branchable_not_mutated` = `validateCollectionIsBranchableNotMutated`, `validate_version_id_not_mutated` = `validateCollectionVersionIDNotMutated`).

**Relation integrity** — the densest cluster, and the one where a missing check most plausibly corrupts data rather than merely diverging:
- [ ] `validateRelationPointsToValidKind`
- [ ] `validateSecondaryFieldsPairUp`
- [ ] `validateSingleSidePrimary`
- [ ] `validateRelationNameSet`
- [ ] `validateRelationalFieldIDType`
- [ ] `validateSelfReferences`

**Views and collection state:**
- [ ] `validateSingleVersionActive`
- [ ] `validateDematerializedViewHasNoData`
- [ ] `validateCollectionID`
- [ ] `validateVersionID`

**Fields, policy, indexes:**
- [ ] `validateCollectionFieldDefaultValue`
- [ ] `validateCollectionDefinitionPolicyDesc` — security-adjacent; a policy description Go rejects being accepted here is worth checking carefully
- [ ] `validateEncryptedIndexes`

## Method for each

1. Read the Go implementation and state precisely what it rejects.
2. Search Rust for equivalent enforcement — **not only `crates/schema`**. Some of these may legitimately be enforced earlier, at SDL parse time in `crates/query-parse`, or during relation construction in `crates/schema/src/relation.rs`. Absence from the validator list is not by itself absence of the check, which is exactly the mistake this issue exists to avoid making 13 times.
3. If genuinely unenforced: construct the schema Go rejects, confirm Rust accepts it, and split out a sub-issue with the repro.
[...]

PR #1488, opened 2026-08-15
fix(schema): validate encrypted index fields

iverc:
Refs #1404.

## Problem

Rust validated encrypted-index directives while parsing SDL, but raw `CollectionVersion` definitions could still reference missing fields or declare more than one encrypted index for the same field. Go rejects both states during collection validation.

## What changed

- Validate encrypted indexes in the shared `CollectionVersion::validate` boundary.
- Reject indexes that reference missing fields.
- Reject duplicate encrypted indexes on one field.
- Add regression coverage for both invalid definitions.

## Validation

- [x] `cargo test -p schema`
- [x] `cargo test -p db collection`
- [x] `cargo clippy -p schema --all-targets -- -D warnings`


vertexclique (approved):

jackzampolin:

PR #1489, opened 2026-08-15
fix(schema): validate relation ID field kinds

iverc:
Refs #1404.

Depends on #1488.

## Problem

Relation finalization skips generating a backing field when one already exists. Rust did not verify that an existing field such as `_authorID` was a `DocID`, so a malformed definition could preserve another scalar kind and be persisted. Go rejects this state through `validateRelationalFieldIDType`.

## What changed

- Validate existing backing fields for non-array relations.
- Return a specific schema error when the backing field is not `DocID`.
- Add regression coverage for a relation backed by a string field.

## Validation

- [x] `cargo test -p schema`
- [x] `cargo test -p db collection`
- [x] `cargo clippy -p schema --all-targets -- -D warnings`
- [x] `git diff --check`

### DIFF
--- a/crates/schema/src/collection.rs
+++ b/crates/schema/src/collection.rs
@@ -330,6 +330,7 @@ impl CollectionVersion {
         self.validate_no_duplicate_index_names()?;
         self.validate_encrypted_indexes()?;
         self.validate_fields()?;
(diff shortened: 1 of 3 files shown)


vertexclique (approved):

jackzampolin:

Issue #1425, opened 2026-08-12
Five Go v1.0.0 HTTP routes are absent from Rust, including /openapi.json


jackzampolin:
Found by executing #1419's route-table comparison against both running binaries, rather than by reading code.

## Method

Started Go v1.0.0 (`3de01484`) and Rust side by side, pulled Go's authoritative route list from its own `GET /openapi.json` (64 method+path pairs), and probed each against Rust.

Route presence was determined with a **method-based discriminator** — an unsupported verb returns `405` when the path exists and `404` when it does not — validated against a known-present route first. This matters: a naive status probe reports `404` for *both* "route absent" and "collection does not exist", and three routes (`/collections/{name}`, `/collections/{name}/document/{docID}`, `/collections/{name}/truncate`) initially looked missing and are in fact present. They are not part of this issue.

## Genuinely absent from Rust

| Method | Go v1.0.0 route | Purpose |
|---|---|---|
| POST | `/ccip` | Chainlink CCIP read interface |
| GET | `/ccip/{sender}/{data}` | CCIP read, GET form |
| GET | `/node/options` | node configuration readback |
| POST | `/p2p/disconnect` | disconnect a peer |
| GET | `/openapi.json` | Go serves 36,940 bytes; Rust 404s |

Path mismatches — Rust has the capability at a different path — are tracked in #1411 rather than here (`acp/policy`, `view`/`views`).

## Assessment per route

These are not equal, and this issue should probably be split once each is judged:

- **`/openapi.json`** is the one I would rate highest. Go publishes a machine-readable contract; Rust publishes none. Any client generating bindings from the spec has nothing to generate from, and it is also what made *this* comparison cheap — the absence is self-perpetuating.
- **`/node/options`** is a readback of node configuration. Related to the unenforced-flag cluster (#1296, #1413, #1414): a client cannot even ask what the node thinks its configuration is.
- **`/p2p/disconnect`** is a real capability gap in P2P management, and `defra-agent` uses P2P heavily.
- **`/ccip`** may well be deliberate — CCIP is a Chainlink integration and may be out of scope for the Rust runtime. If so it should be recorded as an explicit non-goal rather than an unmentioned absence, per #1391's treatment of deliberate divergence.

## Scope check

All five are present at Go v1.0.0 (`3de01484`), taken from that build's own spec, so none are post-tag additions.

PR #1490, opened 2026-08-16
fix(http): add the P2P disconnect route

iverc:
## Problem

The Rust HTTP API does not expose the Go-compatible `POST /api/v0/p2p/disconnect` route, so API clients cannot explicitly disconnect peers through the same interface. This is one of the route parity gaps tracked in #1425.

## What changed

- register `/p2p/disconnect` under both API versions
- accept Go-compatible address arrays, validate each address, and delegate to the existing P2P disconnect operation
- enforce the `P2pPeerDisconnect` permission
- cover route delegation and permission selection

## Validation

- [x] `cargo clippy -p defra-http --all-targets -- -D warnings`
- [x] `git diff --check`

Refs #1425

### DIFF
--- a/crates/http/src/router/routes.rs
+++ b/crates/http/src/router/routes.rs
@@ -168,6 +168,7 @@ pub(crate) fn create_router_with_state_and_body_limits(
         )
         .route("/active-peers", get(handlers::p2p::active_peers)) // Go-compatible
         .route("/connect", post(handlers::p2p::connect)) // Go-compatible
+        .route("/disconnect", post(handlers::p2p::disconnect)) // Go-compatible
         .route("/peers", get(handlers::p2p::list_peers))
         .route("/peers", post(handlers::p2p::connect_peer)) // Legacy
         .route("/replicators", get(handlers::p2p::list_replicators)) // Go uses /replicators
@@ -356,7 +357,8 @@ pub(crate) fn create_router_with_state_and_body_limits(
 mod tests {
     use super::*;
     use crate::mock::{
-        FailingMockP2POperations, MockDocumentAcpOperations, MockQueryExecutor, MockRestOperations,
+        FailingMockP2POperations, MockDocumentAcpOperations, MockP2POperations, MockQueryExecutor,
+        MockRestOperations,
     };
     use crate::router::{
         DocumentAcpOperations, ManageRequester, P2POperations, P2PResult, P2pDocumentInfo,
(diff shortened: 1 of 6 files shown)


crates/http/src/route_permissions_tests.rs
```diff
@@ -121,6 +121,10 @@ mod tests {
             route_permission("/api/v0/p2p/shareable-address", &Method::GET),
             RoutePermission::Required(NodePermission::P2pPeerInfo)
         );
+        assert_eq!(
```
vertexclique: Can you check tests migration to the top level of the crate?
e.g. https://github.com/sourcenetwork/defradb.rs/pull/1498/changes#diff-704fdcd3276ce4a2677dbbeb80eb7205ebbb76555569e6f6283812bdf9683fc9

jackzampolin:

crates/http/tests/route_permissions.rs
```diff
@@ -0,0 +1,549 @@
+//! Exhaustive permission table tests for route permissions.
```
jackzampolin: **[suggestion]** The follow-up commit recreates `crates/http/tests/route_permissions.rs` and deletes the old in-crate copies. That move already landed on main (with extra coverage: `every_mount_form_uses_the_same_permission_as_v0`, `near_miss_version_segments_are_not_folded`, and an exhaustive table that has since drifted, e.g. `DELETE /collections/:name` is `DocumentDelete`). Applying this file as added would conflict with, and can clobber, those tests. Disconnect itself only needs two rows in the existing table.

**Suggestion:** Rebase onto current main. Drop the test-file recreation. Add `POST /api/v0/p2p/disconnect` → `P2pPeerDisconnect` to the in-tree `p2p_routes` test and to `all_registered_routes_return_expected_permission`.

crates/http/src/router/routes.rs
```diff
@@ -509,6 +511,31 @@ mod tests {
         assert_eq!(status, StatusCode::INTERNAL_SERVER_ERROR);
     }
 
+    #[tokio::test]
+    async fn p2p_disconnect_removes_connected_peer() {
```
jackzampolin: **[suggestion]** The only HTTP test disconnects a peer that is already in the mock set. Production `disconnect_peer` is idempotent (libp2p `handle_disconnect` ignores an absent peer; iroh has `disconnect_absent_peer_is_ok`; the mock `retain`s), so hanging up a never-connected address must be 200, matching Go and the manage-relay PeerDisconnect tests. There is also no counterpart to `p2p_collection_add_does_not_ack_when_operation_fails` for a failing `disconnect_peer`.

**Suggestion:** Add a request against a well-formed address that is not connected and assert 200, and a `FailingMockP2POperations` request that asserts a non-success status (400/500 via `map_p2p_bad_request`).

vertexclique (approved):

Issue #1497, opened 2026-08-17
route_permissions table is dead for every parameterized route: it uses :name where axum 0.8 MatchedPath yields {name}


vertexclique:
Found while fixing #1411. This one is **not** a Go-parity gap: it is a Rust-internal defect in the auth middleware's route table. Triage procedure: #1391, question 2.

## The defect

`crates/http/src/route_permissions.rs` matches parameterized routes with **colon** syntax:

```rust
"/api/v0/collections/:name" => match *method { ... }
"/api/v0/collections/:name/document/:docID" => match *method { ... }
"/api/v0/tx/:id" => RoutePermission::IdentityOnly,
```

The lookup key is axum's `MatchedPath` (`crates/http/src/auth_middleware.rs:34-43`), which returns the route template **exactly as registered**. Routes are registered with axum 0.8 **brace** syntax (`crates/http/src/router/routes.rs`):

```rust
.route("/{name}", get(...).post(...).delete(...))
.route("/{name}/document/{docID}", get(handlers::get_document))
```

axum's own test asserts this: `assert_eq!(res.text().await, "/{a}/{b}")` for a router nested at `/{a}` with an inner route `/{b}` (`axum-0.8.9/src/extract/matched_path.rs:176-183`).

So `MatchedPath` yields `/api/v0/collections/{name}` and the table is keyed on `/api/v0/collections/:name`. **They never match.** `normalize_api_version` (line 303) rewrites only the version prefix, not the parameter syntax.

## What actually happens today

Every request to a parameterized route falls through to the `_` arm (`route_permissions.rs:292`):

```rust
_ => {
    tracing::warn!(path = path, method = %method, "Route not in permission table — applying safe default (DocumentRead)");
    RoutePermission::Required(NodePermission::DocumentRead)
}
```

Consequences, in order of how much they matter:

1. **The middleware's permission is wrong for every parameterized route.** `DELETE /collections/{name}` is enforced as `DocumentRead` where the table says `CollectionPatch`; `DELETE /collections/{name}/truncate` as `DocumentRead` where the table says `CollectionTruncate`; `POST /collections/{name}/indexes` as `DocumentRead` where the table says `IndexCreate`, and so on for every `:name`, `:id`, `:docID`, `:index`, `:field` entry.

[...]

Issue #1492, opened 2026-08-17
DELETE /collections/{name} destroys the whole collection where Go deletes only the documents matching a filter


vertexclique:
Found while fixing #1411 by diffing Go's full route table against Rust's router. Triage procedure: #1391.

## Divergence

Go and Rust both serve `DELETE /api/v{0,1}/collections/{name}`. They mean entirely different things.

| | Go v1.0.0 | Rust |
|---|---|---|
| Handler | `DeleteDocumentsWithFilter` (`http/handler_collection.go:511`) | `handlers::delete_collection` (`crates/http/src/router/routes.rs:95`) |
| Request body | `{"filter": {...}}` (`DeleteCollectionRequest`, `http/handler_collection.go:32`) | ignored |
| Effect | deletes the **documents** matching the filter | deletes the **collection and every one of its versions** (`crates/db/src/collection_ops/delete.rs:72`) |
| Response | `{"Count": n, "DocIDs": [...]}` (`client.DeleteResult`, `client/collection.go:198`) | `{}` |

Go's own HTTP client, the reference implementation of the wire contract, builds exactly this request in `Collection.DeleteDocumentsWithFilter` (`http/client_document.go:299-321`): `DELETE` to `apiURL.JoinPath("collections", name)` with the filter as the body.

## Consequence

A Go-compatible client that asks a Rust node to delete a few documents by filter **destroys the entire collection instead**, and gets a success response. There is no error, no warning, and no way for the client to tell.

This is worse than the 404 in #1411. A 404 fails loudly and the caller knows nothing happened. This succeeds and does something far more destructive than what was asked.

## Why the existing audit missed it

#1425 probed this path and recorded it as "present", because its method-based discriminator (an unsupported verb returns 405 when the path exists and 404 when it does not) proves a path **resolves**. It cannot prove the path **means the same thing**. Route-presence parity and route-semantics parity are different properties, and only the first has been checked so far.

## The fix looks cheap

Rust already has a Go-compatible home for collection deletion: `DELETE /api/v0/collections?name=Users&active-only=true`, handled by `delete_collections_by_names` (`crates/http/src/handlers/collections.rs:456`). That is already what the Rust CLI calls (`crates/cli/src/commands/client/http_client/collection.rs:64`), and a workspace-wide search finds **nothing** that calls the parameterized form over HTTP.

[...]

PR #1498, opened 2026-08-17
fix(http): serve Go's wire paths for policy registration and views

### LINKED ISSUE #1411: POST /acp/policy (Rust) does not match Go's wire route POST /acp/document/policy for AddDACPolicy, breaking any Go-compatible client's policy registration call
Found by a systematic Go-parity audit of `http-wire`. Triage procedure: #1391.

## Divergence

Go registers AddDACPolicy at POST /api/v{0,1}/acp/document/policy, and Go's official http.Client (the reference implementation of the wire contract) posts to exactly that path. Rust's router (crates/http/src/router/routes.rs) nests the ACP routes under /acp and registers add_policy at the bare "/policy" sub-path, i.e. POST /api/v{0,1}/acp/policy — there is no /acp/document/policy route registered anywhere in Rust's single production router-construction function (create_router_with_state_and_sync_body_limit, the only call site used by server.rs). All the other ACP paths Go's client calls (acp/document/relationship POST+DELETE, acp/node/relationship POST+DELETE, acp/node/re-enable, acp/node/disable, acp/node/status) do have exact Rust counterparts at matching paths — only document/policy diverges.

## Consequence

Any client written against Go's documented/actual wire API (including Go's own http.Client.AddDACPolicy, or any third-party client mimicking it) that POSTs a policy definition to /api/v0/acp/document/policy against a Rust server gets a 404 Not Found instead of a policy being registered. ACP policy registration over HTTP — a core, security-relevant, advertised feature — silently fails for any Go-compatible client, even though Rust's own internal handler and response shape ({"PolicyID": ...}) are otherwise correctly Go-compatible.

## References

- **Rust:** crates/http/src/router/routes.rs:167 (nested under /acp at line 268, so the live path is /api/v{0,1}/acp/policy); handler doc-comment at crates/http/src/handlers/acp.rs:33 asserting Go-compatibility
- **Go v1.0.0 (`3de01484`):** http/handler_acp.go:338 (router.AddRoute("/acp/document/policy", http.MethodPost, addPolicyDAC, h.AddDACPolicy)); confirmed as the real wire contract by Go's own HTTP client at http/client_acp.go:34 (methodURL := c.http.apiURL.JoinPath("acp", "document", "policy"))

## Verification done

Read the entirety of crates/http/src/router/routes.rs (single production route-registration function, confirmed as the only call site feeding server.rs via crates/http/src/server.rs and crates/http/src/lib.rs) — no /document/policy route exists anywhere in the file, under /acp or otherwise. Also checked crates/http/src/handlers/acp.rs in full; its own doc-comments claim path POST /api/v0/acp/policy, self-consistently confirming no alias exists in the handler-registration wiring.

vertexclique:
Closes #1411.

Go serves three routes Rust answered 404 on, because Rust registered the same handlers at different paths. Fixing only the paths would have left two of the three still broken, so the request contracts are fixed with them.

| Go v1.0.0 | Rust before | Now |
|---|---|---|
| `POST /acp/document/policy` | 404 (served at `/acp/policy`) | serves both |
| `POST /view` | 404 (served at `/views`) | serves both, and accepts Go's `TransformCID` |
| `POST /view/refresh` | 404, then 400 on Go's bodyless request | serves both, with all four of Go's selectors |

Additive on the Rust side: every existing path, body key and client keeps working. Go is untouched.

Three defects the routing fix alone would have shipped:

- **`TransformCID` silently dropped.** Go marshals the view transform under that name; Rust read `Transform` and is not `deny_unknown_fields`, so the view was created without its lens transform and the caller got a 200.
- **Go's own refresh request rejected.** `setDefaultHeaders` sets a JSON content type on every request and `RefreshViews` sends a nil body, which `Json<T>` refuses with 400.
- **A selection silently widened.** `Option<Json<T>>` discards a body sent without a JSON content type, turning a one-view refresh into every view, with no way for the caller to tell.

Rust's CLI now speaks the same wire contract: `view refresh` gained `--collection-id`, `--version-id` and `--get-inactive` to match Go, and the paths come from `defra_http::go_paths` so client, router and tests cannot drift.

Also in this commit, kept together at the operator's direction rather than split:

- route permission tests moved out of `src/` into `tests/`, with the two duplicated exhaustive lists merged into one
- two pre-existing test flakes fixed (`defra-core` identity store, `defra-node` conflict metrics), each proven by stress runs
- pre-existing rustdoc errors fixed across `crypto`, `schema`, `db`, `cli`, `defra-core` and `http`

**Gate:** fmt 0, clippy 0, rustdoc 0 on every touched crate, 5036 passed / 0 failed across 236 binaries. `just doc` workspace-wide is still red on a pre-existing 80-error backlog in `crates/storage`, untouched here.

**Not verified:** the `go_*` halves of the new cross-runtime tests need a Go binary and run in CI. Their `rust_*` halves pass locally.

Six further Go-parity route defects found during this work are filed separately as #1492 to #1497, one of them data-loss class.


vertexclique:
There are bunch of ACP related issues, I have created them to be on par with golang. This work just surfaced them.
cc/ @jackzampolin

crates/db/src/view_ops.rs
```diff
@@ -79,26 +132,22 @@ async fn delete_prefix(store: &NamespaceView, prefix: Vec<u8>) -> Result<()> {
...
         let views_to_refresh: Vec<_> = collections
             .iter()
-            .filter(|col| col.query.is_some() && col.is_materialized && !col.is_embedded_only)
-            .filter(|col| {
-                names_filter
-                    .map(|names| names.contains(&col.name))
-                    .unwrap_or(true)
-            })
+            .filter(|col| is_refreshable_view(col))
+            .filter(|col| options.selects(col))
```
iverc, line 180: A missing `version_id` filters this scan to an empty list and returns `Ok(())`. Go performs a direct version lookup and propagates not-found, so a typo currently looks like a successful refresh. Please validate an explicit version ID before filtering.

crates/db/src/view_ops.rs
```diff
@@ -79,26 +132,22 @@ async fn delete_prefix(store: &NamespaceView, prefix: Vec<u8>) -> Result<()> {
...
+    /// Clears and rebuilds the view cache for each materialized view the
+    /// options select. Without options, every materialized view is refreshed.
+    pub async fn refresh_views(&self, options: RefreshViewsOptions) -> Result<()>
     where
         S: 'static,
     {
-        // Get all collections
-        let collections = self.get_all_active_collections_internal()?;
+        let collections = if options.needs_all_versions() {
+            self.get_all_collection_versions().await?
```
iverc: `needs_all_versions()` can include inactive versions here, but `build_view_cache` still constructs its query runner from active schemas only. Versions share the same cache namespace, so refreshing an inactive renamed version can clear the live cache and then fail to resolve the old name; a same-name version is rebuilt from the active definition instead of the requested one. We should build with the selected version's schema, or avoid inactive selection until the builder supports it.

crates/cli/src/commands/client/http_client/view.rs
```diff
@@ -36,10 +76,9 @@ impl HttpClient {
         self.post_json(&url, &body).await
     }
 
-    pub async fn view_refresh(&self, names: Option<Vec<String>>) -> Result<JsonValue> {
-        let url = format!("{}/api/v0/views/refresh", self.base_url());
-        let body = ViewNamesRequest { names };
-        self.post_json(&url, &body).await
+    pub async fn view_refresh(&self, selectors: &ViewRefreshSelectors) -> Result<JsonValue> {
+        let url = format!("{}{}", self.view_refresh_url(), selectors.query_string());
+        self.post_json(&url, &serde_json::json!({})).await
```
iverc: Go's refresh handler returns 200 with an empty body, but `post_json` always deserializes response JSON. Against a Go node, the refresh succeeds and the CLI then reports an EOF/deserialization error. Could this use the void-response path so both Go's empty response and Rust's `{}` response are accepted?

PR #1502, opened 2026-08-17
fix(p2p): centralize DAG ingestion and merge ownership

jackzampolin:
## Summary

This change gives replicated DAG writes one bounded, observable ownership path:

1. A sender announces one signed current head per scope and peer.
2. The receiver durably registers that root before acknowledging ownership transfer.
3. The paced pending-DAG clock completes missing data through rooted selective CAR recovery.
4. Completed roots enter one globally serialized batch merge writer, regardless of transport or embedded entrypoint.

Sender durability is now document/collection dirty markers plus current-head rederivation, rather than CID or payload delivery ledgers. Stale acknowledgements cannot clear newer work, and legacy retry obligations migrate without being discarded.

The branch also exposes read-only collection-version observation from embedded nodes through the established HTTP control surface. Existing NAC guards and development-mode restrictions remain in force.

## Why

P2P writes previously had overlapping owners. Senders expanded one logical update into ordered per-block PushLogs while receivers also owned durable missing-DAG recovery. Runtime entrypoints could additionally choose different merge concurrency models. That amplified admission, retry, storage, and transaction-conflict pressure and allowed field or metadata blocks to be presented as standalone document heads.

The new boundary is one logical head hint, one durable receiver obligation, one paced recovery owner per root, and one serialized P2P merge writer. The merge writer batches independent roots in a shared transaction while preserving one mutation owner for overlapping state.

Progresses #1116 and resolves the outbound field-head class described by #1450.

## Review map

- `proofs/tla/SyncOwnership*`: obligation conservation, crash/restart, provider qualification and routability, bounded admission, current-head retry, and merge-writer ownership.
- `crates/p2p/src/sync/event_dispatcher.rs`: shared exhaustive classification and bounded admission for transport/host entrypoints.
- `crates/p2p/src/sync/coordinator`: durable register-then-ack, authenticated Iroh origin/hop/provider separation, rooted CAR authority, single-flight recovery, and bounded head-hint delivery.
- `crates/p2p/src/sync/pending_store.rs`: durable receiver obligations, immutable origin binding, and bounded qualified alternate providers.
- `crates/storage/src/stores/peerstore.rs`: presence-only document/collection markers, atomic peer scheduling, fair cursor rotation, and one-way legacy retry migration.
- `crates/p2p-adapter/src/retry.rs`: the single durable marker/rederive sweep shared by CLI, embedded, and defra-node.
- `crates/db-merge/src/push_docs*`: complete current-head frontier rederivation without outbound DAG expansion; marker registration precedes selective-CAR authority installation.
- `crates/db-merge/src/broadcast_mutator` and `crates/p2p/src/sync/replication/loop_runner.rs`: post-commit obligation registration and the serialized batch merge boundary.
- `tools/integration-test/tests` and coordinator tests: deterministic ownership A/B, restart, overload, compatibility, provider binding, and quiescence fences.

## Formal and compatibility boundaries

`SyncOwnership` checks obligation conservation across announce, durable registration, acknowledgement, nack, retry, merge, and crash. Green configurations cover document and collection supersession plus an Iroh relay chain where the immutable signed origin is distinct from a later authenticated direct CAR provider.

Retained red configurations cover volatile registration, payload ledgers, document-only markers, duplicate fetch owners, stale acknowledgement clearing, volatile serving authority, provider rebinding, forged or unroutable origins, root-only relays, recursive-first recovery, every-root dispatch, cancellation on progress, and parallel merge writers. The per-document-writer red configuration keeps the merge-serialization proof non-vacuous.

The model registry labels the SyncOwnership bindings as protocol boundaries. Concrete Rust behavior is gated separately by the repository integration tests; it is not misrepresented as an executable conformance harness.

The deterministic ownership A/B holds topology and admission bounds constant. Its full-DAG arm requires sender nack/re-offer to recover the missing root; its head-hint arm transfers one durable receiver obligation and converges through CAR.

Rust keeps the Go document retry ladder (30s, 1m, 2m, 4m, 8m, 16m, 32m cap) and deterministic bounded jitter. Signed PushLog verification preserves the exact received unsigned CBOR map, including unknown future fields, while retaining rolling old/new signature compatibility.

## Review-driven closure

- Retry state transitions are atomic and never hold a storage writer across network or searchable-encryption work.
- Capacity nacks rotate a durable fair cursor without advancing the failure ladder or hammering adjacent scopes.
- Replay registers every document marker before CAR authority admission, so a full authority table cannot abandon tail documents.
- Remote searchable-encryption and management requests pass through the same admission classifier.
- Telemetry emits both the new `push_marker` label and the legacy `push_ledger` label for one compatibility window.
- The #1154 restart fence now takes a bracketed stable cross-process snapshot before asserting exact ownership conservation.

## Performance evidence

The retry-marker benchmark measures the remaining explicit O(documents × peers) sender persistence boundary:

- 1 document × 1 peer: ~1.69µs
- 1 document × 4 peers: ~7.65µs
- 10 documents × 1 peer: ~12.78µs
- 10 documents × 4 peers: ~71.28µs
- 100 documents × 1 peer: ~218µs
- 100 documents × 4 peers: ~2.20ms

The merge DAG benchmark remains green after serialization/batching (roughly 20µs for depth 1/one field, 164µs at depth 8, and 2.07ms at depth 64; four-field depth 64 is ~5.52ms), with Criterion reporting no regression.

## Validation

Local validation on the pushed tree rebased over current `origin/main`, including the landed P2P adapter compartmentalization:

- all 131 expected TLA green/red verdicts, Lean build, and conformance registry tests;
- 383 P2P, 367 storage, 119 db-merge, 192 db, and 51 P2P-adapter unit tests;
- admission, fairness, admission-restart, bounded-backlog, deep-catchup, push-storm, and the ≥500-document #1154 crash/restart fences;
- both CI-gated Iroh ownership regressions: relay-chain propagation and fanout above the CAR worker bound;
- workspace clippy, formatting, and diff checks;
- the exact previously failing Rust/Go view-path compatibility test.

The old Go-compat run passed all 25 mixed P2P cases and failed only because current main rejected `get_inactive=true` even when the selected view was active. The final standalone compatibility commit now matches the Go inclusion-selector behavior for that active-only case while continuing to reject an actually selected inactive materialized view.

## Non-gating follow-ups

- #1538: durably recover a transient committed head-marker registration failure;
- #1539: remove the unused duplicate `SyncManager::handle_bitswap_complete` path;
- #1537: classify and terminalize deterministic non-`Rejected` pending-DAG merge failures.


jackzampolin:
Post-review non-gating follow-ups are now tracked separately:

- #1538 — durably recover transient committed head-marker registration failures
- #1539 — remove the dead duplicate `SyncManager::handle_bitswap_complete` path
- #1537 — bound deterministic non-`Rejected` pending-DAG merge failures

These are intentionally excluded from the stage-3 correctness scope so #1502 remains focused on sender demotion, durable register-before-ack, bounded provider rotation, and serialized receiver completion.

jackzampolin:
Rebased cleanly onto current `origin/main` at `e27fc944` (including the landed P2P adapter/compartmentalization work) and force-pushed with lease as `c340eaf2`.

The rebased tree is byte-for-byte identical to the independently approved `547a91a0` tree; only commit ancestry changed. Post-rebase validation:

- `proofs/verify-all.sh`: all 130 TLA verdicts matched; Lean and conformance bindings green
- `cargo test -p p2p --lib`: 382/0
- `cargo test -p p2p --test sync_manager_tests`: 37/0
- exact Iroh chain recovery and bounded-fanout tests: 1/0 each
- `cargo fmt --all -- --check` and `git diff --check`: green

PR mergeability currently reports clean. Fresh GitHub checks are being scheduled for the rewritten head.

crates/p2p/src/host/event.rs
```diff
@@ -172,3 +172,26 @@ pub enum HostEvent {
...
+            | Self::BitswapBlockReceived { .. }
+            | Self::DocSyncReply { .. }
+            | Self::BranchableSyncReply { .. }
+            | Self::CarFetchResponse { .. } => DispatchClass::Completion,
+            Self::PushLogRequest { .. }
+            | Self::GossipMessage { .. }
+            | Self::TwoStreamRequest { .. }
+            | Self::DocSyncRequest { .. }
+            | Self::BranchableSyncRequest { .. } => DispatchClass::Admission,
+            _ => DispatchClass::Inline,
```
vertexclique: **Bug: `_ => Inline` puts `PeerConnected` on the sole transport drain, where it can block on a network push.**

`PeerConnected` lands in this arm, and `run_event_dispatcher` runs Inline events with `handler(...).await` directly on the loop (`sync/event_dispatcher.rs:215`). Every runtime's handler awaits `activate_retry_peer` at that point (`embedded/src/node_tasks.rs:73` and `:217`, `cli/src/commands/start/server_p2p/libp2p.rs:211`, `iroh.rs:286`, `defra-node/src/p2p_runtime.rs:378`), which takes `retry_peer_lock(peer).write_arc().await` plus a read-modify-write txn with up to 4 conflict retries (`storage/src/stores/peerstore.rs:563`).

That same lock is held across network I/O: `run_retry_pass` holds `_transition_guard` across `regenerate_and_push_se_artifacts` (`p2p-adapter/src/retry.rs:229` through `:239`), a replicator fan-out with no timeout - the 15s timeout at `:228` wraps only `replay`.

So one peer reconnecting parks the entire transport event loop (gossip, PushLog admission, CAR serving, every peer) for the duration of an SE artifact push. This is `CompletionMode = "BlockingHostEvent"`, which `SyncOwnership.tla:443` marks RED, and whose green comment claims the shipped scheduler "never waits for a request-worker slot ... and always drains blocks, replies, and completions".

Two fixes, both wanted: classify `PeerConnected` / `PeerDisconnected` as `Completion`, and drop `_transition_guard` before the SE push.

crates/p2p/src/transport.rs
```diff
@@ -237,16 +240,59 @@ pub enum TransportEvent<ResponseToken> {
...
+            | Self::BitswapBlockReceived { .. }
+            | Self::DocSyncReply { .. }
+            | Self::BranchableSyncReply { .. }
+            | Self::CarFetchResponse { .. } => DispatchClass::Completion,
+            Self::PushLogRequest { .. }
+            | Self::GossipMessage { .. }
+            | Self::TwoStreamRequest { .. }
+            | Self::DocSyncRequest { .. }
+            | Self::BranchableSyncRequest { .. } => DispatchClass::Admission,
+            _ => DispatchClass::Inline,
```
vertexclique: **Bug: this classifier is byte-identical to `HostEvent::dispatch_class` (`host/event.rs:178-196`), and `_` is fail-open.**

Two copies that must agree. Add a variant to one enum and forget the other and it silently becomes `Inline`, which runs on the sole event drain.

13 variants land in `_` today: `BitswapProgress`, `Listening`, `ManageRequest`, `ManageReply`, `ManageQueryRequest`, `ManageQueryReply`, `PeerConnected`, `PeerDisconnected`, `PeerSubscribed`, `PeerUnsubscribed`, `SEArtifactsReceived`, `SEQueryRequest`, `SEQueryReply`. Four of those are inbound remote requests with no admission control at all.

Suggest one classifier (a `macro_rules!` over the shared variant names), exhaustive, with no `_` arm, so a new variant fails to compile until it is classified.

crates/p2p/src/sync/event_dispatcher.rs
```diff
@@ -0,0 +1,640 @@
...
+                    break;
+                };
+                let class = event.dispatch_class();
+                if class == DispatchClass::Completion {
+                    // Completion producers are themselves bounded (receiver
+                    // fetches, protocol replies). If the fixed completion set
+                    // is full, wait for one owner to finish before retaining
+                    // more work outside the transport channel.
+                    let permit = Arc::clone(&completion_slots)
+                        .acquire_owned()
```
vertexclique, line 245: **Bug: the sizing premise for `MAX_ACTIVE_COMPLETIONS` does not hold.**

This `.await` blocks the drain loop, justified by "Completion producers are themselves bounded (receiver fetches, protocol replies)" against `DEFAULT_MAX_CONCURRENT_DAG_FETCHES = 4`.

But `GossipRawMessage` is also Completion-classed on both transports (`transport.rs:283`, `host/event.rs:183`) and is remote-driven with no pre-admission bound. Its rate limit is checked inside `handle_transport_event` (`event_handler/mod.rs:412`), which runs after the slot is already taken.

Fill 16 and the dispatcher stops draining. That is `CompletionMode = "WorkerSaturated"`, also RED in the model. I am not claiming this is exploitable today, the handlers are fast; what does not hold is the stated bound rationale.

crates/p2p/src/sync/event_dispatcher.rs
```diff
@@ -0,0 +1,640 @@
...
+                        .expect("completion semaphore is never closed");
+                    let completion_handler = handler.clone();
+                    let activity = diagnostics.enter(DispatchActivityKind::Completion);
+                    tasks.spawn(async move {
+                        let _permit = permit;
+                        let _activity = activity;
+                        completion_handler(event, DispatchAdmission::Admitted).await;
+                    });
+                } else if class == DispatchClass::Inline {
+                    handler.clone()(event, DispatchAdmission::Admitted).await;
```
vertexclique, line 256: **Bug: Inline handlers block the sole event drain, and nothing enforces that they are cheap.**

The module doc says "lightweight state transitions drain inline", but the classification is `_ => Inline` (`transport.rs:294`, `host/event.rs:194`), so `PeerConnected`, `ManageRequest`, `ManageQueryRequest`, `SEQueryRequest` and `SEArtifactsReceived` all land here. `PeerConnected` awaits a per-peer write lock and a storage txn; the `Manage*` and `SEQuery*` variants are inbound remote requests with no admission control.

The "Inline is lightweight" contract is not enforced anywhere. Making the classifiers exhaustive is the cheap fix; see the notes on `host/event.rs:194` and `transport.rs:294`.

vertexclique (commented):
Thorough pass over the whole branch, built and tested locally at `c340eaf2` against `e27fc944`.

## Verified green

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all -- -D warnings` (CI's exact command) | clean |
| `cargo check -p p2p --no-default-features --features iroh-transport` | clean |
| `cargo check -p db-merge --no-default-features --features native` | clean |
| `assert-defra-node-graph.sh` | ok (529 / 314 / 418) |
| `cargo test --workspace --exclude integration-test --exclude ffi-test --exclude conformance` | **1872 passed, 0 failed, 59 ignored** across 81 binaries |
| `just doc` | red for `p2p`, `db-merge`, `storage`, but every failure pre-dates this branch; the PR is net -1 broken link (`storage/src/keys/peerstore.rs`) |

Every number in the description that I could check is exact: 382 p2p lib, 37 sync-manager, 119 db-merge, 77 query lib, and 130 TLA runs (43 GREEN + 87 RED, counted from `run-all.sh`). That is unusually disciplined and worth saying out loud.

Not run here: the Iroh integration matrix (232/0), the TLA suite itself, Lean, conformance bindings.

## Summary

The ownership boundary is the right one, and `pending_store.rs` is the cleanest part of the change: bounded conflict retry, atomic `replace_scope_head`, bounded alternates, legacy-decode test. The `CollectionVersionOperations` split is textbook least privilege, with a test asserting DELETE returns 503. The `AnnouncedBlockKind` gate for #1450 (store dependency blocks, never admit them as heads) is the right shape.

What I would want fixed before merge, all commented inline:

- The dispatcher's `_ => Inline` arm puts `PeerConnected` and four inbound remote request types on the sole event drain, where `PeerConnected` blocks on a per-peer write lock that `run_retry_pass` holds across an untimed network fan-out. This is the `BlockingHostEvent` transition `SyncOwnership.tla` marks RED.
- Receiver `AT_CAPACITY` nacks now advance the sender's durable ladder, because `reschedule_persisted_push_retry` was deleted while this PR adds new capacity nacks.
- `clear_retry_peer` lost its atomicity, so a marker can be orphaned with no peer schedule and never swept again. That is `INV_ObligationConservation`.
- The marker sort changed from due-time to lexicographic, so a consumer that breaks after 3 failures starves the tail permanently.
- `registry.rs` declares a `Behavioral` binding for the new family, but `proofs/tests/behavioral/` is untouched and the README still says 19 behavioral tests.

Two changes are not defects but are undisclosed, and belong in the description: `MC_MergeQueue_Green` moved to a global merge lock (and the anti-vacuity probe that guarded exactly that was repurposed), and the retry ladder was truncated from 12h to 32m with its jitter removed.

Also worth a number somewhere: batch commit now does a synchronous durable round-trip per document per replicator, and P2P merges are globally serialized. Neither has a before/after in the PR.

Review-map nit: it points at `crates/db-merge/src/merge_queue.rs`, which exists on neither this branch nor `main`. That claim's real home is `broadcast_mutator/{mod,batch}.rs` plus `sync/replication/loop_runner.rs`.

Housekeeping: four files cross 1000 lines here (`coordinator/mod.rs` 953->1326, `event_handler/mod.rs` 945->1300, `manager/process/pushlog.rs` 996->1356, `iroh/transport.rs` 945->1135), and 8 new `#[cfg(test)] mod` blocks landed in `src/` while `crates/p2p/tests/`, `crates/storage/tests/` and `crates/defra-core/tests/` all exist.

crates/p2p-adapter/src/retry.rs
```diff
@@ -0,0 +1,372 @@
...
+            return;
+        }
+    };
+
+    for (peer_id_str, info_bytes) in peers {
+        let retry_guard = match peerstore.acquire_replicator_retry_guard(&peer_id_str).await {
+            Ok(Some(guard)) => guard,
+            Ok(None) | Err(_) => continue,
+        };
+        drop(retry_guard);
```
vertexclique: **Pitfall: the guard is acquired and immediately dropped.**

`acquire_replicator_retry_guard` documents "The guard must be retained through transport replay and any resulting marker update" (`storage/src/stores/peerstore.rs:119`). Here it is only an existence check, which `has_replicator` says directly and without taking an exclusive per-peer lock.

The consequence is at `:203` below: `peer_retry.bump()` and `update_retry_document` run unguarded, and `update_retry_document` is now an unconditional `txn.set` (`peerstore.rs:537`; the old `update_retry_document_once` compared first). `activate_retry_peer` sets `defer_for(ZERO)` under the write lock, so this unguarded bump can clobber a reconnect activation. It is the disconnected branch, which is exactly where a reconnect is likely.

crates/p2p-adapter/src/retry.rs
```diff
@@ -0,0 +1,372 @@
...
+            continue;
+        }
+
+        let connected = transport.connected_peers().await.unwrap_or_default();
+        if !connected.contains(&peer_id) {
+            // Connectivity is part of the due retry attempt. Do not create a
+            // second two-second redial clock for markers whose ladder has not
+            // elapsed yet.
+            redial_replicator(&peerstore, transport, &peer_id).await;
+            peer_retry.bump();
```
vertexclique: **Pitfall (same finding as the note at `:172`): unguarded blind write of the peer schedule.**

No retry guard is held here. `update_retry_document` (`storage/src/stores/peerstore.rs:537`) is now an unconditional `txn.set`, so this can overwrite a concurrent `activate_retry_peer` that just set `defer_for(ZERO)` on reconnect, leaving the peer deferred despite being reachable again.

crates/p2p-adapter/src/retry.rs
```diff
@@ -0,0 +1,372 @@
...
+                tokio::time::timeout(std::time::Duration::from_secs(15), replay).await;
+            let _transition_guard =
+                match peerstore.acquire_replicator_retry_guard(&peer_id_str).await {
+                    Ok(Some(guard)) => guard,
+                    Ok(None) | Err(_) => break,
+                };
+            match replay_result {
+                Ok(Ok(())) => {
+                    if let Some(repusher) = se_repusher {
+                        repusher
```
vertexclique, line 232: **Bug (second half of the `host/event.rs:194` note): the peer retry guard is held across a network fan-out.**

`_transition_guard` is acquired at `:229` and is still held here. `regenerate_and_push_se_artifacts` pushes SE artifacts to every matching replicator with no timeout - the 15s timeout at `:228` covers only `replay`.

Two consequences:

1. `PeerConnected` for this peer is handled inline on the transport drain (`p2p/src/sync/event_dispatcher.rs:215`) and blocks on this same lock, stalling the whole event loop.
2. If the 1024-slot failure channel fills, `report_push_event` loops forever on `reserve()` (`p2p/src/sync/coordinator/push_worker.rs`, 5s warn but no max attempts) while the recorder waits on this guard. Circular wait with no escape.

The guard is documented as serializing storage transitions. A network push is not one. Suggest dropping it before this call and re-acquiring for the marker update below.

crates/p2p-adapter/src/retry.rs
```diff
@@ -0,0 +1,372 @@
...
+                    }
+                }
+                Ok(Err(error)) => {
+                    tracing::warn!(
+                        doc_id = %marker.doc_id,
+                        %peer_id,
+                        %error,
+                        "retry push failed"
+                    );
+                    marker.retry_info.bump();
```
vertexclique: **Bug: receiver capacity backpressure now advances the durable retry ladder.**

This bumps on any `Err`. `retry_doc` surfaces the nack verbatim - `db-merge/src/push_docs.rs:676` produces `"peer rejected replay after N successful block(s): at capacity: receiver is saturated, back off"` - so an `AT_CAPACITY` nack advances the ladder and counts toward `fast_failures >= 3` five lines down.

This PR deleted the one thing that handled it: `reschedule_persisted_push_retry` and its regression test `capacity_backpressure_does_not_advance_the_failure_ladder` (`p2p/src/sync/mod.rs`, -45). It also adds new capacity nacks: `handle_transport_event_with_admission` returns `AT_CAPACITY_MESSAGE` at `event_handler/mod.rs:300`, `:317`, `:333`, `:350`. So capacity nacks got more common while the code that handled them correctly went away.

`is_at_capacity_message` still exists and is still used on the live path (`push_worker.rs:360`). Suggest restoring the branch here:

```rust
if is_at_capacity_message(&error) {
    marker.retry_info.defer_for(PERSISTED_RETRY_SWEEP_INTERVAL)
} else {
    marker.retry_info.bump()
}
```

crates/p2p-adapter/src/retry.rs
```diff
@@ -0,0 +1,372 @@
...
+    }
+}
+
+async fn finish_peer<S: storage::corekv::Store>(
+    peerstore: &storage::stores::Peerstore<S>,
+    peer_id: &str,
+    complete: bool,
+) {
+    let status = if complete {
+        let _ = peerstore.clear_retry_peer(peer_id).await;
```
vertexclique, line 281: **Bug (companion to the `storage/src/stores/peerstore.rs:672` note): this call is unguarded, while the same call at `:91` is guarded.**

`clear_retry_peer` does its emptiness check and its delete in two separate transactions, so a `register_scope_marker` landing between them orphans the marker with no peer schedule. Holding the peer guard across this call, or making `clear_retry_peer` atomic in one txn, closes it.

Note `clear_retry_peer` cannot take the lock internally without deadlocking the guarded call site at `:91`, so the one-txn fix is the safer of the two.

crates/storage/src/stores/peerstore.rs
```diff
@@ -790,72 +668,18 @@ impl<S: Store> Peerstore<S> {
             .collect())
     }
 
-    /// Stop sweeping a peer once no retry is pending. Dormant newest-head
-    /// watermarks deliberately survive: they cover volatile live work and
-    /// are promoted by `activate_dormant_push_retries` after a restart.
+    /// Stop sweeping a peer once no document or collection marker remains.
     pub async fn clear_retry_peer(&self, peer_id: &str) -> Result<()> {
```
vertexclique, line 696: **Bug: check-then-delete is not atomic, so an obligation can be stranded.**

`load_scope_markers` runs in one read txn and the delete in a separate write txn. `register_scope_marker` skips writing `id_key` when it already exists, so a marker landing between the two txns leaves `/rep/retry/doc/{peer}/{doc}` with no `/rep/retry/id/{peer}`. `get_replicator_retry_peers` iterates `/rep/retry/id/` only (`:638`), so that peer is never swept again until an unrelated failure recreates the key.

Reachable: `finish_peer` calls this without the retry guard (`p2p-adapter/src/retry.rs:283`), while the other call site at `retry.rs:91` does hold it.

This is the invariant `INV_ObligationConservation` claims. The model treats the transition as atomic; the code does not. The old `clear_retry_peer_once` did the scan and the delete in one txn - suggest restoring that shape.

crates/storage/src/stores/peerstore.rs
```diff
@@ -264,478 +261,359 @@ impl<S: Store> Peerstore<S> {
...
+                    doc_id: String::new(),
+                    collection_id: collection_id.to_string(),
+                    scope: super::RetryScope::CollectionCommit,
+                    retry_info: retry_info.clone(),
+                });
             }
         }
-        results.sort_by_key(|retry| retry.retry_info.next_retry_unix);
-        Ok(results)
+        result.sort_by(|a, b| {
```
vertexclique, line 470: **Bug: this ordering starves the tail of the marker set.**

This replaces `results.sort_by_key(|retry| retry.retry_info.next_retry_unix)`, whose deleted comment read: "Retries are returned in next-attempt order so bounded consumers cannot starve later store keys."

All markers for a peer now share one `RetryInfo`, so per-marker backoff is gone, and the consumer breaks after 3 failures or one timeout (`p2p-adapter/src/retry.rs:254` and `:264`). With N dirty docs where the lexicographically-first 3 fail permanently, docs 4..N never replay.

Suggest rotating the start offset per pass, or keeping a per-marker attempt counter so a failing marker falls to the back.

crates/storage/src/stores/peerstore.rs
```diff
@@ -264,478 +261,359 @@ impl<S: Store> Peerstore<S> {
...
+                        oldest.min(info.next_retry_unix)
+                    }),
+            );
         }
-        txn.commit().await?;
-        Ok(is_same_dormant)
+        Ok(stats)
     }
 
     pub async fn update_retry_document(
```
vertexclique: **Pitfall: this lost its compare-and-set.**

`update_retry_document_once` used to check `is_current` before writing. This blind-writes, so a stale snapshot taken from `get_replicator_retry_peers` can overwrite a newer schedule - including the `defer_for(ZERO)` that `activate_retry_peer` (`:556`) writes under the peer write lock.

Note the asymmetry: `activate_retry_peer` takes the lock internally, this does not, and its caller at `p2p-adapter/src/retry.rs:204` holds no guard.

crates/storage/src/stores/peerstore.rs
```diff
@@ -264,478 +261,359 @@ impl<S: Store> Peerstore<S> {
...
-    /// inherit the old peer-level schedule and are rewritten on their next
-    /// failure/update. Retries are returned in next-attempt order so bounded
-    /// consumers cannot starve later store keys.
-    pub async fn get_retry_documents(
-        &self,
-        peer_id: &str,
-    ) -> Result<Vec<super::PersistedPushRetry>> {
-        let fallback = self
+    async fn load_scope_markers(&self, peer_id: &str) -> Result<Vec<super::PushRetryMarker>> {
+        self.migrate_push_retry_markers(Some(peer_id)).await?;
```
vertexclique: **Smell: the migration runs on every read, not once.**

`load_scope_markers` is the read path for every retry pass for every peer, so this opens a write txn and does two full prefix scans even when there is nothing to migrate, collecting all matching keys into `Vec`s first.

`migrate_legacy_push_retries` already runs once at startup (`p2p-adapter/src/retry.rs:304`), which is the right place. Suggest dropping this call; if it has to stay for stores read before that runs, page the scan rather than materializing the keyspace.

crates/storage/src/stores/retry_info.rs
```diff
@@ -3,115 +3,102 @@
...
 use std::time::Duration;
-use std::{hash::Hash, hash::Hasher};
 use web_time::{SystemTime, UNIX_EPOCH};
 
-/// Exponential backoff intervals in seconds, matching Go's seconds-to-hours retry ladder.
-pub const RETRY_INTERVALS_SECS: &[u64] = &[
-    30, 60, 120, 240, 480, 960, 1920, 3600, 7200, 14400, 28800, 43200,
-];
+/// Go-compatible document retry ladder: 30s through a 32-minute cap.
+pub const RETRY_INTERVALS_SECS: &[u64] = &[30, 60, 120, 240, 480, 960, 1920];
```
vertexclique, line 13: **Smell: the ladder cap dropped from 12 hours to 32 minutes, and the PR body does not mention it.**

Was `[30, 60, 120, 240, 480, 960, 1920, 3600, 7200, 14400, 28800, 43200]` with the comment "matching Go's seconds-to-hours retry ladder"; now 7 entries with "Go-compatible document retry ladder". Both cite Go, so one of them is wrong. Could you cite the Go source file and symbol? I could not verify it locally.

Practical effect: a peer that is down for hours is now retried every 32 minutes indefinitely instead of backing off, which cuts against the bounded-admission framing in the description.

crates/storage/src/stores/retry_info.rs
```diff
@@ -142,25 +127,13 @@ impl RetryInfo {
...
             .as_secs();
         let idx = (self.num_retries as usize).min(RETRY_INTERVALS_SECS.len() - 1);
         let cap = RETRY_INTERVALS_SECS[idx];
-        let floor = (cap / 2).max(1);
-        let mut hasher = std::collections::hash_map::DefaultHasher::new();
-        retry_key.hash(&mut hasher);
-        self.num_retries.hash(&mut hasher);
-        let delay = floor + hasher.finish() % (cap - floor + 1);
-        self.next_retry_unix = now + delay;
+        self.next_retry_unix = now + cap;
```
vertexclique: **Smell: the deterministic jitter was removed.**

`bump_for(retry_key)` is gone and this sets exactly `now + cap`. Its deleted doc said the jitter "preserves the exponential cap while preventing a fan-out failure from retrying every peer in lockstep".

Jitter matters more after this change, not less: the schedule is now peer-level rather than per (peer, CID), so a fleet-wide failure aligns every peer's clock and they all wake together. Not in the PR body either.

crates/p2p/src/message/pushlog.rs
```diff
@@ -448,9 +551,48 @@ impl PushLogBroadcast {
...
+            // libp2p replaces this hint with the signed native author.
+            source_peer_id: (!req.sender_id.is_empty()).then(|| req.sender_id.clone()),
+            origin_signature: None,
+            authenticated_source_peer_id: None,
+            authenticated_origin_peer_id: None,
         }
     }
 
+    /// Bytes covered by the Iroh origin signature.
+    pub(crate) fn origin_signing_bytes(&self) -> Result<Vec<u8>, defra_core::cbor::Error> {
```
vertexclique, line 574: **Pitfall: re-serializing the decoded struct means the signature does not cover unknown fields.**

`decode_gossip_payload` deliberately tolerates unknown future CBOR fields, and there is a test for it (`cbor_gossip_payload_tolerates_unknown_future_fields`). The moment any peer adds a serialized field to `PushLogBroadcast`, an older peer decodes it, drops the field, re-encodes here without it, and verification fails. That breaks the exact rolling upgrade this envelope exists to survive.

There is also no domain separator and no topic or nonce binding, so a signed hint stays replayable on any topic indefinitely.

Suggest signing the received bytes with the signature field elided, or an explicit canonical tuple prefixed with a context string.

Minor: clearing `authenticated_source_peer_id` / `authenticated_origin_peer_id` here is a no-op, they are `#[serde(skip)]`.

crates/p2p/src/message/pushlog.rs
```diff
@@ -486,6 +628,14 @@ impl PushLogBroadcast {
                 postcard::from_bytes::<Self>(payload)
```
vertexclique: **Smell: this arm is unreachable.**

`source_peer_id` and `origin_signature` carry `skip_serializing_if`, which postcard honors on the way out but cannot reconstruct on the way in (it is positional). So `postcard::to_allocvec::<PushLogBroadcast>` emits the legacy 5-field shape, `from_bytes::<Self>` always hits an unexpected end, and it falls through to the legacy arm below. `encode_gossip_payload` is CBOR-only, so no producer exists either.

`crates/p2p/tests/message_tests.rs:528` (`test_decode_gossip_payload_from_postcard_broadcast`) therefore now exercises the legacy arm rather than what its name says.

Suggest deleting this arm and renaming that test. Worth a line noting `PushLogBroadcast` is no longer postcard-round-trippable, so nobody reaches for `to_allocvec` later.

crates/db-merge/src/broadcast_mutator/batch.rs
```diff
@@ -347,13 +342,27 @@ impl<S: Store + 'static, B: Blockstore + 'static, T: P2PTransport> MutationBatch
 
         let pending_broadcasts = std::mem::take(&mut *self.pending_broadcasts.lock().await);
         if !pending_broadcasts.is_empty() {
+            let mut marker_errors = Vec::new();
+            for pending in &pending_broadcasts {
+                if let Err(error) = Self::register_pending_static(&self.sync, pending).await {
```
vertexclique, line 351: **Pitfall: this registers one document at a time on the synchronous commit path.**

Each `register_pending_static` calls `coalesce_replicator_push`, which for every replicator does `report_observed_head` -> bounded-channel `reserve().await` plus `durable_rx.await` -> recorder task -> per-peer write lock -> storage txn (`p2p/src/sync/coordinator/broadcast.rs:207-237`, `push_worker.rs:115-160`). `push_jobs` also re-lists replicators per document (`broadcast.rs:240`).

That is O(N docs x R replicators) sequential round-trips before `commit()` returns. The description says ordered batching "amortiz[es] transaction work", but registration is strictly per row, which is the thing batching was supposed to fix.

Suggest one batched registration carrying every `(peer, scope)` for the batch, written in one txn. And since this puts a durable handshake on every mutation commit, a before/after write-latency number belongs in the PR.

crates/db-merge/src/broadcast_mutator/batch.rs
```diff
@@ -347,13 +342,27 @@ impl<S: Store + 'static, B: Blockstore + 'static, T: P2PTransport> MutationBatch
...
+                async move {
                     for pending in pending_broadcasts {
                         Self::broadcast_pending_static(&sync, pending).await;
                     }
-                });
+                },
+            );
+            if !marker_errors.is_empty() {
+                return Err(query::error::QueryError::execution(format!(
+                    "committed batch has undurable P2P head markers: {}",
```
vertexclique, line 366: **Pitfall: `commit()` returns `Err` after the batch has already committed and the broadcast task is spawned.**

The message says so itself. A caller treating `Err` from a commit as "did not commit" will retry and double-apply.

I see this is #1538 and marked non-gating, which is fine, but the method's doc comment should state that an `Err` from here means "committed, markers not durable" so callers do not resolve it the usual way.

crates/db-merge/src/push_docs_common.rs
```diff
@@ -111,27 +150,171 @@ pub(crate) async fn load_latest_composite_head_cids<R: Reader + ?Sized, B: Reade
...
+            heads.push(cid);
         }
-        let _ = iter.close().await;
     }
+    Ok(heads)
+}
 
-    cids
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) async fn resolve_collection_id_for_doc<S: storage::corekv::Store>(
```
vertexclique, line 182: **Smell: O(collections) with a storage read each, per document, on the retry path.**

This scans `list_collections()` and calls `require_persisted_collection_short_id` per collection to reverse one `collection_short_id`. With 100 collections and 1000 dirty docs that is 100k systemstore reads per pass.

A forward index clearly exists. Suggest a reverse lookup (short id -> collection id) rather than a linear scan.

crates/http/src/router/state.rs
```diff
@@ -391,10 +421,22 @@ impl AppStateBuilder {
         mut self,
         collection_mgmt: Arc<dyn CollectionManagementOperations>,
     ) -> Self {
+        self.collection_versions = Some(Arc::new(CollectionVersionsFromManagement(Arc::clone(
```
vertexclique: **Smell: this silently overwrites an explicitly provided `collection_versions`.**

`Server::build` calls `with_collection_versions` (`server.rs:458`) and then `with_collection_mgmt` (`:461`), so whenever both are configured the read-only impl the caller passed is discarded in favour of the management-backed shim. Same behaviour today since both resolve to the same call, but it is an order-dependent builder and a latent bug.

```rust
self.collection_versions.get_or_insert_with(|| {
    Arc::new(CollectionVersionsFromManagement(Arc::clone(&collection_mgmt)))
});
```

Unrelated and positive: the read-only trait split itself is exactly right, and the `SERVICE_UNAVAILABLE`-on-DELETE test is the kind of assertion I like to see on a new surface.

crates/telemetry/src/metrics.rs
```diff
@@ -15,7 +15,7 @@ pub enum RetryLayer {
     HttpAutoCommit,
     EmbeddedExecute,
     Merge,
-    PushLedger,
+    PushMarker,
```
vertexclique, line 18: **Smell: `push_ledger` -> `push_marker` is a silent break for anything watching the old name.**

The label goes out through `as_str()` and the field name through `ConflictMetricsSnapshot`'s serde output. Suggest `#[serde(alias = "push_ledger")]` on the field, or emitting both names for one release.

crates/defra-core/src/signing.rs
```diff
@@ -499,6 +529,39 @@ pub fn set_broadcast_creator_did(did: Option<String>) {
 ///
 /// Returns the DID set by `set_broadcast_creator_did`, or None if no
 /// identity override is active (broadcasts will use the node PeerId).
+#[cfg(not(target_arch = "wasm32"))]
+pub fn get_broadcast_creator_did() -> Option<String> {
+    TASK_BROADCAST_CREATOR_DID
```
vertexclique, line 535: **Smell: `set_broadcast_creator_did` becomes a silent no-op inside a task scope.**

This prefers the task-local whenever a scope is active, so the thread-local setter at `:522` writes state that is then ignored. Nothing is broken today (only `crates/db/src/doc_mutator_tests.rs:1078` still calls it), but it is a trap for the next caller. Suggest making the setter `#[cfg(test)]`, or having it fail loudly when a scope is active.

Separately: the task-local migration is the right fix, a thread-local genuinely cannot survive an `.await` that migrates workers, so the old RAII guard was never sound. But the three #757 regression tests (`broadcast_creator_did_guard_clears_on_panic_unwind`, `..._clears_on_normal_drop`, `..._self_heals_from_leaked_state`) were deleted and only the happy path was replaced. #757 was "the next anonymous mutation would silently broadcast with someone else's identity"; that deserves a direct replacement, for example two sequential mutations on one runtime where the second is anonymous.

proofs/tla/MC_MergeQueue_Green.cfg
```diff
@@ -8,11 +8,12 @@ CONSTANTS
   Dup           <- mcDup
   MaxRetries    = 5
   MaxUserWrites = 2
-  LockMode      = "PerDoc"
+  LockMode      = "GlobalMerge"
```
vertexclique, line 11: **Smell: the green model moved to a global merge lock, and the probe that guarded against exactly this was repurposed.**

The old `MC_MergeQueue_CrossDocParallel` comment: "If this probe ever held (no counterexample), the GREEN serialization result would be suspect (the lock would be a global lock and same-doc serialization would be trivially true)."

The green config now is that global lock, and the probe was rewritten from `NoCrossDocParallel` to `INV_SingleMergeWriter`. Same RED verdict, inverted meaning, and nothing now guards `INV_SameDocSerialized` against vacuity.

If `GlobalMerge` is the intended new boundary that is fine, but it needs three things: a line in the description (the formal section only discusses SyncOwnership), a replacement anti-vacuity probe, and a merge-throughput number, since P2P merges are now globally serialized where they used to run per document.

proofs/tla/MC_MergeQueue_CrossDocParallel.cfg
```diff
@@ -14,4 +13,4 @@ CONSTANTS
   FailMode      = "Closed"
   UserWriteMode = "LockFree"
 INVARIANT INV_TypeOK
-INVARIANT NoCrossDocParallel
+INVARIANT INV_SingleMergeWriter
```
vertexclique: **Smell: this was the anti-vacuity probe for `MC_MergeQueue_Green`.**

Changing the invariant from `NoCrossDocParallel` to `INV_SingleMergeWriter` keeps the RED verdict while inverting what the red certifies, and removes the only check that `INV_SameDocSerialized` is not trivially true under a global lock. See the note on `MC_MergeQueue_Green.cfg:11`.

proofs/src/registry.rs
```diff
@@ -101,6 +101,38 @@ pub const PROPERTIES: &[Property] = &[
         model_ref: "MC_Replicator_Resumable_Green.cfg",
         tiers: &[Behavioral],
     },
+    Property {
+        family: "Sync ownership transfer (head hint / receiver pull)",
```
vertexclique, line 105: **Smell: `tiers: &[Behavioral]` claims a binding that does not exist.**

`proofs/tests/behavioral/` is untouched by this PR (`git diff --name-only e27fc944..HEAD -- proofs/tests/` is empty), and `proofs/README.md` still says "**19 behavioral tests** ... each break-tested for non-vacuity" while the family count goes to 21. Every other family's binding points at a `proofs/tests/behavioral/*.rs` file that exists; this one points at integration tests in another crate that the conformance harness does not drive.

`matrix::every_modeled_family_is_bound` passes on a string match alone, so the guard the README cites does not actually check this.

Either add the behavioral test, or drop the tier to what is genuinely bound so the README stays true.

crates/p2p/src/sync/coordinator/dag_fetcher_tests.rs
```diff
@@ -44,8 +48,16 @@ struct TestTransport {
     dead_providers: Arc<Mutex<HashSet<String>>>,
     skip_serving_syncs: Arc<AtomicUsize>,
     fail_connected_peers: Arc<AtomicBool>,
+    connected_peers: Arc<Mutex<Vec<PeerId>>>,
     cancelled_queries: Arc<Mutex<Vec<u64>>>,
     hang_car_requests: Arc<AtomicBool>,
+    streamed_rooted_blocks: Arc<Mutex<Option<Vec<(Cid, Vec<u8>)>>>>,
```
vertexclique: **Nit: this fails `cargo clippy --all --all-targets -- -D warnings` (`clippy::type_complexity`).**

Introduced by 975bdda0 in this PR. CI's Lint job runs `cargo clippy --all` without `--all-targets`, so it passes CI; `just lint-all-targets` catches it. A `type StreamedBlocks = Arc<Mutex<Option<Vec<(Cid, Vec<u8>)>>>>;` clears it.

The other 7 failures under `--all-targets` are pre-existing and not yours.

justfile
```diff
@@ -21,13 +21,14 @@ jdk_root    := tooling / "jdk"
...
-# TLC 1.8.0 by checksum.
+# TLC 1.7.4 by checksum. The upstream 1.8.0 pre-release asset is rolling and
+# cannot provide a reproducible clean-checkout gate.
 protoc_version := "35.1"
 go_version     := "1.25.12"
 jdk_release    := "jdk-21.0.12+8"
 jq_version     := "1.8.1"
-tla_version    := "1.8.0"
-tla_sha256     := "e22f8ffb4bacdea0a871f444dd94fe5fb0d8013b3388ae39e82e26f852c735d5"
+tla_version    := "1.7.4"
```
vertexclique, line 30: **Nit: TLC pinned down from 1.8.0 to 1.7.4 in the same PR that adds the first liveness properties.**

The reason in the comment is sound, a rolling pre-release asset cannot give a reproducible checksum gate. Two asks: confirm 1.7.4 checks `LIVE_EventualCurrency` / `LIVE_EventualReceiverQuiescence` the way the new `MC_SyncOwnership_Green_Liveness_*` configs expect, and leave a note on #1310, which pinned 1.8.0 deliberately.

edjroz (changes requested):
One blocker, not covered by @vertexclique's pass (no comments on `push_docs.rs`). At `crates/db-merge/src/push_docs.rs:256-260` the `?` on `car_authority.register(...)` exits the whole function, but `observe_push_head` is not written until `:284` — so when the per-peer grant table fills at document K, documents K..N get neither a push nor a durable obligation, and the caller only logs (`crates/p2p-adapter/src/libp2p.rs:613`) before publishing `ReplicatorCompleted`. Reachable on defaults: 4096 grants per peer, slots not freed until `POST_ACK_RECOVERY_WINDOW` (10 min) elapses, replay at 50/s — a first backfill over ~4096 documents to one peer trips it in about 82 seconds. The live push path already conserves the obligation on the identical `None`, and Go records a per-document failure instead of abandoning the tail. Writing the marker before acquiring the document's grants, or making overflow `continue` rather than `?`, should close it; everything else I found is non-blocking and already on your threads.

crates/p2p/src/host/event.rs
jackzampolin: Fixed in cbf92479: both lifecycle events are classified as Completion by the shared exhaustive classifier, and retry storage guards are limited to short atomic transitions and are never held across SE or transport work. The dispatcher and adapter regressions are green.

crates/p2p/src/sync/event_dispatcher.rs
jackzampolin: Fixed in cbf92479: classification is exhaustive and only cheap subscription/listening state remains Inline. Peer lifecycle is Completion; management, SE, CAR, and gossip request work enters bounded Admission or Recovery.

crates/p2p/src/transport.rs
jackzampolin: Fixed in cbf92479: HostEvent and TransportEvent now use one shared macro-based exhaustive classifier with no wildcard arm, so new variants fail compilation until classified. Remote request classes enter bounded Admission.

crates/p2p-adapter/src/retry.rs
jackzampolin: Fixed in cbf92479: reschedule_retry_peer acquires the peer writer and reads/modifies the latest schedule inside one transaction with bounded OCC retry, so reconnect activation or marker registration cannot be overwritten by a stale snapshot.

crates/p2p-adapter/src/retry.rs
jackzampolin: Fixed in cbf92479: the acquire/drop no-op was removed. Retry storage transitions now call atomic peerstore methods that acquire the per-peer writer only while committing state.

crates/p2p/src/sync/event_dispatcher.rs
jackzampolin: Fixed in cbf92479: raw gossip is Admission-classed rather than Completion-classed. Completion capacity is now reserved for bounded lifecycle/reply/completion producers, and the classifier regression asserts this boundary.

crates/p2p-adapter/src/retry.rs
jackzampolin: Fixed in cbf92479: no retry guard survives transport replay or SE regeneration. The guard API documentation now restricts it to short storage transitions, and network work occurs after the durable acknowledgement transition.

crates/p2p-adapter/src/retry.rs
jackzampolin: Fixed in cbf92479: an exact wrapped capacity nack defers to the paced sweep without advancing the Go failure ladder. It also advances the durable fair cursor and stops the peer pass, avoiding adjacent-scope hammering.

crates/storage/src/stores/peerstore.rs
jackzampolin: Fixed in cbf92479: legacy marker migration runs once at retry-loop startup. Ordinary get_retry_documents reads no longer perform migration writes.

crates/storage/src/stores/peerstore.rs
jackzampolin: Fixed in cbf92479: sorted markers rotate from a durable dispatch_cursor, and every reschedule—including capacity deferral—advances the cursor. The bounded-prefix fairness regression covers tail progress.

crates/storage/src/stores/retry_info.rs
jackzampolin: Addressed in cbf92479 and the PR description: 30s through a 32m cap is the approved Go-parity document ladder, sourced from the Go replicator.retryintervals configuration. Tests pin the exact sequence.

crates/storage/src/stores/retry_info.rs
jackzampolin: Fixed in cbf92479: deterministic bounded peer jitter was restored through bump_for, while the underlying retry intervals remain the exact Go document ladder.

crates/storage/src/stores/peerstore.rs
jackzampolin: Fixed in cbf92479: blind schedule writes were replaced by reschedule_retry_peer, which reads and updates current state under the per-peer writer and transaction. Reconnect rung preservation is regression-tested.

crates/p2p-adapter/src/retry.rs
jackzampolin: Fixed in cbf92479: finish_peer now delegates to clear_retry_peer, whose marker scan and conditional schedule deletion are one bounded-retry transaction. Concurrent marker registration cannot be orphaned.

crates/storage/src/stores/peerstore.rs
jackzampolin: Fixed in cbf92479: clear_retry_peer performs marker scan and retry-ID deletion atomically, with bounded transaction-conflict retry. The concurrent marker-registration regression verifies the schedule survives.

crates/p2p/src/message/pushlog.rs
jackzampolin: Fixed in cbf92479 using the first suggested compatibility-safe option: verification preserves the received unsigned CBOR map bytes with only the signature field elided, so unknown future fields remain covered. A new domain prefix was intentionally not introduced because it would break rolling old/new Rust verification.

crates/db-merge/src/broadcast_mutator/batch.rs
jackzampolin: Fixed in cbf92479: the batch commit documentation now explicitly states that this error is post-commit—data is committed but one or more P2P markers are not durable—so callers must not interpret it as rollback. Durable recovery is tracked in #1538.

crates/db-merge/src/push_docs_common.rs
jackzampolin: Fixed for the reported amplification: reverse scope resolution now compares the already loaded collection resolved_root_id, eliminating the per-collection systemstore read. The remaining in-memory collection scan is disclosed and no longer produces 100k storage reads.

crates/p2p/src/message/pushlog.rs
jackzampolin: Fixed in cbf92479: the unreachable new-struct postcard decode arm was deleted and the compatibility test was renamed to state that it exercises the legacy postcard broadcast shape.

crates/db-merge/src/broadcast_mutator/batch.rs
jackzampolin: Addressed without weakening the durability boundary: this PR intentionally retains synchronous O(documents × peers) marker persistence, removes the inaccurate amortization claim, and adds measured benchmarks. The current result is ~2.20ms for 100 documents × 4 peers; any future batching must preserve per-scope crash durability.

crates/telemetry/src/metrics.rs
jackzampolin: Fixed in cbf92479: structured telemetry emits both push_marker and the legacy push_ledger label for a compatibility window, with a regression covering both names.

crates/http/src/router/state.rs
jackzampolin: Fixed in cbf92479: with_collection_mgmt now uses get_or_insert_with for the read-only adapter and therefore preserves an explicitly provided collection_versions implementation.

crates/defra-core/src/signing.rs
jackzampolin: Fixed in cbf92479: the legacy setter is cfg(test), and a sequential scoped-then-anonymous regression proves task-local creator identity cannot contaminate the next mutation. The executor and auto-commit paths now enter the same scoped identity boundary.

proofs/tla/MC_MergeQueue_Green.cfg
jackzampolin: Fixed in cbf92479: NoCrossDocParallel remains an anti-vacuity probe, while MC_MergeQueue_Red_PerDocWriters separately demonstrates the rejected per-document-writer topology. The PR now explicitly documents the global writer and includes merge throughput measurements.

proofs/tla/MC_MergeQueue_CrossDocParallel.cfg
jackzampolin: Fixed in cbf92479: the cross-document anti-vacuity configuration again checks NoCrossDocParallel. The distinct per-document-writer red configuration carries the single-writer counterexample instead of repurposing this probe.

proofs/src/registry.rs
jackzampolin: Fixed in cbf92479: SyncOwnership registry entries are now honestly tiered as Boundary. The README states that the concrete integration gates live in repository CI rather than claiming nonexistent conformance behavioral tests.

crates/p2p/src/sync/coordinator/dag_fetcher_tests.rs
jackzampolin: Fixed in cbf92479: the introduced nested test type now has a StreamedBlocks alias. CI-exact clippy is green; the unrelated pre-existing all-target warnings remain outside this change.

justfile
jackzampolin: Verified and documented: TLC 1.7.4 checks all 131 configurations, including the three eventual-currency/receiver-quiescence liveness greens. I also recorded the immutable-pin requirement and results on #1310: https://github.com/sourcenetwork/defradb.rs/issues/1310#issuecomment-5349195144.

jackzampolin:
Eduardo’s replay-tail blocker is fixed in cbf92479. Existing-document replay now persists each document marker before attempting rooted selective-CAR authority; authority overflow retains that marker and continues marking later documents instead of returning from the whole replay. This preserves obligations for documents K..N at the 4,096-grant bound. The fix is included in the green db-merge/unit, ownership integration, Iroh, and formal validation reported in the updated PR description.

Issue #1161, opened 2026-07-20
fix(p2p): explicit replay capability can reject concurrent public pushes


jackzampolin:
## Problem

`rust_rust_p2p_trust_boundary` can fail when an unsigned public document is created immediately after installing an identity-authorized replicator. The cached explicit-replay capability is attached to the normal two-stream push, and the receiver rejects the document because the capability authorizer DID does not match the unsigned block's effective creator (the source peer ID).

This makes delivery depend on whether the unprivileged gossip registration or capability-bearing two-stream registration wins the race.

## Evidence

CI run: https://github.com/sourcenetwork/defradb.rs/actions/runs/29714196348
Job: https://github.com/sourcenetwork/defradb.rs/actions/runs/29714196348/job/88263858404

Receiver log:

```text
Accepted explicit replay capability ... authorizer_did=did:key:...
DAG fetch complete ... doc_id=bae-319b...
Block merge failed ... explicit replay authorization authorizer 'did:key:...' does not match block creator '12D3KooW...'
```

The Jack-signed protected document in the same test merged successfully; only the unsigned public document was rejected.

## Expected

A replay capability must not taint or block a public live push. Delivery should not depend on gossip/two-stream arrival order. Add a deterministic regression that delays gossip or otherwise forces the capability-bearing two-stream path for a public document.

## Notes

This predates #1160: the capability attachment dates to `e175755f`, while the failing path does not exercise invalid-signature quarantine or pending-DAG root single-flight.

PR #1513, opened 2026-08-18
fix(p2p): align replicator replay triggers

### LINKED ISSUE #1506: iroh and libp2p disagree on replicator replay triggers: capability changes ignored on iroh, filter changes replay all collections
The two transports disagree on when `add_replicator` replays existing docs for an already-registered collection:

- libp2p: `collections_requiring_replay` (`crates/p2p-adapter/src/lib.rs:68`, used at `libp2p.rs:583`) additionally replays a collection whose **explicit-replay capability** changed.
- iroh: the install path (`iroh.rs:424-460`) uses only the inline check — replay on new collection, or on any `ReplicationFilters` inequality (which clears `existing_collection_ids` and replays **every** collection in the install list). `explicit_replay_capabilities` are accepted in the signature but never consulted for replay.

Consequences:
- A capability change that replays on libp2p silently does nothing on iroh.
- On iroh, changing one collection's filter re-replays all collections in the install (each a full datastore prefix scan per `db-merge/src/push_docs.rs:90`), so narrow filter adjustments are disproportionately expensive.

Ask: converge the two paths on one replay-trigger rule (shared helper), and scope filter-change replay to the collections whose filters actually changed.

Found while evaluating filtered replicators for per-session mobile sync; the all-collections replay on filter change is the main reason we rejected filter churn as a hydration mechanism.

iverc:
## Problem

The libp2p and Iroh adapters disagree on when an existing replicator needs replay. Iroh ignores explicit replay capability changes, while either adapter replays every requested collection when only one collection filter changes.

## What changed

- select replay collections with one shared per-collection filter and capability rule
- validate, cache, attach, and receive explicit replay capabilities on Iroh using the same authorization boundary as libp2p
- preserve capability cleanup when replicators or individual collections are removed
- cover scoped filter replay and capability delivery over a real Iroh stream

## Validation

- [x] `cargo test --profile super-dev -p defra-p2p-adapter --all-features`
- [x] `cargo clippy --profile super-dev -p p2p -p defra-p2p-adapter --all-targets --all-features -- -D warnings`
- [x] `git diff --check`

Closes #1506

### DIFF
--- a/crates/p2p-adapter/src/libp2p.rs
+++ b/crates/p2p-adapter/src/libp2p.rs
@@ -428,46 +428,13 @@ impl<B: Blockstore + 'static> P2POperations for P2PAdapter<B> {
         let requested_collections: HashSet<String> = collection_cids.iter().cloned().collect();
         let local_peer_id = self.handle.local_peer_id_cached().to_string();
         let target_peer_id = peer_id.to_string();
-        let mut validated_capabilities = Vec::new();
-
-        if !explicit_replay_capabilities.is_empty() {
-            let expected_authorizer_did = expected_authorizer_did.ok_or_else(|| {
-                P2PError::invalid_input(
-                    "explicit replay capabilities require an authenticated identity",
-                )
-            })?;
-
-            for capability in explicit_replay_capabilities {
-                if !requested_collections.contains(&capability.collection_id) {
-                    return Err(P2PError::invalid_input(format!(
-                        "explicit replay capability collection '{}' was not requested",
-                        capability.collection_id
-                    )));
-                }
-
-                let authorization = p2p::verify_explicit_replay_capability(
-                    &capability.capability,
-                    &local_peer_id,
-                    &target_peer_id,
-                    &capability.collection_id,
-                )
-                .map_err(|error| {
-                    P2PError::invalid_input(format!(
-                        "invalid explicit replay capability for collection '{}': {}",
-                        capability.collection_id, error
-                    ))
-                })?;
-
-                if authorization.authorizer_did != expected_authorizer_did {
-                    return Err(P2PError::invalid_input(format!(
-                        "explicit replay capability authorizer '{}' did not match authenticated identity '{}'",
-                        authorization.authorizer_did, expected_authorizer_did
-                    )));
-                }
-
-                validated_capabilities.push((capability.collection_id, capability.capability));
-            }
-        }
+        let validated_capabilities = crate::validate_explicit_replay_capabilities(
+            explicit_replay_capabilities,
+            expected_authorizer_did,
+            &requested_collections,
+            &local_peer_id,
+            &target_peer_id,
+        )?;
 
         let collections_with_changed_capabilities: HashSet<String> = validated_capabilities
             .iter()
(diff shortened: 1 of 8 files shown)


vertexclique:
Need rebase and conflict fixing. cc/ @iverc

.github/workflows/ci.yml
```diff
@@ -727,9 +727,12 @@ jobs:
             exit 1
           fi
 
-      # A bind failure can reach the harness log watcher before the child-exit
-      # path classifies it for retry. Keep node-spawning suites serial until
-      # the harness handles both readiness failure paths.
+      # Parallelism is bounded at 8 rather than left to libtest's default.
```
jackzampolin: **[suggestion]** The new comment claims integration-test parallelism is bounded at 8 as “the knee of the measured curve,” but every cargo test invocation in this block (basic, core, and p2p) still passes `--test-threads=1`. That text is leftover from a branch state that ran basic/core at 8; after merging main, the commands are fully serial and the comment is both factually wrong and a design-history dump.

**Suggestion:** Drop the 8-thread rationale (or restore `--test-threads=8` on the suites that are actually safe to parallelize). Keep the bind-failure note on the p2p step, which does match `--test-threads=1`.

crates/p2p/src/explicit_replay.rs
```diff
@@ -33,6 +37,81 @@ pub const MAX_CAPABILITY_TTL: Duration = DEFAULT_CAPABILITY_TTL;
...
+        if request.explicit_replay_capability.is_some() {
+            return;
+        }
+
+        let cached = self
+            .capabilities
+            .read()
+            .get(&(target_peer_id.to_string(), request.collection_id.clone()))
+            .cloned();
+        if let Some(cached) = cached.filter(|cached| request.creator == cached.authorizer_did) {
```
jackzampolin: **[suggestion]** The shared `attach` path keeps the `#1161` guard (`request.creator == cached.authorizer_did`) but dropped the WHY that used to live on `P2PHostHandle`. Attaching a collection capability to an unrelated live/public push makes the receiver validate the block against the wrong authorizer and reject otherwise valid replication. That invariant is not obvious from the filter expression alone, and Iroh now uses this same helper on every two-stream send.

**Suggestion:** Restore a short WHY on this creator check (the old `#1161` note is enough) so the next edit does not “simplify” it away.

crates/p2p-adapter/src/libp2p.rs
```diff
@@ -581,6 +542,8 @@ impl<B: Blockstore + 'static> P2POperations for P2PAdapter<B> {
             &effective_collections,
             &collection_cids,
             &existing_collection_ids,
+            &existing_filters,
```
jackzampolin: **[suggestion]** Libp2p now passes per-collection `existing_filters` / `replication_filters` into `collections_requiring_replay`, matching Iroh. The skip-replay debug string still says “same collections and replay capability” and the comment immediately above still describes only capability-change replay. Iroh’s skip log was updated to mention filters; libp2p was not, so a filter-only re-add will look like an idempotent skip in libp2p logs even when it replays.

**Suggestion:** Update the surrounding comment and the skip-replay log to mention filters, the same way `crates/p2p-adapter/src/iroh.rs` now does.

iverc:
> Need rebase and conflict fixing. cc/ @iverc

@vertexclique ✅

vertexclique (approved):

PR #1514, opened 2026-08-18
fix(ffi): expose p2p sync status

### LINKED ISSUE #1504: SyncStatus is unreachable from FFI/embedded clients: no p2p_sync_status binding
`P2POperations::sync_status()` (`crates/http/src/router/traits.rs`) is served over HTTP only (`GET /api/v0/p2p/sync/status`, `crates/http/src/router/routes.rs:130`). `crates/ffi/src/p2p/` has bindings for replicators, collections, documents, sync, and push retry — but no `p2p_sync_status`. `EmbeddedNode` likewise exposes no accessor.

An embedded/FFI client (iOS/Android via the FFI, or a desktop app embedding `defra-node`) therefore cannot read any of the sync observability that `SyncStatus` carries: `push_backlog` with `per_peer` backlog/cooldown/consecutive-failure snapshots, `pending_dags` / `persisted_pending_dags`, quarantine gauges, `next_pending_retry_in_ms`, and the `SyncDiagnostics` counters. These are exactly the fields a client UI needs for a sync indicator ("syncing / stalled / offline"), and mobile is the platform least able to fall back to the HTTP router.

Ask: an FFI binding (`p2p_sync_status` returning the same JSON the HTTP handler serializes) and/or an `EmbeddedNode` accessor. Payload can stay schemaless JSON so the in-flight sync-ownership work (`fix/1116-sync-ownership-stage3`, which adds `car_*` counters and removes `encode_cache_*`) does not need to keep an FFI struct in lockstep.

Context: found while building a mobile session-sync UI against the embedded node; the only current workaround is polling a co-located HTTP router the mobile runtime does not otherwise need.

iverc:
## Problem

Live P2P sync diagnostics are available over HTTP, but embedded FFI clients cannot read them without running a co-located HTTP server.

## What changed

- add a NAC-checked `p2p_sync_status` binding that returns the existing sync-status JSON unchanged
- add `defra_mobile_sync_status` using the node default identity
- cover the exported signatures, live snapshot, no-P2P error path, and generated C header

## Validation

- [x] `cargo test --profile super-dev -p ffi --all-features`
- [x] `cargo clippy --profile super-dev -p ffi --all-targets --all-features -- -D warnings`
- [x] generated the C header and verified both new symbols
- [x] `git diff --check`

Closes #1504


crates/ffi/src/p2p/status.rs
```diff
@@ -0,0 +1,115 @@
...
+                        .await
+                        .map_err(FfiP2PError::from)?;
+                    serde_json::to_string(&status).map_err(|error| {
+                        FfiP2PError::internal(format!(
+                            "failed to serialize P2P sync status: {}",
+                            error
+                        ))
+                    })
+                })
+            })
```
edjroz: The global node-registry read lock is held across `block_on`

`NODES.get()` holds a `parking_lot::RwLock` read guard for the whole closure (`crates/ffi/src/state/registry.rs:39-44`), and the closure here wraps the entire `rt.block_on(...)`. So each call pins that global lock for the duration of `sync_status()`, which takes the backlog mutex, runs `prune_expired_retries`, and walks peers x retries (`crates/p2p/src/sync/push_backlog.rs:780+`).

It matters more for this entrypoint than its neighbours because this is the one meant to be polled on a timer by a mobile status UI, precisely while replication is backed up. A `new_node` / `node_close` writer then queues behind the poll, and parking_lot's eventual fairness makes that queued writer block subsequent readers too.

The safer shape already exists two lines up in `check_nac_for_node` (`crates/ffi/src/nac_check.rs:88-90`): clone the `Arc` out, drop the guard, then `block_on`. `state.p2p` is `Option<Arc<P2PState>>` (`crates/ffi/src/state/mod.rs:115`), so `NODES.get(node_ptr, |state| state.p2p.clone())` is a cheap clone and the rest runs unlocked.

To be clear on scope: ~20 sibling p2p entrypoints do the same thing, so this is not a defect introduced  here and I am not asking you to fix twenty call sites. Your call whether to hoist it just here  or leave it consistent and we can then file it as an issue and address it across the codebase.

crates/ffi/src/p2p/status.rs
```diff
@@ -0,0 +1,115 @@
...
+                };
+
+                rt.block_on(async {
+                    let status = p2p
+                        .system
+                        .ops()
+                        .sync_status()
+                        .await
+                        .map_err(FfiP2PError::from)?;
+                    serde_json::to_string(&status).map_err(|error| {
```
edjroz, line 50: The payload is serialized twice per call.

`P2PAdapter::sync_status` already builds a full `serde_json::Value` tree (`crates/p2p-adapter/src/libp2p.rs:241`), and this line immediately re-serializes it to a string. The intermediate tree is allocated and discarded on every call — which matters a little more here than elsewhere, since this is the endpoint a client polls on a timer.

the `P2POperations` trait shape forces it, and the cheaper path would be a `sync_status_json() -> String` on the trait. Noting it as a follow-up

edjroz (approved):
Left 2 comments, that should be non blocking just stuff we have to surface as follow ups for the team later, as we address the codebase in the future.

The `mobile.rs` is also way pass the 500 lines threshold, this is non-blocking here just something we might want to follow-up on later.

christopherbm (approved):

PR #1526, opened 2026-08-19
fix(schema): align index kind wire envelope

### LINKED ISSUE #1518: Index wire shape: Go writes a Kind + KindDescription envelope; our flat sniffed IndexKindWire misparses it as an ordered index
Merged Go (sourcenetwork/defradb#5096, `client/index.go`) serializes `IndexDescription` as a discriminated envelope:

```json
{"Name": "...", "ID": 1, "Fields": [...], "Kind": 1,
 "KindDescription": {"Algorithm": "HNSW", "Metric": "COSINE", "Dimensions": 3,
                     "HNSW": {"M": 16, "EfConstruction": 128, "EfSearch": 64}},
 "Unique": false}
```

`Kind` (u8: 0 ordered, 1 vector) is the sole authority on the kind. `KindDescription` is raw JSON resolved into the concrete config named by `Kind`. Top-level `Unique` is legacy compat, kept in sync both ways (`Normalize()`). An unknown `Kind` is a hard error (`NewErrUnknownIndexKind`).

Our `IndexKindWire` (`crates/schema/src/index.rs`) still implements the pre-merge PR-branch shape: one flat object, no discriminator, kind sniffed from whether `Algorithm` or `Dimensions` is present. Its comment says "matching Go's `parseIndexKind`", which no longer exists.

## Failure mode: silent, both directions

- Go -> Rust: Go's vector description carries no top-level `Algorithm`/`Dimensions`, so our sniff sees an ordered index and drops the entire vector config.
- Rust -> Go: our flat vector fields sit at the top level, so Go reads `Kind` as absent (= 0 = ordered) and ignores them.

Neither direction errors. A vector index crossing runtimes silently becomes an ordered index, which is worse than a parse failure.

## Scope

- Rewrite the wire form as Go's envelope: `Kind` + `KindDescription` + compat `Unique`, with an unknown `Kind` a hard error rather than a default.
- Our deliberate divergences (`IVFPQ`/`SSG` params, `DOT` metric) move inside the vector `KindDescription`, exactly where `HNSW` sits, so an HNSW-only description stays byte-compatible with Go and a divergent one stays self-contained.
- On-disk compat with our previously stored flat form is not required (pre-1.0; only wire compat is mandated). If a legacy read shim is trivial, fine; otherwise drop the old shape.

## Acceptance

- Fixture round-trip against JSON captured from Go develop at `f73a903f`: HNSW descriptions marshal byte-identically both ways, `Unique` compat included.
- A description with an unrecognized `Kind` fails loudly; misparse-to-ordered is impossible by construction.
- Property test: any `IndexKind` round-trips through the envelope unchanged.

Part of #1517.

iverc:
Closes #1518.

## Problem

Rust serialized index kinds as a flat object and inferred the variant from vector-specific fields. Go now uses a discriminated `Kind` + `KindDescription` envelope, so cross-runtime vector index descriptions could silently decode as ordered indexes.

## What changed

- serialize index kinds with the Go-compatible discriminated envelope
- keep the top-level ordered-index `Unique` field synchronized for compatibility
- reject unknown kind discriminators instead of defaulting to an ordered index
- flatten index kind fields in Go-facing HTTP responses
- add exact JSON/CBOR fixture and round-trip coverage

## Validation

- [x] `cargo test -p schema`
- [x] `cargo test -p defra-http --test vector_index_api`
- [x] `cargo clippy -p schema -p defra-http --all-targets -- -D warnings`
- [x] `git diff --check`


crates/schema/tests/index_kind_tests.rs
```diff
@@ -17,47 +18,36 @@ fn vector_description() -> VectorIndexDescription {
...
 #[test]
-fn the_kind_serialises_with_gos_field_names() {
+fn the_kind_serializes_as_a_discriminated_envelope() {
     let json = serde_json::to_value(IndexKind::Vector(vector_description())).unwrap();
-    assert_eq!(json["Algorithm"], "HNSW");
-    assert_eq!(json["Metric"], "COSINE");
-    assert_eq!(json["Dimensions"], 768);
-    assert_eq!(json["HNSW"]["M"], 16);
-    assert_eq!(json["HNSW"]["EfConstruction"], 128);
-    assert_eq!(json["HNSW"]["EfSearch"], 64);
```
vertexclique, line 30: these config lines for algo must be still be tested for existence/absence. Last time in the meeting we agreed to have `hnsw: {}` for implicit defaults, and `hnsw: { EfConstruction: 45, EfSearch: 56 }` 
cc/ @jsimnz

vertexclique (approved):

PR #1533, opened 2026-08-19
chore(crates): fold query-plan into the query crate (1/3)

edjroz:
## Context

The team is consolidating crates by role. This is the query slice: `query-plan`, `query-parse` and `query-types` fold back into a single `query` crate. This is 1 of 3.

The order is forced and is the reverse of the dependency graph: **plan, then parse, then types**. `query-types` is the leaf, so folding it first would leave `query-parse` and `query-plan` depending on `query`, which depends on them. The leaf folds last.

Every PR in the stack is pure relocation. No behaviour changes, no renames, no dead-code removal.

Rebased onto current `main` (`e27fc944`). An earlier revision of this stack was rebased onto #1474 when that merged first; the conflicts and their resolutions are described in the review thread.

## What changes

**99 files move, nothing outside `crates/query` is touched.**

- `git mv` of `src/{doc_stream.rs,fetcher.rs,mutator.rs,plan/,planner/}` into `crates/query/src/` (94 files, all 100% renames). They land under the exact module names the facade already re-exported, so `query::plan::*`, `query::planner::*`, `query::fetcher::*`, `query::mutator::*` and `query::doc_stream::*` resolve unchanged. No dependent crate needed an edit; none is in the diff.
- `src/txn/` merges **flat** into the existing `crates/query/src/txn/`: `context.rs`, `handle.rs`, `read_access.rs`, `registry.rs` and `result.rs` become siblings of `guard.rs`, the two re-export lists collapse into one, and `txn/mod.rs` ends at **23 lines**. `txn` is the only module-name collision among the 99 moved files.
- The two `#[cfg(test)]` blocks that lived in the two `txn/mod.rs` files move to `txn/tests.rs`, matching the sibling-test-file pattern this crate already uses in `rest/tests.rs`, `runner/commits_tests.rs` and `runner/query/nested_tests.rs`. They stay unit tests, so private access is preserved. All 19 run and pass under `txn::tests`.
- 5 integration tests move to `crates/query/tests/` (`cross_object_read_path`, `scan_short_id_restriction`, `vector_routing`, `vector_routing_hybrid`, `scan_vector_widening`). The only edits are `query_plan::` -> `query::` on `use` lines and one inline path.
- `crates/query/Cargo.toml` absorbs `cursor`, `futures`, `cid`, and promotes `async-lock` from dev to regular; drops `query-plan`. Workspace member removed, `Cargo.lock` regenerated by cargo.
- Docs that named the crate: the `CLAUDE.md` crate-table row, and six `proofs/` files.

**Three changes are not moves:**

1. `runner/executor.rs` loses the wildcard arm on `GetTransactionResult`. Its own comment said it existed because the enum is `#[non_exhaustive]` *in another crate*; that only constrains matches in other crates, so once `result.rs` moved in the three arms became exhaustive and the arm unreachable, which fails `just lint` under `-D warnings`. The enum stays `#[non_exhaustive]`, so external matchers (`crates/db`) are unaffected.
2. `planner/index_selection/filter_to_scan.rs`: two doc examples get their right-hand sides backticked. Rustdoc was reading `[30]` and `["red", "blue"]` as intra-doc links.
3. `plan/type_join/mod.rs`: delete a comment pointing at `crates/query/tests/type_join_tests.rs`, a file that has never existed in this tree.

## Verify

With the pinned Go binary on PATH:

```sh
cargo fmt --all -- --check
just lint
cargo machete
RUSTDOCFLAGS="-D warnings" cargo doc -p query --no-deps
just profile=super-dev test
```

| Gate | Baseline (`main` e27fc944) | This stack |
|---|---|---|
| `cargo fmt --check` / `just lint` / `cargo machete` | pass | pass |
| `cargo doc -p query` | pass | pass |
| integration `--test query` | 70 passed / **1 failed** / 3 ignored | **identical** |
| integration `--test basic` | 34 passed / 0 failed | **identical** |
| integration `--test cursor` | 25 passed / 0 failed | **identical** |

**`main` is currently red and this stack inherits it, unchanged.** `go_view_paths::rust_go_view_paths` fails identically on `e27fc944` and here: `POST /api/v1/view/refresh?get_inactive=true` returns 400 "refreshing inactive collection versions is not supported". Both sides of that contradiction came in with #1498, which added the test asserting 2xx and the `db` validation refusing it. Not caused by this stack and not fixed by it.

Relocation-specific checks:

- **Tests are moved, not rewritten.** Normalized inventory of every `#[test]`/`#[tokio::test]` (names plus bodies, `use` lines stripped): **unchanged from `main` at every commit of the stack**.
- **Nothing outside changed:** `git diff --name-only main..HEAD` touches only `crates/query*`, `Cargo.toml`, `Cargo.lock`, `CLAUDE.md`, `proofs/`.
- **The moves are moves:** `git diff -M --name-status` shows R100 on all 94 source files. Review with `-M` or the diff is unreadable.
- **`just doc`** is red on `main` for 82 unrelated errors in `storage`, `db-index` and `datastore`. `cargo doc -p query` specifically was red on this stack before the fix above and is green now.

## Notes

- **Compile time**, since #670's acceptance criterion was "compile time measurably reduced" and this stack reverses that split. `cargo build -p query --profile super-dev`, dependencies warm, 3 iterations, median, 14-core machine. Each timed build asserts cargo actually compiled the expected crates, and `-j` is swept because the two layouts have different amounts of available parallelism:

  | `-j` | cold split | cold folded | | incremental split | incremental folded | |
  |---:|---:|---:|---|---:|---:|---|
  | 1 | 7.79s | 6.72s | **-14%** | 1.72s | 1.00s | **-42%** |
  | 4 | 3.89s | 4.74s | +22% | 1.26s | 0.97s | -23% |
  | 14 | 3.78s | 4.45s | +18% | 1.34s | 1.03s | -24% |

  Reading: on a cold build the split wins only once it has cores to spend. At `-j1` the folded crate is *faster*, so the split is not doing less work, it is doing it more concurrently. For the incremental edit that a query developer actually feels (touch `mapper/types.rs`, rebuild), the folded crate is faster at every parallelism level; the gap narrows with cores but plateaus by `-j4`, because the split's critical path (`query-types -> query-plan -> query`, with only parse and plan overlapping) is serial.

  vertexclique measured the same experiment on 36 cores and reported +37% cold (same direction as ours) and +74% on the incremental edit, which is the opposite sign from every level measured here. I cannot reproduce that and I no longer have a theory for it: the parallelism explanation predicts the gap narrowing with cores, but it saturates at `-j4` here, so 36 cores should not flip it. Worth comparing method before either number is treated as settled.
- `async-lock` is the one dependency whose *kind* changes (dev -> regular). Plan code uses it outside tests.
- Deliberately **not** in this stack: file splits, dead-code removal, visibility tightening beyond the `pub(crate)` in 3/3, and `proofs/survey/query-*.md` consolidation.


vertexclique (commented):
Reviewed the stack as a whole rather than layer by layer, since the safety property is a stack property. Reproduced the claims independently against `a7e309d3`, not taken from the description:

* 661 test attributes at every commit of the stack and at the base.
* The `pub use` block in `crates/query/src/lib.rs` is byte-identical before and after.
* Blast radius: outside `crates/query*` the stack touches only `Cargo.toml`, `Cargo.lock`, `CLAUDE.md` and `proofs/`.
* Nothing outside the cluster ever depended on `query-plan` / `query-parse` / `query-types`: zero manifests, zero `query_plan::` / `query_types::` / `query_parse::` paths. The facade really was the only door.
* Exactly one module name collides (`txn`) among 149 moved files, so no silent overwrite anywhere else.
* No crate-level `#![...]` attribute was lost; all four crates carried only `[lints] workspace = true`, so lint behaviour is unchanged.
* Dependency merge is a superset in every case (`tokio` non-wasm `rt` to `rt,time`, dev `rt` to `rt-multi-thread`). `Cargo.lock` diff is pure package removal, no version drift.
* No residual `query_plan::` / `query_types::` paths, and no stale crate name in `.github/`, `justfile`, `scripts/` or `tools/`.

`#[non_exhaustive]` on `GetTransactionResult` still earns its keep: `crates/db` matches it at `db/src/migration/tests.rs:753`. The wildcard removal in `runner/executor.rs` is correct and strictly better, because a future variant now fails to compile at the one site that must handle it instead of degrading to a generic error string.

Two things before this lands, then three small inline notes.

## 1. `just doc` fails on the tip

`RUSTDOCFLAGS="-D warnings" cargo doc -p query --no-deps`:

```
error: unresolved link to `30`
  --> crates/query/src/planner/index_selection/filter_to_scan.rs:38:60
   |
38 | /// - `{numbers: {_any: {_eq: 30}}}` -> ExactMatch{values: [30]}
   |                                                            ^^ no item named `30` in scope
```

Pre-existing in `query-plan`, but `cargo doc -p query` was green before this stack and is red after, because the file moved in. The fix is to backtick the right-hand side: `` `ExactMatch{values: [30]}` ``.

Worth adding `just doc` to the Verify block. Separately, and not this PR's problem: `cargo doc` runs nowhere in `.github/`, so the justfile comment "a broken intra-doc link fails CI" is not currently true.

## 2. #670 is open, proposes this exact split, and its acceptance criterion is "compile time measurably reduced"

The stack ships no compile-time number. Measured here on 36 cores, dev profile, incremental on, deps warm, both sides compiling the identical set of binaries, `cargo clean` on the cluster between cold runs, 3 iterations each:

| | split (4 crates) | folded (1 crate) | |
|---|---|---|---|
| cold cluster build | 11.97s | 16.41s | +37% |
| one-line edit in `mapper/types.rs` | 3.28s | 5.69s | +74% |

Caveat on that: this is the query cluster in isolation, which is the developer inner loop, not a full workspace build. On `cargo build --workspace` the other ~40 crates fill the cores and the gap should largely wash out, so I would not expect CI wall-clock to move much.

That is a real cost against real benefits (3 fewer manifests, one canonical import path, about 165 fewer lines of facade and manifest boilerplate). I am not asking to reject it. I am asking for the number to be in the body, and for #670 to be closed with this measurement as the stated justification rather than left open contradicting merged code. An open issue saying "split this for compile time" alongside a merge that unsplits it is the temporal-boundary problem CLAUDE.md warns about.

## Inline

**`crates/query/src/txn/primitives/mod.rs:6`** now says the opposite of its own parent. It reads "The `TransactionGuard` high-level API lives in the `query` crate because it depends on the `QueryExecutor` trait ... without creating a cycle with the runner", while `txn/mod.rs:8` correctly says "`TransactionGuard` lives beside them in `guard`". Same directory, contradictory headers. The `primitives` header should drop the cycle rationale.

**`crates/query/src/txn/`**: I would take the flat merge instead. `primitives` exists so that one file's diff reads as a rename, and it costs a permanent directory level plus the same 11 names re-exported twice (`txn/mod.rs:15` re-exports what `primitives/mod.rs:18` already re-exports). `txn` is the only collision among 149 moved files, so this is one file's worth of one-time diff noise against structure everyone reads forever. Concretely: move `context/handle/read_access/registry/result.rs` straight into `txn/`, fold the module declarations into the existing `txn/mod.rs`, and lift both `#[cfg(test)] mod tests` blocks (231 lines in `mod.rs`, about 100 in `primitives/mod.rs`) into `crates/query/tests/`, which the house rule wants anyway. That leaves `txn/mod.rs` at roughly 55 lines with one re-export list.

**`crates/query/src/plan/type_join/mod.rs:19`**: `// Tests extracted to crates/query/tests/type_join_tests.rs`, but that file does not exist anywhere in the tree, before or after. Pre-existing rot, free to delete while the file is already in the diff.

PR #1534, opened 2026-08-19
chore(crates): fold query-parse into the query crate (2/3)

edjroz:
## Context

Second of three PRs folding the query cluster into a single `query` crate, part of the team's consolidation of crates by role. Stacks on 1 of 3 (`query-plan`). The stack is rebased onto #1474, which merged first; see 1 of 3 for how those conflicts were resolved.

Order is plan, parse, types: the reverse of the dependency graph, because `query-types` is the leaf and must fold last or the remaining crates would depend on `query`, which depends on them.

Pure relocation. No behaviour change.

## What changes

**32 files move. Nothing outside `crates/query` is touched.**

- `git mv` of `src/{query_parse/,schema_gen/,sdl_parse/,select_convert.rs}` into `crates/query/src/` (31 files, all 100% renames). These are the module names the facade already re-exported, so `query::query_parse::*`, `query::sdl_parse::*`, `query::schema_gen::*` and `query::select_convert::*` resolve unchanged.
- `crates/query/src/lib.rs` loses the alias that only existed to work around a crate-vs-module name clash:

```rust
-extern crate query_parse as query_parse_ext;
-pub use query_parse_ext::query_parse;      +pub mod query_parse;
-pub use query_parse_ext::schema_gen;       +pub mod schema_gen;
-pub use query_parse_ext::sdl_parse;        +pub mod sdl_parse;
-pub use query_parse_ext::select_convert;   +pub mod select_convert;
```

  The crate `query-parse` contained a module also called `query_parse`, which is why the alias was needed. Post-fold it is just `crate::query_parse`. Callers already wrote `query::query_parse::...`, so nothing downstream moves.
- `tests/vector_index_sdl.rs` moves to `crates/query/tests/` (`use query_parse::parse_sdl` -> `use query::parse_sdl`), and `crates/query/tests/vector_routing_hybrid.rs` (moved in 1 of 3) switches its one `query_parse::parse_query` call to `query::parse_query`. That call was the last cross-crate dev-dependency edge in the cluster.
- `crates/query/Cargo.toml` absorbs `sha2` and `regex`, drops `query-parse`. Workspace member removed, `Cargo.lock` regenerated. `CLAUDE.md` crate-table row removed.

Six files are modified in total: `lib.rs`, one test, both `Cargo.toml`s, `Cargo.lock`, `CLAUDE.md`.

## Verify

```sh
cargo fmt --all -- --check
just lint
cargo machete
just profile=super-dev test
```

| Gate | Baseline (`main` e27fc944) | This PR |
|---|---|---|
| `cargo fmt --check` / `just lint` / `cargo machete` | pass | pass |
| integration `--test query` | 70 passed / **1 failed** / 3 ignored (pre-existing on `main`, see 1/3) | **identical** |
| integration `--test basic` | 34 passed / 0 failed | **identical** |
| integration `--test cursor` | 25 passed / 0 failed | **identical** |

Relocation-specific checks:

- **Tests are moved, not rewritten.** Normalized inventory (names plus bodies, `use` lines stripped) is unchanged from `main` and from 1 of 3.
- **The parse tests ran where they now live:** `vector_index_sdl` 18 passed, `query_parse_tests` 49 passed.
- **Nothing outside changed:** the diff touches only `crates/query*`, `Cargo.toml`, `Cargo.lock`, `CLAUDE.md`.
- Review with `git diff -M`; without rename detection this reads as 31 deletions and 31 additions.

## Notes

- On `regex = "1"`: an earlier revision of this body called it unpinned "unlike its siblings", which was wrong. There is no `regex` entry in `[workspace.dependencies]` to pin against, and `pg-compat` declares it the same bare way, so this is the house norm rather than an outlier. `chrono` is bare in six crates for the same reason. Adding both to the workspace table is workspace-wide work, not query work.
- Worth crediting the other direction: this stack takes `chrono = { version = "0.4", features = ["serde"] }` from four separate declarations down to one.
- `crates/query/tests/query_parse_tests.rs` is 691 lines, over the 500-line convention. It is not touched anywhere in this stack; the split is its own PR so that it carries its own proof.
- The per-crate report notes two pre-existing defects in `schema_gen` (self-referencing fields render as the literal type `Self`; `<Collection>OrderInput` is always emitted empty). **Both pre-date this PR and are untouched by it** - flagging so they are not misattributed to the fold. They have their own follow-up.
- `proofs/survey/INDEX.md` and `proofs/README.md` drop their `query-parse` crate slice here, matching what 1/3 did for `query-plan`.
- Deliberately not in this PR: file splits, dead-code removal, `proofs/survey/query-parse.md` consolidation.


vertexclique (commented):
Clean layer. The alias removal is the good part: `extern crate query_parse as query_parse_ext` was a 2015-edition idiom carried in a 2021 crate purely to work around the crate-vs-module name clash, and it goes away for free. I verified the public item surface is unchanged by this layer, and that `crates/query/tests/vector_routing_hybrid.rs` was the last cross-crate dev-dependency edge in the cluster.

One correction to the note on `regex = "1"`. There is no `regex` entry in `[workspace.dependencies]` to pin against, and `pg-compat` declares it the same bare way. `chrono` is likewise bare in six crates. So this is the house norm rather than an outlier, and the real follow-up is adding `regex` and `chrono` to the workspace table, which is workspace-wide work rather than query work.

Worth crediting the other direction while you are here: this stack takes `chrono = { version = "0.4", features = ["serde"] }` from four separate declarations down to one.

Stack-level review is on #1533. Nothing in this layer blocks.

PR #1535, opened 2026-08-19
chore(crates): fold query-types into the query crate (3/3)

edjroz:
## Context

Last of three PRs folding the query cluster into a single `query` crate. Stacks on 2 of 3 (`query-parse`). After this there is no `query-*` package left in the workspace: `cargo metadata` lists one `query`.

`query-types` is the leaf of the dependency graph, so it folds last. Folding it earlier would have left `query-parse` and `query-plan` depending on `query`, which depends on them.

Pure relocation, plus one test-file split in a separate commit (briefly its own PR, #1540, folded back here because it stacks on this branch and cannot merge before it).

## What changes

**24 files move; 108 files get a mechanical path rewrite.**

- `git mv` of `src/{collection_provider.rs,doc.rs,document/,error.rs,json_convert.rs,limits.rs,mapper/}` into `crates/query/src/` (24 files, all 100% renames, not one line inside them changed). query-types was a leaf, so nothing inside it referenced a sibling; only its consumers needed rewriting.
- `crates/query/src/lib.rs`: `pub use query_types::{document,error,json_convert,limits,mapper}` become `pub mod`. **`collection_provider` and `doc` are `pub(crate) mod`**, so the fold adds no new public path: both types stay reachable exactly where they already were, through `query::fetcher::{CollectionProvider, StaticCollectionProvider}` and `query::planner::{Doc, DocFields, DocStatus}`. The whole workspace compiles against that, which is the check that those item re-exports were the only real door.
- **Path rewrite:** `query_types::` -> `crate::` across 108 source files, and -> `query::` in 2 integration tests. Almost all of it is `use` lines; the residue is ~50 fully-qualified paths written inline in signatures and bodies. `cargo fmt` then re-sorted the affected `use` lines and re-wrapped two expressions whose lines got shorter. Three stale comments saying "extracted to query-types crate" are deleted.
- `crates/query/Cargo.toml` absorbs `thiserror`, drops `query-types`. Workspace member removed, `Cargo.lock` regenerated. `CLAUDE.md` row removed. `proofs/survey/INDEX.md` drops its `query-types` slice and `proofs/survey/query.md` gains a sentence describing modules instead of sibling crates.

**Separate commit: split `crates/query/tests/query_parse_tests.rs`.** It was 691 lines, over the 500-line convention for test files, and splits at its own two section-marker comments into `query_parse_tests.rs` (232 lines, 17 tests), `query_parse_error_tests.rs` (275, 23) and `query_parse_explain_tests.rs` (186, 9). Test functions are moved verbatim; the only added lines are two file headers and their `use query::parse_query;`. 49 tests before, 49 after.

Its proof is different from the fold's, which is why it is a separate commit: concatenate the three files, strip the two new headers and their import plus the two section markers, and the result is identical to the original line for line.

## Verify

```sh
cargo fmt --all -- --check && just lint && cargo machete
RUSTDOCFLAGS="-D warnings" cargo doc -p query --no-deps
just profile=super-dev test
cargo metadata --no-deps --format-version 1 | jq '[.packages[].name | select(startswith("query"))]'   # ["query"]
cargo test -p query --test query_parse_tests --test query_parse_error_tests --test query_parse_explain_tests
```

| Gate | Baseline (`main` e27fc944) | This PR |
|---|---|---|
| `cargo fmt --check` / `just lint` / `cargo machete` / `cargo doc -p query` | pass | pass |
| integration `--test query` | 70 passed / **1 failed** / 3 ignored (pre-existing on `main`, see 1/3) | **identical** |
| integration `--test basic` / `--test cursor` | 34 / 25 passed, 0 failed | **identical** |

Relocation-specific checks:

- **Tests are moved, not rewritten.** Normalized inventory (names plus bodies, `use` lines stripped) is unchanged from `main` and at every commit of the stack, with one honest exception: `test_invalid_mutation_collection` and `test_variable_in_order_direction` spell `query_types::mapper::...` inline inside the test body, so the prefix rewrite touched those two bodies. Same assertions, same values, only the path prefix differs.
- **Public API preserved:** diffing every `pub use` item line in `lib.rs` before and after gives an identical list, byte for byte. That is what keeps the ~931 `query::` references across the 12 dependent crates resolving.
- **Nothing outside changed:** the diff touches only `crates/query*`, `Cargo.toml`, `Cargo.lock`, `CLAUDE.md`, `proofs/`.
- The path rewrite is compiler-verified: a wrong rewrite does not compile.

## Notes

- The two `use`-line-only test-body edits above are the only places in the stack where a test body changed. Called out explicitly because "tests move, they do not change" is the safety property of the whole exercise.
- Follow-ups now unblocked, none of them in this stack: `proofs/survey/{query-parse,query-plan,query-types}.md` consolidation, oversized-file splits (`sdl_parse/parser_tests.rs` at 1957 lines and `mapper/filter/filter_tests.rs` at 1230 are both larger than anything this stack touches, and both live in `src/`), further visibility tightening, and `mod.rs`-style normalization.
- #670 proposed this split for compile-time reasons and is contradicted by this stack; compile-time numbers are in 1 of 3, and closing or re-scoping that issue is a separate decision.


vertexclique (commented):
Verified the mechanical parts of this layer: the `query_types::` to `crate::` rewrite leaves no residue anywhere in the tree, the only non-rewrite edits are the deleted `query-types` manifest and `lib.rs` plus four stale comments, and the public item surface is unchanged. The two test-body edits you called out are the only ones, and they are prefix-only. Stack-level review is on #1533.

Three notes.

**`proofs/survey/INDEX.md:142-143` and `proofs/README.md:63`** still list `query-parse` and `query-types` as crate slices. #1533 fixed the equivalent `query-plan` reference at `INDEX.md:92`, so the stack is internally inconsistent: same file, same kind of reference, fixed for one crate and left for the other two. One line each. The `survey/query-*.md` files themselves reading as a deferred consolidation is fine, and `survey/query.md` handles it honestly.

**`crates/query/src/lib.rs:21-22`**: I would take the `pub(crate)` you offered. To be precise about the size of it, no new types become reachable: `CollectionProvider` and `StaticCollectionProvider` were already public via `query::fetcher`, and `Doc`, `DocFields`, `DocStatus` via `query::planner`. So this is two new paths to existing items rather than new surface. Still, two paths to the same type is a thing to keep in sync forever in exchange for nothing, and the re-exports keep working from a private module.

**The test split: take your own offer and detach it.** The stack's safety property is "pure relocation, verified by an invariant test count", and the split needs a different proof. I did check it and it holds: concatenating the three files and stripping the headers and the two section markers gives 562 normalized lines identical to the original, and 49 tests split 23 + 17 + 9. But two different proofs for one PR is avoidable.

Separately, if splitting oversized test files is in scope, `sdl_parse/parser_tests.rs` (1957 lines) and `mapper/filter/filter_tests.rs` (1230) are now in the same crate and are 3x and 2x the file that got split. Both also live in `src/` rather than `tests/`. That is a whole follow-up, which is itself the argument for not starting it here.

PR #1540, opened 2026-08-19
test(query): split query_parse_tests.rs at its section markers

edjroz:
## Context

Splits one oversized test file. Separated from the fold stack (#1533, #1534, #1535) at review request, because the fold's safety property is "pure relocation, proven by an unchanged test inventory" and a file split needs a different proof. Stacks on 3 of 3 so the file is already in its final location.

## What changes

`crates/query/tests/query_parse_tests.rs` was 691 lines, over the 500-line convention for test files. It splits at its own two existing section-marker comments into:

| File | Lines | Tests |
|---|---:|---:|
| `query_parse_tests.rs` | 232 | 17 |
| `query_parse_error_tests.rs` | 275 | 23 |
| `query_parse_explain_tests.rs` | 186 | 9 |

Test functions are moved verbatim. The only added lines are two file headers and their `use query::parse_query;`.

## Verify

```sh
cargo test -p query --test query_parse_tests --test query_parse_error_tests --test query_parse_explain_tests
```

49 tests before, 49 after (17 + 23 + 9).

The mechanical proof that nothing changed: concatenate the three files, strip the two new headers and their import plus the two section-marker comments, and the result is identical to the original line for line.

```sh
diff <(git show HEAD~1:crates/query/tests/query_parse_tests.rs | grep -vE '^// (Error path|EXPLAIN) tests$' | grep -v '^$') \
     <(cat crates/query/tests/query_parse_tests.rs crates/query/tests/query_parse_error_tests.rs crates/query/tests/query_parse_explain_tests.rs \
       | grep -vE '^//! Integration tests for the GraphQL query parser: |^use query::parse_query;$' | grep -v '^$' \
       | tail -n +2)
```

## Notes

- Split at the file's own section markers rather than an invented taxonomy, so the grouping matches what the author already had.
- Two larger test files in `src/` remain over the convention and are untouched here: `sdl_parse/parser_tests.rs` (1957 lines) and `mapper/filter/filter_tests.rs` (1230). Moving those is a bigger piece of work with its own argument, since it also raises whether they belong in `tests/` at all.


edjroz:
Folded back into #1535: it stacks on that PR and cannot merge before it, so the separate layer only added a merge step. Content and concatenation proof unchanged.

PR #1542, opened 2026-08-20
docs: codify data path ownership guidance

jackzampolin:
## Summary

- define ownership and callsite mapping rules for shared data paths
- preserve surface-specific policy while centralizing shared mechanisms
- strengthen regression, dependency, fixture, and CI evidence requirements
- expose the same guidance through an AGENTS.md symlink

## Validation

- git diff --check
- verified AGENTS.md is a symlink to CLAUDE.md

The guidance is derived from recurring failure patterns across Claude, Codex, and Grok sessions in the consolidated coding-session store.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,13 +38,108 @@ This codebase is designed for **AI-human pair programming**. Every structural ch
 
 **No commented-out code. No TODO comments (create issues instead). No speculative docs.**
 
-## 3. No Documentation Files
+## 3. Authority, Ownership, and Callsite Discipline
+
+The recurring high-cost failure is a correct rule in one path with a bypass in
+another. Treat data access as an ownership problem, not as local helper
+cleanup.
+
+### Before Changing a Data Path
+
+- **Name the invariant and its owner.** Identify the canonical durable state,
+  the sole writer for each lifecycle transition, and the owner of retry,
+  recovery, and terminal cleanup. Readers and adapters may observe or request
+  transitions; they must not become alternate writers or clocks.
+- **Map every production callsite.** Use `rg` on the changed trait, method, data
+  type, and storage key, then trace the real entrypoints. Check CLI, HTTP and
+  GraphQL, embedded, FFI/WASM, P2P live delivery, replay, and restart recovery
+  as applicable. Include auto-commit, batch, explicit transaction, fast paths,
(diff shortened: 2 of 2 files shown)


vertexclique (commented):
Reviewed briefly, noting this is a draft and docs only.

The renumbering is clean. I grepped for stale references and found none; nothing in the repo cites CLAUDE.md by section number.

### Minor, needs discussion: placement against the document's own stated purpose

`CLAUDE.md:41-135`

95 lines of abstract prose land in a file whose own section 1 says the structure "optimizes for rapid context acquisition" and whose closing goal is "New contributor feels ready to do productive work immediately". The new section 3 contains no file path, crate name, or worked example from this repo, and it is now the first thing a reader hits after the temporal boundaries rule. The content is sound; the placement and density work against the document's stated purpose. Worth deciding whether it belongs ahead of the crate map or after it, and whether each rule can cite the actual incident that produced it.

### Minor

- `CLAUDE.md:43`: section 4 still reads "Only allowed: `README.md`, `CLAUDE.md`, `Cargo.toml` files" while the paragraph directly beneath now permits `AGENTS.md`. The allow list should name the fourth file.
- `CLAUDE.md:45`: "`AGENTS.md` may exist only as a symlink to the nearest `CLAUDE.md`" implies per directory symlinks are sanctioned, but only one `CLAUDE.md` exists, at the repo root. Either drop "nearest" or say explicitly that nested `CLAUDE.md` files are expected.
- `AGENTS.md`: a relative symlink materializes as a one line text file on a Windows checkout without `core.symlinks`. Given that "never maintain a second copy of the guidance" is the rule being encoded, that failure mode is worth a line in the PR description or a `.gitattributes` note.
- Validation gap: the PR body lists `git diff --check` and "verified AGENTS.md is a symlink". Neither checks that the symlink resolves in a fresh clone, which is the case that actually breaks.
- Pre existing but colliding with the section being edited: section 4's "No `docs/` directories" is already contradicted by `proofs/tla/*.md`, `audit/`, and root level `chunked-field-values.md`. Since this PR is rewriting that section, it is the natural moment to reconcile the rule with the tree.

iverc (commented):
is this still alive? sitting as a draft since aug 20 and the base has moved a lot.

on content: the rules themselves are good engineering, but it's ~100 lines of incident lessons landing in the onboarding file, whose stated goal is fast context acquisition for a new contributor. i'd keep the handful that actually recurred (single writer per lifecycle, map every callsite, regression fence for the original failure) and let the rest go. happy either way, just don't want it rotting as a draft.

PR #1543, opened 2026-08-20
feat(node): typed document read/write for embedded nodes

mike7c2:
`EmbeddedNode` read and wrote documents only through GraphQL, whose variables and results are JSON. JSON has no binary type, so an embedder storing bytes had to encode them: the `Blob` convention spells them hex, costing double on every write, in storage, and on the wire, and base64 still costs a third again.

`doc_mutator()` and `doc_fetcher()` expose the seams the query planner already uses, so a `Document` carrying `NormalValue::Bytes` reaches storage as bytes and comes back as bytes. Writes enter where the GraphQL planner enters, so validation, indexing, ACP, and P2P broadcast are unchanged; only the JSON step is skipped.

The fetcher is constructed twice rather than shared. `DocFetcher` has no blanket impl for `Arc`, and a delegating wrapper is how `FetcherWrapper` once lost method overrides silently. Both read the same `Arc<DB>`; only a migration cache is duplicated.

`DocID`, `Document`, `NormalValue`, `DocFetcher`, and `DocMutator` are re-exported so an embedder needs no extra path dependencies. They come from `db` and `query`, which `defra-node` already depends on

### DIFF
--- a/crates/defra-node/src/lib.rs
+++ b/crates/defra-node/src/lib.rs
@@ -60,10 +60,14 @@ pub use config::HttpConfig;
 pub use config::P2PConfig;
 #[cfg(feature = "sourcehub")]
 pub use config::SourceHubConfig;
+/// Typed document access, for values JSON cannot carry — binary above all.
+/// See [`EmbeddedNode::doc_mutator`] and [`EmbeddedNode::doc_fetcher`].
+pub use db::{DocID, Document, NormalValue};
 pub use dense_search::{DenseHybridSearchHit, DenseHybridSearchRequest, DenseHybridSearchResponse};
 pub use events::EventName;
 pub use lens::{LensConfig, LensModule, TransformId};
 pub use query::QueryLimits;
+pub use query::{DocFetcher, DocMutator};
 pub use query::{QueryExecutor, QueryRequest, QueryResponse, TransactionHandle};
 pub use schema::CollectionVersion;
 pub use telemetry::{ConflictMetricsSnapshot, RetryLayerSnapshot};
@@ -166,6 +170,8 @@ pub struct EmbeddedNode {
     rocksdb_stats: Option<storage::RocksDbStatsHandle>,
     #[cfg(feature = "http")]
     txn_cleanup_task: tokio::sync::Mutex<Option<tokio::task::JoinHandle<()>>>,
+    doc_mutator: Arc<dyn query::DocMutator>,
+    doc_fetcher: Arc<dyn query::DocFetcher>,
     #[cfg(feature = "p2p")]
     p2p_ops: Option<Arc<dyn defra_http::P2POperations>>,
     #[cfg(feature = "p2p")]
@@ -485,6 +491,30 @@ impl EmbeddedNode {
     /// or [`Self::execute_request_in_txn`] for document reads and writes.
     pub fn runner(&self) -> &Arc<dyn QueryExecutor> {
         &self.runner
+    }
+
+    /// Write documents with typed field values.
+    ///
+    /// GraphQL is the usual path, but its variables are JSON, which has no
+    /// binary type: bytes have to be hex or base64 to get through, costing
+    /// double or a third again on every write, in storage, and on the wire.
+    /// A [`Document`] carries [`NormalValue::Bytes`] as bytes, and the CBOR
+    /// block encoding keeps them that way.
+    ///
+    /// This is the same seam the GraphQL planner mutates through, so
+    /// validation, indexing, ACP, and P2P broadcast all behave identically —
+    /// only the JSON step is skipped.
+    pub fn doc_mutator(&self) -> &Arc<dyn query::DocMutator> {
+        &self.doc_mutator
+    }
+
+    /// Read documents with typed field values, the counterpart to
+    /// [`doc_mutator`](Self::doc_mutator).
+    ///
+    /// Query results are JSON, which renders [`NormalValue::Bytes`] as a hex
+    /// string. Reading through the fetcher returns the bytes themselves.
+    pub fn doc_fetcher(&self) -> &Arc<dyn query::DocFetcher> {
+        &self.doc_fetcher
     }
 
     /// Access the event bus directly.
@@ -1454,7 +1484,15 @@ impl NodeBuilder {
         let mutator: Arc<dyn query::DocMutator> =
             Arc::new(db::AutoCommitMutator::new(database.clone()));
 
-        // Query runner components
+        let doc_mutator = mutator.clone();
+
+        // Query runner components. The fetcher is built twice rather than
+        // shared: `DocFetcher` has no blanket impl for `Arc`, and a delegating
+        // wrapper is exactly how `FetcherWrapper` once lost method overrides
+        // silently. Both instances read the same `Arc<DB>`; only a migration
+        // cache is duplicated.
+        let doc_fetcher: Arc<dyn query::DocFetcher> =
+            Arc::new(db::LensedAutoCommitFetcher::new(database.clone()));
         let fetcher = db::LensedAutoCommitFetcher::new(database.clone());
         let provider: Arc<dyn query::CollectionProvider> =
             db::DbCollectionProvider::new_arc(database.clone());
@@ -1565,6 +1603,8 @@ impl NodeBuilder {
             rocksdb_stats: None,
             #[cfg(feature = "http")]
             txn_cleanup_task: tokio::sync::Mutex::new(txn_cleanup_task),
+            doc_mutator,
+            doc_fetcher,
             #[cfg(feature = "p2p")]
             p2p_ops,
             #[cfg(feature = "p2p")]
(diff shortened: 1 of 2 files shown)


crates/defra-node/src/lib.rs
```diff
@@ -487,6 +493,30 @@ impl EmbeddedNode {
...
+    /// binary type: bytes have to be hex or base64 to get through, costing
+    /// double or a third again on every write, in storage, and on the wire.
+    /// A [`Document`] carries [`NormalValue::Bytes`] as bytes, and the CBOR
+    /// block encoding keeps them that way.
+    ///
+    /// This is the same seam the GraphQL planner mutates through, so
+    /// validation, indexing, ACP, and P2P broadcast all behave identically —
+    /// only the JSON step is skipped.
+    pub fn doc_mutator(&self) -> &Arc<dyn query::DocMutator> {
+        &self.doc_mutator
```
jackzampolin, line 508: **[bug]** `doc_mutator()` / `doc_fetcher()` return the raw planner seams and skip the request context `EmbeddedNode::execute` always installs. The rustdoc at this method says validation, indexing, ACP, and P2P "all behave identically — only the JSON step is skipped." That is true for indexing and for P2P (the cloned `mutator` Arc is the `BroadcastMutator` when P2P is configured, and `QueryRunner::with_acp` does call `set_document_acp` on that same Arc). It is not true for identity, signing, or DAC:
- Writes: `execute_prepared_request` scopes node identity and `SigningConfig` via `execute_with_signing_context`; GraphQL mutations then wrap the call in `with_scoped_identity` + `scope_broadcast_creator_did` and, for policy-backed collections, register/check DAC in `QueryRunner::execute_single_mutation`. Direct `DocMutator::create` only registers ACP when `get_broadcast_creator_did()` is already set (`AutoCommitMutator::register_created_doc_with_acp`), and `update`/`delete` have no DAC updater/deleter checks at all (those live only in the runner). `get_signing_config()` is also unset, so a node with `with_node_identity_did` produces unsigned blocks — different CIDs than the GraphQL path for the same content.
- NAC: `AutoCommitMutator::create_impl` calls `check_node_access(None, DocumentUpdate)`, which falls back to `current_identity`. With NAC enabled and no ambient identity that is the wildcard, so typed creates fail while `execute()` succeeds as the node.
- Reads: DAC filtering is `PermissionFilterNode` in the planner (`QueryRunner` query path), not `DocFetcher`. `doc_fetcher().get_all` / `get_by_field_value` / `get_commits` return every document, including policy-hidden ones.

**Suggestion:** Do not hand out `&Arc<dyn DocMutator>` / `&Arc<dyn DocFetcher>` as the supported embedder API. Add node-owned methods (or a thin wrapper) that, for the duration of the call, apply the same identity/signing/`broadcast_creator_did` scope as `execute()`, and on reads filter by `DocumentPermission::Read` the way `maybe_wrap_with_acp_filter` does. If a raw escape hatch is still wanted, match `runner()`: document that it bypasses signing and ACP, and delete the "ACP … behave identically" sentence. Cover with a crate test that enables node identity + a DAC policy and asserts typed create registers the owner and typed fetch does not leak unauthorized docs; on a signed node, also assert the typed write CID matches `execute()` for equivalent bytes (e.g. GraphQL Blob under `DEFRA_TEST_BLOB_AS_BYTES`).

crates/defra-node/src/lib.rs
```diff
@@ -1454,7 +1484,15 @@ impl NodeBuilder {
         let mutator: Arc<dyn query::DocMutator> =
             Arc::new(db::AutoCommitMutator::new(database.clone()));
 
-        // Query runner components
+        let doc_mutator = mutator.clone();
+
+        // Query runner components. The fetcher is built twice rather than
```
jackzampolin, line 1489: **[suggestion]** The construction comment narrates an old `FetcherWrapper` override-loss incident and why `DocFetcher` has no `Arc` blanket impl. That is architecture history, not a safety invariant the next reader needs. Duplicating `LensedAutoCommitFetcher` (fresh migration cache, shared `Arc<DB>`) is fine; `clone_for_stream` already does the same thing without the story.

**Suggestion:** Drop the `FetcherWrapper` paragraph. If a comment stays, keep one line: both instances share `Arc<DB>`; the second copy only duplicates the migration cache.

crates/defra-node/src/lib.rs
```diff
@@ -487,6 +493,30 @@ impl EmbeddedNode {
         &self.runner
     }
 
+    /// Write documents with typed field values.
+    ///
+    /// GraphQL is the usual path, but its variables are JSON, which has no
+    /// binary type: bytes have to be hex or base64 to get through, costing
```
jackzampolin, line 499: **[suggestion]** The `doc_mutator` rustdoc spends several lines on JSON having no binary type, hex vs base64 size costs, and CBOR keeping bytes as bytes. That restates the PR pitch and the code (`NormalValue::Bytes` + existing block encoding). Project comments are meant to be short WHY/invariants; `runner()` is the local model (one warning about what the getter skips).

**Suggestion:** Replace with a short contract: typed `Document` values, including `NormalValue::Bytes`, persist without a JSON encode step. Put any ACP/signing caveat next to that contract (see the identity/ACP bug on this method). Same trim on the re-export blurb at line 63 ("binary above all") and the module docs in `tests/typed_document_bytes.rs`.

jackzampolin (changes requested):
## Summary

The PR correctly hangs typed create/fetch off the same `AutoCommitMutator` / `BroadcastMutator` and `LensedAutoCommitFetcher` instances the GraphQL runner already uses, so indexing, block encoding, unique-index checks, and P2P broadcast do follow that seam. The crate test is a real CBOR round-trip (`NormalValue::Bytes` in, `Bytes` back out of a new txn), not an in-memory fake. The dominant risk is the public claim that ACP (and, by implication, identity/signing) is unchanged: `EmbeddedNode::execute` applies node identity, signing thread-locals, DAC registration, and `PermissionFilterNode`, while these getters return the raw planner traits with none of that request context.

Inline notes cover that identity/ACP/signing gap plus two comment trims. The rest of this review is the product direction.

## This is Go's Collection API, not an ORM

What you want is the native document client Go already ships, not a new ORM and not the planner guts.

In Go, embedders do not start at GraphQL. They start at `client.Collection` on `client.Store`:

```go
col, err := db.GetCollectionByName(ctx, "Payload")
doc.Set(ctx, "data", []byte{0xde, 0xad, 0xbe, 0xef})
err = col.AddDocument(ctx, doc, options.WithIdentity(...))
got, err := col.GetDocument(ctx, doc.ID())
```

`Document` / `DocID` / `NormalValue` (including `NewNormalBytes`) are the same object model you re-exported. GraphQL sits **on top** of Collection: the mutation planner's `addNode` calls `GetCollectionByName` + `AddDocument` with identity, encryption, and signing already attached. HTTP (`POST /collections/{name}`, `GET /collections/{name}/document/{docID}`, …), CLI (`defradb client document add`), C bindings, and JS all go through that same interface. Collection methods themselves apply NAC, identity, signing, encryption, DAC register-on-create, and DAC-aware get.

Rust today is inverted. `EmbeddedNode` is GraphQL-first (`execute`). Even `defradb client document add` builds an `add_Payload` GraphQL mutation. `DocMutator` / `DocFetcher` are planner internals — the analogue of Go's internal fetcher, not of `client.Collection`.

Handing out `&Arc<dyn DocMutator>` / `&Arc<dyn DocFetcher>` is the wrong layer. The Go shape to copy is collection-scoped methods that wrap the request context:

```text
node.get_collection("Payload")? -> Collection
collection.create(doc) / get(id) / update(doc) / delete(id)
```

`Document` stays a field bag (`doc.set(...)`). That is still not an ORM. (Go even has an old todo about a `TypedDocument` interface; it was never an object mapper.) If someone later wants `#[derive(Document)]` struct mapping, that is a separate crate on top of Collection.

## If you want this feature

Please duplicate in Rust the features Go's Collection API already offers, rather than exposing the planner traits:

- `AddDocument` / `AddManyDocuments`
- `UpdateDocument` / `SaveDocument`
- `DeleteDocument` / `ExistsDocument`
- `GetDocument`
- filter update/delete (`UpdateDocumentsWithFilter` / `DeleteDocumentsWithFilter`)
- identity, signing, encryption, NAC, and DAC on those methods the way Go's `AddDocument` / `GetDocument` do — not as a later wrap the caller might forget
- HTTP + CLI on that path (Go CLI `document add` calls `col.AddDocument`; Rust CLI currently GraphQL-wraps it)

Then add a Go/Rust integration test in `tools/integration-test` that proves the two Collection APIs actually match. Use `for_each_runtime!` so the same operations run on a Go node and a Rust node, and a mixed `rust_nodes(1).go_nodes(1)` cluster where it matters (same content → same DocID; Bytes stay Bytes; a typed/Collection write CID matches GraphQL Blob for equivalent content, e.g. under `DEFRA_TEST_BLOB_AS_BYTES`). That is the parity bar for this repo: same document content → same document ID, same CRDT ops → same merged state.

The crate test you added is a good local pin for the CBOR round-trip. It is not enough to claim Go Collection parity.

Happy to talk through the Collection-shaped wrapper vs the current getters if you want a second pair of eyes before rewriting.

PR #1544, opened 2026-08-20
chore(db): compartmentalize db, 1/4 error, event, view, database, collection, txn

vertexclique:
## Context

First of four PRs compartmentalizing `crates/db`. The crate is 103 files with four
naming families (`collection_*`, `txn_*`, `*_fetcher`, `*_mutator`) spread across
the top level of `src/`, plus ~8,000 lines of test code living inside `src/`. The
stack ends with `src/` holding `lib.rs` and thirteen directories, and every test
under `crates/db/tests/`.

The destination is kovan-mvcc readiness: an MVCC swap touches how a transaction is
created and retired, how a read resolves a snapshot, and how a write detects
conflict. Today those are smeared across the top level. After the stack they are
`txn/`, `read/` and `write/`.

**Every PR in this stack is pure relocation unless its description says otherwise.**
No behaviour changes.

## What changes

**50 files: `2 A / 22 M / 26 R`.** 18 of the renames are `R100`. Eight of the 22
modified files are consumer path updates in four crates (`cli`, `db-merge`,
`defra-node`, `embedded`); the other 14 are intra-crate path rewrites inside
`crates/db` and need eyes like any other edit.

- `error.rs` -> `error/`, `event_emission.rs` -> `event/emission.rs`,
  `view_ops.rs` -> `view/ops.rs`, and `database.rs` + `dump.rs` + `action.rs` ->
  `database/`.
- Twelve `collection_*` files join the four already in `collection/`;
  `collection_ops/` becomes `collection/ops/`.
- `txn.rs`, `txn_context.rs`, `txn_lens_store.rs`, `txn_registry.rs` -> `txn/`.

Path-invariance is why the consumer churn is small: `foo.rs` -> `foo/mod.rs` does
not change `db::foo`, so `db::error`, `db::database`, `db::collection` and
`db::txn` all resolve unchanged. Only `db::event_emission` (8 sites) and
`db::collection_acp` (3 sites) genuinely move.

**Three changes are not moves:**

1. `database/mod.rs` gains `pub(crate) mod action;` and `pub(crate) mod dump;` --
   a directory module must declare its children.
2. `collection/mod.rs` gains its ten child declarations, and the two pre-existing
   `mod` blocks are consolidated into one alphabetised list.
3. `collection_loader.rs`, `collection_provider.rs` and `collection_ops/mod.rs`
   show larger hunks than a prefix swap: `cargo fmt` re-wrapped expressions whose
   lines grew when `crate::action::` became `crate::database::action::`. No logic
   changed.

## Verify

```sh
cargo fmt --all -- --check
just lint
just check-node-graph
```

|---|---|---|
| `cargo test -p db` | 366 passed / 0 failed / 3 ignored | **identical** |

Relocation-specific:

- **The moves are moves:** `git diff -M --name-status main..HEAD` is `R100` on 18
  files; the rest are `R99`/`R98` where the only delta is the path rewrite plus
  rustfmt re-sorting `use` lines. **Review with `-M` or the diff is unreadable.**
- **Tests are moved, not rewritten.** Normalized inventory of every
  `#[test]`/`#[tokio::test]` (names plus bodies, `use` lines stripped) is unchanged
  from `main`, with one honest exception: two tests in `doc_mutator_tests.rs` spell
  `crate::event_emission::TxnBroadcastEvent` inline in the body, so the prefix
  rewrite touched those bodies. Same assertions, same values, only the path differs.
- **Nothing else changed:** the diff touches only `crates/db/` and the four consumer
  crates whose `db::event_emission` / `db::collection_acp` imports moved.

## Notes

Integration suites were not run locally: this machine has no
`~/.cache/defra-harness`, so the `go_*` parity tests cannot start a Go node. CI
has the pinned binary (`GO_COMPAT_COMMIT` 13c46ec9) and covers them. The Rust-native
half of `--test basic` is 24/24 locally.

## Review follow-ups (jackzampolin)

Both were body arithmetic, no code change:

- "44 are renames, 8 outside `crates/db`" summed past the 50-file total and hid 14
  modified files inside `crates/db`. Replaced with the real `2 A / 22 M / 26 R`
  breakdown above, and the 14 are now called out as needing review rather than
  waved off as renames.
- "the six consumer crates" is four: `cli`, `db-merge`, `defra-node`, `embedded`.

---

| Gate | Result |
|---|---|
| `cargo machete` | pass |

| Artifact | Result |
|---|---|


jackzampolin (approved):
## Approve

I tried to falsify the "pure relocation" claim and could not. Independently reproduced against `origin/main` (9bdb9e98):

- `cargo test -p db --profile super-dev` — **366 passed / 0 failed / 3 ignored on both refs**, identical.
- Rename detection: `18 R100, 3 R099, 1 R098, 2 R097, 2 R095`. Every sub-R100 delta I read is a path-prefix rewrite plus rustfmt re-sorting `use` lines. No hunk changes logic, attributes, or derives.
- Path-invariance argument holds: `db::error`, `db::database`, `db::collection`, `db::txn` all resolve unchanged, which is why 50 files moved and only 8 consumer files changed.
- **Zero back-compat `pub use` aliases were added.** This is the part I care most about for "centralize call sites" — old paths are dead, not double-live, so there is exactly one path to each item. `git grep -E '(crate|db|super)::(event_emission|collection_acp|...)'` at the stack tip returns nothing.

Approving. Two body-arithmetic nits, no code change needed:

- "50 files; 44 are renames, 8 files outside `crates/db`" — 44 + 8 > 50. Actual: `2 A / 22 M / 26 R`. 14 modified files *inside* `crates/db` are unaccounted for in that sentence.
- "the six consumer crates" — it is four: `cli`, `db-merge`, `defra-node`, `embedded`.

Worth correcting only because the stack's bodies are doing real reviewer-routing work, and a reader who trusts "44 are renames" skips 14 files that need eyes.

crates/db/src/txn/mod.rs
```diff
@@ -1,11 +1,11 @@
+use crate::collection::cache::CollectionCache;
```
jackzampolin: **[nit]** The `CollectionCache` import was inserted above the file's existing `///` block, splitting it from the other `crate::collection` import and leaving the module docs no longer leading the file. rustfmt will not rejoin the two `use` lines across that comment.

**Suggestion:** Move `use crate::collection::cache::CollectionCache;` down next to `use crate::collection::{populate_collection_root_id, Collection};` (after the doc comment), matching the pre-move grouping.

jackzampolin (commented):
## Summary

This is a clean mechanical relocation of `db` error/event/view/database/collection/txn modules into nested directories, plus the two claimed extra kinds of edits: new child `mod` declarations and `crate::action::` → `crate::database::action::` (with rustfmt rewraps). Relocated bodies match the old files aside from those path updates; `lib.rs` drops the old crate-root mods without duplicates; every `db::event_emission` (8 sites) and `db::collection_acp` (3 sites) consumer is updated, and crate-root re-exports keep `db::TxnBroadcaster`, `db::DbTransactionRegistry`, ACP helpers, etc. on their previous public paths. Nested module visibilities preserve crate-wide vs public access (`action`/`retriever` stay crate-internal; `collection::acp` and `event::emission` stay public under the new paths). One small import-placement nit in `txn/mod.rs`; no logic or behavior changes found.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

PR #1546, opened 2026-08-20
chore(db): compartmentalize db, 3/4 write, read, block, definition, access, docid

vertexclique:
## Context

Third of four. Stacks on 2/4. The remaining relocation. After this PR `src/`
contains `lib.rs`, thirteen directories, and three test-only files that 4/4 moves out.

## What changes

**115 files: `7 A / 2 D / 48 M / 58 R`.** 23 of the renames are `R100`. 28 of the
modified files are consumer path updates across six crates (`cli`, `db-backup`,
`db-merge`, `defra-node`, `ffi`, `p2p-adapter`).

- **`write/`** -- `auto_commit_mutator/` -> `write/autocommit/`, `doc_mutator.rs` ->
  `write/doc.rs`, `doc_write_queue.rs` -> `write/queue.rs`.
- **`read/`** -- six fetcher families collapse into one compartment:
  `auto_commit_fetcher` -> `read/autocommit/`, `doc_fetcher` -> `read/doc.rs`,
  `versioned_fetcher` -> `read/versioned.rs`, `index_seek` -> `read/seek.rs`,
  `vector_search` -> `read/vector.rs`, `commits_fetcher/` -> `read/commits/`.
  **`lensed_fetcher/` and `lensed_auto_commit_fetcher/` merge under one
  `read/lensed/`** as `lensed/fetcher/` and `lensed/autocommit/`, each with
  `fetch.rs`, `scan.rs`, `migration.rs`, `stream.rs`.
- **`block/`** -- `block_cleanup`, `block_reader`, `block_verify`,
  `commit_priority_index` -> `block/{cleanup,reader,verify,priority}.rs`.
- **`definition/`** -- `schema_loader` -> `definition/loader.rs`, `lens_utils` ->
  `definition/lens.rs`, `patch/` and `migration/` move under it, `json_patch.rs`
  becomes `definition/patch/json.rs`. Leaves are one word:
  `doc_short_id` -> `shortid`, `index_format` -> `format`, `set_migration` -> `set`,
  `version_id` -> `version`.
- **`access/`** -- `kms_adapters`, `nac_guard`, `node_access_checker` ->
  `access/{kms,nac,node}.rs`.
- **`docid/`** -- `doc_id_map.rs` -> `docid/map.rs`. This is the largest consumer
  sweep at 42 sites.

**Four changes are not moves:**

1. `write/autocommit/create.rs` absorbs `create_many.rs`, and `delete.rs` absorbs
   `delete_many.rs`. Merging is what makes the one-word rule work here: the
   alternative was `createmany.rs`. Bodies are concatenated verbatim; the imports
   are the union of the two heads.
2. `write/autocommit/mod.rs`: `pub(super) mod helpers;` becomes `pub(crate)`. Under
   the old flat layout `super` **was** the crate root, so `pub(super)` meant
   `pub(crate)`; nesting narrowed it. This restores the original effective
   visibility rather than widening anything.
3. `definition/patch/mod.rs`: `mod json;` is `pub(crate)`, matching what
   `json_patch` was at the crate root.
4. `read/mod.rs` and `write/mod.rs` are new files (the two compartments had no
   parent module before). Their headers are one line each.

**Proof citations:** 17 occurrences move here -- `doc_write_queue.rs` x11,
`doc_mutator.rs` x4, `auto_commit_mutator/{helpers,batch}.rs` x1 each -- across
`MergeQueue.tla`, `TwoStoreCounter.tla`, `InteractiveTxnCounter.tla`, their three
`_DESIGN.md` files, `lean/README.md`, `proofs/src/registry.rs` and
`proofs/tests/behavioral/partition.rs`.

## Verify

|---|---|---|
| `cargo test -p db` | 366 / 0 / 3 | **identical** |

- **Review with `-M`.** 23 files are `R100`; the rest are `R99`/`R98`/`R96` where
  the delta is the path rewrite and rustfmt's `use`-line re-sort.
- **Tests moved, not rewritten.** Normalized inventory unchanged, with eight honest
  exceptions: tests that spell `crate::doc_id_map::` or `crate::commits_fetcher::`
  inline in the body, so the prefix rewrite touched those bodies. Same assertions.
- The path rewrite is compiler-verified; a wrong rewrite does not compile.

## Review follow-ups (jackzampolin)

Approved with two follow-ups; both are fixed in this PR rather than deferred.

### Arithmetic

"115 files; 81 are renames" was wrong. Corrected above to `7 A / 2 D / 48 M / 58 R`.

### The fourth non-move, now declared and reverted

`block/mod.rs` had `pub mod priority;` where `db-compart-2` had
`mod commit_priority_index;` -- private, at `lib.rs:67`. That was an undeclared
visibility widening in a PR whose contract is "effective visibility preserved
exactly", and the reviewer is right that it is the inverse of the
`pub(super)` -> `pub(crate)` argument made two notes above it.

**Reverted to `mod priority;` here.** It becomes `pub` in 4/4, in the commit whose
whole purpose is widening for out-of-crate tests, where it is measured and declared
with everything else.

### Speculative module docs removed

`read/mod.rs:3` and `write/mod.rs:3` claimed "under an MVCC engine this is where
the visibility check / conflict detection attaches". That is a claim about a design
that does not exist, which CLAUDE.md §2 puts in GitHub issues and §6 calls a
speculative doc. It was also inaccurate: `collection/crud.rs` `*_with_datastore` are
the datastore read/write primitives, they are called from `crates/db-merge/`, and
they sit in neither compartment -- so "every snapshot a query resolves against"
overstated what `read/` holds.

Both headers are now a single accurate line. The MVCC motivation stays in the PR
description, which is its right home.

Also retargeted here: `InteractiveTxnCounter_DESIGN.md:70`'s
`auto_commit_mutator/create_many.rs` anchor, which this PR's merge deletes ->
`write/autocommit/create.rs`.

---

| Gate | Result |
|---|---|
| `cargo machete` | pass |

| Artifact | Result |
|---|---|


crates/db/src/read/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! The read path: every snapshot a query resolves against.
+//!
+//! Under an MVCC engine this is where the visibility check attaches.
```
jackzampolin: This is a claim about a design that doesn't exist yet, in a repo whose CLAUDE.md §2 puts the future in GitHub issues and §6 forbids speculative docs. Same line in `write/mod.rs:3` ("this is where conflict detection attaches").

The MVCC motivation is well argued in the PR description, which is the right home for it — it's reviewer context, not a durable property of the module. When the MVCC work lands, whoever does it will either make these true or have to notice and delete them.

It's also not quite accurate today: `collection/crud.rs:43,257,288,330,391` `*_with_datastore` are the datastore read/write primitives, and they sit outside both the `read/` and `write/` compartments — they're called from `crates/db-merge/`. So "every snapshot a query resolves against" overstates what `read/` actually contains.

Suggest trimming both headers to the first line, which is accurate and useful on its own.

jackzampolin (approved):
## Approve

The largest relocation in the stack, and it holds up. What I verified independently:

- `cargo test -p db` — **366 / 0 / 3**, identical to `origin/main`.
- The `create`/`create_many` and `delete`/`delete_many` merges: I reconstructed both originals from `db-compart-2` and diffed against the concatenation. They differ by **one blank line and two path rewrites**. "Bodies are concatenated verbatim" is accurate.
- The `lensed_fetcher/` + `lensed_auto_commit_fetcher/` merge under `read/lensed/` was the thing I most expected to be cross-wired — four same-named files (`fetch.rs`, `scan.rs`, `migration.rs`, `stream.rs`) in each family, which is exactly where `git diff -M` pairs the wrong source. Content lineage checks out on all eight; nothing swapped.
- Consumers: all path-token rewrites, no semantic change, no aliases. The 42-site `doc_id_map` → `docid::map` sweep is complete — `git grep -E '(crate|db|super)::doc_id_map'` at the tip returns nothing.

Approving. Two things below are worth a follow-up commit but shouldn't hold the merge.

Arithmetic nit, same class as #1544: "115 files; 81 are renames" — actual is `7 A / 2 D / 48 M / 58 R`.

One undeclared non-move: `crates/db/src/block/mod.rs:4` is `pub mod priority;`, but at `db-compart-2` it was `mod commit_priority_index;` — private, at `lib.rs:67`. Every other module in this sweep preserves effective visibility exactly, and the body lists three non-moves, so this is a fourth that didn't get called out. Nothing leaks today (the file holds a private const plus an `impl DB<S>`), and it's the inverse of the `pub(super)` → `pub(crate)` argument made two notes above it.

crates/db/src/read/mod.rs
vertexclique: Hah, it figured out why I wrote that. Which model was this glm?

PR #1545, opened 2026-08-20
refactor(db): compartmentalize db, 2/4 split the transaction registry and adopt one-word names

vertexclique:
## Context

Second of four. Stacks on 1/4. Two changes that are deliberately not folded into a
relocation PR, because neither can prove itself with `git diff -M`.

## What changes

**24 files: `8 A / 1 D / 12 M / 3 R`.**

### 1. `txn/registry.rs` (1,560 lines) becomes `txn/registry/`

Over the ~1000-line cap, and it holds the one genuinely novel concurrency protocol
in the crate. Split into eight files, largest 437:

| file | lines | holds |
|---|---:|---|
| `mod.rs` | 195 | type, constants, `CleanupResult`, constructors, accessors |
| `cleanup.rs` | 155 | `cleanup_stale_transactions`, `start_stale_transaction_cleanup` |
| `lifecycle.rs` | 412 | begin / commit / rollback, counter finalize, the `TransactionRegistry` impl |
| `collections.rs` | 434 | in-txn collection listing, activation, deletion |
| `schemas.rs` | 179 | in-txn schema and migration mutations |
| `lenses.rs` | 114 | in-txn lens transform registration and listing |
| `blocks.rs` | 72 | in-txn block signature verification |
| `docs.rs` | 42 | in-txn document reads |

`cleanup.rs` is the point: `cleanup_stale_transactions` is the subject of
`proofs/tla/TxnRegistry.tla`, and it is now a file the proof can name rather than a
region of a 1,560-line one.

**Proof the split is verbatim.** Strip the lines the split is allowed to add
(the eight `//!` headers, seven `use super::*;`, seven impl wrappers, seven `mod`
declarations, blank separators) and compare the remainder as a multiset against the
original: **1,416 content lines, identical, nothing added or dropped.** The single
exception is `#[path = "registry_tests.rs"]` becoming `#[path = "../registry_tests.rs"]`,
because the test file is now one directory up.

### 2. One-word module names

**Three renames, all `R100`:** `lens_store` -> `lenses`, `crud_datastore` -> `crud`,
`index_ops` -> `index`.

`txn/registry/lenses.rs` (114 lines) and `schemas.rs` (179) are **`A`, not renames** --
293 lines carved out of `registry.rs` by the split above. They are among the files
in this PR that most need reading, not files a reviewer can skip.

Both are pluralised rather than `lens`/`schema` because a child module named `lens`
shadows the `lens` **crate** for its parent, and `txn/registry/mod.rs` imports both.
[...]

PR #1547, opened 2026-08-20
test(db): compartmentalize db, 4/4 move every test into the tests tree

vertexclique:
## Context

Last of four. Stacks on 3/4. Moves every test out of `src/` and into
`crates/db/tests/`, which is what makes `src/` hold `lib.rs` and nothing else.

**This is the PR in the stack that is not a pure move, and the description says so
up front.** A file under `tests/` is a separate crate and sees only `pub` items, so
the move requires widening the internal surface. That widening is a real API change
and is isolated in its own commit so it can be read as an API proposal.

## What changes

**116 files: `39 A / 50 M / 27 R`, in four commits.**

### 1. `chore(db): widen db internals for out-of-crate tests` -- 19 files, visibility keywords only

Roughly 90 items become `pub`: compartment modules (`read`, `txn::registry`,
`collection::{cache,loader,name,...}`, `definition::patch`, ...), module-private
free functions the tests exercised through `use super::*`
(`delete_owned_commit`, `encode_status`, `resolve_cursor_seek_key`,
`sort_commits_go_order`, ...), and several struct fields the tests construct by
literal (`DB::lens_store`, `DB::active_actions`, `AutoCommitDocStream`'s two fields,
`MigrationWriteBack`'s, `MigrationOutcome`'s).

The set was **measured, not guessed**: the tests were moved, and every item the
compiler reported as unreachable was widened, scoped to the specific module the
failing test came from. `db` is a workspace-internal crate with no `cargo publish`
step, so the cost is rustdoc surface rather than a public contract. It is
permanent: those tests can no longer reach a private item.

### 2. `test(db): move every test out of src into the tests tree` -- 93 files

24 inline `#[cfg(test)] mod tests` blocks and 11 attached `tests.rs` / `*_tests.rs`
files become submodules of thirteen per-compartment integration binaries
(`tests/{access,block,collection,database,definition,docid,error,index,read,txn,view,write}/main.rs`),
plus `tests/property_tests.rs`. One binary per compartment rather than one per
file keeps the link cost at 14 targets instead of ~50.

The 17 pre-existing flat files in `tests/` are regrouped into the same
compartments, so the tree is not half-organised.

Both `#[path]` attributes are gone; `grep -rn '#\[cfg(test)\]' crates/db/src`
returns nothing.

**DRY.** Twelve helper fixtures that were duplicated across test files
(`test_schema`, `users_schema`, `test_collection`, `fixture_with_docs`,
`make_test_db_with_bus`, `RecordingStream`, `FailingCloseStream`, ...) are hoisted
once into `tests/common/{schema,fixture,stream,counting_store}.rs`. **Only
byte-identical bodies were hoisted** -- a same-named helper whose body differs
stays where it is, so no test can silently change meaning.

### 4. `ci(proofs):` -- a gate so proof anchors cannot rot silently

`.github/scripts/check-proof-anchors.sh`, wired into `just proofs` (which `just ci`
already runs). Every `*.rs` filename named under `proofs/` or `SURVEY.md` must
resolve to a tracked file. Prefix-free and repo-wide, because the previous check
matched only fully-qualified paths and missed the bare filenames most anchors use.

Verified it fails loud: reverting one anchor by hand makes it exit 1 and print
every occurrence. The 10 pre-existing dead anchors belonging to other crates'
slices are listed in `proofs/anchor-baseline.txt` with a per-entry reason, so the
check fails on anything new rather than being switched off.

## Verify

```sh
cargo fmt --all -- --check
just lint
just check-node-graph
```

|---|---|---|
| `cargo test -p db` | 366 passed / 0 failed / 3 ignored | **366 / 0 / 3** |

Closing assertions:

```sh
find crates/db/src -maxdepth 1 -type f     # exactly crates/db/src/lib.rs
grep -rn '#\[cfg(test)\]' crates/db/src    # nothing
grep -rn '#\[path' crates/db/src           # nothing
# every crates/db/src path cited under proofs/ resolves to a real file
```

## Notes

- Integration suites were not run locally -- see below.
- Integration suites were not run locally -- no `~/.cache/defra-harness` on this
  machine, so the `go_*` parity tests cannot start a Go node. CI has the pinned
  binary. The Rust-native half of `--test basic` is 24/24 locally.

### Compile time: measured, and it is a wash, not a win

The disclaimer is gone -- it should not have been there. But I measured it on this
machine rather than adopting the review's numbers, and I do not reproduce the
conclusion. `--profile super-dev`, 14-core Linux, median of 3, warm:

| | `main` | this tip |
|---|---:|---:|
| test executables | 18 | **14** |
| touch `src/lib.rs` | 3.42s | 3.39s |
| touch a src file that carries inline tests on `main` (`doc_id_map.rs` / `docid/map.rs`) | 3.48s | **3.33s** |
| touch a test file | **0.97s** | 1.25s |
| touch `tests/common/mod.rs` | n/a | 1.99s |
| full `cargo test -p db` | 366/0/3, 2.62s | 366/0/3, 2.65s |

Target count drops 18 -> 14, matching the review. The rest does not. The review's
headline row compares **two different edits** -- editing a src file on `main`
(3.93s) against editing a *test* file on the tip (0.91s). The like-for-like row is
editing the same src file on both sides, and there it is 3.48s -> 3.33s: a ~4%
improvement, because the lib still rebuilds either way. Editing a test file gets
*slower* (0.97s -> 1.25s), since a compartment binary is larger than the flat file
it replaced, and `tests/common/mod.rs` is a genuinely new 13-binary trigger.

So: fewer targets, no inline-test double-compile of the rlib, and otherwise a wash.
That is worth stating plainly rather than either disclaiming it or overselling it.

### File spacing moved rather than shrank -- stated straight

Correct, and worth being direct about. `crates/db/src`: 103 files / 35,008 lines /
4 over 1000 -> **106 / 27,624 / 0 over 1000**. But `crates/db/tests`: 17 / 6,854 /
1 over 1000 -> **65 / 14,284 / 4 over 1000**:

```
1999  tests/index/manager.rs          (was index_manager_tests.rs, 2004 -- untouched)
1216  tests/write/doc_suite.rs
1170  tests/definition/migration_suite.rs
1060  tests/txn/registry_suite.rs
```

Crate-wide that is 5 -> 4. `txn_registry.rs` (in 2/4) is the only file in the whole
stack that was genuinely split; everything else is relocation. CLAUDE.md's 400-line
threshold does not exempt tests, and 27 files in `src/` are still over it. Splitting
those four test files is real work with its own proof obligation and belongs in its
own PR, not smuggled into a relocation stack -- but the stack should not quote the
`src/` number alone, so it no longer does.

### Fixed here

- **Bisect gap.** `27574e96` deleted the three `#[cfg(test)] mod {counting_store,
  limit_pushdown_tests, plan_close_tests};` declarations while the files were still
  in its tree. An orphaned `.rs` compiles to nothing and warns about nothing, so
  **12 test functions silently did not run at that commit** -- `cargo test -p db`
  was green there while running 354, not the 366 the body claimed. The three
  deletions now land in the test-move commit instead. Verified at both ends: the
  widening commit declares all three and reports 366/0/3; the tip declares none and
  the files are gone.
- **Stale assertion message.** `tests/access/coverage.rs:147` routed a failing
  contributor to `nac_coverage_test.rs`, which is what that file used to be called
  -- at exactly the moment the message is meant to be doing work. Now
  `crates/db/tests/access/coverage.rs`. Same class fixed at `coverage.rs:10`,
  `tests/read/routing.rs:6`, `tests/read/lookup.rs:2`.
- **Duplicate fixture.** `tests/read/limit_pushdown.rs` still defined its own
  `users_schema`, byte-identical to the hoisted `tests/common/schema.rs:22`, while
  its sibling in the same binary already imported the shared one. Dropped, imported.
  A full re-scan of the test tree now reports **zero identical helper bodies in 2+
  files**.
- **Stale duplication note.** The "Duplicated from `read::autocommit::tests::
  RecordingStream`" comment in `lensed_autocommit_stream_suite.rs` was left over
  from before the DRY pass; that module no longer exists and the file imports the
  shared struct four lines above. Removed.
- **`SURVEY.md`.** Its four dead paths are fixed in 2/4 rather than here, so no
  intermediate branch in the stack ships a dead anchor.

## Notes

Integration suites were not run locally: this machine has no
`~/.cache/defra-harness`, so the `go_*` parity tests cannot start a Go node. CI has
the pinned binary (`GO_COMPAT_COMMIT` 13c46ec9). The Rust-native half of
`--test basic` is 24/24 locally.

---

| Gate | Result |
|---|---|
| `cargo machete` | pass |

| Artifact | Result |
|---|---|


crates/db/src/lib.rs
```diff
@@ -65,13 +63,9 @@ pub mod event;
 pub use event::emission::{TxnBroadcastEvent, TxnBroadcaster};
 // Index manager extracted to standalone db-index crate.
 pub use db_index as index_manager;
-#[cfg(test)]
-mod limit_pushdown_tests;
-pub(crate) mod read;
+pub mod read;
```
jackzampolin, line 66: The commit split doesn't hold, and it leaves one commit silently under-tested.

> That widening is a real API change and is isolated in its own commit so it can be read as an API proposal.

Commit `27574e96` removes these three declarations (visible in this hunk and the one above):

```
-#[cfg(test)] mod counting_store;
-#[cfg(test)] mod limit_pushdown_tests;
-#[cfg(test)] mod plan_close_tests;
```

but the files are still present in that commit's tree:

```
$ git cat-file -e 27574e96:crates/db/src/limit_pushdown_tests.rs && echo exists
exists
```

An orphaned `.rs` file is not compiled and produces no warning, so at `27574e96` those three files' tests silently do not run — 67 `fn` / `#[test]` items across the three. `cargo test -p db` there is green while running fewer tests than the stated 366 / 0 / 3.

It's transient — all three land under `tests/` in `081763b1`, and nothing is wrong at the tip. But it breaks `git bisect` across this range, and it makes the per-commit gate untrue for the one commit the body asks reviewers to read on its own.

Moving the three `mod` deletions into `081763b1` makes the stated contract true and costs nothing.

crates/db/tests/read/limit_pushdown.rs
```diff
@@ -12,16 +12,21 @@
 //! buffered scans, ...) happens inside its `Iterator` implementation, below
 //! this boundary, and is not observed here.
 
+use crate::common::counting_store::CountingStore;
+use db::AutoCommitMutator;
```
jackzampolin, line 17: This file picks up `common::counting_store` here but not `common::schema` — and it still defines its own `users_schema` at lines 33-45, byte-identical to the hoisted `tests/common/schema.rs:22-34`.

Its sibling module in the *same binary* already uses the shared one: `tests/read/plan_close.rs:8` is `use crate::common::schema::users_schema;`.

So the `read` binary carries two identical copies, one of which is the hoist destination — the precise drift hazard the DRY pass removed everywhere else:

> Twelve helper fixtures that were duplicated across test files … are hoisted once into `tests/common/`.

Dropping lines 33-45 and adding `use crate::common::schema::users_schema;` alongside this import finishes the job.

While in there: `tests/read/lensed_autocommit_stream_suite.rs:30-31` still carries "Duplicated from `read::autocommit::tests::RecordingStream`" above `fn wrapped_stream`, while `:4` imports the shared one and that module no longer exists.

jackzampolin (commented):
## Comment — the build claim is better than you think, plus three fixables

`cargo test -p db` is **366 / 0 / 3** at this tip, identical to `origin/main`. I extracted the test inventory statically as well: name sets match, lost and gained both empty, all `ProptestConfig::with_cases` counts preserved, all `#[cfg(feature = ...)]` gates intact. The move is faithful.

### The unmeasured claim points the wrong way — measure it and take the win

> The test binaries link the full `db` rlib, so `cargo test -p db --no-run` link time will rise. **Not measured.** No claim is made about it.

I measured it. It falls. `--profile super-dev`, M3, warm cache except where noted:

| | `main` | this tip |
|---|---|---|
| test targets | **18** | **14** |
| cold `cargo test -p db --no-run` | 3m21s | 3m20s |
| touch `src/lib.rs` | 4.72s | **3.93s** |
| touch a test file | 0.64s | 0.91s |
| touch a src file carrying inline `#[cfg(test)]` | **3.93s** | n/a — none exist |
| touch `tests/common/mod.rs` | n/a | 3.02s |
| full run | 366/0/3, 7.36s | 366/0/3, 7.31s |

Target count *drops* because 12 compartment binaries replace 17 flat ones, and the per-link dependency closure is unchanged. The row with no counterpart is the real win: on `main`, editing a source file that carries inline unit tests recompiles the 35k-line rlib a second time under `cfg(test)` and relinks everything — ~3.9s. Here, editing a test costs 0.9s and never touches the lib.

The only new many-binary trigger is `tests/common/mod.rs` (13 relinks, 3.0s), still cheaper than main's 18. Lint coverage is unchanged — `git diff origin/main origin/db-compart-4 -- .github/ justfile` is empty and ci.yml already runs `--all-targets`, so the test tree is linted exactly as the inline test code was.

Worth replacing the disclaimer with the table. This is the strongest evidence in the stack for the build-behavior goal, and the PR currently disclaims it.

### File spacing moved rather than shrank

`crates/db/src`: 103 files / 35,008 lines / 4 over 1000 → 106 / 27,629 / **0 over 1000**. Genuine.

But `crates/db/tests`: 17 files / 6,854 lines / 1 over 1000 → 65 / 14,299 / **4 over 1000**:

```
1999  crates/db/tests/index/manager.rs        (was tests/index_manager_tests.rs, 2004 — untouched)
1216  crates/db/tests/write/doc_suite.rs
1170  crates/db/tests/definition/migration_suite.rs
1060  crates/db/tests/txn/registry_suite.rs
```

Crate-wide the count goes 5 → 4, and `txn_registry.rs` is the only file in the stack that was genuinely split. CLAUDE.md's threshold is 400 lines and doesn't exempt tests. Not a merge blocker — but "fix pathological file spacing" is mostly relocation here, and it's worth being straight about that rather than quoting the `src/` number alone.

### A stale path in an assertion message

`crates/db/tests/access/coverage.rs:147` — GitHub shows this file as a pure rename so I can't comment on the line directly. It isn't a code comment, it's the failure message:

```rust
"permission `{perm}` must be in exactly one NAC-coverage category, found {cats:?}. \
 Add it to the appropriate list in nac_coverage_test.rs."
```

`crates/db/tests/nac_coverage_test.rs` is what this file *was*. So: someone adds a `NodePermission`, the test fails, and the message routes them to a path that no longer exists — at exactly the moment the message is supposed to be doing work. Should read `crates/db/tests/access/coverage.rs`.

Same class, lower stakes: `coverage.rs:10`, `tests/read/routing.rs:6`, `tests/read/lookup.rs:2`.

### `SURVEY.md` was missed by the sweep

Untouched by the whole stack, with 4 dead fully-qualified paths:

```
SURVEY.md:29  crates/db/src/collection_ops/delete.rs:68    -> collection/ops/delete.rs
SURVEY.md:33  crates/db/src/txn_context.rs:29              -> txn/context.rs
SURVEY.md:33  crates/db/src/txn_registry.rs                -> txn/registry/
SURVEY.md:42  crates/db/src/collection_ops/mod.rs:56-243   -> collection/ops/mod.rs
```

These carry the `crates/db/src/` prefix, so this PR's closing assertion — "every `crates/db/src` path cited under `proofs/` resolves to a real file" — would have caught them if run repo-wide instead of `-- proofs`. See #1545 for the 26 bare-filename anchors in `proofs/` itself and a prefix-free gate.

crates/db/src/database/mod.rs
```diff
@@ -224,7 +224,7 @@ pub struct DB<S: Store> {
     /// process can exit while an action is in progress and leave that state
     /// behind. The registry provides cancellation-safe mutual exclusion for
     /// operations running in this database instance.
-    pub(crate) active_actions: Arc<crate::database::action::ActionRegistry>,
+    pub active_actions: Arc<crate::database::action::ActionRegistry>,
```
jackzampolin, line 227: **[suggestion]** `DB::active_actions` and `ActionRegistry::active` (`crates/db/src/database/action.rs:48`) went from crate-private to `pub`. `active` is a `Mutex<HashSet<ActionKey>>`, so any workspace crate with `&DB` can `db.active_actions.active.lock().clear()` and drop process-local action claims without going through a lease. The only test that needs this is `finalization_failure_releases_process_local_claim`, which asserts the set is empty after a failed `complete_action`. `DB::lens_store` (`:185`) is the same pattern: tests assign a stub store, but a field write is now a supported mutation of the live transform store. There is already `DB::lens_store()` and `DB::set_kms()`-style setters for neighboring handles.

**Suggestion:** Keep these fields crate-private. For the action test, retry `register_action` after the failed `complete_action` (public API already proves the claim was released), or add `ActionRegistry::is_empty()`. For lens injection, add `DB::set_lens_store` (mirroring `set_kms`) instead of a public field.

crates/db/src/read/autocommit/mod.rs
```diff
@@ -457,9 +457,9 @@ impl<S: Store + 'static> DocFetcher for AutoCommitFetcher<S> {
 /// doesn't stop this struct satisfying `DocStream`'s `MaybeSendSync` bound;
 /// every access goes through `&mut self` via `get_mut`, so it never actually
 /// locks.
-struct AutoCommitDocStream<S: Store + 'static> {
-    inner: Option<Box<dyn DocStream>>,
-    txn: std::sync::Mutex<Option<DbTxn<S>>>,
+pub struct AutoCommitDocStream<S: Store + 'static> {
+    pub inner: Option<Box<dyn DocStream>>,
+    pub txn: std::sync::Mutex<Option<DbTxn<S>>>,
```
jackzampolin: **[suggestion]** `AutoCommitDocStream` and `LensedAutoCommitDocStream` (`crates/db/src/read/lensed/autocommit/stream.rs:42`) were private structs; every field is now `pub`, including `txn: Mutex<Option<DbTxn<S>>>`. Production code only touches `txn` via `get_mut` on `&mut self`. With a public mutex, any crate can `stream.txn.lock().take()` on a shared reference and steal or double-discard the read transaction the stream still owns. Tests only need (1) struct construction wrapping a `RecordingStream` / `FailingCloseStream` and (2) one assertion that close released the txn (`tests/read/autocommit_suite.rs`). `LensedDocFetcher::history_cache` (`crates/db/src/read/lensed/fetcher/mod.rs:63`) is the same leak: the unknown-version test writes the cache directly.

**Suggestion:** Add `pub fn new(...)` constructors (and a `txn_released()` / `insert_history_for_tests()` helper) and keep the fields private. That still lets `tests/` live out of crate without handing every consumer a live txn slot.

crates/db/src/collection/snapshot.rs
```diff
@@ -14,7 +14,7 @@ use crate::collection::Collection;
 /// The snapshot is wrapped in `Arc` internally for efficient cloning and sharing.
 #[derive(Debug, Clone)]
 pub struct CollectionSnapshot {
-    collections: Arc<HashMap<String, Collection>>,
+    pub collections: Arc<HashMap<String, Collection>>,
```
jackzampolin, line 17: **[suggestion]** `CollectionSnapshot` is already in the crate's public re-exports (`lib.rs` `pub use collection::snapshot::CollectionSnapshot`). Its `collections` map was private and is now `pub` solely so `test_snapshot_clone_is_cheap` can `Arc::ptr_eq`. Any downstream crate with `&mut CollectionSnapshot` can now swap the snapshot's map. Cheap Arc clone is an implementation detail, not a behavior the public API needs to expose.

**Suggestion:** Drop the `ptr_eq` assertion (clone + `get`/`len` still covers sharing) or add a crate-internal/`#[cfg(test)]`-free `fn ptr_eq(&self, other: &Self) -> bool`. Leave the field private.

crates/db/src/write/autocommit/helpers.rs
```diff
@@ -66,7 +66,7 @@ pub(crate) async fn write_local_update(
 /// Persist a local CREATE: write the document + maintain indexes, then seed the
 /// CRDT accumulation store for counter fields (#1021). Bundled for the same
 /// by-construction reason as `write_local_update`.
-pub(crate) async fn write_local_create(
+pub async fn write_local_create(
```
jackzampolin: **[suggestion]** `write_local_create`, `write_local_update` (`:47`), `write_branchable_collection_block` (`:12`), `register_created_doc` (`:127`), `register_block_doc_id_mappings` (`:173`), and `ensure_collection_is_active` (`:687`) became `pub`, and `write/autocommit/mod.rs:11` now has `pub mod helpers`. No file under `crates/db/tests` calls any of them; sibling modules still compile with `pub(crate)`. These are the persist path (ACP, events, and txn finalize sit above them). Same for `register_update_event_callback` (`crates/db/src/event/emission.rs:54`) and `pub struct TxnLensStore` (`crates/db/src/txn/lenses.rs:19`, `new` still `pub(crate)`): compiler-driven leftovers from the widen commit, not test requirements after the move.

**Suggestion:** Revert these to `pub(crate)` (and `pub(crate) mod helpers` if nothing in `tests/` imports it). Widen only symbols a `tests/` path actually names.

crates/db/tests/property_tests.rs
```diff
@@ -14,12 +18,6 @@ use storage::index::IndexIterator;
 /// Allocate a distinct doc short ID for index-layer tests. Index entries are
```
jackzampolin: **[nit]** After hoisting `next_test_doc_short_id` into `tests/common/fixture.rs`, its doc comment was left in place and now sits on `test_schema()`, glued to "Generate a test schema with common fields." It describes a function that is no longer here.

**Suggestion:** Delete the orphaned three-line comment (the helper in `fixture.rs` already documents the short-id allocator).

jackzampolin (commented):
## Summary

The test relocation is complete and faithful: 372 tests on both sides, no names lost, no `#[cfg(test)]` left under `crates/db/src`, and `src/` is only `lib.rs` plus compartment directories. Shared fixtures in `tests/common/` match the originals they replaced. The real cost is the visibility widening: several `pub` fields on already-public types (`DB`, streams, `CollectionSnapshot`) expose interior-mutable internals that any workspace crate can smash with a shared reference, and a handful of write-path helpers became `pub` even though no `crates/db/tests` caller needs them.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 1

PR #1549, opened 2026-08-20
test(p2p): move wire tests into the tests tree

iverc:
## Context

This is 2/4 in the P2P test-compartment stack and depends on #1554. It starts moving tests out of production source files without changing the production API.

## Changes

- groups CAR, management, and pubsub wire tests under `tests/wire`
- moves management and searchable-encryption correlator lifecycle tests into the same integration binary
- preserves all 43 tests under `--no-default-features`

## Testing

- [x] `cargo test --profile super-dev -p p2p --no-default-features`


edjroz (approved):

PR #1550, opened 2026-08-20
test(p2p): move libp2p protocol tests into the tests tree

iverc:
## Context

This is 3/4 in the P2P test-compartment stack and depends on #1549. It moves libp2p protocol tests behind a dedicated feature-aware integration target.

## Changes

- groups pubsub RPC topic, envelope, request ID, correlator, and handler tests under `tests/libp2p_protocol`
- moves two-stream protocol, event, and management message tests into the same target
- preserves all 34 tests with `libp2p-transport` enabled and keeps no-default builds feature-clean
- does not widen private host or pending-response APIs

## Testing

- [x] `cargo test --profile super-dev -p p2p --test libp2p_protocol --no-default-features`


edjroz (approved):

PR #1551, opened 2026-08-20
test(p2p): move sync state tests into the tests tree

iverc:
## Context

This is 4/4 in the P2P test-compartment stack and depends on #1550. It moves sync tests that exercise exported behavior without exposing sync internals for test access.

## Changes

- moves peer-state tracking tests under `tests/sync`
- moves libp2p broadcaster construction coverage into the same feature-aware target
- preserves 18 no-default tests and all 19 tests with libp2p enabled
- leaves private queue and coordinator tests for a separate visibility-review layer

## Testing

- [x] `cargo test --profile super-dev -p p2p --test sync --no-default-features`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -692,10 +692,10 @@ jobs:
           "$CACHE_DIR/defradb" version --format json
 
       # Guards against silent drift between the committed Go-parity hex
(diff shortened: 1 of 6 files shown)


edjroz (approved):

PR #1554, opened 2026-08-20
fix(ci): satisfy Rust 1.98 Clippy lints

iverc:
## Context

This is 1/4 in the P2P test-compartment stack. Rust 1.98 introduced new default Clippy diagnostics that caused the workspace lint job to fail before reaching the stack's own checks.

## Changes

- decode IVF-PQ and SSG values through fixed-size byte arrays
- box WebSocket errors at the SourceHub event-consumer boundary
- retain the direct P2P recovery outcome without an unnecessary allocation
- preserve existing runtime behavior and wire representations

## Testing

- [x] `cargo clippy -p defra-wasm --target wasm32-unknown-unknown --no-default-features --all-targets -- -D warnings`
- [x] `cargo test --profile super-dev -p db-index`
- [x] `cargo test -p sourcehub --lib`
- [x] `cargo test -p p2p --lib`

### DIFF
--- a/crates/sourcehub/src/cosmos/event_subscriber.rs
+++ b/crates/sourcehub/src/cosmos/event_subscriber.rs
@@ -106,7 +106,7 @@ fn retry_delay_after_session(current: Duration, session_duration: Option<Duratio
 async fn consume<S>(
     socket: &mut tokio_tungstenite::WebSocketStream<S>,
     cache: &AccessCache,
-) -> Result<(), tokio_tungstenite::tungstenite::Error>
+) -> Result<(), Box<tokio_tungstenite::tungstenite::Error>>
 where
     S: tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin,
 {


edjroz:
Reviewed the full stack (#1554 -> #1549 -> #1550 -> #1551) as one cumulative diff against `main`: 37 files, +1319/-1335.

**Approving.** The test move is clean.

Verified:

- Normalized (whitespace/import-insensitive) diff of all 15 moved `#[cfg(test)]` blocks against their new files: only `use` lines and two `super::super::topic::response_topic` -> `response_topic` call-path rewrites changed. No assertion, fixture, or constant altered.
- Test-function counts match one-for-one across all 15 files. Nothing dropped to make the move compile.
- Every symbol the new integration tests reach was already `pub`. No visibility widened to accommodate the move.
- Feature gating preserved: `libp2p_protocol/main.rs` carries `#![cfg(feature = "libp2p-transport")]`, `sync/main.rs` gates only `mod broadcaster`, `wire/` stays ungated over ungated modules only.
- `tests/wire/pubsub.rs` correctly imports `p2p::message::pubsub::*` rather than `p2p::message::*`. Worth calling out: `message/` carries two distinct `DocSyncRequest`/`DocSyncItem`/`MAX_DOC_IDS` families, and the `p2p::message::*` re-export would have compiled while silently testing the `docsync.rs` types against the Go byte fixtures.
- Lint fixes: `slice::as_chunks` (stable 1.88) is within the workspace `rust-version = "1.91"`; both codec rewrites stay behind the existing `is_multiple_of` guard and are behaviorally identical to the `chunks_exact` versions; the `Box<tungstenite::Error>` return type still flows through `?` and formats via `Display`.

**Non-blocking:** `crates/p2p/src/sync/replication/handlers.rs:63` suppresses `clippy::result_large_err` with `#[allow]`, while the same commit fixes the identical lint in `crates/sourcehub/src/cosmos/event_subscriber.rs:108` by boxing. `Result<RecoveredBlockMetadata, ReplicationResult>` is returned per block during push-driven DAG recovery, so the oversized `Err` variant is moved on every call. Boxing would match the sibling fix; if the `#[allow]` is deliberate, a short `// reason` keeps it from reading as an oversight. Happy for this to land as is.

edjroz (approved):

PR #1556, opened 2026-08-21
chore(db): compartmentalize db, 5/5 fold the six db-* crates into db

vertexclique:
## Context

Fifth of five. Stacks on 4/4. Folds the six `db-*` crates into `db` as modules,
with their tests and benches. After this, `cargo metadata` lists one `db` package
and no `db-*` package exists.

`defra-node` and `defra-core` are untouched -- they are not `db-*` crates and stay
where they are.

## What changes

**264 files: `1 A / 7 D / 104 M / 152 R`.** 152 of them are renames.

| crate | -> module | src | tests | benches |
|---|---|---:|---:|---:|
| `db-backup` | `db::backup` | 4 | - | - |
| `db-blocks` | `db::block::builder` | 10 | - | 1 |
| `db-index` | `db::index` | 54 | 24 | 4 |
| `db-merge` | `db::merge` | 45 | - | 1 |
| `db-nac` | `db::nac` | 5 | - | - |
| `db-search` | `db::search` | 4 | - | - |

`db` ends at **227 src files / 56,848 lines** and **90 test files / 20,659 lines**,
with `src/` still holding `lib.rs` and nothing else.

### Naming

One word throughout, matching the rest of the stack. The three aliases `db` used
to expose its own dependencies under are gone, so each concept has exactly one path:

```
-pub use db_blocks as block_builder;      db::block_builder::  -> db::block::builder::
-pub use db_index  as index_manager;      db::index_manager::  -> db::index::
-pub use db_search as dense_search;       db::dense_search::   -> db::search::
-pub(crate) use db_nac as nac;
```

Two deviations from a literal `db-X -> db::x` mapping, both to avoid a trap:

- **`db-blocks` -> `db::block::builder`, not `db::blocks`.** `db::block` already
  exists (reading, verification, cleanup, priority index) and `db-blocks` is block
  *building* -- the same concept. A sibling `db::blocks` would sit one letter from
  `db::block` and read almost identically, which is the exact hazard this stack has
  been removing. Folding it under `block/` also avoids merging two `lib.rs`/`mod.rs`
  bodies, so it stays a clean rename.
- **`index/index_manager` -> `index/manager`, `search/dense_search` -> `search/dense`**,
  applying the one-word rule inside the folded trees.

### Not moves

1. **`p2p` becomes a non-optional dependency of `db`.** It already was one in
   `db-merge` (`default-features = false`), and nine folded modules use `p2p::`
   without a feature gate because they never needed one. The alternative -- gating
   them behind `db/p2p` -- would silently drop those modules from any build that
   does not enable it, which is what `cargo clippy -p defra-node --features otel`
   caught. `libp2p` stays optional behind `db`'s `p2p` feature, and
   `just check-node-graph` still passes, including the isolated-native assertion
   (now `assert_absent db libp2p`, was `db-merge`).
2. **`db-backup`'s module inception is flattened.** Its `lib.rs` only declared
   `pub mod backup;` and re-exported, so the fold produced `backup/backup/`.
   Collapsed to `backup/`, which is also what `clippy::module_inception` requires.
3. **Five rustdoc warnings fixed** in the folded code (`index/vector/engine/ivfpq`,
   `merge/se/storage`). These were pre-existing -- `just doc` is red on `main` for
   exactly this class -- but `cargo doc -p db --no-deps -D warnings` is a gate this
   stack keeps green, and folding pulled them into it.
4. `db-index`'s 23 test binaries become one `tests/vector/` compartment binary
   rather than 23 new link targets. Test bodies are unchanged; only `mod common;`
   becomes `crate::support::`, matching how the rest of `db/tests` is wired.

### Paths, proofs and build config

Everything naming the old crates is retargeted, not left to rot:

- **31 proof files** (`proofs/tla/*.tla`, `*_DESIGN.md`, `proofs/lean/**`,
  `proofs/survey/*`, `proofs/README.md`) -- every `crates/db-X/src/...` path and
  every `db_X::` reference.
- The six `proofs/survey/db-*.md` slices each gain a one-line banner saying they
  now describe a module of `db` and where it lives. Their content is unchanged --
  the code did not move out from under the model, only its path did.
- `proofs/survey/INDEX.md` and `proofs/README.md`: 40 crates -> 34.
- `CLAUDE.md`: the six rows collapse into the `db` row.
- `justfile`: `cargo bench -p db-index` -> `-p db`, `cargo check -p db-merge
  --no-default-features --features native` -> `-p db`.
- `.github/workflows/ci.yml` and `.github/scripts/assert-defra-node-graph.sh`:
  same two substitutions.

`just check-anchors` (added in 4/4) passes, which is what proves there is no dead
path left anywhere under `proofs/` or `SURVEY.md`.

## Verify

```sh
cargo fmt --all -- --check
just lint
just check-node-graph
just check-anchors
```

| Gate | Before the fold (`db-compart-4`) | This PR |
|---|---|---|
| `cargo fmt --check` / `just lint` / `cargo doc -p db` / `just check-node-graph` / `just check-anchors` | pass | pass |
| `cargo test` summed over `db` + the six `db-*` crates | 705 passed / 0 failed / 15 ignored | **705 / 0 / 15 in one crate** |
| test-name set | 715 names | **identical, 0 lost, 0 gained** |
| `db*` packages in `cargo metadata` | 7 | **1** |

The per-crate baseline it is summed from: `db` 366/0/3, `db-blocks` 26/0/0,
`db-index` 181/0/10, `db-merge` 119/0/2, `db-nac` 7/0/0, `db-search` 6/0/0,
`db-backup` 0/0/0.

## Notes

- Integration suites were not run locally: no `~/.cache/defra-harness` on this
  machine, so the `go_*` parity tests cannot start a Go node. CI has the pinned
  binary.
- Compile-time effect of the fold specifically is **not measured**. 4/4 has the
  measurement for the test-relocation half; this PR changes crate boundaries, which
  is a different question and deserves its own numbers rather than an assumption.

---

| Gate | Result |
|---|---|
| `cargo machete` | pass |

| Artifact | Result |
|---|---|


crates/db/src/lib.rs
```diff
@@ -43,35 +43,36 @@
 /// txn.commit().await?;
 /// ```
 // Backup extracted to standalone db-backup crate (#789). Callers now
```
jackzampolin: **[suggestion]** These comments still describe the pre-fold crate split, and this PR touched them (`db_backup` → `crate::backup`) without dropping the false claim. Same file still says backup/blocks/search/index/NAC were extracted to standalone crates (lines 45–47, 56, 64, 69, 98) and, unchanged but now actively wrong, that consumers “should depend on db-merge directly” (lines 127–128). That last line will send new callers to a crate that no longer exists.

**Suggestion:** Delete the extraction-history comments. The `pub mod` lines already document the layout. If any note is kept, it should say these are modules of `db` (e.g. `db::backup`, `db::merge`) and that consumers depend on `db` only.

crates/db/src/index/mod.rs
```diff
@@ -6,13 +6,13 @@
 //!
 //! Depends only on storage/datastore/schema/document — no dependency
```
jackzampolin: **[suggestion]** Folded crate-root rustdocs still talk as if they were separate packages. `index` claims it “Depends only on storage/datastore/schema/document — no dependency back to `db`, so `db` can depend on this crate cleanly”; `search/mod.rs:14–15` says the same about `query`/`reqwest`; `merge/mod.rs:3–4` says “This crate contains all P2P-related database logic that was previously behind `#[cfg(feature = "p2p")]` in the `db` crate.” All three are now modules of `db`. `index/error.rs:3–11` still says “index_manager” and “lives in the `db` crate.”

**Suggestion:** Rewrite these as module docs (what the module owns). Drop crate-boundary and extraction-epic narration; the `From<crate::index::Error>` impl is now same-crate and does not need a “thin adapter” story.

CLAUDE.md
```diff
@@ -63,14 +63,10 @@ crates/
...
-│                       #   every test lives in db/tests/.
-├── db-backup/          # Database export/import (backup)
-├── db-blocks/          # IPLD block builder for document mutations
-├── db-index/           # Secondary index manager for collections
-├── db-merge/           # P2P merge, broadcast, and document pushing
-├── db-nac/             # Node Access Control state management
-├── db-search/          # Vector search, embeddings, hybrid BM25+dense retrieval
+│                       #   event/, error/, plus the folded former crates:
+│                       #   backup/, block/builder/, index/, merge/, nac/,
+│                       #   search/. src/ holds only lib.rs; every test
```
jackzampolin, line 68: **[suggestion]** This PR restates “src/ holds only lib.rs; every test lives in db/tests/.” while folding in substantial in-src `#[cfg(test)]` modules (`merge/merge_handler/tests.rs`, `merge/browser_sync/tests.rs`, `nac/tests.rs`, `block/builder/{tests,write_*_tests}.rs`, plus others). The crate-list update is otherwise accurate; that sentence is not.

**Suggestion:** Either move those tests under `crates/db/tests/` to match the compartment rule, or drop/qualify the “every test lives in db/tests/” claim (e.g. integration tests in `tests/`, unit tests may stay next to the code).

crates/db/src/block/reader.rs
```diff
@@ -3,7 +3,7 @@
 //! These functions are kept in the `db` crate (not `db-blocks`) because they
```
jackzampolin: **[suggestion]** The module docs still justify keeping these helpers “in the `db` crate (not `db-blocks`)” to avoid a circular dependency. `db-blocks` is gone; the import on line 6 is now `crate::block::builder`. Same class of leftover at `database/mod.rs:12` (“extracted to standalone db-search crate”) next to `pub use crate::search::EmbeddingClientConfig`.

**Suggestion:** Delete the crate-split rationale. A one-liner that this module reads committed documents via `DB`/`DbTxn` is enough if anything is needed.

crates/db/src/lib.rs
```diff
@@ -21,7 +21,7 @@
 /// # Example
 ///
 /// ```ignore
-/// use db::{DB, Collection};
+/// use crate::{DB, Collection};
```
jackzampolin: **[nit]** The crate-level example was changed from `use db::{DB, Collection}` to `use crate::{DB, Collection}`. The block is ` ```ignore `, so this is not a doctest fix; it is what `cargo doc -p db` shows to API consumers, who import `db::`, not `crate::`.

**Suggestion:** Restore `use db::{DB, Collection};`.

jackzampolin (commented):
## Summary

The fold is mechanically correct: the six `db-*` crates leave the workspace, `db` is the single package, there is no `db`↔`p2p` cycle, `libp2p` stays optional behind `db`'s `p2p` feature, and consumers (cli, defra-node, embedded, ffi, p2p-adapter, wasm) plus justfile/CI/node-graph were retargeted. Feature mapping (`db-merge/libp2p-transport` → `db/p2p`, `db-nac/redb` absorbed into `db/redb`) and the vector-test consolidation (23 binaries → `tests/vector/`) look complete; leftover `db_*` / `block_builder` / `index_manager` / `dense_search` crate aliases are gone from Rust sources. Remaining issues are stale crate-boundary comments and copy-pasted `cargo -p db-index` recipes that this PR both introduced into `db` and, in `lib.rs`/`CLAUDE.md`, actively edited without correcting.

## Issue counts by severity

- bugs: 0
- suggestions: 5
- nits: 1

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[suggestion]** crates/db/benches/sift.rs:6 -- Recipes copied with the benches/tests still invoke the deleted `db-index` package. Same stale `-p db-index` appears in `benches/vector_mutations.rs`, `benches/vector_search.rs`, and the ignored vector tests `tests/vector/{ivfpq_recall_baseline,ssg_vs_hnsw,vector_kv_cost,vector_recall_baseline}.rs`. `just bench-vector` was updated to `-p db`; these rustdocs were not. Following them now fails with “package `db-index` not found”, and `--test ivfpq_recall_baseline` is also wrong because those files are modules of the single `vector` test binary.
  - **Suggestion:** Use `cargo bench -p db --bench sift` (and the matching bench names). For ignored tests: `cargo test --release -p db --test vector -- <module> --ignored --nocapture`.

crates/db/benches/sift.rs
```diff
@@ -11,15 +11,15 @@
 //! producing it.
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
-use db_index::vector::store::NodeId;
+use db::index::vector::store::NodeId;
```
jackzampolin, line 14: **[suggestion]** Recipes copied with the benches/tests still invoke the deleted `db-index` package. The rustdoc just above this import is `cargo bench -p db-index --bench sift`. Same stale `-p db-index` appears in `benches/vector_mutations.rs`, `benches/vector_search.rs`, and the ignored vector tests `tests/vector/{ivfpq_recall_baseline,ssg_vs_hnsw,vector_kv_cost,vector_recall_baseline}.rs`. `just bench-vector` was updated to `-p db`; these rustdocs were not. Following them now fails with “package `db-index` not found”, and `--test ivfpq_recall_baseline` is also wrong because those files are modules of the single `vector` test binary.

**Suggestion:** Use `cargo bench -p db --bench sift` (and the matching bench names). For ignored tests: `cargo test --release -p db --test vector -- <module> --ignored --nocapture`.

vertexclique:
Ok nits are fixed. Makes sense. I am merging this once CI is green. cc/ @jackzampolin

PR #1570, opened 2026-08-23
/document crate unit test clean up

### LINKED ISSUE #1532: Crate cleanup: document, embedded, and events

christopherbm:

Validation:
  cargo test -p document — 145 passing, 0 failures.
  cargo clippy -p document --all-targets -- -D warnings — clean.
  cargo fmt -p document applied.

---


edjroz (approved):
Approving.

Verified this is a mechanical relocation rather than a rewrite:

- Zero lines added to any `crates/document/src/*.rs` file. Production code only loses `#[cfg(test)]` blocks.
- `#[test]` / `#[tokio::test]` name set is identical: 140 before, 140 after.
- `assert!` / `assert_eq!` / `assert_ne!` counts unchanged (134 / 139 / 3).
- No `pub` or `pub(crate)` widened to make a moved test compile. Everything the new integration tests reach was already public (`pub mod encoding` at `src/lib.rs:29`; `json_to_normal_value_for_kind` and friends already `pub`).
- `cargo test --profile super-dev -p document -p events -p embedded`: 170 passed on `main`, 170 passed on the stack tip. Coverage parity measured, not assumed.
- `cargo clippy -p document --all-targets -- -D warnings`: clean, so the strip left no orphaned imports.

Three non-blocking notes, all fine as follow-ups:

**Should** - `cargo test -p document --lib` now runs 0 tests, down from 76. CI is unaffected because `ci.yml:344` uses `cargo test --workspace`, which builds the `tests/` targets too. But anyone (or any agent) narrowing to `--lib` locally now gets a false green. Worth a line in the PR body.

**Optional** - `test_cbor_encoding_matches_go` was renamed to `test_cbor_encoding_matches_go_nested_json` (`tests/document_tests.rs:418`). The body is byte-identical and there was no name collision to avoid: `git grep -c 'fn test_cbor_encoding_matches_go' main -- crates/document` returns a single hit at `src/document.rs:478`. So this is a cosmetic change riding along in an otherwise purely mechanical move. Fine either way, just worth naming.

**Optional** - `crates/document/tests/` is now 10 separate integration-test binaries, up from 3. Each is its own compile and link unit. The pattern `CLAUDE.md` documents for `integration-test` is one `[[test]]` binary with submodules. Folding these into `tests/document.rs` with `mod field_tests; mod value_tests;` etc. would cut 7 link steps off every `cargo test -p document`. Follow-up, not a blocker.

PR #1571, opened 2026-08-23
embedded crate unit test clean up

### LINKED ISSUE #1532: Crate cleanup: document, embedded, and events

christopherbm:

Validation:
  cargo test -p embedded — 13 passing, 0 failures.
  cargo clippy -p embedded --all-targets -- -D warnings — clean.
  cargo fmt -p embedded applied.

---


edjroz (approved):
Approving.

- Zero lines added to `crates/embedded/src/node.rs`. The diff is the `#[cfg(test)] mod tests` block leaving, nothing else.
- The moved test lands in the existing `tests/shutdown.rs` alongside four sibling shutdown cases. `#[test]` name set unchanged at 14.
- No visibility widened. `build_with_store` was already re-exported (`src/lib.rs:13`) and `storage::corekv::private::Sealed` was already reachable externally, which is why `SlowCloseStore` still compiles from an integration test.
- `cargo test -p embedded` and `cargo clippy -p embedded --all-targets -- -D warnings`: both clean.
- `crates/embedded/src/node_acp.rs:98` correctly keeps its inline tests, since `create_document_acp` is `pub(crate)`. Good call not to widen it just to relocate a test.

One non-blocking note:

**Should** - `concurrent_shutdown_callers_wait_for_teardown_completion` relies on a single `tokio::task::yield_now()` followed by `assert!(!second.is_finished())` (`tests/shutdown.rs:171-175`). That was already timing-dependent, but it used to run in the `embedded` lib test binary and now shares a process with four other `multi_thread, worker_threads = 2` node-startup cases, including `p2p_shutdown_releases_persistent_store` (libp2p + lark). I could not make it flake (5 passed in 0.61s), so this is not a merge blocker, but the contention profile did change. Worth a follow-up replacing the yield with a deterministic barrier.

PR #1572, opened 2026-08-23
events crate test cleanup

### LINKED ISSUE #1532: Crate cleanup: document, embedded, and events

christopherbm:

Validation:
  cargo test -p events — 12 passing, 0 failures.
  cargo clippy -p events --all-targets -- -D warnings — clean.
  cargo fmt -p events applied.

---

### DIFF
--- a/crates/events/src/channel_bus.rs
+++ b/crates/events/src/channel_bus.rs
@@ -244,7 +244,3 @@ impl Bus for ChannelBus {
         self.closed.load(Ordering::Acquire)
     }
 }
-
-#[cfg(test)]
-#[path = "channel_bus_tests.rs"]
-mod tests;
--- a/crates/events/src/event.rs
+++ b/crates/events/src/event.rs
@@ -377,7 +377,3 @@ impl Message {
         }
     }
 }
-
-#[cfg(test)]
-#[path = "event_tests.rs"]
-mod tests;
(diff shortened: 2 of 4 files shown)


edjroz (approved):
Approving.

- Zero lines added to `src/`. `channel_bus.rs` and `event.rs` each lose only their `#[cfg(test)] #[path = ...] mod tests;` wiring.
- `#[test]` name set unchanged at 13, assertion counts unchanged.
- `src/subscription_tests.rs` correctly stays inline, since `Subscription::new` is `pub(crate)` (`src/subscription.rs:38`). Right call: the usual failure mode for this refactor is widening visibility to make a move compile, and this stack never does that. Worth one line in the PR body so the next reader does not read the leftover as an abandoned sweep.
- `cargo test -p events` and `cargo clippy -p events --all-targets -- -D warnings`: clean.

One non-blocking note:

**Should** - `tests/channel_bus_tests.rs` lost the `channel` feature gate. `mod channel_bus` is gated at `src/lib.rs:27` behind the default `channel` feature, so the old inline tests inherited it. The new integration-test file has no gate and imports `ChannelBus` / `ChannelBusConfig` unconditionally:

```
cargo test -p events --no-default-features --test channel_bus_tests
error[E0432]: unresolved imports `events::ChannelBus`, `events::ChannelBusConfig`
```

One-line fix: `#![cfg(feature = "channel")]` at the top of the file, or `required-features = ["channel"]` on a `[[test]]` entry. Low urgency, since the crate already fails `--no-default-features` on pre-existing errors in `src/subscription_tests.rs` (E0433 `async_channel`, E0599 `Subscription::new`) and no CI job exercises that combination. But if the lib is fixed later, this file will need the gate too.

PR #1574, opened 2026-08-24
fix(http): mount the API router at Go's unversioned /api prefix

### LINKED ISSUE #1495: Unversioned /api/* routes 404 on Rust where Go mounts the full router there
Found while fixing #1411 by reading Go's request mount rather than only its route table. Triage procedure: #1391.

## Divergence

Go mounts its API router at three places (`http/handler.go:116-128`):

```go
mux.Route("/api", func(r chi.Router) {
    r.Use(ApiMiddleware(db, txs, nodeOpts), TransactionMiddleware, AuthMiddleware(allowedOrigins))
    r.Mount("/"+VersionV0, router)   // /api/v0
    r.Mount("/"+Version, router)     // /api/v1
    r.Handle("/*", router)           // everything else under /api
})
```

That third line is an unversioned fallback: `POST /api/collections`, `GET /api/graphql`, and every other route resolve without a version segment.

Rust mounts only the two versioned prefixes (`crates/http/src/router/routes.rs:299-301`):

```rust
root_routes
    .nest("/api/v0", api_routes.clone())
    .nest("/api/v1", api_routes)
```

So every unversioned `/api/...` request 404s on Rust.

## Consequence

A client that omits the version segment works against Go and 404s against Rust. Go's own client always joins `/api/v1` (`http/http_client.go:41`), so this does not break `http.Client` itself; it breaks hand-written clients, curl-based scripts, and anything that took Go's routing at face value.

Lower severity than the other route findings for that reason, but it is a real difference in what the two servers accept, and it is the kind of thing that is trivial now and awkward once someone depends on the 404.

## Fix

One more `.nest("/api", api_routes.clone())` in `create_router_with_state_and_sync_body_limit`, with two things to check:

- **Route conflict.** axum panics at construction on overlapping routes. `/api/v0`, `/api/v1` and `/api`-with-nested-paths need to coexist; nesting the same router at `/api` alongside the two versioned mounts should be fine since the prefixes differ, but it must be proven by a test that builds the router, not by inspection.
- **The permission table.** `route_permissions.rs` keys off axum's `MatchedPath`, and an unversioned mount produces `/api/collections` rather than `/api/v0/collections`. `normalize_api_version` (`crates/http/src/route_permissions.rs:303`) currently folds only `/api/v1/` onto `/api/v0/`, so it would need to fold the bare `/api/` form too. Without that, every unversioned route falls to the `_` arm and is enforced as `Required(DocumentRead)` instead of its real permission. This is the same trap documented in #1411's fix, and it is a privilege downgrade, not a 404.

Add the unversioned form to the exhaustive list in `all_registered_routes_return_expected_permission` when this lands.

vertexclique:
Closes #1495.

Go mounts its whole API router three times (`http/handler.go:116-128`): `/api/v0`, `/api/v1`, and an unversioned `r.Handle("/*", router)`. Rust mounted only the two versioned prefixes, so every unversioned `/api/...` request answered 404.

## What changed

**The mount list has one definition.** `go_paths::API_PREFIXES` already existed as the canonical prefix list, but `routes.rs` hardcoded its own two `.nest()` calls, so the two could drift. The router now folds over `API_PREFIXES`, and the constant gains the bare `/api` entry. Adding a prefix is now a one-line change in one place.

A side effect worth noting: `tests/go_route_paths.rs` iterates `API_PREFIXES`, so its `every_declared_go_path_resolves_under_every_api_version` test picked up coverage of the new mount for free.

**The permission table folds the unversioned form.** This is the part that mattered more than the mount. `route_permissions.rs` keys off axum's `MatchedPath`, and the unversioned mount produces `/api/collections`, not `/api/v0/collections`. `normalize_api_version` folded only `/api/v1/`, so every unversioned route would have fallen to the `_` arm and been enforced as `Required(DocumentRead)`: a privilege downgrade, not a 404. It now folds the bare `/api/` form too, and rejects near-misses (`/api/v10/...`, `/api/v2/...`) rather than rewriting them into the table.

## Tests

New `crates/http/tests/unversioned_api_mount.rs`:
- the unversioned mount serves the same routes as the versioned ones, across every nesting style the router uses
- the plain regression: `GET /api/graphql` is no longer a 404
- the unversioned form keeps its real permission (`CollectionPatch`, `DocumentUpdate`, `DacPolicyAdd`, `CollectionTruncate`) rather than dropping to the default
- exemptions survive the fold, so an unversioned version or ws probe does not start demanding a token
- the router builds, which is what proves the third mount does not trip axum's overlapping-route panic

`tests/route_permissions.rs`: rather than hand-copying an unversioned list into `all_registered_routes_return_expected_permission`, the loop now checks every entry against every prefix in `API_PREFIXES`. That cannot drift when a mount is added. Added `near_miss_version_segments_are_not_folded`.

## Verification

- `just fmt-check`: green
- `just lint`: green (exit 0)
- `just test`: green, 5201 tests passed, 0 failures
- `RUSTDOCFLAGS="-D warnings" cargo doc -p defra-http --no-deps`: green

`just doc` for the whole workspace is red on `main` already, from pre-existing rustdoc errors in `crates/storage` and `crates/datastore` (bare URLs, unclosed `<dyn>` HTML tags, unresolved intra-doc links). None of that is touched here, so the doc gate was run scoped to the crate this PR changes. The workspace doc build is worth its own cleanup PR.

---

## Review round: two assertions removed, four suggestions taken

**The truncate assertion was certifying the `_` arm, as flagged.** I probed axum 0.8 directly and `MatchedPath` is `/api/v0/collections/{name}/truncate` against a colon-keyed table, so no parameterized route matches on any mount. That is #1497 and not this change; the parameterized cases are gone and the file says why rather than re-asserting the default.

**`POST /api/graphql` with a real query is now tested**, across all three mounts, and pinned literally as `/api/graphql` as well. The GET was a 400 for a missing `query` on every mount, so it proved nothing.

Suggestions taken:

- `normalize_api_version` derives from `API_PREFIXES`. One deviation from the suggested form: it takes the **longest** match rather than the first, so the list stays order-independent. First-match rewrites `/api/v0/x` to `/api/v0/v0/x` if `"/api"` is ever ordered ahead of `"/api/v0"`. `/apifoo` is covered by a test.
- Mount parity uses `assert_eq!` across prefixes instead of `!= NOT_FOUND`.
- `the_router_builds_with_every_mount` deleted. Correct that `nest` inlines concatenated paths so there was no overlap panic to guard, and `go_route_paths.rs` already has that test.
- The exemption doc comment described something the harness could not produce. Replaced with a real `route_layer` + NAC test.

Also corrected in this description: the fold **rewrites** near-misses rather than rejecting them, so `/api/v2/collections` becomes `/api/v0/v2/collections` and is safe only because no table key starts with `/api/v0/v`. The earlier text had that backwards.

**Mutation-tested rather than assumed.** Two more tests turned out to have the same weakness the review identified: iterating `API_PREFIXES` means a test stops covering the unversioned mount the moment the constant loses it, so that form is pinned literally now. And the middleware test first written here passed with the fold deliberately broken, because refusing an anonymous caller does not distinguish `DocumentRead` from `DocumentUpdate`; it now runs one caller twice with different grants, so a `DocumentRead` holder must be refused where a `DocumentUpdate` holder is admitted.

Removing `"/api"` from `API_PREFIXES` fails 5 of 7 tests; breaking the fold fails 3.

Re-verified: `just fmt-check` green, `just lint` exit 0, `just test` 5203 passed 0 failed, `cargo doc -D warnings -p defra-http` green.

### DIFF
--- a/crates/http/src/route_permissions.rs
+++ b/crates/http/src/route_permissions.rs
@@ -305,12 +305,32 @@ pub fn route_permission(path: &str, method: &Method) -> RoutePermission {
     }
 }
 
+/// Fold every mount of the API router onto its `/api/v0` key.
+///
+/// `go_paths::API_PREFIXES` mounts one route set three times, so `/api/v1/x`
+/// and the unversioned `/api/x` reach the same handler as `/api/v0/x` and must
+/// carry the same permission. A form that fails to fold here falls to the `_`
+/// arm and is enforced as `DocumentRead`, which is a privilege downgrade
+/// rather than a 404.
 fn normalize_api_version(path: &str) -> Cow<'_, str> {
-    if path == "/api/v1" {
+    let Some(rest) = path.strip_prefix("/api") else {
+        return Cow::Borrowed(path);
+    };
+    if rest == "/v0" || rest.starts_with("/v0/") {
+        return Cow::Borrowed(path);
+    }
+
+    let suffix = match rest {
+        "/v1" => "",
+        _ => rest
+            .strip_prefix("/v1")
+            .filter(|stripped| stripped.starts_with('/'))
+            .unwrap_or(rest),
+    };
+
+    if suffix.is_empty() {
         Cow::Borrowed("/api/v0")
-    } else if let Some(suffix) = path.strip_prefix("/api/v1/") {
-        Cow::Owned(format!("/api/v0/{suffix}"))
     } else {
-        Cow::Borrowed(path)
+        Cow::Owned(format!("/api/v0{suffix}"))
     }
 }
--- a/crates/http/src/route_permissions.rs
+++ b/crates/http/src/route_permissions.rs
@@ -307,30 +307,29 @@ pub fn route_permission(path: &str, method: &Method) -> RoutePermission {
 
 /// Fold every mount of the API router onto its `/api/v0` key.
 ///
-/// `go_paths::API_PREFIXES` mounts one route set three times, so `/api/v1/x`
-/// and the unversioned `/api/x` reach the same handler as `/api/v0/x` and must
-/// carry the same permission. A form that fails to fold here falls to the `_`
-/// arm and is enforced as `DocumentRead`, which is a privilege downgrade
-/// rather than a 404.
+/// `go_paths::API_PREFIXES` is what `routes.rs` mounts the one route set at,
+/// so it is also what has to be folded here. Deriving from the same constant
+/// is the point: a prefix added there without a matching arm here would mount
+/// the whole route set unfolded, and every route under it would fall to the
+/// `_` arm and be enforced as `DocumentRead`.
+///
+/// The longest match wins, so the list stays order-independent: `/api/v0/x`
+/// must fold on `/api/v0` and not on the bare `/api`.
+///
+/// Note this rewrites rather than rejects: `/api/v2/collections` becomes
+/// `/api/v0/v2/collections`, which is safe only because no table key begins
+/// with `/api/v0/v`, and which lands on the `_` arm as it should.
 fn normalize_api_version(path: &str) -> Cow<'_, str> {
-    let Some(rest) = path.strip_prefix("/api") else {
-        return Cow::Borrowed(path);
-    };
-    if rest == "/v0" || rest.starts_with("/v0/") {
-        return Cow::Borrowed(path);
-    }
+    let matched = crate::go_paths::API_PREFIXES
+        .iter()
+        .filter_map(|prefix| {
+            path.strip_prefix(*prefix)
+                .filter(|suffix| suffix.is_empty() || suffix.starts_with('/'))
+        })
+        .max_by_key(|suffix| path.len() - suffix.len());
 
-    let suffix = match rest {
-        "/v1" => "",
-        _ => rest
-            .strip_prefix("/v1")
-            .filter(|stripped| stripped.starts_with('/'))
-            .unwrap_or(rest),
-    };
-
-    if suffix.is_empty() {
-        Cow::Borrowed("/api/v0")
-    } else {
-        Cow::Owned(format!("/api/v0{suffix}"))
+    match matched {
+        Some(suffix) => Cow::Owned(format!("/api/v0{suffix}")),
+        None => Cow::Borrowed(path),
     }
 }
--- a/crates/http/tests/unversioned_api_mount.rs
+++ b/crates/http/tests/unversioned_api_mount.rs
@@ -1,49 +1,111 @@
 //! Go mounts its whole router at `/api` with no version segment
-//! (`r.Handle("/*", router)`, `http/handler.go:127`), where Rust mounted only
+//! (`r.Handle("/*", router)`, `http/handler.go:126`), where Rust mounted only
 //! `/api/v0` and `/api/v1`, so every unversioned request 404'd.
 //!
 //! The mount is driven by `go_paths::API_PREFIXES`, and the permission table
-//! folds all three forms onto the v0 key: a form that resolves but does not
-//! fold is enforced as `DocumentRead`, a privilege downgrade rather than a 404.
+//! folds every prefix in that same list onto the v0 key: a form that resolves
+//! but does not fold is enforced as `DocumentRead`, a privilege downgrade
+//! rather than a 404.
+//!
+//! Parameterized routes are deliberately not asserted here. axum 0.8 yields a
+//! brace `MatchedPath` (`/api/collections/{name}`) while the table is keyed
+//! with colons, so those entries never match on any mount, versioned or not.
+//! That defect is #1497 and is not what this change is about; asserting it
+//! here would only certify the `_` arm.
 
 mod common;
 use common::*;
 
-/// The routes a hand-written client or a curl script is most likely to reach
-/// for, one per nesting style the router uses.
+use axum::{body::Body, http::Request, Router};
+use defra_http::router::{AppStateBuilder, NodeAcpOperations};
+use defra_http::{MockNodeAcpOperations, MockQueryExecutor};
+use std::sync::Arc;
+use tower::ServiceExt;
+
+/// One route per nesting style the router uses, all unparameterized.
 const PATHS: &[(&str, Method)] = &[
-    ("/graphql", Method::GET),
     ("/schema", Method::GET),
     ("/version", Method::GET),
     ("/collections", Method::GET),
-    ("/collections/Users", Method::GET),
     ("/p2p/replicators", Method::GET),
     ("/view/refresh", Method::POST),
+    ("/node/identity", Method::GET),
 ];
 
+/// Every prefix must answer a given request the same way, not merely resolve.
+/// `!= NOT_FOUND` would pass with the unversioned mount answering 405 while
+/// the versioned ones answer 400.
 #[tokio::test]
-async fn the_unversioned_mount_serves_the_same_routes_as_the_versioned_ones() {
+async fn every_mount_answers_a_request_identically() {
+    // Pinned literally: a test that only iterates the constant stops covering
+    // the unversioned mount the moment the constant loses it.
+    assert!(
+        defra_http::go_paths::API_PREFIXES.contains(&"/api"),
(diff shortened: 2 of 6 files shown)


crates/http/tests/unversioned_api_mount.rs
```diff
@@ -0,0 +1,106 @@
...
+            NodePermission::CollectionPatch,
+        ),
+        ("/purge", Method::POST, NodePermission::DocumentUpdate),
+        (
+            "/acp/document/policy",
+            Method::POST,
+            NodePermission::DacPolicyAdd,
+        ),
+        (
+            "/collections/:name/truncate",
```
edjroz: Axum 0.8 yields `MatchedPath` `/api/collections/{name}/truncate` — the router registers `"/{name}/truncate"` (`router/routes.rs:132`). `normalize_api_version` rewrites that to `/api/v0/collections/{name}/truncate`, which isn't a key in the colon-keyed table (`route_permissions.rs:107`), so the `_` arm applies `DocumentRead`.

So this case passes whether or not the fold works — it would also pass with brace-key lookup deleted — and the PR body's claim that unversioned parameterized routes keep their table permission isn't true for any of them. Not a live bypass: the handler re-checks `CollectionTruncate`.

Either drop this case, or switch the table keys to `{name}` here and prove it through `route_layer` (#1497).

crates/http/tests/unversioned_api_mount.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// The regression itself: before this change these were 404s.
+#[tokio::test]
+async fn an_unversioned_request_is_not_a_404() {
+    let (status, body) = Call::post("/api/graphql")
+        .method(Method::GET)
+        .authenticated()
+        .send()
+        .await;
+    assert_ne!(status, StatusCode::NOT_FOUND, "{body}");
```
edjroz, line 113: `GraphqlQueryParams.query` is required (`handlers/graphql/query.rs:169-175`), so this GET is a 400 rather than a success. Measured on this HEAD: `GET /api/graphql` → 400, `GET /api/graphql?query={__typename}` → 200, `POST /api/graphql {}` → 422, `POST` with a query → 200.

Go's contract is `POST /api/graphql` with a query body (`http/handler_store_test.go:111-130`), and that's the URL Go advertises at startup (`node/node_api.go:45`) — but nothing here POSTs GraphQL, and `PATHS` doesn't include it either, so this file passes even if unversioned GraphQL POST were missing entirely.

crates/http/src/route_permissions.rs
```diff
@@ -305,12 +305,32 @@ pub fn route_permission(path: &str, method: &Method) -> RoutePermission {
...
 }
 
+/// Fold every mount of the API router onto its `/api/v0` key.
+///
+/// `go_paths::API_PREFIXES` mounts one route set three times, so `/api/v1/x`
+/// and the unversioned `/api/x` reach the same handler as `/api/v0/x` and must
+/// carry the same permission. A form that fails to fold here falls to the `_`
+/// arm and is enforced as `DocumentRead`, which is a privilege downgrade
+/// rather than a 404.
 fn normalize_api_version(path: &str) -> Cow<'_, str> {
```
edjroz, line 322: This hardcodes the three forms while `routes.rs:347` mounts whatever `API_PREFIXES` holds — so adding `"/api/v2"`, the one-line change the PR body advertises, would mount the route set unfolded and enforce `POST /api/v2/purge` as `Required(DocumentRead)`. Deriving from the same constant is also shorter:

```rust
fn normalize_api_version(path: &str) -> Cow<'_, str> {
    for prefix in go_paths::API_PREFIXES {
        if let Some(suffix) = path.strip_prefix(prefix) {
            if suffix.is_empty() || suffix.starts_with('/') {
                return Cow::Owned(format!("/api/v0{suffix}"));
            }
        }
    }
    Cow::Borrowed(path)
}
```

I compared both against 19 inputs; identical except `/apifoo`, which the current form rewrites to `/api/v0foo`.

Separately, for the PR body: this *rewrites* near-misses rather than rejecting them (`/api/v2/collections` → `/api/v0/v2/collections`), which is safe only because no table key starts with `/api/v0/v`. The outcome is right, the description has it the other way round.

crates/http/tests/unversioned_api_mount.rs
```diff
@@ -0,0 +1,106 @@
...
+async fn the_unversioned_mount_serves_the_same_routes_as_the_versioned_ones() {
+    for (path, method) in PATHS {
+        for prefix in defra_http::go_paths::API_PREFIXES {
+            let (status, body) = Call::post(&format!("{prefix}{path}"))
+                .method(method.clone())
+                .json("{}")
+                .authenticated()
+                .send()
+                .await;
+            assert_ne!(
```
edjroz, line 56: `!= NOT_FOUND` doesn't check what the name promises — the prefixes are never compared to each other, so `/api/view/refresh` answering 405 while `/api/v0/view/refresh` answers 400 passes. 400, 405, 422 and 503 all satisfy it.

`v0_and_v1_resolve_same_route_set` (`src/router/routes.rs:514-528`) already does this properly with `assert_eq!(v1_status, v0_status)`.

crates/http/tests/unversioned_api_mount.rs
```diff
@@ -0,0 +1,106 @@
...
+            RoutePermission::Required(expected),
+            "/api{path} lost its permission"
+        );
+    }
+}
+
+/// The exemptions have to survive the fold too, or an unversioned health or
+/// version probe starts demanding a token.
+#[test]
+fn the_unversioned_form_keeps_its_exemption() {
```
edjroz: Not something this PR broke, but worth closing while you're in here: `Call::send` builds `create_router_with_state` (`tests/common/mod.rs:61-71`), which has no auth layer — production applies it as a `route_layer` (`server.rs:527-530`) — and without NAC `require_permission` is a no-op (`nac_guard.rs:42-44`). So the "starts demanding a token" case this doc comment describes can't occur in this test either way.

One request through `route_layer` with NAC enabled would cover it; `handlers/browser_sync_tests.rs:255-275` is the existing pattern.

crates/http/tests/unversioned_api_mount.rs
```diff
@@ -0,0 +1,106 @@
...
+    assert_eq!(
+        route_permission("/api/graphql/ws", &Method::GET),
+        RoutePermission::Exempt
+    );
+}
+
+/// The router mounts one route set three times; axum panics at construction on
+/// overlapping routes, so this is the proof the third mount is legal.
+#[tokio::test]
+async fn the_router_builds_with_every_mount() {
```
edjroz: This passes with `"/api"` removed from `API_PREFIXES`, so it doesn't prove the third mount exists. `nest` inlines concatenated paths (`axum-0.8.9/path_router.rs:206-243`) rather than registering a catch-all, so there was no overlap panic to guard against — and `the_router_builds` in `go_route_paths.rs:96-99` is already the same test.

edjroz (changes requested):
Request changes — the mount itself is right; two of the new tests assert things that aren't true.

I verified the Go premise at the pinned baseline `13c46ec9`: `mux.Route("/api")` wraps both `Mount`s and `r.Handle("/*", router)` (`http/handler.go:111-126`), and chi rewrites `RoutePath` before the wildcard runs — so Go does serve `/api/graphql`, and mounting at `/api` rather than `/` is the right reading. No axum overlap hazard either: `nest` inlines concatenated paths, so the bare mount can't shadow the versioned ones.

Blocking:

- Unversioned parameterized routes are enforced as `DocumentRead`, not their table permission, so the truncate assertion certifies something production doesn't do.
- `POST /api/graphql` with a real query — the call this PR exists to serve — is never tested.

The rest are suggestions on the same tests. One of them (the auth-middleware note) is a pre-existing gap rather than anything this PR broke, and I've said so inline.

jackzampolin (approved):
Re-reviewed the current head. The prior routing and permission concerns are addressed: real POST /api/graphql coverage, prefix-driven normalization, mount response parity, and enforcement through the real NAC middleware. No further findings.

Issue #1419, opened 2026-08-12
Add cross-runtime HTTP route table and response shape equivalence tests


jackzampolin:
Sub-issue of #1393.

## Gap

No cross-runtime assertion on the HTTP wire contract. A client written against Go DefraDB should work against Rust unchanged; nothing tests that.

#1411 is the case in point: Rust serves `POST /acp/policy` where Go serves `POST /acp/document/policy` (`http/handler_acp.go:338`). A Go-compatible client's policy registration simply 404s. That is mechanically detectable and went unnoticed.

## The test

Cheap and high-yield, in two layers:

**1. Route table equivalence.** Enumerate Go's registered routes (they are declared in `http/handler_*.go` via `router.AddRoute`) and assert Rust serves the same method+path set. A pure diff, no server needed beyond starting both. This alone catches #1411 and every future route drift.

**2. Response shape equivalence** for the routes that matter most:
- success bodies: same JSON keys, same nesting, same types
- **error envelopes**: same shape and same HTTP status per error class. Status codes are contract — a client branching on 404 vs 400 breaks silently when they differ.
- content types and any headers clients depend on

Worth including `/openapi.json` if both serve it: a spec that disagrees with the implementation is its own defect.

## Where

`tools/integration-test/tests/`, `go_*`-named. The harness already starts both binaries, so both layers are within reach of existing infrastructure.

PR #1575, opened 2026-08-24
fix(http): honor Go's collection selectors on GET /collections

### LINKED ISSUE #1494: GET /collections silently ignores Go's name, version_id, collection_id and get_inactive selectors
Found while implementing #1411's `view/refresh` selectors, which are the same selectors. Triage procedure: #1391.

## Divergence

Go's `GET /collections` takes four query parameters and returns the collection definitions they select. Rust's `GET /collections` takes **no query parameters at all** and returns every collection name.

| | Go v1.0.0 | Rust |
|---|---|---|
| Handler | `h.GetCollection` (`http/handler_store.go:1108`) | `handlers::list_collections` (`crates/http/src/handlers/collections.rs:91`) |
| Selectors | `name`, `version_id`, `collection_id`, `get_inactive` | none parsed |
| Selection logic | `getCollections` (`internal/db/collection.go:193-301`) | none |

Go's client sends them from `GetCollections` (`http/client.go:422-448`), driven by `options.GetCollectionsOptions`.

## Consequence

A Go-compatible client asking for one collection by name, or for a specific version, or for inactive versions, gets **the full list of every active collection** instead. The parameters are silently ignored rather than rejected, so the client cannot tell it did not get what it asked for. A client that then indexes the first element of the response is reading an arbitrary collection.

## The selection semantics, for whoever implements this

Go runs two stages, and the precedence in the first is load-bearing. Getting this wrong is easy, so it is written out here.

**Stage 1, candidate set.** A `switch`, first match wins (`internal/db/collection.go:203-256`):

1. `name` set **and** `get_inactive` false: the active collection with that name.
2. else `version_id` set: the collection with that version id, active or not.
3. else `collection_id` set: every version of it when `get_inactive`, otherwise its active version.
4. else: every collection when `get_inactive`, otherwise every active one.

**Stage 2, filter over the candidates** (`internal/db/collection.go:259-285`):

- `version_id` set: keep only `col.VersionID == version_id`
- `name` set: keep only `col.Name == name`
- drop inactive unless `get_inactive` **or** `version_id` was requested
- `collection_set_id` is internal-only and unreachable over HTTP

Two consequences a flat AND over the four selectors gets wrong:

- **`collection_id` is a stage-1 selector only.** When `name` is set with `get_inactive` false, case 1 wins and `collection_id` never applies, so a request carrying a name and a disagreeing collection id returns the named collection on Go rather than nothing.
- **`name` plus `get_inactive` deliberately misses case 1**, falls through to case 4, and is narrowed by the stage-2 name filter. That is how Go reaches an inactive collection by name.

## Already implemented once

The identical selector set landed for `POST /view/refresh` in the #1411 fix, as `db::RefreshViewsOptions` with `needs_all_versions()` and `selects()` in `crates/db/src/view_ops.rs`, with the precedence rules covered in `crates/db/tests/refresh_views_selectors.rs`. Whoever picks this up should reuse that rather than write a second copy: Go itself has one `getCollections` behind both surfaces, and two Rust copies would be a divergence waiting to happen.

`CollectionVersion` already carries every field needed (`name`, `version_id`, `collection_id`, `is_active`, `crates/schema/src/collection.rs:44-116`), and both listing primitives exist (`get_all_active_collections_internal`, `get_all_collection_versions`).

## Also check the response shape

Go returns collection definitions; Rust returns `{"collections": [names]}`. Worth confirming as part of this, though response-shape parity generally belongs to #1419.

vertexclique:
Closes #1494.

`GET /collections` parsed no query parameters at all, so a Go-compatible client asking for one collection by name, or for a specific version, got the full list of every active collection and had no way to tell it did not get what it asked for.

## Reusing the lookup instead of copying it

The issue was explicit that the identical selector set already landed for `POST /view/refresh` and that a second copy would be a divergence waiting to happen. Go has one `getCollections` behind both surfaces, so Rust now has one too:

- `db::RefreshViewsOptions` becomes `db::CollectionSelector` and moves from `view/ops.rs` to `crates/db/src/collection/selector.rs`. It was already Go's collection lookup wearing a view-specific name; it is now named for what it is and lives next to collections. This is a rename across 10 files with no behaviour change.
- `RefreshViewsQuery` becomes `handlers::collection_selector::CollectionSelectorQuery`, shared by `GET /collections` and `POST /view/refresh`, so Go's query string is parsed in one place.
- New `is_unfiltered()` on the selector, which is what lets the unnarrowed listing keep its existing path.

## Two decisions worth reviewing

**The unnarrowed listing stays on `rest.list_collections()`.** This is not just conservatism: that path honours a transaction-scoped collection provider (`current_txn_collection_provider`), and the system store the narrowed path reads does not. Narrowing needs per-version fields the name list does not carry, so it has to read the versions. Both are commented at the call site.

**A narrowed request with no version store configured is an error, not a fallback.** Quietly answering a narrowed request from the unnarrowed listing would be the same silent-widening bug this issue is about, one layer down.

## Response shape

Still `{"collections": [names]}`, not Go's collection definitions. The issue assigns response-shape parity to #1419, and `GET /collections/versions` already serves version-level detail. Names are sorted and deduped, so a collection with two versions is listed once rather than appearing as a repeated name.

## Tests

`crates/http/tests/collection_selectors.rs`, 11 tests over a version set that covers every branch of Go's candidate switch. The two that matter most are the cases a flat AND over the four selectors gets wrong:

- `a_name_outranks_a_disagreeing_collection_id`: `collection_id` is a stage-one selector only, so `?name=Users&collection_id=books-c` returns `Users`, where a flat AND would return nothing.
- `a_name_with_get_inactive_reaches_an_inactive_version`: name plus `get_inactive` deliberately misses the active-by-name arm and is narrowed by the stage-two name filter instead.

Also covered: an unknown name selects nothing, a version id reaches an inactive version, `get_inactive` widens, two versions of one collection are named once, no selectors keep the previous behaviour, a narrowed request without a version store fails loudly, and an unparseable `get_inactive` is a 400 rather than a dropped selector.

The db-level selector tests moved from `tests/view/ops.rs` to `tests/collection/selector.rs` alongside the type, with three added: `only_an_empty_selector_is_unfiltered`, `a_name_with_get_inactive_reaches_an_inactive_version`, and `a_version_id_needs_the_full_listing`.

## Verification

- `just fmt-check`: green
- `just lint`: green (exit 0)
- `just test`: green, 5209 tests passed, 0 failures
- `RUSTDOCFLAGS="-D warnings" cargo doc -p defra-http -p db --no-deps`: green

`just doc` across the whole workspace is red on `main` already (pre-existing rustdoc errors in `crates/storage` and `crates/datastore`), so the doc gate was run scoped to the crates this PR changes.

---

## Follow-up from self-review

**Empty selector values are now a 400.** `?name=` selected nothing under the rules above, while a server treating an empty value as "unset" answers with everything. I could not confirm which Go does, and both silently hand the caller something other than what it asked for, which is the bug this handling exists to fix. Refusing says so instead of guessing. Applies to `name`, `version_id` and `collection_id`, and to `view/refresh` too since both share the query type.

**The selector now takes the cheap listing when it can.** `refresh_views` branches on `needs_all_versions()` and reads the in-memory collection cache when inactive versions are not needed; this handler always scanned every stored version. Two surfaces sharing one selector were deciding differently, which is the divergence the shared type was supposed to prevent.

Fixed by exposing `db::get_active_collection_versions` (the cache read `refresh_views` already used), adding `get_active_collections` to `CollectionVersionOperations` and `CollectionManagementOperations` with a default that filters the full listing so no existing implementation breaks, overriding it in the CLI adapter, and branching in the handler. `a_selector_that_needs_no_inactive_versions_takes_the_cheap_listing` counts which listing was requested, so the cheap path is asserted rather than assumed.

Not fixed, and worth stating plainly: `get_all_collection_versions` still collects the whole `/collection/` prefix scan into a `Vec`. Streaming that needs a new iterator API in `crates/db` and changes a primitive `GET /collections/versions` shares, so it belongs in its own change. What this fixes is that the common case no longer reaches that scan at all.

Re-verified: `just fmt-check` green, `just lint` exit 0, `just test` 5212 passed 0 failed, `cargo doc -D warnings -p defra-http -p db -p cli` green.

---

## Verified against the Go source, and two corrections

Checked against the pinned baseline `13c46ec9`, which is what `GO_COMPAT_COMMIT` names. Two of my earlier answers were wrong.

**The empty-selector 400 was a mistake, and is reverted.** Go keys the selectors off `req.URL.Query().Has("name")`, not off the value being non-empty (`http/handler_store.go:391-402`), so `?name=` is a collection name set to `""`, not an absent selector. `getCollections` then takes the name case, `GetCollectionByName` returns not-found, the error is swallowed and a zero-value collection is appended, and the stage-two filter drops it because it is not active (`internal/db/collection.go:203-206`, `:270-273`). The result is an empty list, which is what this PR did before I added the refusal on top. `into_selector` is infallible again, since it can no longer fail.

**An unknown `version_id` must be a 404, and now is.** Go returns `GetCollectionByID`'s error unconditionally (`internal/db/collection.go:210-215`), and that error is `client.ErrCollectionNotFound` (`internal/db/description/collection.go:47-48`), which `httpStatusFromError` maps to 404 (`http/errors.go:150`). An unknown *name* or *collection_id* is swallowed into an empty list instead, so the asymmetry is deliberate: a typo in a version id must not read as success. This handler returned an empty list for all three. It is also the behaviour `refresh_views` already had on the same selector, so this removes the last inconsistency between the two surfaces.

`?get_inactive=` with an unparseable or empty value stays a 400, matching Go's `strconv.ParseBool` (`http/handler_store.go:403-410`).

Re-verified: `just fmt-check` green, `just lint` exit 0, `just test` 5213 passed 0 failed, `cargo doc -D warnings -p defra-http -p db -p cli` green.

---

## Review round: both blockers fixed

**The 404 fired on the wrong condition.** Go propagates not-found from the direct `GetCollectionByID` lookup, which the active-by-name arm pre-empts, so `?name=Users&version_id=nope` is an empty 200 there and was a 404 here. That is the request `collection describe --collection-name … --version-id …` sends, so it 404'd a working CLI call.

The precedence now lives on the shared selector as `CollectionSelector::resolves_by_version_lookup()`, next to `applies_collection_id()`, and every row of the table in the review is a test:

| Request | Go | Rust now |
|---|---|---|
| `?name=Users&version_id=nope` | `200 []` | `200 []` |
| `?name=Books&version_id=users-v2` | `200 []` | `200 []` |
| `?name=Books&get_inactive=true&version_id=users-v2` | `200 []` | `200 []` |
| `?version_id=nope` | 404 | 404 |

The message is Go's `collection not found`, asserted. This also removes the disagreement with `refresh_views` for the name-plus-existing-version case.

**The two-source split is gone, and my justification for it was wrong.** I traced `scope_txn_collection_provider`: it is applied only around GraphQL execution (`runner/executor.rs:462-471`), never around `rest.list_collections()`, so the transaction scoping the comment claimed does not exist on that path. #1583 is correct and the comment is deleted rather than reworded.

Both branches now read the version store, which is also Go's source (`description.GetActiveCollections`). The divergence described in the review is a regression test: a node holding an inactive P2P-synced collection must not list it from `GET /collections` or select it with `?name=`.

Consequence worth noting for review: `GET /collections` now requires the version store for every request, where the unnarrowed case previously fell back to `rest`. Both production assemblies wire it (`server_http.rs:177`, `defra-node/src/lib.rs:1230`) and `create_router_with_rest` has no in-tree production caller, so this only affects a library user wiring `rest` alone, who now gets a loud "not configured" rather than a listing that disagrees with its own filtered form.

The follow-ups (#1579 through #1583) are left alone as agreed, except #1583, which this removes as a side effect.

Re-verified: `just fmt-check` green, `just lint` exit 0, `just test` 5216 passed 0 failed, `cargo doc -D warnings -p defra-http -p db -p cli` green.

### DIFF
--- a/crates/db/src/view/ops.rs
+++ b/crates/db/src/view/ops.rs
@@ -3,6 +3,7 @@
 //! This module contains operations for refreshing and managing
 //! materialized view caches.
 
+use crate::collection::selector::CollectionSelector;
 use crate::error::{Error, Result};
 use datastore::NamespaceView;
 use schema::CollectionVersion;
@@ -17,71 +18,6 @@ use storage::keys::datastore::ViewCacheKey;
 use tokio::task::JoinHandle;
 #[cfg(not(target_arch = "wasm32"))]
 use tokio::time::{self, Instant, MissedTickBehavior};
-
-/// Options for refreshing materialized views.
-///
-/// The selectors mirror Go's collection lookup (`internal/db/collection.go`
-/// `getCollections`), which is what Go's `RefreshViews` selects with.
-#[derive(Debug, Clone, Default)]
-pub struct RefreshViewsOptions {
-    /// Only refresh views with these names (None = all views)
-    pub names: Option<Vec<String>>,
-    /// Only refresh the view with this collection version id.
-    pub version_id: Option<String>,
-    /// Only refresh views belonging to this collection id.
-    pub collection_id: Option<String>,
-    /// Include inactive collection versions.
-    pub get_inactive: bool,
-}
-
-impl RefreshViewsOptions {
-    /// Create options that refresh all views.
-    pub fn all() -> Self {
-        Self::default()
-    }
-
-    /// Create options that refresh only the named views.
-    pub fn with_names(names: Vec<String>) -> Self {
-        Self {
-            names: Some(names),
-            ..Self::default()
-        }
-    }
-
-    /// Whether inactive versions have to be loaded to answer this selection.
-    ///
-    /// A named version is returned whether or not it is active, so asking for
-    /// one requires the full listing just as `get_inactive` does.
-    pub fn needs_all_versions(&self) -> bool {
-        self.get_inactive || self.version_id.is_some()
-    }
-
-    /// Whether this collection version is selected.
-    pub fn selects(&self, collection: &CollectionVersion) -> bool {
-        let version_matches = self
-            .version_id
-            .as_ref()
-            .is_none_or(|id| &collection.version_id == id);
-        let name_matches = self
-            .names
-            .as_ref()
-            .is_none_or(|names| names.contains(&collection.name));
-        let collection_matches = !self.applies_collection_id()
-            || self
-                .collection_id
-                .as_ref()
-                .is_none_or(|id| &collection.collection_id == id);
-        let visible = self.get_inactive || collection.is_active || self.version_id.is_some();
-
-        version_matches && name_matches && collection_matches && visible
-    }
-
-    /// Go picks candidates by collection id only when neither a name nor a
-    /// version already picked them, so those two take precedence over it.
-    fn applies_collection_id(&self) -> bool {
-        (self.get_inactive || self.names.is_none()) && self.version_id.is_none()
-    }
-}
 
 /// Whether this collection version is a view whose cache can be rebuilt.
 ///
@@ -134,7 +70,7 @@ impl<S: Store> crate::database::DB<S> {
     ///
     /// Clears and rebuilds the view cache for each materialized view the
     /// options select. Without options, every materialized view is refreshed.
-    pub async fn refresh_views(&self, options: RefreshViewsOptions) -> Result<()>
+    pub async fn refresh_views(&self, options: CollectionSelector) -> Result<()>
     where
         S: 'static,
     {
@@ -297,7 +233,7 @@ impl<S: Store> crate::database::DB<S> {
 
                 for name in due_names {
                     if let Err(error) = self
-                        .refresh_views(RefreshViewsOptions::with_names(vec![name.clone()]))
+                        .refresh_views(CollectionSelector::with_names(vec![name.clone()]))
                         .await
                     {
                         tracing::warn!(
--- a/crates/db/src/view/ops.rs
+++ b/crates/db/src/view/ops.rs
@@ -401,6 +401,11 @@ impl<S: Store> crate::database::DB<S> {
             .collections
             .read()
             .map_err(|_| Error::Other("failed to acquire collections lock".to_string()))?;
-        Ok(cache.values().map(|c| c.schema().clone()).collect())
+        Ok(cache
+            .values()
+            .map(|collection| collection.schema())
+            .filter(|schema| schema.is_active)
+            .cloned()
+            .collect())
     }
 }
--- a/crates/http/src/handlers/collections.rs
+++ b/crates/http/src/handlers/collections.rs
@@ -106,7 +106,7 @@ pub async fn list_collections(
 ) -> Result<Json<CollectionsResponse>, HttpError> {
     require_permission(&state, &identity, NodePermission::CollectionGet).await?;
 
-    let selector = query.into_selector()?;
+    let selector = query.into_selector();
     if !selector.is_unfiltered() {
         return list_selected_collections(&state, &selector).await;
     }
@@ -150,10 +150,26 @@ async fn list_selected_collections(
     }
     .map_err(http_error_from_backend_message)?;
 
-    let mut collections: Vec<String> = versions
-        .into_iter()
+    let selected: Vec<&schema::CollectionVersion> = versions
+        .iter()
         .filter(|version| selector.selects(version))
-        .map(|version| version.name)
+        .collect();
+
+    // Go looks a version id up directly and propagates not-found
+    // (`internal/db/collection.go:210-215`, `GetCollectionByID`), where an
+    // unknown name or collection id is swallowed into an empty list. A typo in
+    // a version id must not read as "no such collection version".
+    if let Some(version_id) = &selector.version_id {
+        if selected.is_empty() {
+            return Err(HttpError::NotFound(format!(
+                "no collection version {version_id}"
+            )));
+        }
+    }
+
+    let mut collections: Vec<String> = selected
+        .into_iter()
+        .map(|version| version.name.clone())
         .collect();
     collections.sort();
     collections.dedup();
--- a/crates/http/src/handlers/collections.rs
+++ b/crates/http/src/handlers/collections.rs
@@ -16,6 +16,7 @@ use query::rest::{CollectionDocIdsPage, CollectionDocIdsPagination};
 use serde::Serialize;
 
 use crate::error::{http_error_from_backend_message, HttpError};
+use crate::handlers::collection_selector::CollectionSelectorQuery;
 use crate::identity_extractor::ExtractIdentity;
 use crate::nac_guard::require_permission;
 use crate::router::{AppState, NodePermission};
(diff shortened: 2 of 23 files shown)


crates/http/src/handlers/collections.rs
```diff
@@ -108,6 +129,54 @@ pub async fn list_collections(
...
+    let selected: Vec<&schema::CollectionVersion> = versions
+        .iter()
+        .filter(|version| selector.selects(version))
+        .collect();
+
+    // Go looks a version id up directly and propagates not-found
+    // (`internal/db/collection.go:210-215`, `GetCollectionByID`), where an
+    // unknown name or collection id is swallowed into an empty list. A typo in
+    // a version id must not read as "no such collection version".
+    if let Some(version_id) = &selector.version_id {
```
edjroz: The 404 fires on an empty post-filter set; Go fires it only when `GetCollectionByID` itself misses — i.e. `version_id` set **and** the name-and-active arm did not win (`internal/db/collection.go:202-215`). Case 1 plus a stage-2 miss is an empty 200.

| Request | Go | Rust |
|---|---|---|
| `?name=Users&version_id=nope` | `200 []` | **404** |
| `?name=Books&version_id=users-v1` | `200 []` | **404** |
| `?name=Books&get_inactive=true&version_id=users-v1` | `200 []` | **404** |
| `?version_id=nope` | 404 | 404 ✓ |

Only the last row should 404, and the first is what `defradb client collection describe --collection-name User --version-id bae123` sends (`cli/collection.go:86-100`).

Two smaller things in the same block. The message should be Go's `"collection not found"` (`client/errors.go:29`, mapped at `http/errors.go:150`), not `"no collection version {id}"`. And this makes the handler disagree with `refresh_views` on the same selector — that path errors when the *unfiltered* listing lacks the id (`db/src/view/ops.rs:86-91`), so `?name=Books&version_id=users-v1` is a 404 here and a successful no-op there, where Go runs one `getCollections` for both.

crates/http/src/handlers/collections.rs
```diff
@@ -83,17 +84,37 @@ fn parse_collection_doc_ids_pagination(
...
+    let selector = query.into_selector();
+    if !selector.is_unfiltered() {
+        return list_selected_collections(&state, &selector).await;
+    }
+
+    // The unnarrowed listing stays on the REST path because that one honours a
+    // transaction-scoped collection provider, where the system store below
+    // does not. Narrowing needs per-version fields the name list does not
+    // carry, so it has to read the versions.
     let rest = state
```
edjroz: The unnarrowed branch answers from `rest.list_collections()` → `cache.keys()` (`db/src/collection/ops/lookup.rs:16-22`), and that cache deliberately holds inactive P2P-synced collections (`db/src/merge/merge_handler/definition.rs:169-174`). The narrowed branch below drops them via `visible`.

Failure: on a node that synced `Users` over P2P but never activated it, `GET /collections` returns `["Users"]` while `GET /collections?name=Users` returns `[]`. Go excludes it from both.

Two branches over two sources is the shape that allows this; one source filtered by the selector wouldn't. (The comment justifying the split is #1583 — separate point, same lines.)

edjroz (changes requested):
Request changes. Two things in the new narrowed path need fixing before this lands; everything else I found is filed separately and I'm not asking you to fix it here.

The selector encoding is otherwise right — I traced `collection_id` yielding to name and to version, `name`+`get_inactive` reaching an inactive version, `collection_id`+`get_inactive` widening to all versions, and the empty-value cases (`?name=`, `?collection_id=`, `?version_id=`) against Go's switch at `13c46ec9`, and they agree. The `RefreshViewsOptions` → `CollectionSelector` rename is complete, with no stragglers left in the tree.

**Blocking**, both inline:

- The 404 fires on an empty post-filter set rather than at `GetCollectionByID`, so `?name=Users&version_id=nope` is a 404 where Go answers `200 []`. That's the request `collection describe --collection-name … --version-id …` emits.
- The narrowed and unnarrowed branches read different collection sources, so `GET /collections` and `GET /collections?name=X` disagree on a node with inactive P2P-synced collections.

**Filed as follow-ups, not asked for here.** Most of these are code this PR moves rather than writes — the selector is `RefreshViewsOptions` relocated verbatim, so they already affect `POST /view/refresh` today:

- #1579 — `version_id` escapes the inactive drop when a name already selected the candidate
- #1580 — `get_inactive` rejects Go's `ParseBool` values (`1`, `t`, `TRUE`, …)
- #1581 — malformed selectors return `text/plain`, not Go's JSON error envelope
- #1582 — no test sends `name` and `version_id` together
- #1583 — the `list_collections` comment describes transaction scoping that isn't wired

#1579 and #1582 are worth reading next to each other: the untested combination is exactly where that escape shows up.

jackzampolin (approved):
Re-reviewed through c552e108. Version lookup precedence now matches Go, narrowed and unnarrowed requests share one source, and the active cache path filters inactive P2P-synced schemas. Targeted DB, selector, and view tests pass. No further findings.

crates/db/src/view/ops.rs
```diff
@@ -134,7 +70,7 @@ impl<S: Store> crate::database::DB<S> {
     ///
     /// Clears and rebuilds the view cache for each materialized view the
     /// options select. Without options, every materialized view is refreshed.
-    pub async fn refresh_views(&self, options: RefreshViewsOptions) -> Result<()>
+    pub async fn refresh_views(&self, options: CollectionSelector) -> Result<()>
```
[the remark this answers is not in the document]
vertexclique: This will be resolved through consistency guarantees later and refactoring this part. It is on the TODO list.

Issue #1493, opened 2026-08-17
PATCH /collections/{name} is absent, so Go's UpdateDocumentsWithFilter answers 405


vertexclique:
Found while fixing #1411 by diffing Go's full route table against Rust's router. Triage procedure: #1391. Sibling of #1492, which is the same route path.

## Divergence

Go registers `PATCH /collections/{name}` for `UpdateDocumentsWithFilter` (`http/handler_collection.go:510`). Rust registers `GET`, `POST` and `DELETE` on that path but no `PATCH` (`crates/http/src/router/routes.rs:92-96`), so the request answers **405 Method Not Allowed**.

## The contract Go's client sends

`Collection.UpdateDocumentsWithFilter` (`http/client_document.go:263-296`) sends `PATCH` to `apiURL.JoinPath("collections", name)` with:

```json
{"filter": {...}, "updater": "<update as a JSON string>"}
```

defined as `UpdateCollectionRequest` (`http/handler_collection.go:36-39`), and expects `client.UpdateResult` back:

```json
{"Count": 3, "DocIDs": ["bae-...", "bae-..."]}
```

(`client/collection.go:190`).

## Consequence

Any Go-compatible client calling `UpdateDocumentsWithFilter` against a Rust node gets a 405. Filtered document update is reachable on Rust only through a GraphQL mutation, so there is no HTTP-level equivalent for a client written against Go's API.

Unlike #1492 this one fails loudly, so it is a capability gap rather than a correctness hazard.

## Why it was missed

#1425's route probe used a method discriminator that reports a path as "present" when any verb resolves. `/collections/{name}` resolves for `GET`/`POST`/`DELETE`, so the path was recorded as present and the missing `PATCH` verb never surfaced. Route-presence parity was checked at path granularity, not method granularity.

## Scope

Rust's HTTP layer has no filtered-update path at all today; `crates/http/src/handlers/documents.rs` handles single documents by id. The work is to expose the query engine's existing filtered update over HTTP and return Go's `{Count, DocIDs}` shape.

Best fixed together with #1492: both are Go's filtered-document operations, registered back to back at `http/handler_collection.go:510-511`, and both land on the same route entry in `crates/http/src/router/routes.rs`.

PR #1576, opened 2026-08-24
fix(http): make DELETE /collections/{name} Go's filtered document delete

### LINKED ISSUE #1492: DELETE /collections/{name} destroys the whole collection where Go deletes only the documents matching a filter

vertexclique:
Closes #1492.

Go serves `DeleteDocumentsWithFilter` at `DELETE /collections/{name}` (`http/handler_collection.go:511`) and its own client posts `{"filter": ...}` to exactly that path (`http/client_document.go:299-321`). Rust dropped the collection and every one of its versions there, and answered success.

So a Go-compatible client asking to delete two documents destroyed the collection, got a 200, and had no way to tell. That is worse than a 404: it succeeds, and does something far more destructive than what was asked.

## What changed

`DELETE /collections/{name}` now deletes the documents matching the filter and returns Go's `{"Count": n, "DocIDs": [...]}`. Dropping a collection keeps its existing route, `DELETE /collections?name=Users&active-only=true`, which is already what the Rust CLI calls; nothing in the workspace called the parameterized form over HTTP.

New `RestOperations::delete_documents_with_filter`, implemented as a `delete_<Collection>(filter: ...)` mutation, which is the same path the single-document delete already took.

**One result extractor, not two.** `delete_document` had its `_docID` extraction written inline. The filtered path needs the same shape, so that logic moved into a shared `mutation_doc_ids(result, collection, op)` and both call it. Two copies would have drifted the moment either result shape changed.

## An absent filter is refused, not treated as match-all

A missing, null, or unparseable `filter` returns 400 and deletes nothing.

I could not confirm Go's behaviour for a null filter: there is no Go checkout available in this environment, so this is stated as an open question rather than as verified parity. The asymmetry decided it. If Go refuses too, refusing costs a client one clear error message. If Go treats nil as match-all, guessing that way here deletes every document in the collection, which is the same silent-destruction class this issue is about. Refusing is recoverable; guessing is not.

This is the one thing worth checking on the Go side before merge, and the reason is recorded at `required_filter` in `handlers/documents.rs`.

## Permission changes

**`DELETE /collections/:name` moves from `CollectionPatch` to `DocumentDelete`**, because the route no longer patches a collection. Leaving it would gate document deletion behind a schema permission and refuse a legitimate `DocumentDelete` holder.

**`DELETE /collections` gains an explicit `CollectionPatch` arm.** This one is a pre-existing gap that a test in this PR caught: the route had no `Method::DELETE` arm and fell to the `_` default, so the outer gate read it as `CollectionGet`, a read permission, while `delete_collections_by_names` independently enforced `CollectionPatch`. Not exploitable, since the handler re-checks, but the table was weaker than the handler it fronts. Fixed here because this PR is re-splitting exactly this pair of delete routes; happy to lift it into its own PR if you would rather.

## Tests

`crates/http/tests/filtered_document_delete.rs`, 8 tests:

- a filter deletes only the matching documents, and the non-matching one survives
- **the regression itself**: the collection is still standing after a filtered delete
- the response uses Go's capitalised `Count` / `DocIDs`, since Go's result fields carry no json tags and a client reading `Count` off a lowercase `count` sees nothing
- a filter matching nothing is a successful no-op, not an error
- an empty body, `{}`, a null filter and non-JSON are each refused, and each is asserted to have deleted nothing
- an unknown collection is not a success
- dropping a collection still works on its own route
- both delete routes carry the permission they should

## Verification

- `just fmt-check`: green
- `just lint`: green (exit 0)
- `just test`: green, 5203 tests passed, 0 failures
- `RUSTDOCFLAGS="-D warnings" cargo doc -p defra-http -p query --no-deps`: green

`just doc` across the workspace is red on `main` already (pre-existing rustdoc errors in `crates/storage` and `crates/datastore`), so the doc gate was run scoped to the crates this PR changes.

#1493 adds `PATCH` on this same route entry and is stacked on this branch to avoid a merge conflict on the same lines.

---

## Follow-up from self-review: a GraphQL injection, and the file split

**`json_to_graphql_input` wrote JSON object keys into the mutation raw.** Probed, not reasoned about:

```
input : {"a) { _docID } } mutation { delete_Other(filter: {": 1}
output: {a) { _docID } } mutation { delete_Other(filter: {: 1}
```

A key carrying `)`, `{` or `}` closes the argument list and lets the caller append operations of its own. This is pre-existing, since `create_document` and `update_document` already fed request-body JSON through the same encoder, but this PR adds a new untrusted path to it (the `filter`) from a new route, so it is fixed here.

Keys must now be GraphQL `Name`s (`[_A-Za-z][_0-9A-Za-z]*`) or the request is refused. Refused rather than escaped: a key that is not a `Name` cannot address a schema field, so nothing legitimate is turned away. `json_to_graphql_input` and every builder became fallible.

Covered by: hostile keys at any nesting depth and inside arrays, `build_filtered_delete_mutation` refusing them at the builder the real implementation calls, ordinary field names still encoding, and string *values* staying free-form because a quoted `StringValue` cannot escape.

One test did not survive its own review. I first wrote this as an HTTP-level test, and it failed: `MockRestOperations` does its own filter matching and never builds a GraphQL document, so the encoder is never reached through the mock. A passing version would have been asserting mock behaviour while looking like it proved the boundary. It was replaced with tests against the real builder and encoder.

**`operations.rs` was 565 lines, over the 400-line guideline.** GraphQL document construction moved to `crates/query/src/rest/gql.rs`, which is also what puts the trust boundary in one file. The builders never used `self`, so they became free functions. `operations.rs` is now 429 lines, `gql.rs` 148. `mutation_doc_ids` stayed behind: it parses results rather than building documents.

Re-verified: `just fmt-check` green, `just lint` exit 0, `just test` 5208 passed 0 failed, `cargo doc -D warnings -p defra-http -p query` green.

---

## The nil-filter question is now answered from the Go source

Checked against the pinned baseline `13c46ec9`.

**Refusing a nil filter is what Go does**, so this is verified parity rather than the safe guess described above. `DeleteDocumentsWithFilter` passes `request.Filter` straight through (`http/handler_collection.go:40-61`), and `makeSelectionPlan` type-switches on it (`internal/db/document_update.go:161-183`): `string`, `immutable.Option[request.Filter]` and `map[string]any` are handled, and everything else, `nil` included, hits `default:` and returns `NewErrUnsupportedFilterType`. `httpStatusFromError` has no arm for it, so it falls through to the 400 default (`http/errors.go:196`). An absent or null filter is a 400 on both runtimes.

An empty *string* filter is `ErrEmptyFilter` on Go, also a 400.

The `{Count, DocIDs}` response shape is confirmed too: `client.DeleteResult` and `client.UpdateResult` carry no json tags (`client/collection.go:189-204`), so they marshal capitalised.

**One divergence found and not closed.** Go's switch also accepts a filter given as a *string*, parsed with `NewFilterFromString`. Rust encodes a string filter as a GraphQL string literal, so it fails rather than filtering. Supporting it would mean splicing caller-supplied text directly into the mutation, which is the injection hole this PR closes, so it wants a deliberate decision rather than a quiet passthrough. Filed as a known gap: it fails loudly, so it is a capability difference, not a silent wrong answer.

---

## Review round: the tests were green on a language the server rejects

All three blockers fixed. The filter-language one was the important catch and both halves of it were real.

**The mock and production disagreed, so every test here certified a rejected request.** Confirmed: `filter_impl.rs:312` requires an operator object, so `{"name":"Alice"}` was a 400 on the live path, while Go's connor reads a bare scalar as equality (`internal/connor/eq.go:63-76`) and deletes Alice.

Fixed on both sides:

- The mock now calls `query::Filter::from_conditions(..).matches_json_object(..)`, the production engine, so the two cannot diverge by construction.
- Implicit-eq now matches connor. `{name: "Alice"}` and `{name: {_eq: "Alice"}}` are one filter, as on Go. The change is purely additive: it only accepts input that previously errored.
- That normalization is **one** function used by both field-condition sites (`filter_impl.rs` and `eval_relation.rs`). Two copies is how the mock and production drifted apart in the first place.
- Arrays are deliberately still refused: connor gives them a meaning other than equality, and guessing it is worse than an error.

**String filters work.** Parsed by a new `query::parse_filter_string`, which wraps the source in a throwaway operation and runs the real parser, so the result is a JSON object validated like any other filter. Not interpolated, as asked. An empty string is Go's `ErrEmptyFilter`, and an unsupported type is Go's `ErrUnsupportedFilterType`; both 400.

**The guard test could not detect its own regression.** Correct: the listing does not go through collection management, and the mock's drop was a no-op. `MockCollectionManagementOperations` now records dropped names and the test asserts management is never reached. Verified by actually re-wiring the route back to the destructive handler: **9 of 11 tests fail**, including the guard.

Also done: dropped the single-call-site `field` parameter and the `pub(crate)` on `required_filter`.

**Not done, and worth its own decision.** No test here runs `RestOperationsImpl`, so the failure mode described on #1577 (renaming the `input` argument would 500 the route with the suite still green) remains uncovered. Pointing the mock at the production filter engine closes the filter-language divergence but not this. One test through a real node would cover this route and the PATCH together, and `tools/integration-test/` has no coverage of either today. Flagged rather than folded in, since it is a new integration binary rather than a change to this one.

Re-verified: `just fmt-check` green, `just lint` exit 0, `just test` 5211 passed 0 failed, `cargo doc -D warnings -p defra-http -p query` green.

### DIFF
--- a/crates/http/src/handlers/documents.rs
+++ b/crates/http/src/handlers/documents.rs
@@ -9,10 +9,12 @@
 //! Go DefraDB behavior. Go returns only HTTP 200 status with no body.
 
 use axum::{
+    body::Bytes,
     extract::{Path, State},
     http::StatusCode,
     Json,
 };
+use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
 
 use crate::error::HttpError;
@@ -207,6 +209,96 @@ pub async fn delete_document(
                 doc_id = %doc_id,
                 error = %e,
                 "Failed to delete document"
+            );
+            Err(e.into())
+        }
+    }
+}
+
+/// Go's `DeleteCollectionRequest` (`http/handler_collection.go:32`).
+#[derive(Debug, Deserialize)]
+pub struct DeleteDocumentsRequest {
+    pub filter: Option<JsonValue>,
+}
+
+/// Go's `client.DeleteResult` / `client.UpdateResult` (`client/collection.go`).
+///
+/// Go's fields carry no json tags, so they marshal capitalised. A client
+/// reading `Count` off a lowercase `count` sees nothing.
+#[derive(Debug, Serialize, PartialEq, Eq)]
+pub struct DocumentsResult {
+    #[serde(rename = "Count")]
+    pub count: usize,
+    #[serde(rename = "DocIDs")]
+    pub doc_ids: Vec<String>,
+}
+
+impl From<Vec<String>> for DocumentsResult {
+    fn from(doc_ids: Vec<String>) -> Self {
+        Self {
+            count: doc_ids.len(),
+            doc_ids,
+        }
+    }
+}
+
+/// Read the `filter` a filtered mutation must have.
+///
+/// A missing or null filter is refused rather than treated as match-all. Go's
+/// behaviour for that case could not be confirmed against its source here, and
+/// the failure mode of guessing wrong is deleting or rewriting every document
+/// in the collection, which is exactly the silent-destruction bug this route
+/// is being fixed for. Refusing is recoverable; guessing is not.
+pub(crate) fn required_filter(body: &Bytes, field: &str) -> Result<JsonValue, HttpError> {
+    let parsed: JsonValue = serde_json::from_slice(body)
+        .map_err(|e| HttpError::BadRequest(format!("invalid request body: {e}")))?;
+
+    match parsed.get(field) {
+        Some(JsonValue::Null) | None => Err(HttpError::BadRequest(format!(
+            "'{field}' is required; send a filter object to select the documents to act on"
+        ))),
+        Some(filter) => Ok(filter.clone()),
+    }
+}
+
+/// Delete every document matching a filter.
+///
+/// DELETE /api/v0/collections/{name}
+///
+/// This is Go's `DeleteDocumentsWithFilter` (`http/handler_collection.go:511`),
+/// which its own client calls by `DELETE`ing this path with `{"filter": ...}`
+/// (`http/client_document.go:299-321`). Rust used to drop the collection and
+/// every one of its versions here, and answer success, so a Go-compatible
+/// client asking to delete a few documents destroyed the collection instead.
+///
+/// Dropping a collection lives on `DELETE /api/v0/collections?name=...`.
+///
+/// Requires `DocumentDelete` permission when NAC is enabled.
+pub async fn delete_documents_with_filter(
+    State(state): State<AppState>,
+    identity: ExtractIdentity,
+    Path(collection): Path<String>,
+    body: Bytes,
+) -> Result<Json<DocumentsResult>, HttpError> {
+    require_permission(&state, &identity, NodePermission::DocumentDelete).await?;
+
+    let filter = required_filter(&body, "filter")?;
+
+    let rest = state
+        .rest
+        .as_ref()
+        .ok_or_else(|| HttpError::Internal("REST operations not configured".into()))?;
+
+    match rest
+        .delete_documents_with_filter(&collection, &filter, identity.did())
+        .await
+    {
+        Ok(doc_ids) => Ok(Json(doc_ids.into())),
+        Err(e) => {
+            tracing::warn!(
+                collection = %collection,
+                error = %e,
+                "Failed to delete documents with filter"
             );
             Err(e.into())
         }
--- a/crates/http/src/handlers/documents.rs
+++ b/crates/http/src/handlers/documents.rs
@@ -242,22 +242,37 @@ impl From<Vec<String>> for DocumentsResult {
     }
 }
 
+/// Parse a filtered mutation's request body.
+fn request_body(body: &Bytes) -> Result<JsonValue, HttpError> {
+    serde_json::from_slice(body)
+        .map_err(|e| HttpError::BadRequest(format!("invalid request body: {e}")))
+}
+
 /// Read the `filter` a filtered mutation must have.
 ///
-/// A missing or null filter is refused rather than treated as match-all. Go's
-/// behaviour for that case could not be confirmed against its source here, and
-/// the failure mode of guessing wrong is deleting or rewriting every document
-/// in the collection, which is exactly the silent-destruction bug this route
-/// is being fixed for. Refusing is recoverable; guessing is not.
-pub(crate) fn required_filter(body: &Bytes, field: &str) -> Result<JsonValue, HttpError> {
-    let parsed: JsonValue = serde_json::from_slice(body)
-        .map_err(|e| HttpError::BadRequest(format!("invalid request body: {e}")))?;
-
-    match parsed.get(field) {
-        Some(JsonValue::Null) | None => Err(HttpError::BadRequest(format!(
-            "'{field}' is required; send a filter object to select the documents to act on"
-        ))),
-        Some(filter) => Ok(filter.clone()),
+/// Go types it as `any` and accepts three forms
+/// (`internal/db/document_update.go:163-183`): a `map`, GraphQL source as a
+/// `string`, and nothing else. Its own HTTP client marshals whatever the
+/// caller handed it (`http/client_document.go:299-321`), and the JS client
+/// always sends a string, so both forms arrive in practice.
+///
+/// A missing or null filter is refused, which is parity rather than caution:
+/// nil reaches Go's `default:` arm as `ErrUnsupportedFilterType`, a 400. An
+/// empty string is Go's `ErrEmptyFilter`, also a 400.
+///
+/// The string form is parsed into conditions rather than pasted into the
+/// mutation, so it is validated exactly like a filter that arrived as an
+/// object.
+fn required_filter(request: &JsonValue) -> Result<JsonValue, HttpError> {
+    match request.get("filter") {
+        Some(JsonValue::Object(conditions)) => Ok(JsonValue::Object(conditions.clone())),
+        Some(JsonValue::String(source)) => {
+            query::parse_filter_string(source).map_err(|e| HttpError::BadRequest(e.to_string()))
+        }
+        Some(JsonValue::Null) | None => Err(HttpError::BadRequest(
+            "'filter' is required; send a filter object to select the documents to act on".into(),
+        )),
+        Some(_) => Err(HttpError::BadRequest("unsupported filter type".into())),
     }
 }
 
@@ -282,7 +297,7 @@ pub async fn delete_documents_with_filter(
 ) -> Result<Json<DocumentsResult>, HttpError> {
     require_permission(&state, &identity, NodePermission::DocumentDelete).await?;
 
-    let filter = required_filter(&body, "filter")?;
+    let filter = required_filter(&request_body(&body)?)?;
 
     let rest = state
         .rest
--- a/crates/http/src/mock/rest.rs
+++ b/crates/http/src/mock/rest.rs
@@ -15,6 +15,16 @@ struct MockDocument {
     data: JsonValue,
 }
 
+/// Whether every key in `filter` equals the same key in `data`.
+fn matches_filter(data: &JsonValue, filter: &JsonValue) -> bool {
+    match filter.as_object() {
+        Some(fields) => fields
+            .iter()
+            .all(|(key, want)| data.get(key).is_some_and(|got| got == want)),
+        None => false,
+    }
+}
+
 /// Mock REST operations for testing collection and document handlers.
 #[derive(Debug)]
 pub struct MockRestOperations {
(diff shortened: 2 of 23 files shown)


crates/http/src/mock/rest.rs
```diff
@@ -15,6 +15,16 @@ struct MockDocument {
     data: JsonValue,
 }
 
+/// Whether every key in `filter` equals the same key in `data`.
+fn matches_filter(data: &JsonValue, filter: &JsonValue) -> bool {
```
edjroz: This is flat key/value equality, but production requires an operator object:

```rust
// query/src/mapper/filter/filter_impl.rs:309-312
let ops = value.as_object()
    .ok_or_else(|| QueryError::invalid_filter("field condition must be object"))?;
```

So the two languages are mutually exclusive against this mock:

| Filter | Mock | Production | Go |
|---|---|---|---|
| `{name: "Alice"}` (what the tests send) | Count 1 ✅ | **400** `field condition must be object` | **200**, Alice deleted |
| `{name: {_eq: "Alice"}}` (real syntax) | Count 0 ❌ | 200, Alice deleted | 200, Alice deleted |

There's no shape that passes both, so every test in `filtered_document_delete.rs` certifies a request the server rejects — on a destructive route. `tools/integration-test/` has no coverage of this route either, and `RestOperationsImpl` is only constructed at `cli/src/commands/start/server_query.rs:88`.

The middle column is also a parity bug in its own right: `makeSelectionPlan` takes `map[string]any` (`internal/db/document_update.go:179-180`) and connor's `eq` treats a raw scalar as equality (`internal/connor/eq.go:64-68`), so `{"name":"Alice"}` deletes Alice on Go.

`query` is already a dependency (`crates/http/Cargo.toml:22`), so this can be `Filter::from_conditions(obj.clone()).matches_json_object(&doc.data)` — then decide whether implicit-eq wraps to `_eq` like connor or 400s with a pinned message, and add one test through `RestOperationsImpl` or a real node.

crates/http/src/handlers/documents.rs
```diff
@@ -212,3 +214,93 @@ pub async fn delete_document(
...
+}
+
+/// Read the `filter` a filtered mutation must have.
+///
+/// A missing or null filter is refused rather than treated as match-all. Go's
+/// behaviour for that case could not be confirmed against its source here, and
+/// the failure mode of guessing wrong is deleting or rewriting every document
+/// in the collection, which is exactly the silent-destruction bug this route
+/// is being fixed for. Refusing is recoverable; guessing is not.
+pub(crate) fn required_filter(body: &Bytes, field: &str) -> Result<JsonValue, HttpError> {
```
edjroz: Go accepts three filter forms here and this accepts one. `makeSelectionPlan` (`internal/db/document_update.go:163-183`):

- `case string:` → `NewFilterFromString`; `""` → `ErrEmptyFilter` (`"filter cannot be empty"`)
- `case map[string]any:` → conditions
- `default:` → `ErrUnsupportedFilterType`, 400

The Go client types `filter` as `any` and marshals it straight through (`http/client_document.go:299-321`), and Go's own integration tests pass GraphQL-syntax strings (`tests/integration/mutation/delete/with_filter_action_test.go:46`).

Failure: `DeleteDocumentsWithFilter(ctx, "{name: {_eq: \"Alice\"}}")` sends `{"filter":"{name: {_eq: \"Alice\"}}"}`. Go: 200, docs deleted. Rust: 400, nothing deleted. Same for `{"filter":""}` — Go says `"filter cannot be empty"`, this says `"filter must be an object"`.

Worth doing here rather than later since the route is new and this is the payload Go's client produces. Please parse and validate rather than splicing the string into the mutation — passing it as a GraphQL variable through the existing `execute_mutation_with_identity_and_vars` avoids the escaping question entirely.

Minor, same line: `field: &str` is `"filter"` at its only call site and this is `pub(crate)` with a same-file caller.

crates/http/tests/filtered_document_delete.rs
```diff
@@ -0,0 +1,193 @@
...
+    let result: Value = serde_json::from_str(&body).unwrap();
+    assert_eq!(result["Count"], json!(1));
+    assert_eq!(result["DocIDs"], json!(["bae-123"]));
+
+    assert_eq!(doc_ids(router).await, vec!["bae-456"], "Bob must survive");
+}
+
+/// The bug itself: this route used to destroy the collection.
+#[tokio::test]
+async fn a_filtered_delete_leaves_the_collection_standing() {
```
edjroz: This test can't detect the bug it exists for.

The `GET /api/v0/collections` it checks goes through `MockRestOperations` (`handlers/collections.rs:91-103`), while the pre-change handler called `collection_mgmt.delete_collection` — whose mock is a no-op returning `Ok(())` (`mock/collections.rs:102-104`).

Revert the route to `handlers::delete_collection` and rerun: still 200, still `["Users", ...]`. It passes.

Needs either a mock that actually drops the collection, or an assertion that collection management was never invoked.

edjroz (changes requested):
Request changes. The route change itself is right, and the response contract matches Go — I verified `client.DeleteResult` marshals `Count`/`DocIDs` with `DocIDs` initialised to `[]` rather than null, the nil-filter 400, the DAC ordering on the filtered path (filter resolved to doc ids before the two-phase check), and that nothing in-tree was relying on the old drop-the-collection behaviour.

What blocks is that the tests don't exercise the path they claim to, on a route that destroys data:

- **No filter shape is valid on both the mock and production.** The tests send `{"name":"Alice"}`, which the live path 400s; the real syntax `{name: {_eq: "Alice"}}` makes the mock match nothing. The suite is green on a language the server rejects.
- **Go accepts string filters and its own client library sends them.** The new handler 400s them.
- **`a_filtered_delete_leaves_the_collection_standing` still passes if the route is wired back to `delete_collection`**, so the one test guarding the original bug can't detect its return.

jackzampolin (approved):
Re-reviewed through 11a4edf5. Object, scalar, and string filter forms now share production semantics; the regression test detects collection deletion; and the real-node DELETE contract covers Count, DocIDs, and persisted removal. No further findings.

PR #1577, opened 2026-08-24
fix(http): serve Go's filtered document update on PATCH /collections/{name}

### LINKED ISSUE #1493: PATCH /collections/{name} is absent, so Go's UpdateDocumentsWithFilter answers 405

vertexclique:
Closes #1493.

**Stacked on #1576 (issue #1492).** Both issues land on the same route entry and the same handler file, so independent branches would conflict on merge. Review only the last commit; the base retargets to `main` once #1576 lands.

Go registers `UpdateDocumentsWithFilter` at `PATCH /collections/{name}` (`http/handler_collection.go:510`), back to back with the filtered delete. Rust registered `GET`, `POST` and `DELETE` on that path but no `PATCH`, so the request answered 405 and filtered update had no HTTP-level equivalent at all: it was reachable only through a GraphQL mutation.

Unlike #1492 this failed loudly, so it is a capability gap rather than a correctness hazard.

## What changed

`PATCH /collections/{name}` takes `{"filter": ..., "updater": "..."}` and returns Go's `{"Count": n, "DocIDs": [...]}`.

New `RestOperations::update_documents_with_filter`, built as an `update_<Collection>(filter: ..., input: ...)` mutation. It reuses the `mutation_doc_ids` extractor that #1576 factored out, so all four document-mutation paths agree on the result shape by construction.

## The `updater` encoding

Go types this as a `string` holding JSON (`UpdateCollectionRequest`, `http/handler_collection.go:36-39`), which is what its client sends, so that is the primary path and it is tested directly.

A plain JSON object is also accepted. A hand-written client is likelier to send one than to double-encode, and refusing it would buy nothing. The leniency does not widen into accepting nonsense: a non-object updater such as `"[1,2]"` is refused.

## Neither half may be guessed

A missing or null `filter` and a missing or null `updater` are both 400s. A missing filter would rewrite every document in the collection, and a missing updater would report an update that never happened. Same reasoning as the filter handling in #1576, and the shared `required_filter` helper is the same function.

## Permission

`PATCH /collections/:name` is registered as `DocumentUpdate`, matching the `POST` on the same path. A new verb on an existing path gets its own `MatchedPath` arm, and without an entry it would fall to the `_` default and be gated as a read.

## Tests

`crates/http/tests/filtered_document_update.rs`, 8 tests:

- **the gap itself**: the route is no longer a 405
- an updater sent as a JSON string is applied, only to matching documents, and a non-matching document is asserted untouched
- an updater sent as an object is applied
- the response uses Go's capitalised `Count` / `DocIDs`
- a filter matching nothing updates nothing and reports `Count: 0`
- nine malformed bodies are each refused, and each is asserted to have changed nothing
- an unknown collection is not a success
- the route is gated as `DocumentUpdate`

## Verification

- `just fmt-check`: green
- `just lint`: green (exit 0)
- `just test`: green, 5211 tests passed, 0 failures
- `RUSTDOCFLAGS="-D warnings" cargo doc -p defra-http -p query --no-deps`: green

`just doc` across the workspace is red on `main` already (pre-existing rustdoc errors in `crates/storage` and `crates/datastore`), so the doc gate was run scoped to the crates this PR changes.

---

## Follow-up from self-review

Rebased onto the hardened base in #1576. `build_filtered_update_mutation` moved into `crates/query/src/rest/gql.rs` with the other builders and became fallible, so both caller-supplied objects on this route, the filter and the updater, are refused when a key could escape the GraphQL document. `a_filtered_update_refuses_a_hostile_key_in_either_half` pins both halves and the ordinary encoding.

Re-verified after the rebase: `just lint` exit 0, `just test` 5217 passed 0 failed.

---

## Verified against the Go source

Checked against the pinned baseline `13c46ec9`. `updater` really is a `string` holding JSON (`http/handler_collection.go:36-39`, passed to `col.UpdateDocumentsWithFilter(ctx, request.Filter, request.Updater, ...)` at `:81`), so the primary encoding this PR implements is correct, and a nil filter is a 400 on Go as well (see #1576).

**One divergence found and not closed.** Go runs `fastjson.Parse(updater)` and branches on the parsed type (`internal/db/document_update.go:73-87`): an **array** is a JSON Patch update, an **object** is a merge, anything else is `client.ErrInvalidUpdater`. Rust has no JSON Patch path, so this handler refuses arrays with a 400 where Go would apply a patch.

That is a capability gap that fails loudly rather than a silent wrong answer, but it is not full parity and is called out here rather than papered over. Object updaters, which is what a merge update sends and what Go's own client produces from a `map`, work.

---

## Review round: the patch claim was wrong, and the 400 is now deliberate

**Correcting this description first, because it was the actual error.** It said this is where "Go would apply a patch". At `13c46ec9` Go does no such thing: `if isPatch { // todo }` is empty, and the loop still calls `c.update` and increments the count (`internal/db/document_update.go:135-152`). Go answers 200 with a non-zero `Count` and no field change, which is what its own `TestUpdateWithPatch_DoesNothing` is named for.

**The 400 stays, now as a stated choice rather than mistaken parity.** Matching Go would mean telling a caller that N documents were updated while touching none of them. Reporting work that did not happen is worse than refusing, so this route says it cannot apply a JSON Patch instead of copying a no-op that lies about its result. The test carries that reasoning and asserts nothing changed.

**String filters** are resolved by the rebase onto #1576: `required_filter` parses them before the builder is reached, so nothing is interpolated. Covered here for PATCH as well, alongside the bare-scalar form.

**The mock filter language** is fixed in #1576, and the tests in this file moved to the real syntax in the same rebase, as requested.

**The unfalsifiable-tests point stands and is not fixed here.** Nothing in this file executes `RestOperationsImpl`, so the `input`-argument-rename failure mode remains uncovered. That needs one test through a real node covering this route and the DELETE together; raised on #1576 as a shared follow-up rather than duplicated into two mock suites.

Re-verified after the rebase: `just fmt-check` green, `just lint` exit 0, `just test` 5223 passed 0 failed, `cargo doc -D warnings -p defra-http -p query` green.

### DIFF
--- a/crates/http/tests/filtered_document_update.rs
+++ b/crates/http/tests/filtered_document_update.rs
@@ -1,0 +1,247 @@
+//! `PATCH /collections/{name}` is Go's filtered document update.
+//!
+//! Go registers `UpdateDocumentsWithFilter` here
+//! (`http/handler_collection.go:510`), back to back with the filtered delete,
+//! and its client sends `{"filter": ..., "updater": "..."}` to this path
+//! (`http/client_document.go:263-296`). Rust registered `GET`, `POST` and
+//! `DELETE` on this path but no `PATCH`, so the request answered 405 and
+//! filtered update was reachable only through a GraphQL mutation.
+//!
+//! Unlike the filtered delete this failed loudly, so it was a capability gap
+//! rather than a correctness hazard.
+
+use std::sync::Arc;
+
+use axum::{
+    body::{to_bytes, Body},
+    http::{header::HOST, Method, Request, StatusCode},
+    Router,
+};
+use defra_http::route_permissions::{route_permission, RoutePermission};
+use defra_http::router::{AppStateBuilder, NodePermission};
+use defra_http::{MockQueryExecutor, MockRestOperations};
+use query::rest::RestOperations;
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+fn router() -> Router {
+    let state = AppStateBuilder::new(Arc::new(MockQueryExecutor::new()))
+        .with_rest(Arc::new(MockRestOperations::new()) as Arc<dyn RestOperations>)
+        .build();
+    defra_http::create_router_with_state(state)
+}
+
+async fn send(router: Router, method: Method, uri: &str, body: &str) -> (StatusCode, String) {
+    let response = router
+        .oneshot(
+            Request::builder()
+                .method(method)
+                .uri(uri)
+                .header(HOST, "localhost:9181")
+                .header("content-type", "application/json")
+                .body(Body::from(body.to_string()))
+                .unwrap(),
+        )
+        .await
+        .expect("router should respond");
+    let status = response.status();
+    let body = to_bytes(response.into_body(), usize::MAX).await.unwrap();
+    (status, String::from_utf8_lossy(&body).into_owned())
+}
+
+async fn patch_with(router: Router, body: &str) -> (StatusCode, String) {
+    send(router, Method::PATCH, "/api/v0/collections/Users", body).await
+}
+
+async fn document(router: Router, doc_id: &str) -> Value {
+    let (status, body) = send(
+        router,
+        Method::GET,
+        &format!("/api/v0/collections/Users/document/{doc_id}"),
+        "",
+    )
+    .await;
+    assert_eq!(status, StatusCode::OK, "{body}");
+    serde_json::from_str(&body).unwrap()
+}
+
+/// The gap itself: this used to be a 405.
+#[tokio::test]
+async fn the_route_is_no_longer_a_405() {
+    let (status, body) = patch_with(
+        router(),
+        r#"{"filter":{"name":{"_eq":"Alice"}},"updater":"{\"age\":31}"}"#,
+    )
+    .await;
+    assert_ne!(status, StatusCode::METHOD_NOT_ALLOWED, "{body}");
+    assert_eq!(status, StatusCode::OK, "{body}");
+}
+
+/// Go types `updater` as a string holding JSON, which is what its client
+/// sends, so that encoding has to work.
+#[tokio::test]
+async fn an_updater_sent_as_a_json_string_is_applied() {
+    let router = router();
+    let (status, body) = patch_with(
+        router.clone(),
+        r#"{"filter":{"name":{"_eq":"Alice"}},"updater":"{\"age\":31}"}"#,
+    )
+    .await;
+    assert_eq!(status, StatusCode::OK, "{body}");
+
+    let result: Value = serde_json::from_str(&body).unwrap();
+    assert_eq!(result["Count"], json!(1));
+    assert_eq!(result["DocIDs"], json!(["bae-123"]));
+
+    assert_eq!(document(router.clone(), "bae-123").await["age"], json!(31));
+    assert_eq!(
+        document(router, "bae-456").await["age"],
+        json!(25),
+        "a non-matching document must not be touched"
+    );
+}
+
+/// A hand-written client is likelier to send an object than to double-encode.
+#[tokio::test]
+async fn an_updater_sent_as_an_object_is_applied() {
+    let router = router();
+    let (status, body) = patch_with(
+        router.clone(),
+        r#"{"filter":{"name":{"_eq":"Alice"}},"updater":{"age":31}}"#,
+    )
+    .await;
+    assert_eq!(status, StatusCode::OK, "{body}");
+    assert_eq!(document(router, "bae-123").await["age"], json!(31));
+}
+
+/// Go's `UpdateResult` fields carry no json tags, so they marshal capitalised.
+#[tokio::test]
+async fn the_response_uses_gos_field_names() {
+    let (_, body) = patch_with(
+        router(),
+        r#"{"filter":{"name":{"_eq":"Alice"}},"updater":"{\"age\":31}"}"#,
+    )
+    .await;
+    let result: Value = serde_json::from_str(&body).unwrap();
+    assert!(result.get("Count").is_some(), "{body}");
+    assert!(result.get("DocIDs").is_some(), "{body}");
+}
+
+#[tokio::test]
+async fn a_filter_matching_nothing_updates_nothing() {
+    let router = router();
+    let (status, body) = patch_with(
+        router.clone(),
+        r#"{"filter":{"name":{"_eq":"Nobody"}},"updater":"{\"age\":31}"}"#,
+    )
+    .await;
+    assert_eq!(status, StatusCode::OK, "{body}");
+
+    let result: Value = serde_json::from_str(&body).unwrap();
+    assert_eq!(result["Count"], json!(0));
+    assert_eq!(document(router, "bae-123").await["age"], json!(30));
+}
+
+/// Neither half may be guessed. A missing filter would rewrite the whole
+/// collection; a missing updater would claim an update that never happened.
+#[tokio::test]
+async fn an_incomplete_request_is_refused() {
+    let bodies = [
+        "",
+        "{}",
+        "not json",
+        r#"{"updater":"{\"age\":31}"}"#,
+        r#"{"filter":null,"updater":"{\"age\":31}"}"#,
+        r#"{"filter":{"name":{"_eq":"Alice"}}}"#,
+        r#"{"filter":{"name":{"_eq":"Alice"}},"updater":null}"#,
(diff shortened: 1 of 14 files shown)


crates/query/src/rest/gql.rs
```diff
@@ -146,3 +146,16 @@ pub(super) fn build_filtered_delete_mutation(
...
     ))
 }
+
+pub(super) fn build_filtered_update_mutation(
+    collection: &str,
+    filter: &JsonValue,
+    updater: &JsonValue,
+) -> RestResult<String> {
+    Ok(format!(
+        r#"mutation {{ update_{collection}(filter: {filter}, input: {updater}) {{ _docID }} }}"#,
```
edjroz, line 156: `json_to_graphql_input` encodes a string filter as a GraphQL `String`, so Go's own client payload gets rejected downstream.

Go's filter is `any` and a string is first-class (`internal/db/document_update.go:163-175` → `NewFilterFromString`). Go's HTTP client marshals whatever it's handed (`http/client_document.go:277-280`), the JS client *always* passes a string (`js/client_document.go:196-209`), and Go's own filter-update tests use GraphQL source as a string (`tests/integration/mutation/update/with_filter_action_test.go:164-167`):

```json
{"filter":"{name: {_eq: \"John\"}}","updater":"{\"name\": \"Eric\"}"}
```

That produces `update_Users(filter: "{name: …}", input: {name: "Eric"})`, which `query_parse/filters.rs:21-39` rejects with `"filter must be an object"`. Go: 200, renamed to Eric. Rust: 400, unchanged.

The splicing design is #1576's, so the fix probably belongs there — but this builder is new, so flagging it where it's introduced. Please don't fix it by interpolating the string in raw; that's the hole `json_to_graphql_input` exists to close (`gql.rs:31-35`). Passing the filter as a GraphQL variable through `execute_mutation_with_identity_and_vars` handles both this and the escaping question.

crates/http/tests/filtered_document_update.rs
```diff
@@ -0,0 +1,196 @@
...
+    let bodies = [
+        "",
+        "{}",
+        "not json",
+        r#"{"updater":"{\"age\":31}"}"#,
+        r#"{"filter":null,"updater":"{\"age\":31}"}"#,
+        r#"{"filter":{"name":"Alice"}}"#,
+        r#"{"filter":{"name":"Alice"},"updater":null}"#,
+        r#"{"filter":{"name":"Alice"},"updater":"not json"}"#,
+        r#"{"filter":{"name":"Alice"},"updater":"[1,2]"}"#,
```
edjroz: This row asserts 400 for an array updater, but Go answers 200.

Go doesn't apply JSON Patch: an array sets `isPatch`, the apply branch is `// todo` (`internal/db/document_update.go:135-137`), and the loop still falls through to `c.update(ctx, doc)` with `results.Count++` at `:151`. Go's own test is named for it — `TestUpdateWithPatch_DoesNothing` (`with_filter_action_test.go:120-152`): updater `[{"name": "Eric"}, {"name": "Sam"}]`, name stays John, HTTP 200.

Concretely, `{"filter":{"name":{"_eq":"John"}},"updater":"[{\"name\":\"Eric\"}]"}` is Go 200 with `Count >= 1` and no field change; here it's a 400.

Rust refusing a Go *no-op* is the mild direction, so I'd take either fix — match Go's 200, or keep the 400 deliberately and say so. What needs correcting either way is the PR description: it says this is where "Go would apply a patch", and at `13c46ec9` Go does no such thing. Right now that row pins the divergence as intended behaviour.

crates/http/tests/filtered_document_update.rs
```diff
@@ -0,0 +1,196 @@
...
+    Router,
+};
+use defra_http::route_permissions::{route_permission, RoutePermission};
+use defra_http::router::{AppStateBuilder, NodePermission};
+use defra_http::{MockQueryExecutor, MockRestOperations};
+use query::rest::RestOperations;
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+fn router() -> Router {
```
edjroz, line 27: Everything in this file runs against `MockRestOperations` through `create_router_with_state`, which carries no auth middleware (`server.rs:529-532` is only `ServerBuilder::router`), so `RestOperationsImpl::update_documents_with_filter` (`rest/operations.rs:430-452`) is never executed. That makes several assertions here unfalsifiable:

- Revert the builder or the impl method to `Ok(vec![])` and **every test in this file still passes**. Only `rest/tests.rs:335-346` would notice, and that only compares a formatted string.
- Remove the PATCH route entirely and `an_unknown_collection_is_not_a_success` still passes, because 405 ≠ 200.
- Rename the update input argument from `input` (`query_parse/mutations.rs:66`) and the suite stays green while the route 500s on `missing key 'update_Users'`.

One test through `RestOperationsImpl` or a real node — Go's client body, asserting `Count`, `DocIDs` and the stored field — closes this and #1576's equivalent together. Worth one integration test for the pair rather than two mock suites.

Related: `create_router_with_state` also means the mock ignores `_identity` (`mock/rest.rs:266-271`), so nothing here covers document ACP on this route. The GraphQL path does two-phase ACP (`runner/mutation.rs:291-342`); this just doesn't prove the HTTP path reaches it.

edjroz (changes requested):
Request changes, though most of what I'd flag here isn't yours — it arrives with #1576 and is better fixed there.

Reviewed against the pinned Go baseline `13c46ec9`. The parity contract is right and I checked it line by line: route registration, `UpdateCollectionRequest{Filter any; Updater string}`, the capitalised `Count`/`DocIDs` marshalling, the nil-filter 400 (so `required_filter`'s refusal is parity, not the guess its doc comment still calls it), and `{"filter":{}}` as match-all on both runtimes. I also traced the empty-match hazard and it doesn't exist — a filter matching nothing yields `Some(vec![])` and `plan/mutation/update.rs:341-343` takes the explicit-ids branch rather than falling through to match-all, so a no-match filter can't rewrite the collection. The reuse of #1576's seam is genuine, and splitting `request_body` out of `required_filter` also drops the single-call-site `field` parameter I flagged over there.

Three things specific to this delta, inline.

**Gated on #1576.** The mock's `matches_filter` is unchanged here, so the "HTTP tests are green on a filter language production rejects" problem is inherited rather than introduced — I've raised it on #1576 and it should be fixed there. Once it is, the tests in this file need moving to the real syntax at the same time, which is why I'd rather this land after #1576 than alongside it.

jackzampolin (approved):
Re-reviewed the restacked head through 8cf83e91. The Go-compatible PATCH body now has real-node coverage for Count, DocIDs, persisted updates, and non-matching documents, alongside the base DELETE contract. The JSON Patch divergence is explicit. No further findings.

PR #1578, opened 2026-08-24
fix(p2p): recover transient marker writes

### LINKED ISSUE #1538: p2p: durably recover committed head-marker registration failures
## Context

Follow-up from #1502.

The sender-demotion path registers a presence/dirty marker before admitting a head hint. If marker registration fails after the local document commit is already durable, the failure is observable through an error and the `head_hints_failed_local` counter, but there is no durable retry obligation when the marker itself was not written.

That leaves a narrow window in which a committed head may never be announced to an otherwise healthy replicator.

## Required behavior

- Preserve fail-closed, bounded sender behavior.
- Ensure a transient marker-write failure for an already-committed head becomes durably recoverable.
- Do not restore payload/CID-valued delivery ledgers or add an unbounded in-process retry loop.
- Retain marker-plus-rederive semantics: recovery must re-read current document/collection heads.
- Expose bounded retry exhaustion/terminal disposition through existing structured status and tracing.

## Acceptance

- Inject a transient marker-store failure after a successful local commit.
- Show that the current head is eventually represented by a durable scope marker and scheduled on `/rep/retry/id/{peer}`.
- Show restart preserves the obligation.
- Show repeated failures terminate observably without queue growth.
- Cover both document and collection scopes where applicable.

iverc:
## Context

A committed document or collection head can encounter a transient storage failure while registering its required presence marker. Previously, one failed write left no durable retry schedule even when the store recovered immediately afterward.

Closes #1538.

## Changes

- retry transaction conflicts and transient I/O or backend failures within the existing four-attempt bound
- keep recovery inside the idempotent presence-only marker transaction, without adding a payload/CID ledger or background queue
- briefly pace native storage retries while keeping WASM attempts bounded and dependency-free
- report terminal exhaustion through the existing retry telemetry, tracing, and returned storage error
- cover document and collection markers, persistence through a fresh peerstore handle, and bounded exhaustion

## Testing

- [x] `cargo test -p storage`
- [x] `cargo test -p defra-p2p-adapter`
- [x] `cargo clippy -p storage --all-targets -- -D warnings`
- [x] `cargo clippy -p defra-wasm --target wasm32-unknown-unknown --no-default-features --all-targets -- -D warnings`
- [x] `git diff --check`

### DIFF
--- a/crates/storage/src/stores/peerstore.rs
+++ b/crates/storage/src/stores/peerstore.rs
@@ -15,6 +15,8 @@ use std::sync::{Arc, Mutex, OnceLock, Weak};
 use tracing;
 
 const PUSH_RETRY_TXN_MAX_ATTEMPTS: usize = 4;
+#[cfg(not(target_arch = "wasm32"))]
+const PUSH_MARKER_IO_RETRY_DELAY: std::time::Duration = std::time::Duration::from_millis(10);
 const LEGACY_RETRY_COMMIT_PREFIX: &str = "/rep/retry/commit/";
 
 fn legacy_retry_commit_prefix(peer_id: Option<&str>) -> Vec<u8> {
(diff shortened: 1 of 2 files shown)


vertexclique (approved):

vertexclique:
this looks neat so can go directly now.

PR #1584, opened 2026-08-25
fix(schema): enforce relation integrity

iverc:
## Context

This is 1/3 in the schema-definition validation stack and addresses relation-integrity gaps identified in #1404. Rust accepted invalid relation definitions through collection creation paths that Go rejects.

## Changes

- resolve named, collection-ID, `Self`, and relative-`Self` relation targets against new and persisted definitions
- reject unknown or removed targets, missing counterparts, dual-primary relations, and self-relations declared without `Self`
- enforce the checks across regular, atomic, and explicit-transaction schema creation
- add focused schema and database boundary regressions

## Testing

- [x] `cargo test -p schema`
- [x] `cargo test -p db --test collection`
- [x] `cargo test -p db --test txn`
- [x] `cargo clippy -p schema -p db --all-targets -- -D warnings`
- [x] `git diff --check`

### DIFF
--- a/crates/db/src/collection/ops/create.rs
+++ b/crates/db/src/collection/ops/create.rs
@@ -291,6 +291,13 @@ impl<S: Store> crate::database::DB<S> {
     ) -> Result<CollectionVersion> {
         self.check_node_access(None, acp::nac::NodePermission::CollectionPatch)
             .await?;
+        let existing = self.get_all_collection_versions().await?;
+        schema::definition_validation::validate_new_collections_with_existing(
+            std::slice::from_ref(&schema),
+            &existing,
+        )
+        .map_err(Error::Other)?;
+
         let name = schema.name.clone();
         let mut txn = self.new_txn(false).await?;
 
@@ -351,6 +358,10 @@ impl<S: Store> crate::database::DB<S> {
     ) -> Result<Vec<CollectionVersion>> {
         self.check_node_access(None, acp::nac::NodePermission::CollectionPatch)
             .await?;
+        let existing = self.get_all_collection_versions().await?;
+        schema::definition_validation::validate_new_collections_with_existing(&schemas, &existing)
+            .map_err(Error::Other)?;
+
         // Track old collection_id -> new collection_id mappings.
         // When views are created, create_collection_with_txn regenerates CIDs to include
         // query source data. Sibling schemas' relation fields may reference the old CIDs
--- a/crates/db/src/txn/registry/schemas.rs
+++ b/crates/db/src/txn/registry/schemas.rs
@@ -149,6 +149,12 @@ impl<S: Store + 'static> DbTransactionRegistry<S> {
 
         schema::definition_validation::validate_new_collections(&collections)
             .map_err(|e| Error::Other(format!("failed to validate schema: {}", e)))?;
+        let existing = self.db.get_all_collection_versions().await?;
+        schema::definition_validation::validate_new_collections_with_existing(
+            &collections,
+            &existing,
+        )
+        .map_err(|e| Error::Other(format!("failed to validate schema: {}", e)))?;
 
         let mut finalized = Vec::new();
         for collection in collections {
--- a/crates/schema/src/definition_validation/global_validators.rs
+++ b/crates/schema/src/definition_validation/global_validators.rs
@@ -4,6 +4,160 @@ use super::helpers::*;
 use super::DefinitionState;
 use crate::{FieldKind, ScalarKind, ORPHAN_COLLECTION_ID};
 
+fn format_relation_kind(kind: &FieldKind) -> String {
+    let (name, is_array) = match kind {
+        FieldKind::Named { name, is_array } => (name.clone(), *is_array),
+        FieldKind::Relation {
+            collection_id,
+            is_array,
+        } => (collection_id.clone(), *is_array),
+        FieldKind::SelfRef {
+            relative_id,
+            is_array,
+        } => {
+            let name = if relative_id.is_empty() {
+                "Self".to_string()
+            } else {
+                format!("Self-{relative_id}")
+            };
+            (name, *is_array)
+        }
+        _ => return format_field_kind(kind),
+    };
+
+    if is_array {
+        format!("[{name}]")
+    } else {
+        name
+    }
+}
+
+/// Matches Go's validateRelationPointsToValidKind.
+pub(super) fn validate_relation_points_to_valid_kind(
+    new_state: &DefinitionState,
+    old_state: &DefinitionState,
+) -> Vec<String> {
+    let mut errs = Vec::new();
+    for collection in &new_state.collections {
+        for field in collection.relation_fields() {
+            if new_state
+                .collection_for_kind(collection, &field.kind)
+                .is_some()
+            {
+                continue;
+            }
+
+            if let Some(removed) = old_state.collection_for_kind(collection, &field.kind) {
+                errs.push(format!(
+                    "cannot remove a collection while another field references it. Removed: {}, ReferencedBy: {}, Field: {}",
+                    removed.name, collection.name, field.name
+                ));
+            } else {
+                errs.push(format!(
+                    "no type found for given name. Field: {}, Kind: {}",
+                    field.name,
+                    format_relation_kind(&field.kind)
+                ));
+            }
+        }
+    }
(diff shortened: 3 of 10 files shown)


crates/db/src/collection/ops/create.rs
```diff
@@ -291,6 +291,13 @@ impl<S: Store> crate::database::DB<S> {
     ) -> Result<CollectionVersion> {
         self.check_node_access(None, acp::nac::NodePermission::CollectionPatch)
             .await?;
+        let existing = self.get_all_collection_versions().await?;
```
jackzampolin: **[bug]** `create_collection_inner` (and `create_collections_atomic_inner` at line 361) loads existing definitions with `get_all_collection_versions()`, which scans every `/collection/id/` record, active and inactive. Go's `addCollections` feeds `description.GetActiveCollections(ctx)` into `validateNewCollection` — name-pointer entries from the current txn, active only. After a migration, an inactive version still has the old name, so `collection_for_kind`'s Named fallback (`collections.iter().find(|col| col.name == *name)`) will treat that leftover as a valid target. A new collection can then relate to a renamed/inactive type that Go would reject with "no type found for given name". The same snapshot is also iterated by `validate_secondary_fields_pair_up`, so an inactive version with a dangling secondary can fail an unrelated create.

**Suggestion:** Pass only active collections into `validate_new_collections_with_existing`, matching `GetActiveCollections`. Filter `existing` with `is_active` (or use the active name index), and leave the all-version scan on `validate_collection_changes` where Go's patch path uses `GetCollections`.

crates/db/src/txn/registry/schemas.rs
```diff
@@ -149,6 +149,12 @@ impl<S: Store + 'static> DbTransactionRegistry<S> {
 
         schema::definition_validation::validate_new_collections(&collections)
             .map_err(|e| Error::Other(format!("failed to validate schema: {}", e)))?;
+        let existing = self.db.get_all_collection_versions().await?;
```
jackzampolin: **[bug]** `add_schema_in_txn_with_acp` already holds the write txn, then ignores it and calls `self.db.get_all_collection_versions()`, which opens a separate read txn on committed state. Go's `GetActiveCollections` reads `datastore.CtxMustGetTxn(ctx)`. Uncommitted deletes in this txn are invisible, so `delete_collections_in_txn` followed by `add_schema_in_txn` of a type that relates to the deleted collection still sees the target (`list_collections()` / the store snapshot) and accepts the schema; commit then leaves a dangling relation. The inverse also applies: collections created earlier in the same txn are missing from `existing`, so a later in-txn schema that relates to them is rejected (today this is partly masked because `known_types` also comes from the process cache via `list_collections()` at line 142).

**Suggestion:** Build `existing` from the current txn's systemstore, the same way `get_collections_in_txn` already does. Prefer active versions only (Issue 1). `known_types` should come from that same txn-visible set so parse and relation validation agree.

crates/schema/src/definition_validation/global_validators.rs
```diff
@@ -4,6 +4,160 @@ use super::helpers::*;
...
+    for collection in &new_state.collections {
+        for field in collection.relation_fields() {
+            if new_state
+                .collection_for_kind(collection, &field.kind)
+                .is_some()
+            {
+                continue;
+            }
+
+            if let Some(removed) = old_state.collection_for_kind(collection, &field.kind) {
```
jackzampolin: **[suggestion]** When a relation target is missing from `new_state` but still resolvable in `old_state`, this emits "cannot remove a collection while another field references it. Removed: …". Go's `validateRelationPointsToValidKind` has no `oldState` branch and always returns `NewErrFieldKindNotFound` ("no type found for given name. Field: …, Kind: …"). Collection delete already has its own "cannot remove a collection while another field references it" check in `delete_collection_version` / `delete_collections_in_txn`. On patch, dropping or replacing a referenced collection now produces a different error string than Go, which breaks error-parity tests.

**Suggestion:** Always report `no type found for given name. Field: {}, Kind: {}` here, as Go does. Keep the remove-while-referenced message on the delete paths that already own it.

crates/schema/src/field.rs
```diff
@@ -165,6 +165,7 @@ impl FieldDescription {
             );
         }
 
+        // Matches Go's validateRelationNameSet.
```
jackzampolin: **[suggestion]** The new `// Matches Go's validateRelationNameSet.` comment (and the sibling `// Matches Go's validateRelationalFieldIDType.` at `crates/schema/src/collection.rs:367`) restates provenance on existing checks. They do not explain a non-obvious invariant, and the mapping already lives in the `/// Matches Go's …` docs on the new validators.

**Suggestion:** Drop both `// Matches Go's …` comments. Leave the checks in place.

crates/schema/src/definition_validation/mod.rs
```diff
@@ -110,6 +149,31 @@ pub fn validate_new_collections(new_collections: &[CollectionVersion]) -> Result
...
 
+/// Validates relation integrity for new collections alongside persisted definitions.
+pub fn validate_new_collections_with_existing(
+    new_collections: &[CollectionVersion],
+    existing_collections: &[CollectionVersion],
+) -> Result<(), String> {
+    let collections = new_collections
+        .iter()
+        .chain(existing_collections)
+        .cloned()
```
jackzampolin: **[nit]** `validate_new_collections_with_existing` clones every new and existing `CollectionVersion` into a Vec, then `DefinitionState::new` clones them again into `collections`, `collections_by_id`, and the active maps.

**Suggestion:** Build one `Vec` (or iterate once) and let `DefinitionState::new` be the single clone, or have `DefinitionState` borrow the slice. Same pattern is already expensive; this call site adds another full copy on every schema create.

vertexclique:
Reviewed against the Go validators this claims parity with (`internal/db/definition_validation.go` @ `870a59c5`). One blocker, one unflagged behaviour change, two notes.

### 1. `validate_self_references` does not match Go, though the doc comment says it does

Go has **two** loops. The first (`definition_validation.go:1020`) resolves the field kind against `col` and compares `otherDef.CollectionID == col.CollectionID`. The second (`:1040`) resolves against `newState.activeCollectionsByName[col.Name]` and compares `otherDef.CollectionID == newState.collectionsByID[col.VersionID].VersionID`.

This implementation collapses both into one loop with an extra disjunct:

```rust
if target.collection_id == collection.collection_id
    || target.collection_id == collection.version_id
```

Two problems with the second clause. It compares a `collection_id` against a `version_id`, which is not a comparison Go makes anywhere. And it drops the active-collection resolution that Go's second loop is specifically there to perform, so a non-active version resolves differently than it would in Go.

It is also reachable as a false positive: any collection whose `collection_id` happens to equal another collection's `version_id` gets rejected with "must specify 'Self' kind for self referencing relations".

If this is a deliberate stand-in because collection IDs are not finalised at creation time, that is defensible, but it needs a comment saying so and the doc comment should stop claiming a straight match. Otherwise please reproduce Go's second loop.

### 2. The four validators are registered twice, and one registration changes patch behaviour

They are appended to `GLOBAL_VALIDATORS` **and** duplicated into the new `RELATION_VALIDATORS`. Two consequences:

- On any path that runs both, each relation error appears twice in the joined error string.
- `GLOBAL_VALIDATORS` runs on the update/patch path, not just creation. The PR description says the checks are enforced "across regular, atomic, and explicit-transaction schema creation" and does not mention patch. If patch enforcement is intended, it is worth saying so and adding a patch regression, since that is the riskier path for existing deployments. If it is not intended, the `GLOBAL_VALIDATORS` registration should come out.

### 3. `validate_single_side_primary` has a context-free message

```rust
errs.push("relation can only have a single field set as primary".to_string());
```

No collection, field, or relation name. Every other message added in this PR carries context, and on a large schema this one gives the user nothing to go on.

### 4. Cost on every schema create

`validate_new_collections_with_existing` does `get_all_collection_versions()` and then `.chain(existing).cloned().collect::<Vec<_>>()`, cloning every persisted collection on each create, and `collection_for_kind` falls back to an O(n) linear scan per relation field. Fine at ten collections, less so at a thousand. Taking references instead of cloning looks like a small change.

Tests and the parity comments elsewhere in the diff look good. Item 1 is the one I would want resolved before merge.

vertexclique (approved):

PR #1585, opened 2026-08-25
fix(schema): validate field defaults

iverc:
## Context

This is 2/3 in the schema-definition validation stack and depends on #1584. Invalid field defaults could be accepted with a schema and fail only when a document attempted to use them.

## Changes

- validate defaults with the same schema-aware JSON conversion used by document creation
- reject malformed or type-incompatible defaults during collection creation
- apply the same validation after JSON schema patches
- cover supported scalar defaults and invalid create/patch cases

## Testing

- [x] `cargo test -p db --test collection`
- [x] `cargo test -p db --test definition`
- [x] `cargo test -p query`
- [x] `cargo clippy -p query -p db --all-targets -- -D warnings`
- [x] `git diff --check`

### DIFF
--- a/crates/db/src/collection/validation.rs
+++ b/crates/db/src/collection/validation.rs
@@ -1,6 +1,31 @@
 use super::*;
 
 impl Collection {
+    /// Matches Go's validateCollectionFieldDefaultValue.
+    pub(crate) fn validate_default_values(schema: &CollectionVersion) -> Result<()> {
+        for field in &schema.fields {
+            let Some(default) = &field.default_value else {
+                continue;
+            };
+            let value =
+                query::plan::mutation::json_to_normal_value_with_kind(default, Some(&field.kind));
+            let value = value.map_err(|error| {
+                Error::Other(format!(
+                    "default field value is invalid. Collection: {}, Inner: {}",
+                    schema.name, error
+                ))
+            })?;
+
+            if !is_value_compatible_with_kind(&value, &field.kind) {
+                return Err(Error::Other(format!(
+                    "default field value is invalid. Collection: {}, Inner: Field '{}' has incompatible type",
+                    schema.name, field.name
+                )));
+            }
+        }
+        Ok(())
+    }
+
     /// Validate a document against this collection's schema.
     ///
     /// Returns an error if the document contains fields with incorrect types.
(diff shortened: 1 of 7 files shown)


crates/db/src/collection/validation.rs
```diff
@@ -1,6 +1,31 @@
...
 impl Collection {
+    /// Matches Go's validateCollectionFieldDefaultValue.
+    pub(crate) fn validate_default_values(schema: &CollectionVersion) -> Result<()> {
+        for field in &schema.fields {
+            let Some(default) = &field.default_value else {
+                continue;
+            };
+            let value =
+                query::plan::mutation::json_to_normal_value_with_kind(default, Some(&field.kind));
+            let value = value.map_err(|error| {
```
jackzampolin: **[suggestion]** Conversion failures are wrapped as `default field value is invalid. Collection: {name}, Inner: {error}` without `field.name`. The compatibility branch below includes the field name, but DateTime/Float/array conversion errors from `json_to_normal_value_with_kind` do not. SDL still accepts any string as a DateTime `@default`; those now fail here, and a schema with multiple DateTime defaults would not identify which field was invalid.

**Suggestion:** Include `field.name` in the conversion `map_err` the same way the incompatible-type branch does, e.g. `Inner: Field '{field}': {error}`.

vertexclique:
Looks good. Coverage is right and the wiring is minimal.

Worth recording why the atomic path is covered, since it is not obvious from the diff: `validate_default_values` sits in `create_collection_with_txn`, and `create_collections_atomic_inner` calls that per schema, so atomic creation is covered by construction rather than only by the passing positive test. Patch is covered separately in `definition/patch/validate.rs`.

One parity note rather than a defect. Go's `validateCollectionFieldDefaultValue` (`internal/db/definition_validation.go:1203`) validates by attempting real document construction:

```go
_, err := client.NewDocFromMap(ctx, map[string]any{}, col)
if err != nil && !errors.Is(err, client.ErrMissingRequiredField) {
```

so it catches anything document construction rejects, not only kind incompatibility. This PR reimplements the check as convert-then-`is_value_compatible_with_kind`. Same intent, and arguably a clearer error, but the two can drift as document construction gains validation. A line in the doc comment noting the deliberate difference would stop a future reader assuming they are equivalent.

Dropping `#[allow(dead_code)]` to make `json_to_normal_value_with_kind` load-bearing is the right call, and reusing the same conversion the create path uses is what makes this trustworthy.

vertexclique (approved):

PR #1586, opened 2026-08-25
fix(view): protect populated cache on dematerialize

### LINKED ISSUE #1404: Audit: Rust registers 25 of Go v1.0.0's 39 schema definition validators

iverc:
## Context

This is 3/3 in the schema-definition validation stack and depends on #1585. A metadata-only patch could dematerialize a populated view while leaving its cache data behind.

## Changes

- reject materialized-to-embedded transitions while the view cache contains data
- reuse the existing collection-data check at the shared patch decision point
- continue allowing empty views to dematerialize and embedded views to rematerialize
- add a regression covering the populated-cache transition

## Testing

- [x] `cargo test -p db --test definition`
- [x] `cargo test -p db`
- [x] `cargo test -p schema -p query`
- [x] `cargo clippy -p schema -p query -p db --all-targets -- -D warnings`
- [x] `git diff --check`

Closes #1404

### DIFF
--- a/crates/db/src/definition/patch/mod.rs
+++ b/crates/db/src/definition/patch/mod.rs
@@ -168,6 +168,17 @@ impl<S: Store> crate::database::DB<S> {
         // same index/head/cache namespace as the rest of the collection history.
         new_schema.root_id = old_schema.root_id;
 
+        // Matches Go's validateDematerializedViewHasNoData.
+        if old_schema.is_materialized
+            && !new_schema.is_materialized
+            && self.collection_has_data(&old_schema).await?
+        {
+            return Err(Error::InvalidPatch(format!(
+                "cannot dematerialize a materialized view that has data, first truncate it and then try again. Name: {}, VersionID: {}",
+                old_schema.name, old_schema.version_id
+            )));
+        }
+
         // Handle in-place updates (deactivation, IsActive-only, or PreviousVersion/Transform-only).
         // These don't create a new schema version - they update the existing one.
         let is_isactive_only_change = is_active_explicitly_set
--- a/crates/db/tests/definition/patch.rs
+++ b/crates/db/tests/definition/patch.rs
@@ -6,6 +6,7 @@ use schema::FieldKind;
 use schema::ScalarArrayKind;
 use storage::backends::MemoryStore;
 use storage::corekv::Key;
+use storage::keys::datastore::ViewCacheKey;
 use storage::keys::systemstore::LensConfigKey;
 
 async fn agent_response_db() -> DB<MemoryStore> {
@@ -288,6 +289,72 @@ async fn patch_view_query_creates_new_version() {
         Some(original.version_id())
     );
     assert_eq!(db.get_all_collection_versions().await.unwrap().len(), 3);
+}
+
+#[tokio::test]
+async fn patch_dematerialization_requires_an_empty_view() {
+    let db = DB::new(MemoryStore::new()).unwrap();
+    let users = query::parse_sdl("type Users { name: String }").unwrap();
+    db.create_collections_atomic(users).await.unwrap();
+
+    let mut views = query::parse_sdl("type UserView { name: String }").unwrap();
+    let select = query::parse_query("query { Users { name } }").unwrap();
+    views[0].query = Some(schema::QuerySource::new(query::select_to_go_json(
+        &select[0],
+    )));
+    views[0].is_materialized = true;
+    db.create_collections_atomic(views).await.unwrap();
+    let dematerialized = db
+        .patch_collection(
+            "UserView",
+            r#"[{"op":"replace","path":"/UserView/IsMaterialized","value":false}]"#,
+            None,
+        )
+        .await
+        .unwrap();
+    assert!(!dematerialized.is_materialized);
+
+    let view = db
+        .patch_collection(
+            "UserView",
+            r#"[{"op":"replace","path":"/UserView/IsMaterialized","value":true}]"#,
+            None,
+        )
+        .await
+        .unwrap();
+    assert!(view.is_materialized);
+
+    let txn = db.new_txn(false).await.unwrap();
+    let cache_key = ViewCacheKey::new(view.root_id, 0).bytes();
+    txn.datastore()
+        .unwrap()
+        .set(&cache_key, b"cached")
+        .await
+        .unwrap();
+    txn.commit().await.unwrap();
+
+    let txn = db.new_txn(true).await.unwrap();
+    assert!(txn
+        .datastore()
+        .unwrap()
+        .get(&cache_key)
+        .await
+        .unwrap()
+        .is_some());
+    txn.force_discard().unwrap();
+
+    let error = db
+        .patch_collection(
+            "UserView",
+            r#"[{"op":"replace","path":"/UserView/IsMaterialized","value":false}]"#,
+            None,
+        )
+        .await
+        .unwrap_err();
+
+    assert!(error.to_string().contains(
+        "cannot dematerialize a materialized view that has data, first truncate it and then try again"
+    ));
 }
 
 #[tokio::test]


crates/db/src/definition/patch/mod.rs
```diff
@@ -168,6 +168,17 @@ impl<S: Store> crate::database::DB<S> {
         // same index/head/cache namespace as the rest of the collection history.
         new_schema.root_id = old_schema.root_id;
 
+        // Matches Go's validateDematerializedViewHasNoData.
+        if old_schema.is_materialized
+            && !new_schema.is_materialized
+            && self.collection_has_data(&old_schema).await?
+        {
+            return Err(Error::InvalidPatch(format!(
+                "cannot dematerialize a materialized view that has data, first truncate it and then try again. Name: {}, VersionID: {}",
```
jackzampolin: **[bug]** The new error tells the caller to truncate the view and retry, matching Go v1.0.0 (`errDematerializePopulatedView`). In Go v1.0.0, `collection.Truncate` deletes `keys.ViewCacheKey{CollectionShortID}` (`internal/db/collection_truncate.go`), which is what makes that recovery work. Rust `truncate_collection` / `truncate_collection_metadata` only drop `/d/`, `/del/`, `/v/`, index, and head prefixes — not `ViewCacheKey::collection_prefix`. After this check lands, a materialized view with cache data cannot be dematerialized by following the error: truncate leaves the cache, `collection_has_data` still returns true, and the next patch fails the same way. The only working workaround is emptying the source collection and calling `refresh_views` (which clears then rebuilds), which is not what the error says.

**Suggestion:** Either make truncate delete `ViewCacheKey::collection_prefix(short_id)` as Go does, so the advertised recovery actually unblocks dematerialize, or change the error to name a recovery path that exists on Rust today. Do not ship the Go wording until truncate honors it.

crates/db/tests/definition/patch.rs
```diff
@@ -290,6 +291,72 @@ async fn patch_view_query_creates_new_version() {
...
+    let error = db
+        .patch_collection(
+            "UserView",
+            r#"[{"op":"replace","path":"/UserView/IsMaterialized","value":false}]"#,
+            None,
+        )
+        .await
+        .unwrap_err();
+
+    assert!(error.to_string().contains(
```
jackzampolin: **[suggestion]** `patch_dematerialization_requires_an_empty_view` does cover the two cache states (never-populated dematerialize succeeds, then a hand-inserted `ViewCacheKey` is refused). It never asserts the recovery the error string promises: truncate, then dematerialize again. It also never populates the cache through `refresh_views`, so it would not have caught Issue 1 — `truncate_collection` after `b"cached"` would still leave that key, and the test would keep passing.

**Suggestion:** After the refusal, call `truncate_collection("UserView")` and assert dematerialize then succeeds (this should fail today). Optionally also add a path that creates a source doc, `refresh_views`, and then refuses dematerialize, so the prefix check is exercised against real cache items rather than a dummy value.

crates/db/src/definition/patch/mod.rs
```diff
@@ -168,6 +168,17 @@ impl<S: Store> crate::database::DB<S> {
         // same index/head/cache namespace as the rest of the collection history.
         new_schema.root_id = old_schema.root_id;
 
+        // Matches Go's validateDematerializedViewHasNoData.
+        if old_schema.is_materialized
+            && !new_schema.is_materialized
+            && self.collection_has_data(&old_schema).await?
```
jackzampolin: **[suggestion]** Go v1.0.0 runs this validator against `datastore.CtxMustGetTxn(ctx)` — the same transaction that later persists the schema change. This check calls `collection_has_data`, which opens a fresh read txn and discards it, after which the in-place metadata write opens a second write txn. `patch_collection_with_migration` holds `collection_write_guards` on the view, but `refresh_views` does not take that lock (it only takes a `REFRESH_DATASTORE` action lease, which patch does not hold). A concurrent refresh can therefore commit cache rows after the empty-cache check and before the dematerialize write, leaving the populated cache this PR is meant to protect.

**Suggestion:** Perform the cache-prefix existence check inside the write transaction that stores the dematerialized schema (or take the collection write lock in `refresh_views` / `build_view_cache` so it cannot interleave with this patch).

vertexclique:
Looks good, and the ordering detail is right in a way that is easy to get wrong.

The guard sits after `new_schema.root_id = old_schema.root_id`, which matters because `collection_has_data` keys the view cache prefix off `root_id` (`collection/ops/version.rs:518`). The `!version.is_materialized` early return in that helper is unreachable from here since the branch already requires `old_schema.is_materialized`, so the two agree rather than fighting.

One note, probably acceptable but worth a comment in the code. `collection_has_data` opens its own read transaction (`self.new_txn(true)`), so it sees committed state only. A patch issued inside a transaction that has written view cache data without committing would not see that data and would be allowed through. Narrow seam, but it is a real time-of-check gap and the next reader will not spot it from the call site.

The regression is nicely built: it exercises dematerialize on empty, rematerialize, then the populated case, so it pins the allowed transitions rather than just the rejected one.

vertexclique:
Btw this will be fixed by regolith backend automatically later.

vertexclique (approved):

PR #1593, opened 2026-08-26
fix(p2p): replace stale cached send paths on inbound reconnect

jackzampolin:
## Summary
- prefer each newly authenticated inbound Iroh connection as the bidirectional send path for its peer and ALPN
- replace stale cached QUIC handles after a peer restarts before transport liveness expires
- cover the cache replacement with a focused Iroh regression test

## Why
A restarted mobile peer can reconnect with the same endpoint identity and ticket while the remote still considers the old QUIC path connected. The remote accepts the new inbound connection, but outbound PushLog requests continue using the stale cached handle and time out. Since QUIC streams are bidirectional, the authenticated inbound connection is the freshest valid send path.

## Test
`cargo test -p p2p --features iroh-transport authenticated_inbound_connection_replaces_cached_send_path -- --nocapture`

### DIFF
--- a/crates/db/src/merge/mod.rs
+++ b/crates/db/src/merge/mod.rs
@@ -41,8 +41,8 @@ pub use peer_identity::{
 };
 #[cfg(not(target_arch = "wasm32"))]
 pub use push_docs::{
-    push_existing_docs, push_existing_docs_with_config, retry_collection_commit, retry_doc,
-    PushExistingDocsSeOptions,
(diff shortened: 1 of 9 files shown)


jackzampolin:
Closing after end-to-end falsification. The focused cache unit passed, but the same-environment native cold run still produced repeated 30s PushLog timeouts and the warm run still stalled, so this behavior change is not independently justified as the fix. The branch and traces remain available; any transport follow-up should start from an isolated two-node reproduction with before/after evidence.

PR #1613, opened 2026-08-27
fix(query): duplicate relation selections and grouped _docID leak

### LINKED ISSUE #1597: Duplicate relation selection returns null, and `_docID` leaks into grouped output
## What

Selecting the same relation twice produces two joins where one result comes back `nil`, and a relation filter forces `_docID` into the child mapping without stripping it before render, so it appears in output that never asked for it. This is a divergence found out of ffi-testing, affecting 3 tests in `query/one_to_many`. Suspected locations are the selection-join and group-rendering paths, but these were **inferred from the queries and failure output, not traced** — confirm before editing.

edjroz:
Closes #1597. Stack 3/5, base `edjroz/1594-relation-order-panic`.

## What

Selecting the same relation twice returns `null` for one occurrence: the parser assigned each duplicate its own mapping index, joins wrote to the first, the renderer read the last. A relation filter also forced `_docID` and filter fields into grouped child output that never requested them.

The leak came from `filter_only.rs` not descending into `GROUP` when checking "already joined?", so a second filter-only join clobbered the real selection join's mapping.

## What changed

- Duplicate selections now merge at parse time on the **parsed structure** (`Requestable` equality) in all three arms — field, fragment spread, inline fragment — matching the layer where Go dedups.
- This closes four routes to the same null: plain duplicates, `...F` spreads, inline fragments, and spec-identical selections with reordered arguments; a skipped duplicate consumes no mapping index.
- Filter-only joins consume the same collected selection set as selection joins (passed in from `apply_joins`), so `GROUP`-nested relations are recognized; the relation filter still applies via `parent_relation_filter_for_child`.
- Added `relation_rendering_1597.rs` (6 tests): the three mirrored Go tests plus the fragment/reordered-args repros.

## Verification

- All repros red-first (`"author": null`, leaked `"_docID"`), green after; the explain test asserts the duplicated request plans byte-identically to the single-selection request.
- Discriminating probe for the clobber: filtering on `age` while selecting `name` rendered `{"_docID": …, "age": 65}` — the filter join's mapping.
- `cargo test -p query`, clippy, query + basic integration areas all pass.
- Out of scope, to be filed separately: Go's `groupNode`/`pipeNode` explain nesting (reproduces with a single selection too) and two remaining leak variants needing a merge-don't-clobber fix in `filter_relation.rs`.

### DIFF
--- a/crates/query/src/mapper/filter/filter_impl.rs
+++ b/crates/query/src/mapper/filter/filter_impl.rs
@@ -13,7 +13,7 @@ use crate::limits::DEFAULT_MAX_FILTER_DEPTH;
 ///
 /// Conditions map field names to their filter values.
 /// Supports nested conditions for related objects.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Filter {
     /// Parsed filter conditions
     conditions: Map<String, JsonValue>,
--- a/tools/integration-test/tests/query/relation_rendering_1597.rs
+++ b/tools/integration-test/tests/query/relation_rendering_1597.rs
@@ -155,6 +155,65 @@ async fn duplicate_relation_selection_plans_one_join_test(cluster: TestCluster) 
     );
 }
 
+/// Mirrors Go `TestQueryOneToManyWithGroupByAndFilterOnParentRelation`: the `_docID`
+/// the relation filter needs internally must not leak into the rendered GROUP items.
+async fn group_by_with_parent_relation_filter_test(cluster: TestCluster) {
+    let node = cluster.client(0);
+    add_schema(&node);
+
+    let john = add_author(&node, "John Grisham", 65, true);
+    let voltaire = add_author(&node, "Voltaire", 327, true);
+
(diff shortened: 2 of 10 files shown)


crates/query/src/mapper/filter/filter_impl.rs
vertexclique, line 21: PartialEq must be only covers over conditions not the other elements here.

tools/integration-test/tests/query/relation_rendering_1597.rs
```diff
@@ -0,0 +1,321 @@
...
+        duplicated,
+        serde_json::json!({
+            "explain": {
+                "operationNode": [
+                    {
+                        "selectTopNode": {
+                            "selectNode": {
+                                "typeIndexJoin": {
+                                    "typeJoinMany": {
+                                        "root": {"scanNode": {}},
```
vertexclique, line 81: @jsimnz Is this full scan? Do we have cardinality metric here normally on go side? or do we want that at some point?

vertexclique (approved):
Leaving ornamental comment and a question to John

PR #1614, opened 2026-08-27
fix(query): honor groupBy on aggregate targets

### LINKED ISSUE #1595: `groupBy` is silently dropped on aggregate targets
## What

`COUNT(X: {groupBy: [field]})` counts documents rather than groups, with no error or warning — the argument is parsed and discarded. This is a divergence found out of ffi-testing, affecting 5 tests across `query/simple` and `query/one_to_many`. The aggregate target type has no field to carry `groupBy`, so there is nowhere for the parser to put it even if it were read. The true blast radius is wider than the failure count: one test passes only because grouping by two fields over three documents happens to produce three groups.

edjroz:
Closes #1595. Stack 4/5, base `edjroz/1597-relation-rendering`.

## What

`_count(X: {groupBy: [field]})` counts documents instead of groups — the argument is parsed and silently discarded. `AggregateTarget` had no field to carry it, both target parsers dropped the key, and every runner path counted items; one Go-mirrored test passed only by coincidence.

## What changed

- `AggregateTarget` gains `group_by`; both parsers (GraphQL object and JSON-variable) fill it.
- COUNT counts distinct group keys on the top-level, planner, and relation paths; keys are JSON-typed and control-character-safe (stricter than Go's collidable `%v_` concatenation), and missing fields share one `null` group like Go.
- Grouped fields are injected into document/child mappings so keys are populated, mirroring Go's mapper.
- Relation-kind group fields rewrite to `_<name>ID`; array relations error with Go's message (`cannot group by array or object field. Field: …`).
- A grouped target never shares a join — sharing made the answer depend on selection order. Deliberate divergence: Go's `Targetable.equal` ignores `GroupBy` and silently degrades to the plain count; Rust returns the correct group count in both orders. Upstream issue candidate.
- Underscore group fields (`_docID`, rewritten `_authorID`) are fetched via a private `__agg_` slot stripped before the response — no leak into rendered output.
- Added `aggregate_groupby_1595.rs` (14 tests): the five affected Go tests mirrored exactly, plus order-independence, rewrite/error, `_docID` grouping, leak guard, and Go's zero/empty cases.

## Verification

- Red-first: document counts where group counts were expected (e.g. ages 32/32/19 → 3, expected 2); order-dependence reproduced live (`b` = 1 or 3 by field position).
- Leak guard proves `published { name }` stays `name`-only alongside `COUNT(published: {groupBy: [_docID]})`.
- Full workspace gate at stack head: fmt clean, workspace clippy `-D warnings` exit 0, 5270 unit tests pass, query area 64 rust_* / basic at baseline.
- Known gaps for follow-up issues: grouped-target `limit`/`offset` slices documents not groups; SUM/AVG accept `groupBy` where Go rejects at validation; the relation-path rewrite leg lacks a discriminating fixture.

### DIFF
--- a/crates/query/src/runner/query/aggregate.rs
+++ b/crates/query/src/runner/query/aggregate.rs
@@ -7,8 +7,8 @@ use serde_json::Value as JsonValue;
 use std::sync::Arc;
 
 use crate::document::{documents_to_plan_docs, DocumentMapping};
-use crate::error::Result;
-use crate::mapper::{GroupBy, Requestable, Select};
+use crate::error::{QueryError, Result};
+use crate::mapper::{Requestable, Select};
 use crate::planner::{Doc, Planner};
 use crate::txn::TransactionRegistry;
 
@@ -24,19 +24,20 @@ use super::super::{DocFetcher, QueryRunner};
 fn distinct_group_count<'a>(
     docs: impl Iterator<Item = &'a Doc>,
     mapping: &DocumentMapping,
-    group_by: &GroupBy,
+    group_fields: &[String],
 ) -> i64 {
-    let indexes: Vec<Option<usize>> = group_by
-        .fields
+    let indexes: Vec<Option<usize>> = group_fields
         .iter()
         .map(|name| mapping.first_index_of_name(name))
         .collect();
     docs.map(|doc| {
         indexes
             .iter()
-            .map(|index| match index {
-                Some(i) => doc.get(*i).unwrap_or(&JsonValue::Null).to_string(),
-                None => JsonValue::Null.to_string(),
+            .map(|index| {
+                index
+                    .and_then(|i| doc.get(i))
+                    .unwrap_or(&JsonValue::Null)
+                    .to_string()
             })
             .collect::<Vec<_>>()
             .join("\u{1}")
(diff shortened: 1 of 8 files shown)


crates/query/src/runner/query/aggregate.rs
```diff
@@ -6,17 +6,46 @@ use schema::CollectionVersion;
...
+            .iter()
+            .map(|index| {
+                index
+                    .and_then(|i| doc.get(i))
+                    .unwrap_or(&JsonValue::Null)
+                    .to_string()
+            })
+            .collect::<Vec<_>>()
+            .join("\u{1}")
+    })
```
vertexclique, line 44: This is N x M materialization. Please uze zip then fold to get that accumulator value.

crates/query/src/runner/query/relation_aggregate.rs
```diff
@@ -561,6 +567,33 @@ impl<F: DocFetcher + 'static, R: TransactionRegistry> QueryRunner<F, R> {
     }
 }
 
+/// Number of distinct `group_by` keys across relation items.
+///
+/// An item missing a grouped field is keyed as null, so all such items share
+/// one group. A single relation is fetched under its ID field, so `author`
+/// falls back to `_authorID` (see `GroupBy::resolved_fields`).
+fn distinct_group_count(items: &[&JsonValue], group_by: &crate::mapper::GroupBy) -> i64 {
+    items
```
vertexclique, line 576: same as above btw. mat is happening.

vertexclique (commented):
In flight mat is happening in two places.

Issue #1596, opened 2026-08-27
Related-document ordering is reversed


edjroz:
## What

Child relation lists arrive in reverse order, so `limit`/`offset` windows select different rows and float accumulation differs. This is a divergence found out of ffi-testing, proven arithmetically rather than inferred: Go sums `4.9 + 4.5 + 4.2 = 13.600000000000001` where Rust returns `13.6`, and two independent limit/offset cases confirm the reversal. The aggregate code is correct — it operates on already-wrongly-ordered input from the join planner, so the fix belongs in the join, not the aggregate.

vertexclique:
> Go sums 4.9 + 4.5 + 4.2 = 13.600000000000001 where Rust returns 13.6

this is totally ok. Machine epsilon problems are actually very ok and can be checkable with [f64::EPSILON](https://doc.rust-lang.org/std/f64/constant.EPSILON.html)

PR #1615, opened 2026-08-27
fix: preserve f64 bits through JSON parsing

edjroz:
Stack 5/5, base `edjroz/1595-aggregate-groupby`. No issue filed — found during #1596's verification.

## What

Any stored f64 needing more than 15 significant digits collapsed to a nearby double: a created `13.600000000000001` stored and returned the bits of `13.6`. Go round-trips exactly (`strconv.ParseFloat`), so JSON-created documents diverged from Go — and from our own GraphQL-inline create path, which parsed correctly.

The cause is serde_json's default number parser, off by up to 1 ULP without the `float_roundtrip` feature. The collapse happened at ingest, **before content addressing**, so affected documents carried docIDs that diverge from Go's.

## What changed

- Enabled `float_roundtrip` on the workspace `serde_json` dependency (one line in `Cargo.toml`); feature unification fixes every entry point — HTTP ingest, document parsing, the CLI's response re-parse — at once.
- Added `tools/integration-test/tests/query/float_precision.rs` (3 tests): exact create round-trip, `_eq` filter match on the precise value, the SUM case.
- Tightened `join_order_1596.rs`'s SUM assertion from `13.6` to Go's exact `13.600000000000001`, dropping the stale ceiling comment.

## Verification

- Discriminating probe: the same string through `str::parse` vs serde_json gives different bits (`…3334` vs `…3333`); equal with the feature enabled.
- Pre-fix probes pinned the loss to ingest, not render: `_eq: 13.6` matched a JSON-created doc while `_gt: 13.6` returned nothing; CBOR storage, index keys, SUM arithmetic, and render are all bit-exact.
- **DocID impact:** identical content gets a different docID before/after (`bae-fa7ac32d-…` → `bae-e48938ce-…`) for >15-digit floats; the new IDs are the Go-parity ones. Pre-1.0, no migration concern. `cargo test -p document` and the basic area pass.
- Full gate: query area 78 passed / 0 rust failures, basic at baseline, workspace tests exit 0, workspace clippy clean.
- Follow-up candidate (not here): `cursor`/`ffi-test`/`integration-test` hardcode `serde_json = "1.0"` — safe today only via feature unification.

### DIFF
--- a/tools/integration-test/tests/query/float_precision.rs
+++ b/tools/integration-test/tests/query/float_precision.rs
@@ -1,0 +1,80 @@
+//! A stored `Float` must round-trip its exact f64 bits. `13.600000000000001`
+//! is a different double from `13.6`, and Go DefraDB returns it verbatim.
+
+use crate::one_to_many_common::{add_author, add_book, add_schema};
+use integration_test::{DefraClient, TestCluster};
+
+/// `4.9 + 4.5 + 4.2` accumulated in insertion order.
+const PRECISE: f64 = 13.600000000000001;
+
+fn add_rated(node: &DefraClient) {
+    node.schema_add("type Rated { name: String rating: Float }")
+        .expect("add Rated schema");
+    node.collection_create(
+        "Rated",
+        r#"{"name": "precise", "rating": 13.600000000000001}"#,
+    )
+    .expect("create Rated doc");
+}
+
+async fn create_round_trip_test(cluster: TestCluster) {
+    let node = cluster.client(0);
+    add_rated(&node);
+
+    let result = node
+        .query("query { Rated { name rating } }")
+        .expect("query Rated");
+
+    assert_eq!(
+        result["Rated"],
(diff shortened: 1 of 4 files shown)


tools/integration-test/tests/query/float_precision.rs
```diff
@@ -0,0 +1,80 @@
+//! A stored `Float` must round-trip its exact f64 bits. `13.600000000000001`
+//! is a different double from `13.6`, and Go DefraDB returns it verbatim.
+
+use crate::one_to_many_common::{add_author, add_book, add_schema};
+use integration_test::{DefraClient, TestCluster};
+
+/// `4.9 + 4.5 + 4.2` accumulated in insertion order.
+const PRECISE: f64 = 13.600000000000001;
```
vertexclique, line 8: I left a comment for this in the issue. I think we must be machine epsilon flexible over the network.

we can extend the tests with assuming epsilon flip will not happen.

vertexclique (approved):

PR #1622, opened 2026-08-28
feat(storage): Default `Regolith` and shed every other backend

vertexclique:
Replaces every storage backend with [regolith](https://github.com/sourcenetwork/regolith) 0.1.2.

## What changed

**One engine.** `lark`, `redb`, `fjall`, `rocksdb`, `leveldb` and the separate in-memory store are gone, along with the feature flags that selected between them across eight crates. regolith is the store on server, embedded and browser; in-memory is the same engine on `MemEnv`, not a second implementation. The browser client moves to regolith on OPFS.

**The shadow concurrency control is gone.** `backends/shared.rs` carried a second OCC implementation layered over backends that had none. regolith validates its own transactions, so it was deleted.

**Reads borrow instead of copying.** `Reader::get` and `KvPair::value` return `Bytes` over regolith's refcounted slices, so a read hands back a pointer rather than a copy.

**Concurrent appends to a branchable collection no longer conflict.** Two writers used to delete the same head key, which regolith refuses at every isolation level, so one append was lost instead of forming a sibling head. A writer now only writes keys naming its own block and the head set is derived on read. Modelled in `proofs/tla/HeadSet.tla` and `proofs/lean/HeadSet/Core.lean`, with a RED configuration for each defect.

## Numbers

Same corpus through the full node on both branches: 14 sessions, 3,440 messages, 1,870 actions, 11.9 MiB of payload.

| | main (lark) | this branch (regolith) | |
|---|---:|---:|---|
| store on disk | 143.3 MiB | **113.2 MiB** | 21% smaller |
| `defra` binary | 53.51 MB | **42.96 MB** | 20% smaller |
| open | 67 ms | **25 ms** | |
| seed the corpus | 23.82 s | 23.84 s | unchanged |

Query latency is a wash and **regolith is slower on four of six cases**: it wins the two heaviest queries (62.5 ms vs 65.4 and 66.9 ms - because of durability and consistency guarantees) and loses the lighter ones by 0.6-2.9 ms. One machine, one run, five samples per case; the light-case differences are small enough that they need repeated runs before anyone should trust the direction.

## Upstream

Five gaps found doing this were fixed in regolith 0.1.2 (sourcenetwork/regolith#185): a cursor bug that made `scan_stream` return rows the caller had seeked past, byte-identical concurrent writes reported as conflicts, `Options::memory()`, reverse iteration over a transaction with pending writes, and a streaming-scan benchmark.

## Known

The `#[ignore]`d `identical_delta` test is gone: 0.1.2 fixes it. Superseded head keys are reclaimed by a separate transaction amortised over 16 appends, so a branchable collection carries a small constant of stale keys rather than one per mutation.

binary-size
store-size
open-and-seed
query-latency-p95
query-latency-p50

peak-rss


crates/cli/src/config/types.rs
```diff
@@ -142,13 +142,23 @@ impl std::str::FromStr for KeyringBackend {
...
+    /// aliases, so a config file written by an older binary keeps
+    /// loading instead of failing to parse on a value that no longer
+    /// selects anything.
     #[default]
-    Lark,
-    RocksDb,
-    #[serde(alias = "badger")]
-    Redb,
+    #[serde(
+        alias = "regolith",
```
iverc: do we still need other backends here?

crates/cli/src/config/types.rs
```diff
@@ -247,11 +254,12 @@ impl std::str::FromStr for DatastoreType {
 
     fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
         match s.to_lowercase().as_str() {
-            "lark" => Ok(DatastoreType::Lark),
-            "redb" | "badger" => Ok(DatastoreType::Redb),
+            // The old backend names all resolve to regolith rather than
+            // failing, so an existing config file keeps starting.
+            "regolith" | "lark" | "redb" | "badger" | "fjall" | "rocksdb" => {
```
iverc: same here

crates/cli/tests/config_types_tests.rs
```diff
@@ -98,23 +98,23 @@ fn test_log_output_display_roundtrip() {
...
         "redb".parse::<DatastoreType>().unwrap(),
-        DatastoreType::Redb
+        DatastoreType::Regolith
     );
     assert_eq!(
         "memory".parse::<DatastoreType>().unwrap(),
         DatastoreType::Memory
     );
     assert_eq!(
         "rocksdb".parse::<DatastoreType>().unwrap(),
```
iverc: same here

.github/workflows/release.yml
edjroz: The rename to unsuffixed artifacts missed the docker job's two consumers. The build job now uploads `defra-${{ matrix.platform }}` (line 97), but these download steps still ask for the old -full-suffixed names:

341:  name: defra-full-linux_amd64
346:  name: defra-full-linux_arm64

`download-artifact@v4 `fails on a missing exact name → the docker job fails → the release gate needs.docker.result == 'success' (line 373) blocks → a tag push produces no GitHub Release and no image. PR CI can't surface this because `release.yml` only triggers on v* tags, so the green checks here don't cover it.

crates/blockstore/src/lib.rs
```diff
@@ -205,7 +205,7 @@ impl<S: Store + 'static> Blockstore for DefraBlockstore<S> {
 
         // Populate cache on miss, convert Vec<u8> → Bytes
         if let Some(data) = result {
-            let bytes = Bytes::from(data);
+            let bytes = data;
             self.cache.lock().put(*cid, bytes.clone());
```
edjroz: This caches the borrowed `Bytes` from the zero-copy read, `Reader::get returns Bytes::from_owner(DbSlice)`, a refcount on the engine memory the value lives in: a 16 KiB SSTable block, or a whole memtable arena. 

Regolith's own docs warn a held slice "keeps that block resident even after the block cache evicts it… expensive to hold forever; call to_vec if the value must outlive the read." 

The LRU is capped at 1M entries (line 65), so the refcount so the 16Kib page can't be freed, which will keep rising until the LRU has 1M entries 100k entries grab ~1.6GiB.

we likely have to limit the cache size, or add some policies to handle it or this can get out of control.

crates/storage/src/backends/regolith/iterator.rs
```diff
@@ -0,0 +1,425 @@
...
+        });
+        let reverse_end = reverse_end.map(|end| match (end, self.end.as_ref()) {
+            // A seek past the range's end is clamped back into it.
+            (Some(end), Some(limit)) if end > *limit => Some(limit.clone()),
+            (end, _) => end,
+        });
+        let from = match &anchor {
+            Anchor::RangeStart => self.start.as_deref(),
+            Anchor::At(key) | Anchor::After(key) => Some(key.as_slice()),
+        };
```
edjroz: This lower bound needs the same start clamp the read-only arm gets from `floor()` (line 157): here the raw seek key goes straight into the scan, so a forward seek below start in a writing txn returns keys outside the iterator's bounds.

crates/storage/src/backends/regolith/store.rs
```diff
@@ -0,0 +1,265 @@
...
+        if self.inner.closed.swap(true, Ordering::AcqRel) {
+            return Ok(());
+        }
+        self.await_quiescence().await?;
+        self.inner
+            .db
+            .db()
+            .close()
+            .map_err(|error| Error::Backend(format!("failed to close regolith: {error}")))
+    }
```
edjroz, line 256: `close()` does not call `persist()`, which is what writes the database back to OPFS in mirror mode. We could call `self.persist().await?`to mitigate this.

edjroz (changes requested):
This is looking super solid.

I'm requesting changes focusing on the LRU item since it can lead to items not being freed from memory.

iverc (approved):

PR #1628, opened 2026-08-31
refactor(db): share auto-commit transaction lifecycle

### LINKED ISSUE #653: Extract shared transaction ceremony from auto_commit_mutator
## Summary

Transaction setup in `auto_commit_mutator/` is copy-pasted 30+ times across `create.rs`, `update.rs`, `delete.rs`, `delete_many.rs`, and `batch.rs`. Each file is 150-225 lines of near-identical ceremony with only the actual mutation logic varying.

## Details

The repeated ceremony follows this exact pattern in every file:

1. `get_collection_or_err()`
2. `new_txn(false)` → `map_err`
3. `txn.datastore()` → `map_err`
4. `IndexManager::from_collection()` → `map_err`
5. Actual mutation (the only part that varies)
6. `txn.blockstore()` → `map_err`
7. `txn.headstore()` → `map_err`
8. `write_document_blocks()` / `write_delete_block()`
9. `write_collection_block()` (if branchable)
10. `txn.commit()` → `map_err`
11. `emit_update_events()`
12. Error path: `txn.discard()`

Files have already drifted slightly — `create.rs` has embedding generation, `delete.rs` doesn't. This divergence will worsen over time.

## Expected Behavior

Each mutation operation should be a short closure (~10 lines) that only contains the varying logic. The transaction lifecycle ceremony should be defined once.

## Suspected Cause

LLM-generated code tends to duplicate patterns file-by-file rather than abstracting them. Higher-order async functions in Rust are harder to write than in Go (closure must be `Send`, future must be `Pin<Box<...>>`), so the LLM took the path of least resistance.

## Acceptance Criteria

- [ ] Extract a shared `with_mutation_txn()` (or similar) higher-order async function
- [ ] `create.rs`, `update.rs`, `delete.rs`, `delete_many.rs` each reduced to ~20-30 lines of mutation-specific logic
- [ ] All integration tests still pass (`cargo test -p integration-test`)
- [ ] No behavior change — same error messages, same commit/discard semantics

## Files

- `crates/db/src/auto_commit_mutator/create.rs`
- `crates/db/src/auto_commit_mutator/update.rs`
- `crates/db/src/auto_commit_mutator/delete.rs`
- `crates/db/src/auto_commit_mutator/delete_many.rs`
- `crates/db/src/auto_commit_mutator/batch.rs`
- `crates/db/src/auto_commit_mutator/helpers.rs`

iverc:
## Context

Single-document create, update, and delete repeated the same auto-commit transaction setup and commit/discard handling. Keeping those paths separate made lifecycle fixes easy to apply inconsistently.

## Changes

- centralize implicit mutation transaction creation and commit/discard finalization
- route create, update, and delete pre-commit failures through the shared lifecycle
- preserve operation-specific errors, no-op delete commits, ACP registration, and post-commit event timing
- reuse the transaction setup helper for existing multi-document entry points without changing their commit models

## Testing

- [x] `cargo test -p db --test write`
- [x] `cargo check -p db --target wasm32-unknown-unknown --no-default-features`

Closes #653

### DIFF
--- a/crates/db/src/write/autocommit/create.rs
+++ b/crates/db/src/write/autocommit/create.rs
@@ -44,9 +44,7 @@ impl<S: Store + 'static> AutoCommitMutator<S> {
         // No per-doc write guard for creates: the DocID is derived from the
         // genesis block inside the txn, so no identity exists to guard yet.
         // The DocID-mapping duplicate check is the gate.
-        let txn = self.db.new_txn(false).await.map_err(|e| {
-            query::error::QueryError::execution(format!("failed to create txn: {}", e))
-        })?;
+        let txn = self.new_mutation_txn().await?;
 
         // Acquire store views up front (dropped before commit); the mutation
         // itself runs in an async block so errors fall through to the discard.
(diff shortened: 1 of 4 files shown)


vertexclique:
Can we use existing transaction mechanism now regolith exposed @iverc ?

vertexclique (approved):

PR #1630, opened 2026-09-01
fix(db): simplify parent marker check

iverc:
## Context

The current stable clippy flags the manual optional-parent predicate in block head traversal. The equivalent standard-library helper keeps the intent direct and prevents the lint from blocking otherwise unrelated branches.

## Changes

- replace the negated `is_some_and` parent check with `Option::is_none_or`
- preserve the existing missing-marker and mismatched-parent behavior

## Testing

- [x] `cargo clippy --profile super-dev -p db --all-targets -- -D warnings`


mike7c2 (approved):
Unexpected De Morgan Theorum!

LGTM

edjroz (approved):

Issue #1510, opened 2026-08-17
Audit the Rust-specific harness accommodations


edjroz:
Sub-issue of #1392. Depends on #1395.

## Problem

The harness patch that makes the FFI oracle run is not one kind of edit. Reading all 14 shared-file changes, they
fall into three groups: registration plumbing that mirrors the existing `CClientType`, generic improvements that
never name Rust in the type system, and **Rust-specific accommodations**. The third group changes what the oracle
reports.

| File | Edit | Concern |
|---|---|---|
| `tests/integration/signature/utils.go` | `case state.RustFFIClientType:` hex-encodes the expected signature, because "Rust FFI returns signature.value as hex-encoded string" | A real Rust/Go divergence absorbed by the harness instead of reported. Go returns bytes. |
| `tests/integration/events.go` | event timeout `30 * eventTimeout` → `60 *` | **Not gated on client type** — changes Go's own test behaviour for every client |
| `tests/action/sync_branchable_collection.go` | pre-populates expected merge events for the FFI client, because "the FFI merge poller delivers events asynchronously… the Go client processes merges synchronously" | Compensates for a real behavioural difference |
| `tests/integration/utils.go` | extra `reconnectPeers(s)` after node restart | Compensates for missing reconnect behaviour |
| `tests/integration/db_setup.go` | disables Go's P2P host under rustffi; routes `PeerInfo` through the wrapper | Architecturally necessary, not a mask |

#1391 has an explicit vocabulary for this: a divergence where Go is wrong is coded `Go Status = Defect upstream`
and pinned by a test; a divergence where Rust is wrong is a parity gap and a 1.0 blocker. It warns that
"triaging them naively as 'Rust differs from Go, therefore parity gap' is the single most likely way to get this
wrong."

These accommodations are triaged **neither way**. They have lived in a fork-branch patch that no review process
looks at, and they turn failures into passes. A test that passes because the matcher was widened is worse
evidence than a test we wrote, because nothing in the output reveals it happened.

## Goal

- [ ] Every accommodation in `03-rust-accommodations.patch` enumerated: file, edit, and the divergence it absorbs
- [ ] Each one triaged per #1391 as `Defect upstream` or as a parity gap
- [ ] Each parity gap filed as its own 1.0 blocker sub-issue of #1392
- [ ] Each `Defect upstream` pinned by a test, so it fails when upstream converges
[...]

PR #1632, opened 2026-09-01
fix(query): encode commit signatures as base64

iverc:
## Context

Commit signatures are byte values on the wire, but the Rust GraphQL response encodes them as hexadecimal text while Go clients expect standard base64. The FFI compatibility revision also predates the corresponding Go client support.

This depends on #1630, which carries the shared clippy cleanup.

## Changes

- encode commit signature values as standard base64
- update the GraphQL signature regression expectation
- pin the Go FFI client to the audited compatibility revision
- make the missing-client error independent of a development branch name

## Testing

- [x] `cargo clippy --profile super-dev -p db -p ffi-test -p defra-version --all-targets -- -D warnings`
- [x] `cargo test --profile super-dev -p db --test query signature -- --nocapture`
- [x] `cargo test --profile super-dev -p ffi-test`
- [x] `cargo test --profile super-dev -p defra-version`

Refs #1510

### DIFF
--- a/crates/db/src/read/commits/conversion.rs
+++ b/crates/db/src/read/commits/conversion.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 
+use base64::Engine;
 use cid::Cid;
 use defra_core::block::{Block, Signature};
 use document::Document;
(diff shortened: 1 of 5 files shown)


mike7c2 (approved):
LGTM!

Substitution of hex for base64 is a simple change

The test asserts base64 is used from here forwards.

PR #1633, opened 2026-09-01
ci: run encryption and SourceHub Go compatibility tests

### LINKED ISSUE #1402: go-compat CI never runs the encryption and sourcehub cross-runtime tests
Sub-issue of #1393, found by the coverage inventory.

## Problem

The go-compat job selects tests by name (\`-- go_\`) across a fixed binary list (\`.github/workflows/ci.yml:585-588\`):

\`\`\`
--test basic --test query --test acp --test nac --test backup --test p2p --test identity
\`\`\`

`encryption` and `sourcehub` are separate `[[test]]` binaries and are absent from that list, yet both contain `go_*` cross-runtime tests. These seven are written, maintained, and never executed by the gate whose purpose is to run them:

- `go_rust_kms_interop`
- `go_to_rust_encrypted_p2p_replication`
- `go_rust_encrypted_p2p_replication_rust_writer`
- `go_rust_se_cross_node_rust_owner`
- `go_to_rust_se_cross_node`
- `go_circuit_breaker_trip_recovery`
- `go_policy_cache_ttl_expiry`

Cross-runtime encryption and searchable-encryption interop therefore has **zero** CI enforcement today, despite the tests existing.

## Fix

Add `--test encryption --test sourcehub` to the invocation.

Expect this to surface real failures rather than pass quietly — these have not run in CI, so nothing has been keeping them honest. Any failure found is itself a parity finding and should be triaged per #1391 (a post-v1.0.0 Go behavior is out of 1.0 scope; anything else is a blocker).

`sourcehub` additionally needs `sourcehubd` available on the runner, which is why it may have been omitted originally. If that dependency cannot be met in CI, the honest alternative is to mark those two tests `#[ignore]` with a reason, so the tree stops implying coverage it does not have.

## Also worth fixing while here

`--test basic` and `--test nac` are in the list but contain zero `go_*` tests, so they contribute nothing and make the invocation read as though basic CRUD and NAC were cross-runtime tested. Either remove them or, better, add the missing coverage — tracked in #1393.

iverc:
## Context

The Go compatibility job does not run the encryption or SourceHub integration suites, leaving those cross-runtime paths outside the merge gate. The tests also contain a developer-specific Go binary path instead of using the binary prepared by CI.

This depends on #1630, which carries the shared clippy cleanup.

## Changes

- add the encryption and SourceHub suites to the Go compatibility job
- resolve the Go `defradb` binary from `PATH`
- keep local runs configurable through the existing harness environment

## Testing

- [x] `cargo clippy --profile super-dev -p integration-test --all-targets -- -D warnings`

Closes #1402

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -471,6 +471,7 @@ jobs:
         run: >-
           cargo test --profile super-dev -p integration-test
           --test basic --test query --test acp --test nac --test backup --test p2p --test identity
+          --test encryption --test sourcehub
           -- --test-threads=1 go_
           --skip go_acp_link_collection_nonexistent_policy_reject
           --skip go_acp_link_collection_nonexistent_resource_reject
--- a/tools/integration-test/tests/encryption/cross_runtime_p2p.rs
+++ b/tools/integration-test/tests/encryption/cross_runtime_p2p.rs
@@ -12,17 +12,13 @@
 //! exercise Go↔Rust KMS wire-compat over libp2p: the reader must fetch the DEK
 //! from the writer's KMS over the `encryption` gossip topic (bare CBOR
 //! `FetchEncryptionKeyRequest`/`Reply`, ECIES-wrapped key blocks) and decrypt.
-//! These require the Go KMS binary at `GO_KMS_BINARY`.
+//! These require the Go `defradb` binary on `PATH`.
 
 use std::path::Path;
 use std::time::{Duration, Instant};
 
 use integration_test::identity::generate_identity;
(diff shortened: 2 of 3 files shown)


mike7c2 (approved):
This is good as is BUT - it could be taken further

Consider scenarios where we may be looking at a couple of different go binaries, to run them one after the other here we would need to twiddle our path back and forth or similar.

The previous env var approach allows us to do N runs quickly with different binaries by adding an extra env var to invocation. Achieving the same via path is a little more awakward.

If we don't intent to use these tests against go it's not so important.

edjroz (approved):

Issue #1356, opened 2026-08-10
Background loops block on uninterruptible sleeps: no in-flight cancellation, so shutdown stops tasks by abort instead of by signal


vertexclique:
## Summary

Background loops and long waits across `crates/p2p` block on uninterruptible
`tokio::time::sleep`, and the work they drive takes no cancellation parameter.
Shutdown therefore stops tasks by aborting the handle (or not at all) rather
than by signalling them, so a task can hold an `Arc` into the coordinator and
through it the storage handle for up to one full interval after `shutdown()`
returns.

#1309 is the acute instance of this. This issue tracks the general pattern,
because #1309's fix only covers two of the sites.

## The Go reference

Go does not have this pattern. Every long-lived loop selects on `ctx.Done()`,
and cancellation is threaded into the work itself.

`internal/db/p2p/replicator.go:445`, the direct sibling of our sweeps:

```go
func (p *P2P) handleReplicatorRetries(ctx context.Context) {
	for {
		select {
		case <-ctx.Done():
			return
		case <-time.After(retryLoopInterval):
			p.retryReplicators(ctx)   // ctx passed into the work
		}
	}
}
```

`internal/se/coordinator_retry.go:52` has the identical shape with a ticker.

Where the goroutine holds the store, Go additionally waits for it, and names
the hazard explicitly (`internal/db/db.go:70`):

```go
// recoveryWG tracks the background index build worker. Close waits on it before tearing down
// storage so the worker does not outlive the rootstore.
recoveryWG sync.WaitGroup
```

and in `Close()` (`db.go:466`):

```go
db.ctxCancel()
// Wait for the background recovery goroutine to exit before tearing down storage.
// The goroutine observes cancellation via db.ctx and will stop promptly.
db.recoveryWG.Wait()
```

So the Go contract is cancel, then wait. We currently do neither uniformly.

## Sites

All verified as production code (sites inside `#[cfg(test)]` modules are
excluded).

### 1. Unbounded loops with no cancellation arm

| Site | Interval | Notes |
|---|---|---|
| `crates/p2p/src/sync/coordinator/mod.rs:534` | 60s | `run_pending_dag_resync`. Covered by #1309. |
| `crates/p2p/src/sync/coordinator/mod.rs:550` | 2s | `run_pending_dag_retry_clock`. Covered by #1309. |
| `crates/p2p/src/host/handle.rs:942` | 50ms | |
[...]

PR #1634, opened 2026-09-01
fix(p2p): cancel pending sync work during shutdown

iverc:
## Context

Several P2P background tasks can outlive their owning operation. Partial setup failures can leave a swarm running, while pending resync and block-fetch tasks can continue after shutdown begins.

This is 1/2 in the P2P reliability stack and depends on #1630.

## Changes

- roll back partially initialized P2P resources when setup fails
- retain and cancel pending resync task handles
- cancel pending block fetches during shutdown
- add focused cleanup and cancellation coverage

## Testing

- [x] `cargo clippy --profile super-dev -p defra-core -p db -p embedded -p p2p --all-targets -- -D warnings`
- [x] `cargo test --profile super-dev -p p2p --lib`
- [x] `cargo test --profile super-dev -p embedded failed_p2p_setup_rolls_back -- --nocapture`

Refs #1356

### DIFF
--- a/crates/embedded/src/node_p2p.rs
+++ b/crates/embedded/src/node_p2p.rs
@@ -63,6 +63,21 @@ async fn shutdown_libp2p_host(handle: &p2p::P2PHostHandle, host_task: tokio::tas
     }
 }
 
+#[cfg(feature = "iroh")]
+async fn shutdown_iroh_endpoint(
+    transport: &p2p::iroh::IrohTransport,
+    endpoint_task: tokio::task::JoinHandle<()>,
+) {
+    use p2p::transport::P2PTransport;
+
+    if let Err(error) = transport.shutdown().await {
+        tracing::debug!(%error, "failed to signal Iroh endpoint during setup rollback");
+    }
+    if let Err(error) = endpoint_task.await {
+        tracing::debug!(%error, "Iroh endpoint task failed during setup rollback");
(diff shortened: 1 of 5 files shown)


crates/embedded/src/node_p2p.rs
```diff
@@ -182,15 +182,38 @@ where
...
+    });
+
+    // Receiver's re-arm loop (#1116 stage 2): dispatches due pending roots
+    // at a tight cadence. Sibling of the resync sweep above.
+    let coordinator_for_retry_clock = coordinator.clone();
+    let pending_dag_retry_task = tokio::spawn(async move {
+        coordinator_for_retry_clock
+            .run_pending_dag_retry_clock(std::time::Duration::from_secs(2))
+            .await;
+    });
```
edjroz, line 230: `setup_iroh` has this same pair at [490-503](crates/embedded/src/node_p2p.rs#L490-L503), still above its `PubsubKeyTransport::new` at [517](crates/embedded/src/node_p2p.rs#L517), so that `?` orphans both tasks holding `Arc<SyncCoordinator>` → `Arc<S>` and the 2s retry clock pins the store open:

```
failed to open regolith: I/O error: database directory is already locked for read-write access: .../data.regolith/LOCK
```

edjroz (changes requested):

PR #1636, opened 2026-09-01
test(crdt): compare document block CIDs with Go

### LINKED ISSUE #1416: Add cross-runtime block CID equivalence tests for ordinary document lifecycle
Sub-issue of #1393. Closes the highest-value gap the coverage inventory found.

## Why this one first

#1405 reports that Rust computes a field delta's priority from a **document-wide** headstore max (`crates/db-blocks/src/write.rs:221-230`, scanning `HeadstoreDocKey::document_prefix`), while Go derives each field's height from **that field's own** head-set (`internal/core/block/store.go:85-93`). Priority is DAG-CBOR-encoded into the delta and hashed into the block CID.

If that holds, the same document operations produce different block CIDs on the two implementations — which breaks P2P interop wholesale, since content addressing is the entire basis of block exchange. It has gone unnoticed because `--test basic` contains **zero** `go_*` tests.

## The test

In the dual-binary suite, driving both the Go and Rust binaries through identical operations and comparing block CIDs directly:

1. Create a two-field document. Compare the composite CID and both field block CIDs.
2. Update field A only. Compare again.
3. **Update field B only** — the field untouched since create. Compare again.

Step 3 is the discriminating one. Per #1405's trace, Go should stamp field B's block with priority 2 (its own last head was 1) while Rust stamps 3 (document-wide max). Different priority, different CID.

Then extend to the cases most likely to diverge for the same reason:
- alternating A/B/A/B updates over several rounds
- a counter field alongside LWW fields (counter priority feeds the same path)
- delete and recreate

## Assertions

Compare **actual CID strings**, not just document values. A value-level assertion passes even when the DAG has diverged, which is exactly how this class hides. `_commits { cid height }` gives both per block.

## Where

`tools/integration-test/tests/basic/`, named `go_*` so the go-compat job picks it up. `basic` is already in the CI binary list (`.github/workflows/ci.yml:586`) and currently contributes nothing to that job — this would make its inclusion meaningful.

Would have caught: **#1405**, and would catch any future regression in priority, head selection, or delta encoding.

iverc:
## Context

Rust and Go can accept the same document operation while producing different immutable blocks or CIDs. Existing integration coverage checks behavior within each runtime but does not compare the encoded outputs directly.

This is 1/3 in the data parity stack and depends on #1630.

## Changes

- run equivalent document mutations against Rust and Go nodes
- compare generated block bytes and CIDs across runtimes
- cover representative scalar, optional, relation, and collection values

## Testing

- [x] `cargo clippy --profile super-dev -p integration-test --all-targets -- -D warnings`

Closes #1416

### DIFF
--- a/tools/integration-test/tests/basic.rs
+++ b/tools/integration-test/tests/basic.rs
@@ -4,6 +4,8 @@ mod batch_mutations;
 mod collection_delete_4657;
 #[path = "basic/collection_management.rs"]
 mod collection_management;
+#[path = "basic/cross_runtime_cids.rs"]
+mod cross_runtime_cids;
 #[path = "basic/document_lifecycle.rs"]
 mod document_lifecycle;
 #[path = "basic/multi_collection.rs"]
--- a/tools/integration-test/tests/basic/cross_runtime_cids.rs
+++ b/tools/integration-test/tests/basic/cross_runtime_cids.rs
@@ -1,0 +1,216 @@
+use integration_test::{DefraClient, TestCluster};
+use serde_json::Value;
+
+#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
+struct Commit {
+    field_name: String,
+    height: i64,
+    cid: String,
+}
+
+fn doc_id(response: &Value, mutation: &str) -> String {
+    response[mutation][0]["_docID"]
+        .as_str()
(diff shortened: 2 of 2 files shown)


mike7c2 (approved):
🔥 LGTM!

PR #1638, opened 2026-09-01
fix(http): match permissions to axum route templates

### LINKED ISSUE #1497: route_permissions table is dead for every parameterized route: it uses :name where axum 0.8 MatchedPath yields {name}

iverc:
## Context

Permission rules use colon-prefixed path parameters while axum reports matched paths with brace-prefixed parameters. Requests to parameterized routes therefore miss their configured permission entry.

This is 1/2 in the HTTP routing stack and depends on #1630.

## Changes

- express permission paths using axum's matched-route syntax
- cover authorization lookup for parameterized collection and document routes
- update the route coverage expectation to include the corrected permission paths

## Testing

- [x] `cargo clippy --profile super-dev -p defra-http --all-targets -- -D warnings`

Closes #1497

### DIFF
--- a/crates/http/src/route_permissions.rs
+++ b/crates/http/src/route_permissions.rs
@@ -27,7 +27,8 @@ pub enum RoutePermission {
 /// Look up the permission requirement for a route.
 ///
 /// The `path` parameter is the Axum `MatchedPath` template string
-/// (e.g., `/api/v0/collections/:name/:docID`). The `method` is the HTTP method.
+/// (e.g., `/api/v0/collections/{name}/document/{docID}`). The `method` is the
+/// HTTP method.
 ///
 /// Unknown routes return `Required(DocumentRead)` as a safe default:
 /// when NAC is enabled, this blocks unauthenticated access.
@@ -66,10 +67,10 @@ pub fn route_permission(path: &str, method: &Method) -> RoutePermission {
         // Transactions (permissions enforced per-operation within tx)
         // =====================================================================
         "/api/v0/tx" => RoutePermission::IdentityOnly,
-        "/api/v0/tx/:id" => RoutePermission::IdentityOnly,
-        "/api/v0/tx/:id/lens" => RoutePermission::Required(NodePermission::CollectionPatch),
-        "/api/v0/tx/:id/collections" => RoutePermission::Required(NodePermission::CollectionGet),
-        "/api/v0/tx/:id/schema" => RoutePermission::Required(NodePermission::CollectionPatch),
+        "/api/v0/tx/{id}" => RoutePermission::IdentityOnly,
+        "/api/v0/tx/{id}/lens" => RoutePermission::Required(NodePermission::CollectionPatch),
+        "/api/v0/tx/{id}/collections" => RoutePermission::Required(NodePermission::CollectionGet),
+        "/api/v0/tx/{id}/schema" => RoutePermission::Required(NodePermission::CollectionPatch),
 
         // =====================================================================
         // Collections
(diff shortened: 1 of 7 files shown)


mike7c2 (approved):
LGTM

Clean, well supported by tests

edjroz (approved):

PR #1639, opened 2026-09-01
test(schema): enforce validation parity with Go

### LINKED ISSUE #1420: Add cross-runtime schema validation parity tests (same SDL accepted/rejected)
Sub-issue of #1393.

## Gap

No cross-runtime assertion that the two implementations accept and reject the **same schemas**. Schema validation is a correctness boundary: a schema Rust accepts and Go rejects is a state the database will then have to live with forever.

Filed defects here: #1403 (Rust accepts a patch adding a non-nillable field to a populated collection; Go rejects it) and #1407 (the WASM client skips definition validation entirely). #1404 tracks the broader finding that Rust registers 25 of Go's 39 definition validators.

## The test

Table-driven, both binaries, same SDL in, compare accept/reject:

- One case per Go validator — the list is in `internal/db/definition_validation.go:135-187`. Each should have a schema that violates it and an assertion that **both** implementations reject it.
- The same for patches, not just initial schema creation: #1403 is a patch-path defect, and the patch path has its own validator list in Go.
- Assert on **outcome, not message text**. Error strings will differ and that is acceptable; accept-vs-reject is the contract.
- Include the passing cases too — a validator that is too strict in Rust would reject a schema Go accepts, which is equally a divergence and easier to miss.

This is the natural mechanical companion to the #1404 audit: as each validator there is confirmed or dismissed, it gets a row here.

## Where

`tools/integration-test/tests/basic/` or a new `schema` area, `go_*`-named. If new, add it to the go-compat CI list (#1402).

iverc:
## Context

Schema validation is implemented independently in Rust and Go, but the shared integration suite does not compare their accepted and rejected schema states. Small validator differences can therefore ship without a parity signal.

This is 1/2 in the schema parity stack and depends on #1630.

## Changes

- add stateful cross-runtime schema validation cases
- compare table-driven valid and invalid schema inputs
- align validator behavior exposed by the new parity cases

## Testing

- [x] `cargo clippy --profile super-dev -p schema -p integration-test --all-targets -- -D warnings`
- [x] `cargo test --profile super-dev -p schema`

Closes #1420

### DIFF
--- a/tools/integration-test/tests/basic/cross_runtime_schema_validation.rs
+++ b/tools/integration-test/tests/basic/cross_runtime_schema_validation.rs
@@ -1,0 +1,381 @@
+use std::fmt::Display;
+
+use integration_test::{DefraClient, TestCluster};
+
+#[path = "cross_runtime_schema_validation/stateful.rs"]
+mod stateful;
+
+struct AddCase {
+    validator: &'static str,
+    sdl: &'static str,
+    accepted: bool,
+}
+
+struct PatchCase {
+    validator: &'static str,
+    sdl: &'static str,
+    patch: &'static str,
+    accepted: bool,
+}
+
+fn assert_outcome<RT, RE, GT, GE>(
+    case: &str,
+    expected: bool,
+    rust: &Result<RT, RE>,
+    go: &Result<GT, GE>,
+) where
+    RE: Display,
+    GE: Display,
+{
+    let rust_accepted = rust.is_ok();
+    let go_accepted = go.is_ok();
+    assert_eq!(
+        rust_accepted,
+        go_accepted,
+        "{case}: Rust and Go outcomes differ; Rust={:?}, Go={:?}",
+        rust.as_ref().err().map(ToString::to_string),
+        go.as_ref().err().map(ToString::to_string),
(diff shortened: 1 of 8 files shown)


tools/integration-test/tests/basic/cross_runtime_schema_validation.rs
```diff
@@ -0,0 +1,381 @@
...
+        rust.as_ref().err().map(ToString::to_string),
+        go.as_ref().err().map(ToString::to_string),
+    );
+    assert_eq!(
+        rust_accepted,
+        expected,
+        "{case}: unexpected shared outcome; Rust={:?}, Go={:?}",
+        rust.as_ref().err().map(ToString::to_string),
+        go.as_ref().err().map(ToString::to_string),
+    );
```
edjroz, line 45: I think we aren't evaluating errors here, which could lead to negative cases passing. this is because both sides are reduced to `is_ok()` without checking for errors.

We could add an enum `Outcome` to be able to discern which cases would have an error and evaluate both sides? I'll leave this as a comment you can take it or leave it.

edjroz (approved):

PR #1642, opened 2026-09-01
test(http): enforce Go wire contract parity

### LINKED ISSUE #1419: Add cross-runtime HTTP route table and response shape equivalence tests

iverc:
## Context

Route coverage confirms that an endpoint exists, but not that Rust and Go return compatible statuses, JSON shapes, or GraphQL metadata for the same request.

This is 2/2 in the HTTP contract stack and depends on #1637.

## Changes

- execute representative HTTP requests against Rust and Go nodes
- compare response status and JSON structure across runtimes
- cover concrete paths derived from the pinned route contract
- align GraphQL response metadata exposed by the parity cases

## Testing

- [x] `cargo clippy --profile super-dev -p cli -p defra-http -p query -p integration-test --all-targets -- -D warnings`

Closes #1419

### DIFF
--- a/crates/http/src/handlers/graphql/meta.rs
+++ b/crates/http/src/handlers/graphql/meta.rs
@@ -9,7 +9,7 @@ use crate::router::{AppState, NodePermission};
 
 /// Health check response.
 pub async fn health_check() -> impl IntoResponse {
-    (StatusCode::OK, "Healthy")
+    Json("Healthy")
 }
 
 /// Version endpoint handler.
--- a/crates/query/src/executor.rs
+++ b/crates/query/src/executor.rs
@@ -85,7 +85,6 @@ impl QueryRequest {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct QueryResponse {
     /// Query result data (null if errors occurred).
-    #[serde(skip_serializing_if = "Option::is_none")]
     pub data: Option<JsonValue>,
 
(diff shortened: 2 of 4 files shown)


mike7c2 (approved):
LGTM

PR #1643, opened 2026-09-01
fix(http): normalize trailing slashes before routing

### LINKED ISSUE #1496: Trailing slashes are not normalized, so /api/v0/collections/ 404s where Go serves it
Found while fixing #1411, and measured against a router built exactly as Rust builds it. Triage procedure: #1391.

## Divergence

Go normalizes trailing slashes before routing (`http/handler.go:105-111`):

```go
// Normalize trailing slashes so that, for example, `/collections` and
// `/collections/` resolve to the same route instead of the latter missing
// chi's router and returning a bare 404. StripSlashes rewrites the routing
// path in place (no redirect), preserving the request method and body.
mux.Use(middleware.StripSlashes)
```

Rust applies no such layer, and axum does not normalize by default. Measured against a router with the same nesting shape:

```
POST /api/v0/view          -> 200 OK
POST /api/v0/view/         -> 404 Not Found
POST /api/v0/views         -> 200 OK
POST /api/v0/views/        -> 404 Not Found
```

## Consequence

Any client that appends a trailing slash gets a 404 on Rust and a normal response on Go. Go's comment shows this is deliberate on their side, and specifically that they consider the bare 404 the wrong answer.

Rust is at least self-consistent today: every route behaves the same way, so this is a uniform difference rather than a per-route surprise. The #1411 fix added a test (`a_trailing_slash_behaves_the_same_on_both_paths` in `crates/http/tests/go_route_paths.rs`) pinning that the new Go-compatible mounts do not diverge from the Rust ones here, so this issue is the only remaining gap.

## Fix

`tower-http` provides `NormalizePathLayer::trim_trailing_slash()`, and `tower-http` is already a dependency (`crates/http/Cargo.toml:49`).

One thing to get right: the layer must be applied so it rewrites the path **before** routing, which for axum means wrapping the finished `Router` in the layer rather than adding it with `Router::layer` (a `Router::layer` runs after routing has already failed). The usual shape is:

```rust
ServiceBuilder::new()
    .layer(NormalizePathLayer::trim_trailing_slash())
    .service(router)
```

Two things to verify when it lands:

- `MatchedPath` still resolves to the registered template, since `route_permissions.rs` depends on it. A path rewritten before routing should match normally, but it must be proven by a test rather than assumed.
- The root routes (`/health-check`) and both API version mounts are covered, since Go applies `StripSlashes` to the outermost mux, ahead of everything.

iverc:
## Context

Equivalent HTTP paths with a trailing slash currently bypass normal route matching. Normalization must happen before routing so method handling, request bodies, and matched-path permissions retain their existing semantics.

This is 2/2 in the HTTP routing stack and depends on #1638.

## Changes

- normalize non-root trailing slashes before axum route matching
- preserve request methods and bodies through normalization
- keep normalized parameterized routes connected to their permission entries
- cover root, API, collection, and method-sensitive paths

## Testing

- [x] `cargo clippy --profile super-dev -p defra-http --all-targets -- -D warnings`

Closes #1496

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -145,7 +145,7 @@ graphql-parser = "0.4"
 # HTTP
 axum = { version = "0.8.9", features = ["ws"] }
 tower = { version = "0.5", features = ["timeout", "limit"] }
-tower-http = { version = "0.6", features = ["cors", "trace"] }
+tower-http = { version = "0.6", features = ["cors", "normalize-path", "trace"] }
 hyper = "1.9.0"
 
 # Error handling
--- a/crates/http/src/server.rs
+++ b/crates/http/src/server.rs
@@ -11,8 +11,9 @@ use axum::Router;
 use tokio::net::TcpListener;
 use tower::limit::ConcurrencyLimitLayer;
 use tower::timeout::TimeoutLayer;
-use tower::ServiceBuilder;
+use tower::{Layer, ServiceBuilder};
 use tower_http::cors::CorsLayer;
+use tower_http::normalize_path::NormalizePathLayer;
 use tower_http::trace::TraceLayer;
 
 use query::executor::QueryExecutor;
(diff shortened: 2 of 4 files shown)


mike7c2 (approved):
LGTM

edjroz (approved):

PR #1644, opened 2026-09-01
test(query): cover cross-runtime lens migrations

### LINKED ISSUE #1421: Add cross-runtime lens migration tests, including the HTTP patch registration path
Sub-issue of #1393.

## Gap

Lens migrations have Rust-side tests (`query/lens`, `query/lens_persistence`) but no cross-runtime coverage, and two filed defects live on that path:

- **#1415** — `PatchCollection` parses the `Migration` payload at `crates/http/src/handlers/collections.rs:150` and never reads it. Lens migrations are simply never registered when patching a schema. Go calls `db.PatchCollection(ctx, message.Patch, message.Migration, ...)` (`http/handler_store.go:127-158`).
- **#1406** — the migration path's JSON-to-NormalValue coercion turns a Blob field's hex text into `NormalValue::Bytes` of the ASCII hex characters, where Go keeps Blob as a String throughout (`client/document.go:272`).

Both are on the same surface, and neither would be caught by a Rust-only lens test, because a Rust-only test can be self-consistently wrong.

## The test

Both binaries, same schema, same lens transform, same documents:

- Register a migration **through the HTTP patch path specifically**, then assert it is actually registered and applied. #1415 is precisely a case where the API accepts the request and does nothing, so assert on observable effect rather than on the request succeeding.
- Migrate documents containing every scalar kind — **Blob especially**, given #1406, plus JSON, DateTime, and nillable variants — and compare migrated field values across implementations.
- Compare the resulting document CIDs, not only values: a migration that produces the right value via the wrong encoding still forks the DAG.
- Cover the reindex-after-migration path (`crates/db/src/migration/reindex.rs`), which is a second consumer of the same coercion helper.

The existing WASM lens binary used by `p2p_iroh` tests may be reusable as the transform.

## Where

`tools/integration-test/tests/query/`, `go_*`-named — `query` is already in the go-compat CI list.

iverc:
## Context

Schema validation parity does not cover the data transformation performed by lens migrations. Rust and Go need to produce the same migrated rows, commit history, and document identifiers for equivalent schema patches.

This is 2/2 in the schema parity stack and depends on #1639.

## Changes

- run equivalent HTTP lens migrations against Rust and Go
- compare migrated scalar values and query results
- verify commit history and document CID parity after migration
- cover representative scalar type transformations

## Testing

- [x] `cargo clippy --profile super-dev -p schema -p integration-test --all-targets -- -D warnings`
- [x] `cargo test --profile super-dev -p schema`

Closes #1421

### DIFF
--- a/tools/integration-test/tests/query.rs
+++ b/tools/integration-test/tests/query.rs
@@ -40,6 +40,8 @@ mod join_order_1596;
 mod json_missing_key;
 #[path = "query/lens.rs"]
 mod lens;
+#[path = "query/lens_cross_runtime.rs"]
+mod lens_cross_runtime;
 #[path = "query/lens_persistence.rs"]
 mod lens_persistence;
 #[path = "query/lens_reindex_secondary_index_979.rs"]
--- a/tools/integration-test/tests/query/lens_cross_runtime.rs
+++ b/tools/integration-test/tests/query/lens_cross_runtime.rs
@@ -1,0 +1,285 @@
+use std::process::Command;
+
+use integration_test::{DefraClient, TestCluster};
+use serde_json::{json, Value};
+
+struct ScalarCase {
+    collection: &'static str,
+    kind: &'static str,
+    required: &'static str,
+    other_required: &'static str,
+    optional: &'static str,
+    expected_required: Value,
+    expected_optional: Value,
+    indexed: bool,
+}
+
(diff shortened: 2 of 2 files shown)


mike7c2 (approved):
LGTM

edjroz (approved):

PR #1651, opened 2026-09-01
fix(query): resolve a subscription's root through fragments

mike7c2:
A subscription whose root selection is a fragment opens a live event stream
and then delivers nothing, for as long as it stays connected.

## The bug

The spec allows a subscription exactly one root field (5.2.3.1), and that rule
applies to the selection **after** fragments are expanded. So both of these are
legal GraphQL, selecting one field each:

```graphql
subscription { ...F }  fragment F on Subscription { User { name } }
subscription { ... on Subscription { User { name } } }
```

`subscription_to_scoped_query` required a bare `Selection::Field` and rejected
them. It runs once per event, from inside an already-open stream, where
`graphql_sse` can only `tracing::warn!` and `continue`. So the caller gets a
healthy `200 text/event-stream`, the connection stays open, and no event is
ever delivered — with the only evidence in the server's log.

Measured against an unpatched node, writing one matching document:

```
inline fragment root     0 events
named fragment root      0 events
plain root (control)     1 event
```

Both spellings take that path, so this is a class of input rather than one
exotic query. Anything that subscribes at start-up and then waits — an edge
device being the obvious case — passes a smoke test and goes deaf in the field.

## The fix

`subscription_to_scoped_query` resolves the root before scoping it, expanding
fragment spreads and inline fragments the way `query_parse::parser` already
does: the same cycle detection, and inline type conditions ignored for the same
stated reason, that DefraDB has no interface or union types. Fragment
definitions are collected in a first pass, because a definition may appear
after the operation that spreads it.

The stream is also validated before it opens. Scoping already refused a root it
could not resolve, but only from inside the stream. `validate_subscription`
runs the same resolution at subscribe time, so a document that can never be
scoped is answered while the caller is still listening. In practice the parser
rejects such documents earlier — this is the guard that keeps a future
disagreement between the two expansions from reintroducing the silent stream,
and `validation_agrees_with_scoping` asserts they match across the known
shapes.

The FFI subscription path had the same shape — a per-event scoping below a
subscribe that could not fail — so it takes the same check and returns the
error rather than handing back a handle that never yields.

A root resolving to no field or to several is still refused. The parser catches
that case and reports it with its reason, for plain and fragment spellings
alike; `a_root_resolving_to_two_fields_is_reported` pins it so the two paths
stay distinct.

## Reviewing

Two commits, deliberately ordered so the failure is reproducible:

| | on `test(query)` | on `fix(query)` |
|---|---|---|
| `a_named_fragment_root_delivers_like_a_plain_one` | **FAILS** | ok |
| `an_inline_fragment_root_delivers_like_a_plain_one` | **FAILS** | ok |
| `a_root_resolving_to_two_fields_is_reported` | ok | ok |
| `single_root_subscription_delivers_as_a_control` | ok | ok |

**The first commit fails by design.** If CI runs per-commit it will go red
there; the branch tip is green.

The two passing tests are what make the two failures readable. The control runs
the same apparatus against a subscription the node accepts, so "0 events"
cannot be dismissed as a broken test; the two-field case pins the neighbouring
path that already behaved correctly.

Verified by rebuilding the CLI at each commit and running against a live node,
not by reasoning about the diff.

## Checks

- `cargo test -p integration-test --test query` — no non-Go failures
- `cargo test --workspace` — 222 test binaries, 0 failures
- `cargo clippy --all -- -D warnings`, `cargo fmt --all --check` — clean

## Not in this PR

`subscription_to_scoped_query` re-parses the query *text* on every event, for
every open subscription. Parsing once at subscribe time would be a worthwhile
follow-up: measured on a debug node, commit-to-frame latency goes from 0.8 ms
with no other subscribers to 3.9 ms with 100 idle ones, because every
subscription re-runs its own query for every update regardless of whether it
matches.


vertexclique:
The fix works for every root spelling I threw at it (named, inline, nested spread), and validating up front on both the HTTP and FFI paths is the right call: turning a silent dead stream into a 400 is the whole point.

One edge case: `collect_root_fields` counts raw fields rather than grouping by response key, so it rejects some legal GraphQL. Checked against `bfdc96c2`:

```
subscription { ...A ...B }
fragment A on Subscription { User { name } }
fragment B on Subscription { User { name } }
  => Err("subscription must have exactly one root field, found 2")

// diamond, A -> C and B -> C
subscription { ...A ...B }
fragment A on Subscription { ...C }
fragment B on Subscription { ...C }
fragment C on Subscription { User { name } }
  => Err("subscription must have exactly one root field, found 2")
```

5.2.3.1 runs the check over `CollectFields`, which groups by response key, so two spreads selecting the same field are one entry and both of these are valid. Counting distinct keys (`alias` else `name`) instead of `fields.len()` covers it.

Not a regression, and arguably a strict improvement: before this PR those same queries opened a stream and delivered nothing, whereas now they at least fail loudly. So a follow-up rather than something to hold the PR for.

Cycle detection looks right, by the way. The `visiting.remove()` after recursion is what keeps the diamond above from being misreported as circular, and a real `A -> B -> A` is still caught.

mike7c2:
Both findings verified against `bfdc96c2` before acting on them.

**@coderabbitai — the recursion bound: valid, and worse than a lint.** Fixed in `97a82911`.

It is not a theoretical exhaustion. A chain of *distinct* fragments never repeats a name, so cycle detection never fires, and each spread costs a stack frame:

```
chain len   1000: ok
chain len  10000: thread has overflowed its stack / SIGABRT
```

Root expansion now refuses a chain deeper than 32 — far above any legitimate document, far below what exhausts the stack — while keeping the circular and unknown-fragment errors intact. Tests cover both the refusal and that a chain within the limit still resolves.

**@vertexclique — the response-key grouping: correct, reproduced exactly.** Both your cases fail on `bfdc96c2` as you describe, and `collect_root_fields` counting `fields.len()` instead of distinct response keys is the cause.

Agreed it is a follow-up rather than a blocker, and for a reason worth stating: counting distinct keys fixes the *validation* but not the *scoping*. When two spreads select the same key with different sub-selections — `...A` = `User { name }`, `...B` = `User { age }` — `CollectFields` merges them to `User { name age }`. Counting keys and then taking the first field would silently drop `age`, which is worse than today's loud rejection. So the real fix is a merge, not a count, and it deserves its own change.

Thanks for checking the cycle behaviour too — the `visiting.remove()` placement is exactly what separates the diamond from a real `A -> B -> A`.

---

**Separately, and more seriously than either finding: `main` has the same unbounded recursion on a path that has nothing to do with subscriptions.**

`parse_selection_to_selects` recurses through fragment spreads the same way, and every GraphQL request reaches it — `graphql_required_permission` parses the document to decide the required permission. A plain query is enough:

```
query chain 1000: body 34803 bytes
query chain 1000: thread has overflowed its stack / SIGABRT
```

A 34 KB request body aborts the process, and it is *more* fragile than the subscription path (1,000 is already fatal there, versus 10,000 here). That is a pre-existing availability bug in the parser, not something this PR introduces, and it wants its own fix rather than being folded in here. Flagging it now rather than after this merges — happy to raise it separately.

vertexclique (approved):

Issue #1653, opened 2026-09-01
Make `p2p/libp2p-transport` optional in `crates/ffi` and `crates/embedded`


edjroz:
## Context

#1345's original prerequisite list included making `p2p/libp2p-transport` optional in ffi/embedded (the db-merge half already exists as `db`'s optional `p2p` feature). The lean-embedded variant shipped without it because that variant keeps libp2p; this issue carries the remaining hygiene so the whole workspace matches the defra-node graph contracts (#1398-#1400): transports are features, never hard-wired deps.

## Scope

- `crates/ffi/Cargo.toml:76` and `crates/embedded/Cargo.toml:24,30`: move `p2p/libp2p-transport` (and the adapter's `libp2p` feature) behind a `libp2p` feature, default on, mirroring `defra-node`.
- cfg-gate libp2p types in `embedded`'s public API (`Libp2pConfig`, `ManagedP2PSystem`, libp2p re-exports) and their ffi consumers.
- `db` merge path needs only libp2p-identity + multiaddr (shim-proven, binary-size round 2); no db change expected.
- Extend `.github/scripts/assert-defra-node-graph.sh` with no-libp2p combos for ffi/embedded.

## Non-goals

- No new release artifact. An iroh-only FFI variant remains undecided (#1337 named no consumer for it); this issue only makes one possible.
- No change to any shipped variant's graph: `libp2p` is default-on, and `--no-default-features` release invocations add it explicitly.

## Acceptance criteria

- [ ] `cargo check -p ffi --no-default-features --features native` (lean) and default builds unchanged.
- [ ] A no-libp2p ffi/embedded combo compiles and its graph contract asserts pass.
- [ ] Existing release artifacts byte-identical (release size report ~0 delta).

---

Part of #1327. Follow-up severed from #1345's prerequisite list by design decision (D2).

Issue #1345, opened 2026-08-07
Mobile-lean FFI variant


edjroz:
## Context

A libp2p-only, no-lens, sourcehub-optional FFI variant lands around 20-21 MiB, roughly 35% below the shipped dylib `[hypothesis, from individually measured components]`:

- iroh 6.9 MiB `[measured]`
- wasmtime/lens ~2 MiB `[measured via cargo-bloat]`
- sourcehub live code up to ~3.5 MiB `[bounded by the cli release-full delta]`

## Prerequisites

- Make `sourcehub` optional: `crates/ffi/Cargo.toml:76`, `crates/embedded/Cargo.toml:29`.
- Make `p2p/libp2p-transport` optional: `crates/ffi/Cargo.toml:78`, `crates/embedded/Cargo.toml:25`, `crates/db-merge/Cargo.toml:69`. The p2p crate already has the feature.

## Scope boundary

The default Go-interop variant is unchanged. The lean variant knowingly gives up SourceHub ACP, lens migrations, and one transport, and that must be visible to consumers.

## Acceptance criteria

- [ ] New variant builds in CI alongside the existing variants, which stay unchanged.
- [ ] Measured size verifies the 20-21 MiB estimate **before** any public claim.
- [ ] Capability differences documented in the release matrix.

---

Part of #1327. Effort: M. **Blocked by #1337.** Informed by #1340.

jackzampolin:
**Triage: `Post-1.0`.** Procedure: #1391.

`Go Status: Not in Go`

Explicitly scoped to leave the default Go-interop FFI variant unchanged ('The default Go-interop variant is unchanged'); it adds a new opt-in lean variant that trades away SourceHub ACP, lens migrations, and one transport. Blocked by #1337 (undecided) and informed by #1340 (unmeasured baseline).

<sub>Go reference: Go v1.0.0's cbindings/ directory is a single, undifferentiated set of C bindings (node_new.go, collection_add.go, p2p_*.go, etc.) with no lean/full variant split, no iroh, no lens/wasmtime, and no SourceHub-optional toggle to mirror.</sub>

iverc:
Unblocked by #1337: keep the existing Iroh-capable iOS XCFramework unchanged and make this an additive lean mobile variant for consumers that do not need Iroh. Embedded Rust users continue to opt into Iroh through crate features rather than FFI artifacts.

PR #1660, opened 2026-09-02
feat(ffi): make sourcehub an optional, default-on feature

edjroz:
Part of #1345. First of a three-PR stack (this → lean release artifacts → #1653); merge bottom-up.

## Problem

The ffi crate hard-wires `sourcehub`, both directly and through its `embedded` dependency line, so no FFI build can exclude on-chain ACP. The lean variant (#1345) needs it optional, while the default Go-interop build must keep an identical dependency graph and C ABI.

## What Changed

- `sourcehub` is now a default-on ffi feature (`["dep:sourcehub", "embedded/sourcehub"]`); the hard-wired dep lines are gone.
- SourceHub internals are cfg-gated at the single `resolve_embedded_config` seam; struct fields and the mobile JSON schema stay unconditional.
- Configuring SourceHub in a build without the feature fails with an explicit error naming the missing feature (contract test committed red first).
- `assert-defra-node-graph.sh` gains lean-ffi asserts (libp2p in; iroh/wasmtime/sourcehub/cosmrs out) plus default-ffi presence asserts.
- `release.yml` selections name `sourcehub` explicitly so every shipped artifact keeps its exact graph; ci.yml's minimal group covers the lean selection.

## Verification

- [x] `cargo tree` byte-identical for the default, generic-release, and Apple selections (before vs after)
- [x] cbindgen header unchanged (empty diff, 2063 lines)
- [x] `cargo test -p ffi` 167 passed; lean selection 168 passed incl. the refusal test
- [x] `cargo clippy -D warnings` clean on both selections; `cargo fmt` clean
- [x] `assert-defra-node-graph.sh` passes with the new asserts (red before the change)
- [x] `Cargo.lock` untouched


crates/ffi/Cargo.toml
```diff
@@ -16,9 +16,10 @@ bench = false
 crate-type = ["staticlib", "cdylib", "rlib"]
 
 [features]
-default = ["native", "wasmtime-runtime"]
+default = ["native", "wasmtime-runtime", "sourcehub"]
```
iverc, line 19: we should rename it to "vera" here too

iverc (approved):
@edjroz one comment regarding Vera rename, LGTM otherwise

PR #1681, opened 2026-09-03
fix(sourcehub): Vera proto package, bech32 prefix, and a separate Orbis service key

vertexclique:
Vera (github.com/sourcenetwork/vera, formerly SourceHub) renamed its proto package from `sourcehub.*` to `vera.*` and its bech32 account prefix from `source` to `vera`. The SourceHub client on main still speaks the old names, so every ACP transaction and query against a current Vera chain is unroutable and every derived address is wrong.

This branch makes `--document-acp-type source-hub` work against Vera again, and splits the key a node authenticates to an Orbis ring with from the key it signs chain transactions with.

## What changed

**Wire names.** `/vera.acp.*` message type URLs, the `/vera.acp.Query/VerifyAccessRequest` ABCI path, the `/sourcenetwork/vera/acp/...` REST paths, and the `vera` bech32 prefix. The event decoder matches `vera.acp.` message types for cache invalidation.

**A separate Orbis service key: `--signer-orbis-identity`.** The ring authenticates a Sign request with a JWT whose algorithm follows the key type and accepts EdDSA only, while `--identity` is also the key that signs Vera transactions and must be secp256k1. One key cannot serve both roles: a node with a secp256k1 identity presents an ES256K token and the ring refuses it with `unknown variant ES256K`. The new flag carries the ed25519 key for the ring and defaults to `--identity` when unset, so an ed25519-identity node is unaffected.

## Tests

- `crates/cli/src/commands/start/mod.rs`: key type follows key length, the `0x` prefix is transparent, other lengths are refused.
- `crates/cli/tests/start_tests.rs`: the node identity and the Orbis service identity are separately parsed, land on different DIDs and key types, and the service identity is absent when the flag is.
- `cargo test -p cli --features orbis,sourcehub` and `cargo test -p sourcehub` pass (46 sourcehub tests).
- All four `cli` feature sets CI checks are clean: default, `--no-default-features`, `--no-default-features --features release-full`, and `--features otel`.

## Verified end to end

Against a live `verad` devnet (Vera main, ddcb612), a 3-node Orbis ring at threshold 2 and 35 DefraDB cells, in `sourcenetwork/backbone`'s `tests/gents_cloud` suite ([backbone#30](https://github.com/sourcenetwork/backbone/pull/30)): 13 of 13 scenarios passed, 32 tenants provisioned, 223 chain transactions committed with none failed. Policies and relationships apply on Vera, documents register with the creating DID as owner, ring-signed blocks replicate and verify on a peer, revocation propagates in 9 ms on an event-subscribed cell, and all 32 ordered cross-tenant reads are denied. Per-tenant cost measured at 6.06 s to provision and 71 MiB resident.

That run also needs the companion orbis-rs fix ([orbis-rs#264](https://github.com/sourcenetwork/orbis-rs/pull/264)), without which the ring advertises a key its signatures do not verify under.

## Not included

The create path still sends the ring no ACP tuple: `create_access_decision` returns `Ok(None)` for the Cosmos provider, so a create is authorised by DefraDB's own check rather than by the ring. That is unchanged behaviour and is tracked separately.

### DIFF
--- a/crates/cli/src/commands/start/mod.rs
+++ b/crates/cli/src/commands/start/mod.rs
@@ -404,14 +404,9 @@ impl StartArgs {
         Ok(())
     }
 
-    /// Parse the user identity from the --identity flag.
-    ///
-    /// The identity flag should contain a hex-encoded private key.
-    /// Key type is auto-detected from byte length:
-    /// - 64 bytes -> Ed25519
-    /// - 32 bytes -> secp256k1
     /// Parse `--signer-orbis-identity`, the key used only to authenticate to
-    /// the Orbis ring.
+    /// the Orbis ring. It is separate from `--identity` because the ring
+    /// accepts an EdDSA token while the chain needs a secp256k1 signer.
     #[cfg(feature = "orbis")]
     pub fn parse_orbis_service_identity(
         &self,
@@ -424,6 +419,12 @@ impl StartArgs {
         Ok(Some(std::sync::Arc::new(identity)))
     }
 
+    /// Parse the user identity from the --identity flag.
+    ///
+    /// The identity flag should contain a hex-encoded private key.
+    /// Key type is auto-detected from byte length:
+    /// - 64 bytes -> Ed25519
+    /// - 32 bytes -> secp256k1
     pub fn parse_user_identity(&self) -> Result<Option<std::sync::Arc<identity::RawIdentity>>> {
         let hex_key = match &self.identity {
             Some(key) => key,
--- a/crates/cli/src/commands/start/mod.rs
+++ b/crates/cli/src/commands/start/mod.rs
@@ -361,6 +361,8 @@ impl StartArgs {
             })?,
         };
 
+        Self::require_ed25519_service_key(service_identity.key_type())?;
+
         let endpoint = self.signer_orbis_endpoint.as_ref().ok_or_else(|| {
             Error::InvalidConfig("--signer-orbis-endpoint required for orbis signer".into())
         })?;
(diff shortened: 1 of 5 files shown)


crates/cli/src/commands/start/mod.rs
```diff
@@ -334,13 +350,18 @@ impl StartArgs {
...
+        let service_identity = match self.parse_orbis_service_identity()? {
+            Some(identity) => identity,
+            None => user_identity.as_ref().cloned().ok_or_else(|| {
+                Error::InvalidConfig(
+                    "--identity or --signer-orbis-identity is required when \
+                     --signer-type=orbis (service key signs JWTs for Orbis auth)"
+                        .into(),
+                )
+            })?,
+        };
```
edjroz, line 362: > When --signer-type=orbis uses a secp256k1 identity, reject it before OrbisClient::new. identity::new_token_with_custom_claims encodes secp256k1 identities as ES256K, but the Orbis ring requires an EdDSA bearer token.

Confirmed,  twith the flag unset a Vera node's  --identity is always secp256k1, so the documented fallback is the failing path for the exact configuration this PR enables, and the node boots clean because OrbisClient::new mints no JWT and only dies at the first ring Sign; a 32-byte ed25519 seed passed to the new flag is inferred as secp256k1 at mod.rs:449-457 and lands in the same place, and one `key_type() != Ed25519` guard before OrbisClient::new covers both.

edjroz (changes requested):
For the requested change I think you can, Guard in setup_orbis_signer (mod.rs:362, 3 lines). Rejects a non-ed25519 service key before OrbisClient::new, so the node refuses to boot instead of dying at first Sign. Covers both entry paths at their convergence point.

edjroz (approved):

Issue #634, opened 2026-03-30
Multi-model embedding support for @embedding directive


jackzampolin:
## Summary

The `@embedding` directive currently supports a single embedding model. We need multi-model embedding support to enable hybrid retrieval (BM25 + dense vector search) across different use cases and collections.

## Context

Building agentic search capabilities on top of DefraDB that combine existing BM25 full-text search with dense vector retrieval via reciprocal rank fusion. This is informed by recent research on agentic multi-hop retrieval (Chroma Context-1) showing that hybrid BM25+dense search is the right base primitive for search agents.

## Requirements

- **Multiple embedding providers**: Support configuring different embedding models per field/collection (e.g., local models for sovereign compute, API-based for higher quality)
- **Provider abstraction**: Clean interface for embedding providers (local inference, OpenAI, Anthropic, etc.)
- **Schema integration**: `@embedding` directive should support provider, model, URL, and field configuration:
  ```graphql
  content_v: [Float32!] @embedding(provider: "local", model: "nomic-embed-text", url: "http://localhost:8080", fields: ["content"])
  ```
- **Batch embedding generation**: Efficient bulk embedding for historical data re-import
- **Similarity search query**: GraphQL query support for vector similarity (cosine, dot product) alongside existing BM25

## Current State

- `@fulltext` with BM25 works well across multiple collections
- `@embedding` directive exists in design but only supports a single model configuration
- No vector similarity search query support yet

## Use Cases

- Hybrid search (BM25 + dense) with reciprocal rank fusion for agentic retrieval
- Semantic search over coding sessions, conversations, and document collections
- Multi-hop retrieval where keyword search misses semantic connections

jackzampolin:
**Triage: `1.0 — Go parity` / Parity gap.** Procedure: #1391.

`Go Status: Has at v1.0.0`

[...]

PR #1682, opened 2026-09-03
test(db): cover per-field embedding providers

iverc:
## Context

#1631 added provider-specific embedding requests, but its regression exercised Ollama in isolation. #634 also requires multiple embedding configurations to coexist without sharing provider or model state.

## Changes

- exercise OpenAI-compatible and Ollama embeddings on different fields of one document
- verify each request uses its field-specific model and payload contract
- verify each response is decoded into the correct field, including Ollama normalization

## Testing

- `cargo test --profile super-dev -p db --test search embedding -- --nocapture`
- `cargo clippy --profile super-dev -p db --test search -- -D warnings`
- `git diff --check`

Refs #634

### DIFF
--- a/crates/db/tests/search/embedding.rs
+++ b/crates/db/tests/search/embedding.rs
@@ -125,17 +125,25 @@ async fn set_embedding_skips_when_effective_config_is_missing() {
 }
 
 #[tokio::test]
-async fn set_embedding_uses_ollama_contract() {
+async fn set_embedding_uses_each_field_provider_and_model() {
     use axum::{routing::post, Json, Router};
     use serde_json::{json, Value};
     use tokio::net::TcpListener;
 
-    async fn embed(Json(request): Json<Value>) -> Json<Value> {
+    async fn openai(Json(request): Json<Value>) -> Json<Value> {
         assert_eq!(
             request,
-            json!({"model": "nomic-embed-text", "prompt": "hello\n"})
+            json!({"model": "text-embedding-3-small", "input": "hello\n"})
         );
-        Json(json!({"embedding": [3.0, 4.0]}))
+        Json(json!({"data": [{"embedding": [1.0, 0.0]}]}))
+    }
+
+    async fn ollama(Json(request): Json<Value>) -> Json<Value> {
+        assert_eq!(
+            request,
+            json!({"model": "nomic-embed-text", "prompt": "world\n"})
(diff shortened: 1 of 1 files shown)


vertexclique (commented):
**Verdict: has issues.** The test is real, but one of its two stated claims is not actually pinned.

What it covers well: per field provider (`input` versus `prompt`), model, url on distinct routes, source field (`content` and `summary`, so a cross wired `fields` trips the other handler), target field, and the two response decoders (`/data/0/embedding` versus `/embedding`, so an openai body fed to the ollama pointer fails to decode). Not a happy path assertion. No sleeps, and the listener is already listening before `axum::serve` accepts so there is no bind race.

### Minor: the Ollama normalization claim is only half pinned

`crates/db/tests/search/embedding.rs:137`

The OpenAI response is `[1.0, 0.0]`, already unit length. `call_embedding` normalizes only for ollama (`crates/db/src/search/embedding.rs:229-231`). So `content_v == [1.0, 0.0]` passes whether or not the openai branch normalizes, and the claim "verify each response is decoded into the correct field, including Ollama normalization" pins the Ollama half only. A non unit vector, say `[3.0, 4.0]` expecting `[3.0, 4.0]`, makes the test fail if `normalize_embedding` is ever moved out of the provider guard.

### Minor: assertion failures surface as transport errors

`crates/db/tests/search/embedding.rs:132-146`

The request shape assertions run inside axum handlers on a `tokio::spawn`ed server whose `JoinHandle` is only `abort()`ed. A failed `assert_eq!` kills the connection task; the test then fails on the `.unwrap()` at line 185 with a transport error rather than the payload diff. Recording requests into a captured `Mutex<Vec<Value>>` and asserting after the call gives the same coverage with a readable failure.

### Minor: a fallback that reaches the real API

`crates/db/tests/search/embedding.rs:161-171`

If `resolve_embedding_config` ever stopped preferring `embedding.url`, the empty `EmbeddingClientConfig` falls through to `DEFAULT_OPENAI_URL` and the test makes a real request to `api.openai.com`. It still fails, but on network state.

### Claims

"multiple embedding configurations to coexist without sharing provider or model state": the one thing that genuinely is shared node level state, `EmbeddingClientConfig::api_key`, which is openai only (`embedding.rs:212-214`), is never exercised because the test's key is empty.

PR #1683, opened 2026-09-03
fix(db): preserve unique winners during reindex

### LINKED ISSUE #1308: Boot-time reindex (materialize_collection/bulk_index) hard-fails on docs the merge path deliberately parked unindexed
## Asymmetry

The merge path resolves live unique-index conflicts deterministically (#1111): `db_index::index_manager::save_resolving_unique_conflict` gives the index entry to the lexicographically smallest public DocID and persists the loser unindexed, so replicas converge instead of wedging the merge. That is the documented, intentional behavior.

But the reindex path does not apply the same rule. `DB::materialize_collection` → `materialize_collection_inner` (crates/db/src/migration/reindex.rs) does `remove_all` + `IndexManager::bulk_index` over **every** persisted doc, and `bulk_index` (crates/db-index/src/index_manager/mod.rs) uses plain `index.save()` per doc. The merge-parked loser re-raises `UniqueConstraintViolation`, and the whole collection's materialization fails:

```
storage error: can not index a doc's field(s) that violates unique index.
```

## Impact (observed live, source-inc/gents#984)

An embedder that calls `materialize_collection` at every boot (gents' `ensure_migrations`) hard-fails forever after a single P2P merge parks one stale foreign doc — the home is bricked until a full reset. The merge tolerates a state the reindex refuses: one subsystem's graceful outcome is another's poison pill.

gents now works around this defensively (source-inc/gents PR forthcoming): it classifies the unique-violation failure, skips that collection's eager pass (the failed txn rolls back, so the merge-time winner's index state survives), warns with the parked docIDs, and boots. But the workaround relies on string-matching the error message and leaves the collection's eager materialization permanently skipped while the parked doc exists.

## Proposed fix

`bulk_index` (or a merge-consistent variant used by `materialize_collection_inner`) should resolve unique conflicts with the same deterministic-winner rule as `save_resolving_unique_conflict`: smallest public DocID keeps the entry, the loser is skipped (left unindexed) and reported. The rule is order-independent, so a full reindex would deterministically converge to exactly the state the merge left behind — instead of erroring on it.

Secondarily: a typed error (rather than the shared message string) would let embedders classify this without string matching.

iverc:
## Context

Replication resolves live unique-index conflicts deterministically by keeping the document with the smallest public DocID indexed and persisting the other document without an index entry. Collection materialization rebuilt the same index through the strict bulk path, so a state accepted by replication could make every later materialization fail with a unique-constraint violation.

## Changes

- rebuild unique indexes through the same deterministic conflict resolver used by replication
- retain the existing bulk path for non-unique indexes
- verify materialization succeeds after a replicated unique conflict and preserves the canonical winner

## Testing

- `cargo test --profile super-dev -p db --test merge`
- `cargo clippy --profile super-dev -p db --test merge -- -D warnings`
- `git diff --check`

Closes #1308

### DIFF
--- a/crates/db/src/index/manager/mod.rs
+++ b/crates/db/src/index/manager/mod.rs
@@ -25,7 +25,7 @@ use storage::keys::IndexIDSequenceKey;
 pub mod doc_source;
 pub use doc_source::{DocumentSource, SliceSource};
 
-/// How a live unique conflict was resolved during merge (#1111).
+/// How a live unique conflict was resolved during merge or index rebuild (#1111/#1308).
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum MergeConflictOutcome {
     /// No conflict, or the stale-entry heal handled it.
@@ -498,6 +498,47 @@ impl IndexManager {
         })
     }
 
+    /// Rebuild an index using the same deterministic unique-conflict rule as replication.
+    pub(crate) async fn bulk_index_resolving_unique_conflicts(
+        &self,
+        datastore: &NamespaceView,
+        systemstore: &NamespaceView,
+        index_name: &str,
+        documents: &[(u64, Document)],
+        schema: &CollectionVersion,
+    ) -> Result<()> {
+        let index = self
+            .indexes
+            .get(index_name)
+            .ok_or_else(|| Error::Other(format!("index '{}' not found", index_name)))?;
+        let mut mutable_datastore = datastore.clone();
+
+        for (doc_short_id, doc) in documents {
+            if *doc_short_id == 0 {
+                continue;
+            }
+            let doc_id = doc
+                .id()
+                .ok_or_else(|| Error::InvalidDocument("document must have an ID".to_string()))?
+                .to_string();
+            let value_sets = self.extract_index_values(doc, index.description(), schema)?;
+            for values in &value_sets {
+                self.save_resolving_unique_conflict(
+                    &mut mutable_datastore,
+                    systemstore,
+                    index,
+                    *doc_short_id,
+                    &doc_id,
+                    values,
+                )
+                .await
+                .map_err(Error::Storage)?;
+            }
+        }
+
+        Ok(())
+    }
+
     /// Save an index entry, healing stale unique conflicts (#1111/#700).
     ///
     /// A unique violation whose existing entry points at a **deleted or
@@ -550,8 +591,7 @@ impl IndexManager {
         }
     }
 
-    /// Save an index entry on the MERGE path, resolving live unique conflicts
-    /// deterministically instead of failing the merge (#1111).
+    /// Save an index entry while resolving live unique conflicts deterministically (#1111/#1308).
     ///
     /// A CRDT merge cannot preserve cross-replica uniqueness: two nodes that
     /// each locally accepted the same unique value must still converge when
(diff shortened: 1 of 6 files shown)


crates/db/src/index/manager/mod.rs
```diff
@@ -498,6 +498,54 @@ impl IndexManager {
...
+                    &mut mutable_datastore,
+                    systemstore,
+                    index,
+                    *doc_short_id,
+                    &doc_id,
+                    values,
+                )
+                .await
+                .map_err(Error::Storage)?;
+            }
```
vertexclique, line 571: Make this concurrent and race through with `join_all!`

vertexclique (changes requested):

vertexclique (commented):
**Verdict: looks sound.** The algorithm is correct; two scope questions worth an explicit decision.

I traced the reduction. After `remove_all`, each document goes through `save_resolving_unique_conflict`, which on a violation resolves the holder's node local short id to its public DocID and gives the entry to the lexicographically smaller DocID via `save_blind`. Because `save_blind` is a single overwrite of the value key (`crates/storage/src/index/unique.rs:148-149`), the fold is a min reduction and is order independent: for docIDs A<B<C, all six arrival orders end with A holding the entry. Same rule and same code path as the merge, so a full reindex converges to exactly the state the merge left. `conflicting_doc_id` is a point `get` so it reads the same transaction's pending writes and the in txn `remove_all` is visible. Skipping `build_if_needed` is correct, since `index_type.rs:137` makes it a no op for `Unique`. Deleted documents cannot be reindexed, as `reindex.rs:137` passes `show_deleted = false`.

### Major: every unique violation is now silently unindexed, not just merge parked ones

`crates/db/src/index/manager/mod.rs:515-546`

The new path converts every unique violation during materialization into "loser is silently unindexed", not only the merge parked ones #1308 describes, and there is no signal at this point that distinguishes them.

Scenario: a lens transform lowercases `email`, mapping doc `bafy...aaa` (`A@x`) and doc `bafy...zzz` (`a@x`) onto the same value. Before this PR, `materialize_collection` failed loudly with `UniqueConstraintViolation` and the migration was rejected. After, it returns `Ok`, and `bafy...zzz` is dropped from an index the schema still declares `@unique`, so every later `email: {_eq: "a@x"}` misses it. It is logged at `tracing::error!`, but the operation reports success.

### Minor

- `crates/db/src/definition/migration/reindex.rs:246` with `manager/mod.rs:507`: the new entry point takes `documents: &[(u64, Document)]`, the whole collection in RAM, rather than the pull based `DocumentSource` the repo built to avoid exactly this (`crates/db/src/collection/stream.rs:217` documents the motive; `crates/ffi/src/index/create.rs:201` already uses `bulk_index_from` with `BackfillSource`). The `migrated_docs` Vec is pre existing, but the unique path is now locked to the non streaming shape.
- `manager/mod.rs:527-529`: `doc_short_id == 0` is skipped with no count and no log, and the function returns `Result<()>` rather than `BulkIndexResult`. A materialization that skipped documents is indistinguishable from one that indexed everything.

### Claims

"Closes #1308": the primary ask is delivered, but the issue's secondary ask, a typed error rather than the shared message string so embedders can classify without string matching, is not in the diff. Separately, index creation remains strict (`crates/ffi/src/index/create.rs:201` to `bulk_index_from` to `index.save`), so `CREATE UNIQUE INDEX` on a collection holding a merge parked duplicate still hard fails. That may be intentional; the PR does not say.

The test at `crates/db/tests/merge/merge_handler_tests.rs:3057` genuinely fails without the fix and pins the winner as `min(doc_a_id, doc_b_id)` rather than "one of them". Gap: no three way conflict and no permutation test, so the order independence property the whole design rests on is not pinned.

PR #1684, opened 2026-09-03
fix(cli): load configured secret files

iverc:
## Context

`--secret-file` was parsed and persisted but never loaded. This differed from Go and made secrets configured through the file unavailable at runtime.

## Changes

- load the selected env file during configuration
- ignore missing files while reporting malformed or unreadable files
- cover successful, missing, and invalid files

## Testing

- `cargo test --profile super-dev -p cli --test config_tests`
- `cargo clippy --profile super-dev -p cli --tests -- -D warnings`

Refs #1422

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2092,6 +2092,7 @@ dependencies = [
  "defra-version",
  "dirs",
  "document",
+ "dotenvy",
  "events",
  "futures",
  "getrandom 0.2.17",
@@ -3887,6 +3888,12 @@ dependencies = [
  "thiserror 2.0.18",
  "uuid",
 ]
+
+[[package]]
+name = "dotenvy"
+version = "0.15.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
 
 [[package]]
 name = "dtoa"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2092,7 +2092,6 @@ dependencies = [
  "defra-version",
  "dirs",
  "document",
- "dotenvy",
  "events",
  "futures",
  "getrandom 0.2.17",
@@ -3888,12 +3887,6 @@ dependencies = [
  "thiserror 2.0.18",
  "uuid",
 ]
-
-[[package]]
-name = "dotenvy"
-version = "0.15.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
 
 [[package]]
 name = "dtoa"
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -48,6 +48,7 @@ telemetry = { path = "../telemetry", default-features = false }
 
 # CLI
 clap = { version = "4.5", features = ["derive", "env"] }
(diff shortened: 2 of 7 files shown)


vertexclique (commented):
**Verdict: has issues.** The core fix is real and Go faithful: missing file ignored, malformed file loud, no override precedence, and errors carry only path and line with the secret absent from the message (asserted by the test). The problems are in when it loads and in the hand rolled parser.

### Major: DEFRA_ vars in the secret file are ignored

`crates/cli/src/config/mod.rs:96`

`load_secret_file()` runs after `Cli::parse()` has already resolved every clap `env = "DEFRA_*"` binding, 24 global plus 16 start flags. Go resolves lazily via viper `AutomaticEnv` after `godotenv.Load`, so it honours them.

Scenario: a `.env` containing `DEFRA_API_ADDRESS=0.0.0.0:9181`. The Go node binds all interfaces; the Rust node silently stays on `127.0.0.1`. Only lazily read vars work, such as `DEFRA_KEYRING_SECRET` via `crates/cli/src/commands/mod.rs:39` and the embedding key via `start/server.rs:89`. Not mentioned in the description.

### Major: env::set_var inside a multi threaded runtime

`crates/cli/src/config/secret_file.rs:60`

This executes inside `#[tokio::main]`'s multi threaded runtime (`crates/cli/src/main.rs:15-31`) whose workers are already spawned. Concurrent `getenv` from another thread is undefined behaviour on glibc.

The three new tests (`crates/cli/tests/config_tests.rs:165,181,191`) also call `Config::load` without `#[serial]` while sibling tests in the same binary read env through `Config::resolve_rootdir`. The repo's own convention for env mutating tests is `#[serial]` (`crates/keyring/src/lib.rs:71-79`).

### Major: no variable expansion

`crates/cli/src/config/secret_file.rs:67`

godotenv expands `$VAR` and `${VAR}` on both unquoted and double quoted values. `DEFRA_KEYRING_SECRET=${BASE}-prod` gives the literal string here and the expanded string in Go, so a keyring written by a Go node cannot be opened by the Rust node from the same `.env`. It surfaces as a decrypt failure, not a config error.

### Minor parser divergences from godotenv

- `secret_file.rs:33`: only `=` is accepted; godotenv also accepts `:`. A YAML style `.env` Go loads makes the Rust node exit with `invalid secret file ... at line N`.
- `secret_file.rs:22`: line by line parsing rejects multi line quoted values, which godotenv supports. A PEM blob in `.env` is a hard startup failure on Rust only.
- `secret_file.rs:30`: `strip_prefix("export ")` matches only a literal space; godotenv accepts any space character, so `export\tKEY=v` errors here.
- `secret_file.rs:98-101`: dead code, since when `character != '\\'` the previous line has already set `escaped = false`. Backslash escaping is also applied inside single quotes, which POSIX and godotenv do not do, so `X='a\'` becomes a startup error instead of the value `a\`.

### Minor

- `crates/cli/tests/config_tests.rs:165-207`: roughly 90 lines of parser, three end to end tests. Nothing covers quoting, escapes, `export`, inline comments, or the key charset, so every divergence above is untested.
- `secret_file.rs:10`: no permission or ownership check, and values are promoted to process env (`/proc/<pid>/environ`, inherited by children). Go does not check either, so this is parity consistent. Worth noting that the default `.env` is resolved against the working directory, matching Go's `ConfigPaths`, so running `defra` inside an unrelated project directory imports that project's `.env`, including `OPENAI_API_KEY`, which `config.embedding.api_key_env` reads by default.

PR #1685, opened 2026-09-03
fix(p2p): stop connection polling when host closes

iverc:
## Context

Connection polling discarded host command-channel errors and could wait until the caller timeout after the host had already closed.

## Changes

- propagate terminal host errors immediately
- cover polling against a closed host

## Testing

- `cargo test --profile super-dev -p p2p --features test-utils --test host_tests connection_poll_stops_when_host_is_closed`
- `cargo clippy --profile super-dev -p p2p --features test-utils --test host_tests -- -D warnings`

Refs #1356

### DIFF
--- a/crates/p2p/src/host/handle.rs
+++ b/crates/p2p/src/host/handle.rs
@@ -877,10 +877,9 @@ impl P2PHostHandle {
     ) -> Result<()> {
         let deadline = tokio::time::Instant::now() + timeout;
         loop {
-            if let Ok(connected) = self.connected_peers().await {
-                if connected.contains(&peer_id) {
-                    return Ok(());
-                }
+            let connected = self.connected_peers().await?;
+            if connected.contains(&peer_id) {
+                return Ok(());
             }
             if tokio::time::Instant::now() >= deadline {
                 return Err(Error::ConnectionTimeout(peer_id.to_string()));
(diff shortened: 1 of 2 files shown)


vertexclique (approved):

vertexclique (commented):
**Verdict: looks sound.**

The change is correct and matches the shape the iroh transport already uses (`crates/p2p/src/iroh/transport.rs:195`). `connected_peers()` can only fail terminally: `send` fails only when the receiver is gone, and `crates/p2p/src/host/command_handler/mod.rs:47-52` never drops the `ConnectedPeers` responder without replying, so no transient error is being converted into a hard failure. Downstream classification is unchanged, since `ChannelSend` and `ConnectionTimeout` are both `is_connection_like()` (`crates/p2p/src/error.rs:337-343`).

### Minor

- `crates/p2p-adapter/src/libp2p.rs:314-317`: the same swallow, `.unwrap_or(false)`, survives one frame up in `connect_peer`. A closed host makes `already_connected` read `false` and the code dials anyway. Self correcting, since the dial fails too, but it is the identical defect this PR is fixing, left in place.
- `crates/p2p/tests/host_tests.rs:171-188`: asserts only `result.is_err()`, so it would pass if the poll returned any error, including a wrong one. It does fail before the fix, since the old loop burns the 60s window against the 1s outer timeout, so it is a real regression test, just a loose one.

Note for merge order: #1689 is based on this branch rather than main, so it cannot land first.

PR #1686, opened 2026-09-03
test(cli): inventory config flag enforcement

### LINKED ISSUE #1422: Build a config-enforcement test class: every flag must demonstrably do something

iverc:
## Context

A configuration flag can map into `Config` while still having no runtime effect. Several options were also omitted from the existing all-flags test.

## Changes

- assert every config-backed global and start flag
- classify flags by their runtime enforcement boundary
- compare the inventory against the argument surface generated by Clap

## Testing

- `cargo test --profile super-dev -p cli --features release-full --test config_tests --test start_tests`
- `cargo clippy --profile super-dev -p cli --features release-full --tests -- -D warnings`

Closes #1422

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -48,3 +48,7 @@ fn should_profile(cli: &Cli) -> bool {
 fn enable_telemetry(cli: &Cli) -> bool {
     matches!(&cli.command, Command::Start(_))
 }
+
+#[cfg(test)]
+#[path = "../tests/unit/main_tests.rs"]
+mod tests;
--- a/crates/cli/tests/config_tests.rs
+++ b/crates/cli/tests/config_tests.rs
@@ -1,11 +1,15 @@
 //! Tests for the Config struct and configuration loading
 
+use std::collections::BTreeSet;
 use std::fs;
 use std::path::PathBuf;
 
+use clap::CommandFactory;
 use cli::cli::{Cli, Command};
 use cli::commands::VersionArgs;
-use cli::config::{Config, DatastoreType, KeyringBackend, LogFormat, LogLevel, LogOutput};
+use cli::config::{
+    AcpDocumentType, Config, DatastoreType, KeyringBackend, LogFormat, LogLevel, LogOutput,
+};
 use cli::error::Error;
 
 /// Helper to create a minimal Cli for testing apply_cli_flags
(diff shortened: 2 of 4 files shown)


vertexclique (commented):
you might need to update this with what i also added for gents-cloud tests. flags block.

vertexclique (commented):
**Verdict: has issues.** This is an inventory with a tripwire rather than enforcement, and two of the classifications it ships are false at its own base commit.

### Critical: the inventory asserts coverage that does not exist

`crates/cli/tests/start_tests.rs:392`, `HTTP_START_FLAGS`, `pubkeypath` and `privkeypath`

The group comment says "Consumed by start/server_http.rs". On this PR's base neither path reaches the served listener; the only consumer is redaction in `start/server_http/node_options.rs:90-91`. That is exactly the "accepted then ignored" defect this PR exists to surface, asserted here as covered.

Failure scenario: a reviewer takes the green test as proof TLS is wired and ships with `--pubkeypath/--privkeypath`, and the API serves cleartext. #1702 is the actual fix.

### Major: `default-key-type` is misclassified the same way

`crates/cli/tests/start_tests.rs:353`, `STORAGE_START_FLAGS`

The group header claims each group "names the runtime boundary that consumes those fields", but on this base nothing reads `config.datastore.default_key_type`; `start/mod.rs:539` only writes it. #1700 is the fix.

### Major: the mechanism cannot catch what the body claims

`crates/cli/tests/start_tests.rs:484-516` and `crates/cli/tests/config_tests.rs:256-278`

The only machine checked property is that the set of clap long flags equals the union of hand maintained constants. Nothing links a name in `CONFIG_BACKED_*` to any assertion, so a developer who adds `--foo`, drops it on the floor, and appends `"foo"` to the list gets a green suite. That is a test that passes when the logic is wrong. Real enforcement would derive the list from per flag round trip assertions rather than from a parallel constant.

### Minor

- `start_tests.rs:438-446`: `DIRECT_START_FLAGS` invokes `orbis_service_identity_is_separate_from_the_node_identity`, which is itself `#[test]` at line 300, so it runs twice. Extract a helper.
- `start_tests.rs:462-482`: `identity_flag_selects_signing_key` covers 32 byte secp256k1 plus one non hex rejection; the documented 64 byte ed25519 auto detect branch (`start/mod.rs:96-99`) is untested.
- Merge order: the body says "Closes #1422" while #1700 and #1702 are the PRs that make two of its listed flags true. Merging this first closes the issue while `pubkeypath`, `privkeypath` and `default-key-type` are still inert. `RUNTIME_EXEMPT_START_FLAGS` is honestly labelled; the `CONFIG_BACKED_*` comments need the same honesty.

Checked and not findings: the `#[cfg(test)] #[path = "../tests/unit/main_tests.rs"]` pattern matches existing use in `crates/http/src/server.rs:723-725`, and `tests/unit/` and `tests/start/` are subdirectories so cargo will not also auto build them. The real gain is `test_apply_to_config_all_flags` now setting `durability`, `p2p_transport`, `pg_address` and five `acp_*` flags that were previously `None`.

Issue #1438, opened 2026-08-12
Flaky: go_rust_burst_bidirectional times out converging; may be masking #1401/#1405


jackzampolin:
`write_contention::go_rust_burst_bidirectional` failed with:

```
burst writes did not converge on both nodes within 60s
```

## It is intermittent, with both outcomes recorded today

| Run | Branch | Time | Result |
|---|---|---|---|
| 31618124588 | `edjroz/1333-dedup-sysinfo-thiserror` (#1383) | 20:19:50 | **ok** (~3s) |
| 31619368962 | `ci/merge-queue-support` (#1427) | 21:04:57 | **FAILED** (60s timeout) |

Both branches sit on the same `main`; #1427 changes only `ci.yml` triggers and cannot affect runtime behaviour. So this is the same code passing and failing 45 minutes apart.

## Two candidate explanations, and they are not equally benign

**1. Load.** The fleet is heavily oversubscribed — see #1429, where `studio-1` was measured at load 50-65 on 32 cores with a non-CI tenant consuming ~12 cores. A 60-second convergence budget on a contended host is a plausible false negative, and the failing job ran on `studio-2` while four other suites were in flight.

**2. A real intermittent convergence defect.** This is a *mixed Rust↔Go* burst test, and two cross-runtime convergence defects were found today that could produce exactly this signature:

- **#1401** — Rust emits `nonce = 0` for the first write of a counter field on an existing document, so two equal concurrent increments become byte-identical blocks and dedup to one. A burst of writes is precisely the shape that triggers it.
- **#1405** — Rust derives field delta priority from a document-wide head max where Go uses the field's own head-set, which diverges block CIDs and can change LWW arbitration outcomes.

Both are load- and ordering-dependent, which is what an intermittent convergence timeout looks like from the outside.

## Why this should not simply be re-run

If explanation 2 is right, this test is the only automated signal we have for a real cross-runtime convergence bug, and treating it as noise discards that signal. The distinguishing evidence is cheap to collect and nobody has collected it: **what the two nodes actually converged to when it failed.** A value mismatch points at #1401/#1405; identical values with a slow tail points at load.

## Suggested next step

Capture per-node state on failure rather than asserting only convergence-within-60s. `DEFRA_E2E_KEEP=1` retains server logs under `target/e2e/.../logs/`, which would show whether the DAGs diverged or merely lagged.

Filed while triaging #1427, where this failure appeared. Not caused by that PR.

PR #1687, opened 2026-09-03
test(p2p): report burst convergence state

iverc:
## Context

Burst convergence timeouts currently report only document counts, making replication lag, loss, and state divergence indistinguishable.

## Changes

- query and sort the complete document state on both nodes
- include both states in convergence timeout failures

## Testing

- `cargo test --profile super-dev -p integration-test --test p2p --no-run`
- `cargo clippy --profile super-dev -p integration-test --test p2p -- -D warnings`

Refs #1438

### DIFF
--- a/tools/integration-test/tests/p2p/write_contention.rs
+++ b/tools/integration-test/tests/p2p/write_contention.rs
@@ -70,6 +70,19 @@ fn count_docs(cluster: &TestCluster, idx: usize, collection: &str) -> usize {
         .unwrap_or(0)
 }
 
+fn query_record_docs(cluster: &TestCluster, idx: usize) -> Result<Vec<serde_json::Value>, String> {
+    let response = cluster
+        .client(idx)
+        .query("query { Record { _docID origin idx } }")
+        .map_err(|error| format!("{error:#}"))?;
+    let mut docs = response["Record"]
+        .as_array()
+        .cloned()
+        .ok_or_else(|| format!("unexpected query response: {response}"))?;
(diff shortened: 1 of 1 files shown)


vertexclique (commented):
**Verdict: looks sound** as a diagnostic change. The description overstates what it detects.

The loop is correctly structured: it queries first and only sleeps on a miss, so a fast machine exits on iteration one; the 60s deadline is checked every pass; and the assertion carries both node dumps plus the query errors when a query itself failed. Asserting `== total` on the same snapshot that satisfied `>= total` is slightly stronger than the old re-query after `poll_until`. It still fails when replication loses or duplicates a document.

No flakiness is introduced: the 500ms fixed interval matches the `poll_until` it replaced and sits under an enforced deadline, so a slow machine still gets its full 60s.

### Minor

- `tools/integration-test/tests/p2p/write_contention.rs:291-308`: the PR says the change makes state divergence distinguishable, but nothing compares `node0_docs` to `node1_docs`, only lengths are asserted. Two nodes holding 15 records each with different `origin` and `idx` values pass. The `sort_by` at line 82 therefore affects no assertion and is purely a readability aid for the failure dump. An `assert_eq!(node0_docs, node1_docs)` in the converged branch is what the description implies and would cost one line.
- `write_contention.rs:305`: the message hardcodes "within 60s" while the deadline comes from `Duration::from_secs(60)` at line 291. Two places that must agree.
- `write_contention.rs:73-83`: `query_record_docs` hardcodes `Record` while the neighbouring `count_docs` takes the collection as a parameter; the two now overlap.

PR #1688, opened 2026-09-03
test(conformance): retire report-only parity probes

### LINKED ISSUE #1394: epic: parity probes assert automatically, not manually
Gate for the `1.0 — Go parity` milestone. Scope and procedure: #1391.

## Oracle

`proofs/tests/behavioral/parity.rs` — runs the same operations on Go and Rust
nodes, including mixed clusters, and asserts Rust converges to the value Go
converges to.

## Problem

The probes are `#[ignore]` by default and require a harness-compatible Go
`defradb` on PATH, so they run manually if at all. A parity assertion nobody
runs is not an oracle.

Some probes also only REPORT rather than assert (`parity_samedoc_*`,
`parity_delete_update_*`) — they print each node's converged state for human
comparison. Those cannot gate anything in their current form.

## Done when

The asserting probes (`parity_counter_3node_*`, `parity_mixed_fields_3node_*`,
`parity_lww_tie_partition_*`, `parity_indexed_lww_*`) run automatically in the
go-compat job and are green.

## Work

1. Opt the go-compat CI step into the asserting probes.
2. Decide, per REPORT-only probe, whether it becomes an assertion or is retired.
   Reporting probes that nobody reads are dead weight.
3. Keep `parity_unique_twins_*` (#1134) as a known-divergence pin, not a
   convergence contract — Go's unique-index twin merge rejects inside the merge
   transaction and the sender treats the rejection as success. That divergence
   is deliberate and stays.

Known open divergences already tracked: #1043 (mixed counter storm, Go peers
double-apply from an identical DAG — `Go Status = Defect upstream`).

iverc:
## Context

Obsolete ignored probes used fixed sleeps, swallowed setup failures, and only printed results. Maintained parity scenarios already provide asserting CI coverage.

## Changes

- remove report-only parity probes
- retain asserting and intentional characterization tests
- update the module coverage description

## Testing

- `cargo test --profile super-dev -p conformance --test tla_conformance parity:: -- --list`
- `cargo clippy --profile super-dev -p conformance --test tla_conformance -- -D warnings`

Closes #1394


vertexclique (commented):
**Verdict: has issues.** The stated rationale is accurate and the substituted coverage is real for two of the three scenarios. Two cross implementation scenarios go away with nothing left in the tree that runs them.

I verified the CI claim: `.github/workflows/ci.yml:496-505` already runs the asserting probes in the go-compat job, and the filter list never included `parity_samedoc_*` or `parity_delete_update_*`. So no CI signal is lost and "Closes #1394" is defensible. The stated defects are real: `.ok()` on every `p2p_connect` and `p2p_replicator_set` swallowed setup failures, and the probes used `sleep(3)` and `sleep(10)` rather than deadline polling.

### Major: cross implementation delete versus update has no remaining coverage

`proofs/tests/behavioral/parity.rs:645-700`, the deleted `run_delete_update_parity`, `parity_delete_update_go_go` and `parity_delete_update_mixed`

The two remaining delete and update tests are Rust only: `proofs/tests/behavioral/partition.rs:1174 convergence_delete_update_race_preserves_tombstone` and `proofs/tests/behavioral/bughunt.rs:200 bughunt_delete_update_race` both build `rust_nodes`. `partition.rs` has no `go_nodes` call at all; `bughunt.rs` uses `go_nodes` once, at line 702, for the merge storm control.

Scenario: Go resolves a concurrent delete versus update as tombstone wins and Rust as update revives, or the reverse. A mixed cluster reaches permanent, unrecoverable state divergence on a deleted document, and no test in the repo would run that comparison. This is the one removed scenario the retained probes do not subsume, so "maintained parity scenarios already provide asserting CI coverage" is true for same doc fields and counters but not for this.

### Major: no mixed cluster survives a restart anywhere

`proofs/tests/behavioral/parity.rs:511-533`, the deleted `mixed_disk_cluster`, and `:444`, the deleted `restart_node` call

After this PR `parity.rs` contains zero `restart_node` calls and zero `with_node_store` calls, so no mixed Go/Rust cluster survives a process restart. `parity_lww_tie_partition_mixed` is a partition then heal, not a restart, and the retained restart coverage in `partition.rs` and `bughunt.rs` is Rust only.

Restart exercises a distinct path: persisted DAG reload, replicator re-target, peer id stability across a process boundary. `mixed_disk_cluster` was the only helper building a mixed cluster with per node native stores (`with_node_store(0,"regolith")` and `with_node_store(1,"badger")`), which is the only way to make a mixed cluster persistent at all, so a future mixed restart test has to re-derive it.

### Minor: an investigation note is being deleted with no home

`proofs/tests/behavioral/parity.rs:524-525, 553-557`

Those lines are the only record in the repository of the `--no-keyring` ephemeral peer id artifact ("Go's delete by address 404s, leaving a stale replicator pushing to the dead peer id"). `grep -rn "ephemeral peer-id"` across the tree returns only these deleted lines. Worth keeping somewhere before it goes.

### For the record

`run_samedoc`'s live case and `run_counter_parity` are genuinely subsumed, and by stronger tests. `run_mixed_fields_3node_probe` (`parity.rs:644`) asserts an exact converged `(name, views)` on a live 3 node mesh with DAG convergence gating, in rust/rust, go/go and mixed variants. That part of this PR is a clear improvement over `eprintln!`.

Suggestion rather than objection: both Major findings would close by converting `parity_delete_update_mixed` and one `parity_samedoc_mixed_restart_*` into asserting tests instead of deleting them, or, if that is out of scope here, by filing a follow up issue before merge so the gap is written down rather than dropped.

vertexclique (approved):

PR #1689, opened 2026-09-03
fix(p2p): stop KMS retries when transport closes

### LINKED ISSUE #1356: Background loops block on uninterruptible sleeps: no in-flight cancellation, so shutdown stops tasks by abort instead of by signal

iverc:
## Context

KMS subscriber and publish retries treated terminal transport closure like temporary peer unavailability, delaying shutdown for the full retry window.

## Changes

- stop subscriber polling on transport errors
- retry publishing only when peers are temporarily unavailable
- cover requests against a closed transport

## Testing

- `cargo test --profile super-dev -p p2p --features libp2p-transport --lib`
- `cargo clippy --profile super-dev -p p2p --features libp2p-transport --all-targets -- -D warnings`

Closes #1356

### DIFF
--- a/crates/p2p/src/kms/pubsub_transport.rs
+++ b/crates/p2p/src/kms/pubsub_transport.rs
@@ -328,6 +328,7 @@ impl<T: P2PTransport> PubsubKeyTransport<T> {
                 Ok(_) => {}
                 Err(e) => {
                     debug!(error = %e, "topic_peers query failed while awaiting KMS subscriber");
+                    return;
                 }
             }
             if tokio::time::Instant::now() >= deadline {
@@ -355,7 +356,9 @@ impl<T: P2PTransport> PubsubKeyTransport<T> {
                 .await
             {
                 Ok(_) => return Ok(()),
-                Err(e) if tokio::time::Instant::now() < deadline => {
+                Err(ref e @ crate::error::Error::GossipSubPublish(ref message))
+                    if message == "InsufficientPeers" && tokio::time::Instant::now() < deadline =>
+                {
                     debug!(topic = %topic, error = %e, "publish not yet ready; retrying");
                     tokio::time::sleep(SUBSCRIBER_POLL_INTERVAL).await;
                 }
@@ -556,7 +559,7 @@ mod tests {
     use crate::{QueryId, ReplicatorInfo};
     use async_trait::async_trait;
     use parking_lot::Mutex;
-    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
     use std::time::Duration;
 
     type PublishedRawMessages = Arc<Mutex<Vec<(String, Vec<u8>)>>>;
@@ -576,6 +579,7 @@ mod tests {
         subscriber_visible_after: usize,
         publish_attempts: Arc<AtomicUsize>,
         published: PublishedRawMessages,
+        closed: Arc<AtomicBool>,
     }
 
     impl RacyTransport {
@@ -588,11 +592,16 @@ mod tests {
                 subscriber_visible_after,
                 publish_attempts: Arc::new(AtomicUsize::new(0)),
                 published: Arc::new(Mutex::new(Vec::new())),
+                closed: Arc::new(AtomicBool::new(false)),
             }
         }
 
         fn subscriber_known(&self) -> bool {
             self.topic_peers_calls.load(Ordering::SeqCst) >= self.subscriber_visible_after
+        }
+
+        fn close(&self) {
+            self.closed.store(true, Ordering::SeqCst);
         }
     }
 
@@ -631,6 +640,9 @@ mod tests {
             Ok(Vec::new())
         }
         async fn topic_peers(&self, _topic: DefraTopic) -> Result<Vec<PeerId>> {
+            if self.closed.load(Ordering::SeqCst) {
+                return Err(Error::ChannelSend);
+            }
             let n = self.topic_peers_calls.fetch_add(1, Ordering::SeqCst) + 1;
             if n >= self.subscriber_visible_after {
                 Ok(vec![self.peer.clone()])
@@ -649,6 +661,9 @@ mod tests {
         }
         async fn publish_raw(&self, t: String, d: Vec<u8>) -> Result<MessageId> {
             self.publish_attempts.fetch_add(1, Ordering::SeqCst);
+            if self.closed.load(Ordering::SeqCst) {
+                return Err(Error::ChannelSend);
+            }
             if self.subscriber_known() {
                 self.published.lock().push((t, d));
                 Ok(MessageId::new("ok".to_string()))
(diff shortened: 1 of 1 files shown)


vertexclique (commented):
**Verdict: has issues.**

### Major: narrowing the predicate also drops genuine backpressure

`crates/p2p/src/kms/pubsub_transport.rs:359-363`

Narrowing the retry predicate to the exact string `"InsufficientPeers"` also drops `gossipsub::PublishError::AllQueuesFull`, which is transient backpressure rather than a terminal failure.

Scenario: a node with many gossipsub peers whose per peer send queues are momentarily full. `handle_publish_raw` (`crates/p2p/src/host/command_handler/pubsub.rs:98-102`) to `map_publish_error` yields `Error::GossipSubPublish("AllQueuesFull(8)")`. That no longer matches, so `send_request` returns `kms::Error::Internal` on the first attempt instead of retrying inside the existing 10s window. The requester reports the key unavailable and the document stays undecryptable.

The same regression hits the responder side at line 304, in `dispatch_incoming`'s `_response` reply, where a dropped reply is only `warn!` logged and the peer's request times out.

Classify on retryability rather than on one string.

### Minor: two independent classifiers

`crates/p2p/src/kms/pubsub_transport.rs:359` versus `crates/db/src/merge/broadcast_mutator/broadcast.rs:26`

There are now two independent classifiers of "is this publish retryable": `message == "InsufficientPeers"` here and `err_str.contains("InsufficientPeers")` there, with different policies, a fixed 100ms poll to a deadline versus exponential backoff with a max attempt cap. Per the one canonical implementation rule these should be one function; as written they will drift.

### Claims versus diff

The description says KMS retries no longer treat terminal closure as temporary unavailability. The background republish loop at `pubsub_transport.rs:486-497` is untouched: it still calls `publish_raw` every `REPUBLISH_INTERVAL` for the full 5s `RESPONSE_TIMEOUT` after a terminal `ChannelSend`, logging only `debug!("...will retry")`. Bounded and off the caller's critical path, but the fix is narrower than the description implies.

Trivial: the doc comment at `pubsub_transport.rs:318-322` still describes only two exits; `wait_for_subscriber` now also returns on any `topic_peers` error.

The test at roughly line 803 is real: `assert_eq!(publish_attempts, 1)` fails before the fix.

Note for merge order: this branch is based on #1685 rather than main.

Issue #1392, opened 2026-08-12
epic: ffi-test green against the Go v1.0.0 test suite


jackzampolin:
Gate for the `1.0 — Go parity` milestone. Scope and procedure: #1391.

## Oracle

`tools/ffi-test` runs Go DefraDB's own test suite against the Rust FFI library.
Go's tests are the source of truth; we do not hand-port them.

## Done when

The suite is green, or every remaining failure is closed out as one of:

- a **post-v1.0.0 Go behavior** — out of 1.0 scope by definition (#1391), or
- a **deliberate divergence** where Go is wrong, pinned by a test and coded
  `Go Status = Defect upstream`.

Anything else is a 1.0 blocker and gets a sub-issue here.

## Blocked on #1395

The Go shim that makes this oracle run (`tests/clients/rustffi/`, ~8,600 lines)
is not in this repo and not on Go `develop` — it lives on fork branches of the
Go repo. Until #1395 lands, a clean clone cannot run this suite, so "green"
cannot be independently reproduced.

## First task

Re-establish the current number. The last recorded rate was ~95%+, but that
predates recent drift, and a stale pass rate is not evidence. Run the suite,
enumerate the failures, and file one sub-issue per distinct failure — not per
failing test, since one defect usually fails several.

PR #1694, opened 2026-09-05
fix(db): traverse commits from live collection heads

iverc:
## Context

Collection head keys retain superseded heads. `_commits` treated every stored key as a traversal root, so historical commits could appear before the current head and consume the requested depth.

## Changes

- resolve live collection heads using the existing supersession markers
- start commit traversal from those heads, preserving ancestor traversal
- cover a retained parent and child with `depth: 1`, asserting only the live child is returned

## Testing

- `cargo test --profile super-dev -p db --test read commits` (6 passed)
- `cargo clippy --profile super-dev -p db -p defra-p2p-adapter --tests -- -D warnings`
- `git diff --check`

Refs #1510

### DIFF
--- a/crates/db/src/read/commits/mod.rs
+++ b/crates/db/src/read/commits/mod.rs
@@ -9,7 +9,7 @@ mod delta;
 use async_lock::Mutex as TokioMutex;
 use cid::Cid;
 use document::Document;
-use std::collections::HashSet;
+use std::collections::{BTreeSet, HashSet};
 use std::str::FromStr;
 use std::sync::Arc;
 use storage::corekv::{IterOptions, Store};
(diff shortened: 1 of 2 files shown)


vertexclique (commented):
**Verdict: looks sound.**

Key encoding verified: `HeadstoreColKey::bytes` (`crates/storage/src/keys/headstore.rs:179`) is `format!("/c/{}/{}", collection_id, cid)`, ASCII decimal, so `from_utf8` plus `strip_prefix("/c/")` plus `split_once('/')` is correct, and prefix `/c/` cannot match the `/cs/` marker range. Superseded parents remain reachable as ancestors because `record_supersedes` writes markers and the child head key in the same transaction (`block/builder/collection.rs:79-95`, `merge/merge_handler/collection.rs:384`, `:756`), so no reader observes markers without their child. Document heads use eager delete, so the `/d/` branch has no equivalent defect and scoping the fix to `/c/` is right.

### Minor: the head range is now read twice

`crates/db/src/read/commits/mod.rs:361-364`

The first pass over `/c/` reads every head value only to derive the set of collection ids, then `live_collection_heads` rescans the same range per collection. On a store where `prune_superseded_heads` has fallen behind, and the head range grows one key per mutation, `_commits` now reads that range twice and materializes each value. `.with_keys_only(true)` on the first iterator costs one call.

### Minor, uncertain: a swallowed error now has a worse consequence

`crates/db/src/merge/merge_handler/collection.rs:756` and `:384`

Both sites write markers first, then `let _ = headstore.set(col_key, ...)` or warn and continue. If markers for parents P1 and P2 commit but child C's head key write fails silently, then before this change `_commits` still returned P1 and P2 as roots and the history stayed visible; now both read as superseded and the collection reports zero live heads, so `_commits` returns nothing while the blocks sit in the blockstore. Pre existing, but the fix turns a false extra into a false empty.

### Minor: the issue reference is wrong

"Refs #1510" points at *Audit the Rust specific harness accommodations*, a sub issue of #1392 about the Go FFI oracle harness patch, which has nothing to do with commit traversal. Worth correcting so the trail is real.

The test at `crates/db/tests/read/commits_suite.rs:112-160` does fail without the fix, since both head keys become roots and `depth: 1` yields 2. Gaps: one collection, one marker; head values are written as `&[]` so `max_priority` decoding is never exercised; no coverage of two markers naming the same parent, which is the duplicate marker skip in the merge join, or of a marker whose parent has no head key, the replicated child first case `heads.rs` explicitly calls out.

PR #1695, opened 2026-09-05
fix(p2p): wait for branchable sync completion

iverc:
## Context

Branchable collection sync could return after a quiet period while advertised heads were still being fetched or merged. Callers could then read an incomplete collection immediately after a successful sync.

## Changes

- track the advertised head CIDs and wait for every outstanding head to merge
- skip already-merged heads and ignore duplicate or unrelated merge events
- return an error if the deadline expires or the event bus closes before completion
- cover delayed merges, partial completion, the shared deadline, and already-complete syncs

## Testing

- `cargo test --profile super-dev -p defra-p2p-adapter --lib` (61 passed)
- `cargo clippy --profile super-dev -p db -p defra-p2p-adapter --tests -- -D warnings`
- `git diff --check`

Stacked on the collection-head traversal fix. Refs #1510; the harness accommodation audit remains open.

### DIFF
--- a/crates/p2p-adapter/Cargo.toml
+++ b/crates/p2p-adapter/Cargo.toml
@@ -44,6 +44,7 @@ tracing.workspace = true
 zeroize = "1"
 
 [dev-dependencies]
+tokio = { workspace = true, features = ["test-util"] }
 crypto = { path = "../crypto" }
 data-encoding = "2"
 iroh.workspace = true
--- a/crates/p2p-adapter/src/libp2p.rs
+++ b/crates/p2p-adapter/src/libp2p.rs
@@ -78,7 +78,7 @@ async fn wait_for_branchable_merges(
     }
 }
 
-async fn unmerged_doc_sync_heads<B, I>(blockstore: &B, heads: I) -> HashSet<cid::Cid>
+async fn unmerged_heads<B, I>(blockstore: &B, heads: I) -> HashSet<cid::Cid>
 where
     B: Blockstore,
     I: IntoIterator<Item = cid::Cid>,
@@ -951,7 +951,7 @@ impl<B: Blockstore + 'static> P2POperations for P2PAdapter<B> {
                 return Err(error);
             }
         };
-        let mut pending_heads = unmerged_doc_sync_heads(coord.blockstore().as_ref(), heads).await;
+        let mut pending_heads = unmerged_heads(coord.blockstore().as_ref(), heads).await;
 
         while !pending_heads.is_empty() && start.elapsed() < overall_timeout {
             match tokio::time::timeout(std::time::Duration::from_millis(100), sub.recv()).await {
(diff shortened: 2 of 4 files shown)


vertexclique (commented):
**Verdict: has issues.** The new wait can turn a completed sync into a hard failure, and it covers one of three paths.

### Critical: a fully synced collection can report a timeout

`crates/p2p-adapter/src/libp2p/branchable_sync.rs:9-37` with `crates/embedded/src/node_tasks.rs:449-462`

`wait_for_heads` clears a head only when a `MergeComplete` carrying that exact CID arrives. An advertised head that becomes merged as an ancestor of another peer's newer head never produces such an event.

Interleaving: peer A advertises head `H_a`; peer B, one commit behind, advertises `H_b`, an ancestor of `H_a`. `unmerged_heads` returns both. The DAG fetch for `H_a` merges the whole chain, marking `H_b` merged. The separate fetch for `H_b` then emits `BlockReceived{cid: H_b, doc_id: ""}` (`event_handler/branchable_sync.rs:308` passes an empty doc_id), the replication loop returns `Skipped{reason: "already merged", doc_id: ""}`, and `node_tasks.rs:449-451` publishes `MergeComplete` only when `!doc_id.is_empty()` or `reason == "no linked composites needed merging"`. Neither holds. `H_b` stays pending, the call blocks the full 30s and returns `timeout ... 1 heads remain unmerged` on a collection that is fully synced.

This is user visible through `crates/http/src/handlers/p2p/collections.rs:121`, the CLI, and `crates/ffi/src/p2p/sync.rs:111`.

### Major: two of three implementations still return early

- `crates/p2p-adapter/src/iroh.rs:888-921` fires the requests into `tokio::spawn` and returns `Ok(())` immediately, waiting for nothing. The stated problem is unaddressed on the iroh transport.
- `crates/p2p-adapter/src/libp2p.rs:1065`, the non-coordinator fallback, still calls `wait_for_branchable_merges`, which breaks on a 3s idle gap and returns `()` unconditionally.
- `crates/p2p-adapter/src/libp2p.rs:958-973`, the doc-sync loop directly above, still reports success on timeout, which is the defect being fixed. The shared helper was renamed (`unmerged_doc_sync_heads` to `unmerged_heads`) as if to unify the two, then only one was unified.

### Major: the timeout message asserts a cause it has not verified

`crates/p2p-adapter/src/libp2p/branchable_sync.rs:31-34`

`ChannelBus` drops messages when a subscriber's 4096 slot buffer fills (`crates/events/src/channel_bus.rs:130-140`) and records it in `Subscription::dropped_count()`. `wait_for_heads` never reads it, so a dropped `MergeComplete` yields "N heads remain unmerged" when the heads did merge. The counter is available and should gate the wording.

### Minor

- `crates/p2p-adapter/src/libp2p.rs:1032`: `.filter_map(|head| Cid::try_from(...).ok())` discards malformed advertised heads with no log, so a peer sending garbage shrinks the pending set and the sync reports success having waited for fewer heads than were advertised.
- `crates/p2p-adapter/src/libp2p.rs:1037`: `(start + overall_timeout).into()` converts a `std::time::Instant` to a `tokio::time::Instant`. Correct in production, meaningless under a paused clock, and the four unit tests construct `tokio::time::Instant` directly so this conversion is never exercised.

The four tests in `crates/p2p-adapter/tests/libp2p/branchable_sync.rs` target `wait_for_heads` in isolation and are good. Nothing tests the integration seam where the defect lives: that the CIDs harvested from `reply.heads` are the same CIDs that later appear in `MergeComplete.cid`.

PR #1697, opened 2026-09-05
fix(p2p): stop Iroh endpoints when command owners drop

iverc:
## Context

Dropping the last Iroh transport handle leaves the endpoint running: the closed command channel disables its `select!` arm instead of entering cleanup.

## Changes

- Stop the endpoint when its last command sender is dropped.
- Reject queued and new commands before network teardown begins.
- Cover dropping the sender before startup and dropping the last transport clone.

## Testing

Both lifecycle regressions fail before the fix and pass afterward.

- [x] `cargo test --profile super-dev -p p2p --features iroh-transport --test iroh_lifecycle_tests` (2 passed)
- [x] `cargo clippy --profile super-dev -p p2p --features iroh-transport --tests -- -D warnings`

Refs #1356

### DIFF
--- a/crates/p2p/src/iroh/endpoint.rs
+++ b/crates/p2p/src/iroh/endpoint.rs
@@ -176,6 +176,10 @@ pub async fn spawn_endpoint(
     Ok((command_tx, event_rx, replicators, task))
 }
 
+#[cfg(test)]
+#[path = "../../tests/unit/iroh_endpoint_lifecycle.rs"]
+mod lifecycle_tests;
+
 /// Main event loop processing incoming connections, gossip, and commands.
 async fn run_event_loop(
     endpoint: Endpoint,
@@ -224,12 +228,19 @@ async fn run_event_loop(
     }
 
     'endpoint: loop {
+        if command_rx.is_closed() {
+            break;
+        }
         tokio::select! {
             cmd = command_rx.recv() => {
                 let Some(cmd) = cmd else { break };
                 let mut pending_cmd = Some(cmd);
                 let mut processed = 0;
                 while let Some(cmd) = pending_cmd.take() {
+                    // Closing a channel does not discard its buffered commands.
+                    if command_rx.is_closed() {
+                        break 'endpoint;
+                    }
                     let should_shutdown = handle_command(
                         cmd,
                         &resources,
--- a/crates/p2p/src/iroh/endpoint.rs
+++ b/crates/p2p/src/iroh/endpoint.rs
@@ -225,7 +225,8 @@ async fn run_event_loop(
 
     'endpoint: loop {
         tokio::select! {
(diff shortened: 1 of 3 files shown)


vertexclique (commented):
**Verdict: looks sound**, with two convention and observability nits.

The diagnosed bug is real and I confirmed it in the pre change source: `Some(cmd) = command_rx.recv()` makes `select!` permanently disable that branch once the channel closes, and `else => break` never fires because `endpoint.accept()` stays enabled, so the endpoint task ran forever holding the QUIC socket.

The `drop(command_rx)` at `endpoint.rs:292` is a genuine improvement too. Previously a command issued during the roughly 1s teardown window sat in the buffer, or blocked on a full channel, with nobody to answer; now it fails fast with `ChannelSend`, which callers already treat as connection like. All production setup paths (`crates/defra-node/src/p2p_runtime.rs:199`, `crates/embedded/src/node_p2p.rs:485`, `crates/cli/src/commands/start/server_p2p/iroh.rs:46`) retain the transport, so the new exit condition will not fire spuriously.

### Minor

- `crates/p2p/src/iroh/endpoint.rs:227,237`: buffered commands are dropped with no log, and the caller sees only `Error::ChannelReceive`, indistinguishable from a crash. Interleaving: two tasks each queue a command and drop their transport clone; the endpoint is inside `handle_command(...).await` for command 1, and on return `is_closed()` is true, so command 2, say `RemoveReplicatorCollections`, vanishes. In production every `send_command` holds a sender across its own `rx.await`, so this is confined to teardown, but a `debug!` with the discarded count would satisfy fail loud. The in src unit test explicitly asserts this discard, so it is intended behaviour, just unlogged.
- `crates/p2p/src/iroh/endpoint.rs:175-177` with `crates/p2p/tests/unit/iroh_endpoint_lifecycle.rs`: a new convention, a `#[cfg(test)] #[path = "../../tests/unit/..."] mod` that pulls a file out of `tests/` into the library build. Every existing `#[path]` in the repo stays inside `tests/`, and this crate's own in src tests are plain files under `src/` (`sync/coordinator/access_tests.rs`, `broadcast_tests.rs`, `two_stream_tests.rs`). It also creates `tests/unit/`, which is not a cargo target. #1701 introduces the same pattern with `tests/coordinator/`. Worth picking one of the two existing patterns rather than a third.
- `crates/p2p/tests/iroh_lifecycle_tests.rs:22`: the 5s budget covers `gossip.shutdown()`, capped at 1s, plus `endpoint.close()`, which has no timeout in `run_event_loop`. With relay and discovery disabled it should be instant; flagging only as a possible CI flake source. Uncertain.
- `crates/p2p/src/iroh/endpoint.rs:294-320` (pre existing, in touched code): every routine teardown step logs at `warn!`. Per the log level rule these are info or debug; a clean shutdown should not emit five warnings.

PR #1698, opened 2026-09-05
fix(p2p): bound Iroh connection waits and cancel on close

iverc:
## Context

`poll_until_connected` checks its timeout only between requests. A full command queue or missing reply can keep it waiting indefinitely, and channel closure does not interrupt its polling sleep.

## Changes

- Apply one deadline to the entire connection wait, including command sends and replies.
- Stop immediately when the endpoint closes its command receiver.
- Add paused-clock tests for blocked sends, missing replies, short deadlines, and channel closure, while preserving successful connections and transport errors.

## Testing

Five new regressions fail before the fix and pass afterward.

- [x] `cargo test --profile super-dev -p p2p --features iroh-transport --lib --test iroh_lifecycle_tests` (384 unit tests and 2 lifecycle tests passed)
- [x] `cargo clippy --profile super-dev -p p2p --features iroh-transport,libp2p-transport --tests -- -D warnings`

Refs #1356

### DIFF
--- a/crates/p2p/src/iroh/transport.rs
+++ b/crates/p2p/src/iroh/transport.rs
@@ -191,16 +191,21 @@ impl P2PTransport for IrohTransport {
     }
 
     async fn poll_until_connected(&self, peer_id: &PeerId, timeout: Duration) -> Result<()> {
-        let start = std::time::Instant::now();
-        loop {
-            let peers = self.connected_peers().await?;
-            if peers.iter().any(|p| p.as_str() == peer_id.as_str()) {
-                return Ok(());
+        let wait_for_peer = async {
+            loop {
+                let peers = self.connected_peers().await?;
+                if peers.contains(peer_id) {
+                    return Ok(());
+                }
+                tokio::time::sleep(Duration::from_millis(50)).await;
             }
-            if start.elapsed() >= timeout {
-                return Err(Error::ConnectionTimeout(peer_id.to_string()));
+        };
+        tokio::select! {
+            biased;
+            _ = self.command_tx.closed() => Err(Error::ChannelSend),
+            result = tokio::time::timeout(timeout, wait_for_peer) => {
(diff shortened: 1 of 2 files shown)


vertexclique (commented):
**Verdict: looks sound**, with one half applied fix.

The core change is correct: wrapping the whole poll loop, send plus reply plus sleep, in `tokio::time::timeout` genuinely fixes the unbounded hang, since the old code only checked `start.elapsed()` between requests and `connected_peers()` awaits a oneshot the endpoint may never answer. The five new tests do fail on the old code. I traced each: the full queue and pending reply cases blocked forever, the short deadline case slept past the deadline, and both closure cases returned `Pending`.

### Major: the libp2p sibling keeps the identical defect

`crates/p2p/src/host/handle.rs:873-890`

`HostHandle::poll_until_connected` still checks the deadline only between requests, and additionally swallows the error with `if let Ok(connected) = self.connected_peers().await`, so it can neither time out nor report why.

Scenario: `crates/p2p-adapter/src/libp2p.rs:330` calls it with a 10s timeout. If the host command queue is saturated or the host task drops a reply, `connected_peers()` awaits forever and the 10s bound is never reached, which is the exact bug this PR fixes on the iroh side. Two implementations of one contract now disagree.

### Minor

- `crates/p2p/src/iroh/transport.rs:205`: `biased` puts `command_tx.closed()` ahead of the success branch, so a peer that is connected reports `Error::ChannelSend` ("channel send error") when the endpoint shuts down concurrently. Nothing was sent, so the message misdescribes the cause. `Error::ConnectionClosed` style wording, or checking the peer list once before giving up, would be honest.
- `crates/p2p/src/iroh/transport.rs:195` (pre existing, unchanged): the iroh impl propagates the first `connected_peers()` error via `?`, aborting the wait, while the libp2p impl retries through transient errors. Divergent semantics behind one trait method.

Description matches the diff. "Five new regressions" is accurate; the other two tests preserve existing behaviour.

PR #1700, opened 2026-09-05
fix(cli): initialize node identity from the configured key type

iverc:
## Context

`--default-key-type` was saved in configuration but never used at startup. Nodes without an explicit `--identity` did not load or generate a database identity, and the HTTP identity endpoint fell back to the unrelated P2P peer key.

## Changes

- Load or create `node-identity-key` using the configured key type. Keep existing identities when the default changes, migrate legacy raw keys without changing their DID, and reject malformed keys without replacing them.
- Generate an ephemeral identity in development mode with `--no-keyring`; otherwise keep anonymous startup when neither a keyring nor an explicit identity is available.
- Wire the node identity into the database, P2P, and HTTP signing while preserving explicit `--identity` precedence and keeping caller permissions separate.
- Cover persistence, key types, malformed keys, signing, `--no-signing`, and anonymous requests. Encryption defaults and the existing secp256r1 signing guard are unchanged.

## Testing

- [x] CLI library and integration tests with default features and `release-full`, including six identity-loader tests and six live-node regressions.
- [x] Strict CLI clippy across all targets with default features and `release-full`.
- [x] Rust identity integration suite: 23 passed.
- [x] Rust ACP integration suite: 109 passed.
- [x] Rust P2P replication scenarios: 2 passed.
- [x] `cargo fmt --all --check` and `git diff --check`.

Integration checks used the rebuilt binary, an isolated root directory, and `RUST_LOG=info` so the harness could detect readiness.

Refs #1422

### DIFF
--- a/crates/cli/src/commands/start/mod.rs
+++ b/crates/cli/src/commands/start/mod.rs
@@ -1,6 +1,7 @@
 //! Start command implementation
 
 mod node;
+mod node_identity;
 mod p2p;
 mod run;
 mod server;
--- a/crates/cli/src/commands/start/node.rs
+++ b/crates/cli/src/commands/start/node.rs
@@ -164,7 +164,7 @@ impl Node {
         config: &Config,
         peer_keypair: Option<p2p::Keypair>,
         user_identity: Option<std::sync::Arc<identity::RawIdentity>>,
-        node_identity_did: Option<String>,
+        node_identity: Option<Arc<identity::RawIdentity>>,
         se_key: Option<[u8; 32]>,
     ) -> Result<ServerSetup> {
         info!("Using unified ACP store (namespace isolated in main database)");
@@ -179,7 +179,7 @@ impl Node {
             user_identity,
             acp_store,
             zanzibar_store,
-            node_identity_did,
+            node_identity,
             se_key,
         )
         .await
@@ -195,17 +195,12 @@ impl Node {
         info!("Root directory: {}", config.rootdir.display());
         info!("Data directory: {}", config.data_path().display());
 
-        // Initialize peer keypair from keyring (if P2P enabled and keyring not disabled)
-        let (peer_keypair, node_identity_did) =
-            if !config.net.p2p_disabled && !config.keyring.disabled {
-                let (kp, did) = Self::init_peer_key(&config)?;
-                (Some(kp), Some(did))
-            } else if !config.net.p2p_disabled {
-                info!("Keyring disabled, using ephemeral peer identity");
-                (None, None)
-            } else {
-                (None, None)
-            };
+        let node_identity = super::node_identity::resolve(&config, user_identity.clone())?;
+        let peer_keypair = if !config.net.p2p_disabled && !config.keyring.disabled {
+            Some(Self::init_peer_key(&config)?)
+        } else {
+            None
+        };
 
         // Load the cluster-shared searchable-encryption key from the keyring
         // (Go's getOrCreateSearchableEncryptionKey). `None` when SE or the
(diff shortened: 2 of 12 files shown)


vertexclique (commented):
**Verdict: needs discussion.** `load_or_create` is a faithful port of Go's `getOrCreateIdentity`: same `"<type>:" + raw` encoding, ed25519 raw of 64 bytes matching Go, legacy migration, `Zeroizing` on persist, and no key material in any error, which `invalid_stored_identity_is_not_replaced_or_exposed` asserts. The PR changes more than it says.

### Major: the diff changes DocID stability, and the description does not mention it

`crates/query/src/runner/plan.rs:222-230` with `tools/integration-test/tests/backup/restore.rs:106-129`

The description covers only identity wiring. The diff also changes the query planner and rewrites the backup test from "Verify doc IDs match originals (CID stability)" to "signed DocIDs may change; original IDs must still resolve as aliases."

Enabling the node identity by default enables commit signing by default, which changes the genesis block and therefore the DocID a JSON re-import produces. That contradicts the project's own parity statement, "Same document content maps to the same document ID", and is not stated anywhere in the PR.

### Major: --identity short circuits the node identity, diverging from Go

`crates/cli/src/commands/start/node_identity.rs:15-17`

`node-identity-key` is never created or loaded, and the operator's ACP key becomes the DB node identity, the P2P identity and the commit signing key. Go does the opposite: `--identity` only sets the context identity (`cli/start.go:72`) while `opts.DB().SetNodeIdentity(getOrCreateIdentity(kr, cfg))` runs unconditionally whenever the keyring is enabled (`cli/start.go:177-180`).

Consequence: `/api/v0/node/identity` reports the admin DID where Go reports the keyring node DID, and that admin DID then hits the node identity DAC bypass at `crates/db/src/collection/acp.rs:70-73`. This is not a regression, it preserves pre PR Rust behaviour, but the description frames it as intended precedence without flagging the divergence.

### Minor

- `node_identity.rs:27`: an invalid `--default-key-type` is rejected only when a key must be generated. Once `node-identity-key` exists, `--default-key-type=garbaje` is silently accepted forever, and the PR's own tests rely on this. Go has the same hole so it is parity consistent, but "validate at the trust boundary, once" argues for parsing it in `apply_to_config` (`start/mod.rs:539`) like every other typed flag.
- `node_identity.rs:26`: `resolve` now opens the keyring on every `start` without `--no-keyring`. A node run as `--no-p2p --no-searchable-encryption` without `DEFRA_KEYRING_SECRET` used to start and now exits with "keyring secret not set". Matches Go, but it is a startup breaking change absent from the description.
- `node_identity.rs:59`: legacy detection is `encoded.len() == 32`; Go's is "no `:` byte anywhere" (`cli/start.go:468-477`). Rust is strictly safer, since a legacy key containing `0x3a` migrates here and errors in Go, and Rust never overwrites a malformed entry while Go does. Worth confirming the divergence is deliberate. Marking uncertain.
- `server_http/node_options.rs:11,29,128`: `_user_identity_present` carries a leading underscore but is still read under `#[cfg(feature = "sourcehub")]`, and the call site is now four positional bools. Use a small struct or `cfg_attr`.
- `crates/query/src/runner/plan.rs:227`: `filter_map(|doc| doc.doc_id()...)` silently drops docs with no `doc_id` from the substituted filter set. Behaviourally equivalent to before, but it is a silent drop in a filter that gates which documents are returned.

Verified clean: the substitution is reachable only when `select.doc_ids.is_some()`, and in both producers that is the `get_by_ids` branch, so the index scan `ScanSource::Docs` path cannot reach it and the docID predicate is not weakened. `SelectNode::with_doc_ids` is explain only, so leaving it on the raw input IDs is correct. `RawIdentity` and `DbOptions` `Debug` impls print only the DID.

PR #1701, opened 2026-09-05
fix(p2p): wait for coordinator shutdown cleanup

iverc:
## Context

Coordinator shutdown returned immediately to concurrent callers and aborted timed-out tasks without joining them. Cancelling the first shutdown caller could also detach the tasks being drained, leaving them holding resources after shutdown appeared complete.

## Changes

- Share teardown completion across callers and keep cleanup running if a caller is cancelled.
- Keep one 500 ms graceful-drain budget, then abort and join remaining tasks before reporting completion.
- Spawn and register background work under the same lock so shutdown cannot miss a newly started task.
- Add deterministic cleanup regressions, rejected-task coverage, and a registration/shutdown race test.

## Testing

The three cleanup regressions fail before the fix and pass afterward.

- [x] P2P unit tests: 334 with default features; 453 with both transports.
- [x] Strict P2P clippy across all targets with default features and both transports.
- [x] Embedded shutdown integration tests: 6 passed, including persistent-store reopening.
- [x] Rust-to-Rust write contention integration tests: 3 passed.
- [x] Full-feature CLI build.
- [x] `cargo fmt --all --check` and `git diff --check`.

Refs #1356

### DIFF
--- a/crates/p2p/src/sync/coordinator/mod.rs
+++ b/crates/p2p/src/sync/coordinator/mod.rs
@@ -71,7 +71,7 @@ use acp::DocumentACP;
 use blockstore::Blockstore;
 use cid::Cid;
 use parking_lot::Mutex;
-use tokio::sync::{Notify, OwnedSemaphorePermit, Semaphore};
+use tokio::sync::{watch, Notify, OwnedSemaphorePermit, Semaphore};
 use tokio::task::JoinHandle;
 
 use crate::bitswap::{AccessMode, ReplicatorRegistry};
@@ -265,6 +265,7 @@ struct SyncShutdownState {
     /// remains the single source of truth and stays a plain atomic because it
     /// is read on hot paths.
     shutdown_notify: Notify,
+    shutdown_complete: watch::Sender<bool>,
     background_tasks: Mutex<Vec<JoinHandle<()>>>,
     non_authoritative_broadcast_slots: Arc<Semaphore>,
     non_authoritative_broadcast_high_water: AtomicUsize,
@@ -354,6 +355,7 @@ impl SyncShutdownHandle {
             inner: Arc::new(SyncShutdownState {
                 is_shutting_down: AtomicBool::new(false),
                 shutdown_notify: Notify::new(),
+                shutdown_complete: watch::channel(false).0,
                 background_tasks: Mutex::new(Vec::new()),
                 non_authoritative_broadcast_slots: Arc::new(Semaphore::new(
                     NON_AUTHORITATIVE_BROADCAST_TASK_LIMIT,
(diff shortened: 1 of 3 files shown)


vertexclique (commented):
**Verdict: has issues.** The direction is right and the hardest part is correct, but the graceful budget no longer bounds the whole drain.

I traced the registration race and it is genuinely closed. `begin_shutdown()` sets the flag before spawning the drain, and both `spawn_task` (`mod.rs:424-437`) and `spawn_pending_dag_fetch` check that flag while holding the same mutex the drain must acquire, so either the task is in the vector when the drain takes it or the spawn is rejected. No lost task window. `watch::wait_for` also evaluates against the current value before waiting, so a post completion `shutdown()` returns immediately.

### Major: the join phase is unbounded

`crates/p2p/src/sync/coordinator/mod.rs:582-589`

The 500ms budget now bounds only the graceful phase. After the deadline the code aborts every remaining handle and then awaits each one with no bound. `JoinHandle::abort()` only lands at the task's next yield point, so a background task inside a long synchronous section (a regolith commit, a large DAG or CAR walk, a `parking_lot` lock held by another thread) parks the drain for as long as that section runs.

No caller wraps this in a timeout: `crates/cli/src/commands/start/run.rs:162`, `crates/defra-node/src/p2p_runtime.rs:69`, `crates/embedded/src/node_p2p.rs:221` and `:532`, `crates/embedded/src/lib.rs:245`. The sibling `await_endpoint_task` (`p2p_runtime.rs:118`) does have a 5s guard.

Concrete: a push worker is inside a blocking blockstore write in `run_push_job` when the 500ms deadline expires; `pending.await` blocks until that write returns, and a SIGTERM shutdown that previously returned in 500ms now hangs. The old code aborted without joining, so this is a new failure mode.

### Major: completion can be lost silently

`crates/p2p/src/sync/coordinator/mod.rs:410-422`

Completion is published only by a detached `tokio::spawn` whose `JoinHandle` is discarded. If that task never reaches `send_replace(true)`, from a panic inside the drain or from never being polled during runtime teardown, every `shutdown()` caller parks on `wait_for` forever and silently: the `watch::Sender` lives in the same `Arc<SyncShutdownState>` the caller holds, so it never closes and `wait_for` can never return `Err`.

Publishing completion from a drop guard would release waiters on cancellation or panic, and keeping the handle would surface the panic.

### Minor

- `mod.rs:576-580`: `Ok(Err(_))` is matched and discarded in the graceful phase, so a background task that panicked during normal operation is swallowed at shutdown. `JoinError::is_panic()` is a bug, not a cancellation.
- `mod.rs:1006` and `push_worker.rs:174`: `spawn_task`'s `bool` is dropped at both sites while `spawn_background_task` logs its rejection. For `spawn_push_workers` a rejection mid loop yields a partial pool while `backlog.worker_count()` still reports the full count, so jobs queue against fewer consumers than assumed.
- `mod.rs:440-460`: `try_acquire_non_authoritative_broadcast_slot` bumps the high water mark before `spawn_task` can reject, so the reported high water counts a task that never ran.
- `crates/p2p/tests/coordinator/shutdown.rs:14-30`: `concurrent_shutdown_callers_wait_for_task_completion` proves only "not ready within one `yield_now()`". It does fail before the fix, but it would still pass if `shutdown()` returned after any short delay rather than after the task finished. The other three are strong, asserting via `Weak::upgrade()` that the future was dropped.
- New convention: a `#[cfg(test)] #[path = "../../tests/coordinator/..."]` module pulling a file out of `tests/` into the library build. #1697 introduces the same pattern with `tests/unit/`. This crate's existing in src tests are plain files under `src/`. Worth picking one of the two existing patterns rather than a third.

The description is accurate as written. "Keep one 500ms graceful drain budget, then abort and join remaining tasks before reporting completion" is literally true but understates that the join phase itself is unbounded, which is the one thing I would resolve before merge.

PR #1702, opened 2026-09-05
fix(http): honor configured TLS certificates

iverc:
## Context
The certificate flags were accepted and made CLI clients select HTTPS, but the server still listened in plaintext. Invalid certificate paths also allowed startup to succeed.

## Changes
- Serve HTTPS when both certificate paths are configured, preserving plain HTTP otherwise.
- Validate the certificate chain and matching key before starting stores or background tasks.
- Support HTTP/1.1 and HTTP/2 and close established TLS connections when the server is cancelled.
- Cover flag and config-file wiring, rejected plaintext and untrusted certificates, malformed credentials, and shutdown.

## Testing
The missing-certificate regression reproduced the original bug before the fix.

- [x] CLI and HTTP unit/integration targets: 457 passed, one existing ignored test.
- [x] Five live TLS regressions in both default and release-full + Iroh builds.
- [x] Existing Rust document-lifecycle integration test.
- [x] Strict clippy for all CLI/HTTP targets, default and release-full + Iroh.

Refs #1422

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1378,6 +1378,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "axum-server"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1df331683d982a0b9492b38127151e6453639cd34926eb9c07d4cd8c6d22bfc"
+dependencies = [
+ "arc-swap",
+ "bytes",
+ "either",
+ "fs-err",
+ "http 1.4.0",
+ "http-body 1.0.1",
+ "hyper 1.9.0",
+ "hyper-util",
+ "pin-project-lite",
+ "rustls 0.23.37",
+ "rustls-pki-types",
+ "tokio",
+ "tokio-rustls 0.26.4",
+ "tower-service",
+]
+
+[[package]]
 name = "backon"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2086,6 +2108,7 @@ dependencies = [
  "portpicker",
  "query",
  "rand 0.9.2",
+ "rcgen",
  "replication-filter",
  "reqwest 0.12.28",
  "schema",
@@ -3468,6 +3491,7 @@ dependencies = [
  "async-stream",
  "async-trait",
  "axum",
+ "axum-server",
  "base64 0.22.1",
  "cid",
  "crypto",
@@ -3483,6 +3507,7 @@ dependencies = [
  "lens",
  "query",
  "rand 0.8.5",
+ "rustls 0.23.37",
  "schema",
  "serde",
  "serde_json",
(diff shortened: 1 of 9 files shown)


vertexclique (commented):
**Verdict: looks sound on the main path**, with one parity gap.

Verified end to end: the cert reaches the served listener, since `Server::run` dispatches to `TlsConfig::serve` (`crates/http/src/server.rs:691-694`) and `Node::new` applies it after the store match so both `memory` and `regolith` get it (`node.rs:284-286`). A bad cert fails loudly before stores or background tasks (`node.rs:194`), covered by `invalid_tls_configuration_fails_before_serving` for missing, bad cert, bad key and mismatched key, asserting non zero exit and that "DefraDB node started" never appears. No half configured pair can slip through (`config/sections.rs:155-158` reached via `Config::load` then `validate`), and no cleartext fallback exists once `tls_enabled()` is true. No secret leak: `TlsConfig` has no `Debug`, and the worst key derived text in an error is a malformed PEM header line echoed by `rustls-pki-types`, not key bytes. Provider ambiguity is handled explicitly, ALPN advertises h2 and http/1.1, and `rustls`'s `tls12` feature is on so Go era TLS 1.2 clients still connect. Shutdown closes established keep alive TLS connections and is tested. The logged scheme matches the served scheme.

### Major: no auto detection of rootdir/certs

`crates/cli/src/commands/start/node.rs:194` with `crates/cli/src/config/mod.rs:230-250`

Go auto detects `<rootdir>/certs/server.crt` and `server.key` when neither path is configured (`cli/config/config.go:211-241`, called from `LoadConfig`). Rust has no equivalent; `server.crt` appears nowhere in the tree.

Scenario: an operator migrating a Go deployment drops the pair in `<rootdir>/certs/` and starts with no TLS flags. Go serves HTTPS, Rust serves cleartext with no warning. This is the remaining silent cleartext path in a PR titled "honor configured TLS certificates".

### Minor

- `crates/pg-compat/src/lib.rs:50`: with TLS configured, `--pg-address` still binds a plaintext Postgres listener. No Go parity obligation, but a node the operator believes is TLS only exposes a cleartext data port. Either reject the combination or document it.
- `crates/defra-node/src/lib.rs:1188-1205`: a second hand rolled bind and serve loop bypasses `defra_http::Server::run` and can never carry TLS. Pre existing duplication that this PR makes divergent in capability.
- `crates/http/Cargo.toml:51-52` and `crates/cli/Cargo.toml:133-134`: `axum-server`, `rustls` and `rcgen` pinned inline while `axum` and `tokio` are workspace managed. `rustls` already appears in the graph via reqwest and quinn; keep the version in `[workspace.dependencies]`.
- `crates/http/src/server.rs:119-220`: `tls: None` is the fifth field copied verbatim across four near identical constructors. Delegate three of them to one.
- `crates/cli/tests/tls_tests.rs`: no live case for a partial pair, `--pubkeypath` without `--privkeypath`. `ApiConfig::validate` and `config_sections_tests.rs:132-137` cover it statically, but the startup failure behaviour is absent from the suite that covers the other four failure modes.

PR #1703, opened 2026-09-05
fix(p2p): drain Iroh tasks without late registrations

iverc:
## Context
Iroh shutdown drained a snapshot of task handles while running tasks could still register children. Those children escaped cleanup. Block-sync tasks also kept only abort handles, so shutdown could not join them.

This follows #1698 in the Iroh lifecycle stack.

## Changes
- Spawn and register work under one lock, and close registration before draining.
- Use Tokio JoinSet to abort and join tracked tasks, including block syncs.
- Join subscription readers before finishing endpoint teardown.
- Cover late registration, cleanup-spawned work, concurrent spawning, and a full event queue with active subscriptions and syncs.

## Testing
Both late-registration regressions fail before the fix and pass afterward.

- [x] P2P library tests with both transports: 459 passed.
- [x] Iroh endpoint lifecycle tests: three passed.
- [x] Live Iroh replication and persistent-node restart test.
- [x] Full-feature CLI build with Iroh.
- [x] Strict P2P clippy for all targets with both transports.

Refs #1356

### DIFF
--- a/crates/p2p/src/iroh/endpoint.rs
+++ b/crates/p2p/src/iroh/endpoint.rs
@@ -6,12 +6,13 @@
 //! - Commands from the `IrohTransport` facade
 
 use std::collections::{HashMap, HashSet};
+use std::future::Future;
 use std::sync::Arc;
 
 use iroh::{Endpoint, EndpointId};
 use iroh_gossip::net::Gossip;
 use tokio::sync::{mpsc, oneshot};
-use tokio::task::JoinHandle;
+use tokio::task::{AbortHandle, JoinHandle, JoinSet};
 use tracing::{debug, warn};
 
 use crate::bitswap::ReplicatorRegistry;
@@ -31,6 +32,10 @@ use super::peer_map::{parse_endpoint_id, PeerMap};
 use super::protocols;
 
 const MAX_COMMAND_BATCH: usize = 16;
+
+#[cfg(test)]
+#[path = "../../tests/iroh/task_shutdown.rs"]
+mod task_shutdown_tests;
 
 /// Shared handles to the endpoint's long-lived state, cloneable into spawned
 /// tasks (dials, incoming connections, gossip heals).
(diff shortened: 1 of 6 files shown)


vertexclique (commented):
**Verdict: has issues.** The late registration fix is correct. It silently removed the drain's enforced deadline.

The registration close is right: `spawned_tasks.lock().take()` in `shutdown_tracked_tasks` and `tasks.as_mut()?` in `spawn_task` make close versus spawn atomic under one lock, and rejected work drops its captured future. The guard is released before the `future` parameter drops, so a `Drop` side re-spawn does not deadlock, which is why the `SpawnOnDrop` test passes.

### Major: the 5s drain bound is gone

`crates/p2p/src/iroh/endpoint.rs:88-94`

The old drain had a hard 5s budget with a `warn!` naming the remaining task count on expiry. `JoinSet::shutdown().await` aborts and then joins every task with no deadline. `abort()` only takes effect at an await point, so any tracked task in a synchronous stretch (a long CBOR encode, a blockstore call, a contended `parking_lot` section) stalls endpoint teardown for as long as it runs, with no log at all.

`crates/defra-node/src/p2p_runtime.rs:118` backstops this at 5s, but `crates/embedded/src/node_p2p.rs:88` (`shutdown_iroh_endpoint`) does a bare `endpoint_task.await`. An embedded node's setup rollback can now hang indefinitely where the inner 5s cap previously guaranteed return.

This is not mentioned anywhere in the PR body.

### Major: the same unbounded join for gossip readers

`crates/p2p/src/iroh/endpoint.rs:293-294`

`for reader in readers { let _ = reader.await; }` is the same unbounded join for the aborted gossip subscription readers, and it runs inside the block whose `warn!` at line 298 claims "tracked spawned tasks drained". Neither wait can report that it did not finish.

### Minor

- `crates/p2p/src/iroh/endpoint_commands.rs:659-665`: when `spawn_task` returns `None` the `reply` oneshot is dropped with no log, so the caller surfaces a generic `Error::ChannelReceive` instead of a shutdown specific reason. This is the only one of roughly 20 call sites that inspects the `Option`; `handle_publish` and `handle_publish_raw` (`endpoint_commands.rs:1087`, `1137`) return `Ok(message_id)` regardless of whether the publish task was accepted. Unreachable today because `handle_command` cannot run after the event loop exits, but the false success shape is one refactor away from being real.
- `crates/p2p/tests/iroh/task_shutdown.rs:74`: `while !completed.is_finished() { yield_now().await; }` is an unbounded spin with no deadline; if the invariant broke, the test hangs rather than failing.
- `crates/p2p/tests/iroh_lifecycle_tests.rs:57-88`: I doubt this fails on the old code. The old path aborted block sync tasks via `active_syncs` and the runtime reaps them during the roughly 1s `gossip.shutdown()` and `endpoint.close()` awaits that follow, so `events.is_closed()` would very likely already be true. Reasonable smoke test, probably not a guard for the "block syncs are now joined" claim. Marking this uncertain.

The two late registration tests are genuinely good and do fail on the old code: the old `mem::take` left the shared `Vec` live, so a late `track_task` pushed a `JoinHandle` that kept the task and its `Arc` alive.

PR #1704, opened 2026-09-05
fix(crdt): accept 4-byte Float32 deltas in CompositeDAG

Ayush7614:
Follow-up to #883 (Float32 counter support).

**Bug:** the direct `Counter` path stores Float32 deltas as 4 bytes (`counter.rs`), but both `CompositeDAG` validation gates (`apply_field_delta` and `merge` pre-validation in `crates/crdt/src/composite.rs`) hardcoded `data.len() != 8`. Any Float32 counter field merged through the document-level path — the path real writes take — was rejected with `expected 8 bytes`, and an 8-byte payload for a Float32 field passed validation only to be silently truncated by `data[..4]`.

**Fix:** branch the expected length on `NumericKind` (4 for Float32, 8 for Int64/Float64) in both gates.

**Tests:** added `test_composite_float32_counter_accepts_4_bytes` and `test_composite_counter_rejects_wrong_length` to `crates/crdt/tests/composite_tests.rs`.

**Verified:**
- `cargo test -p crdt` — all pass
- `cargo clippy -p crdt --all-targets -- -D warnings` — clean

### DIFF
--- a/crates/crdt/src/composite.rs
+++ b/crates/crdt/src/composite.rs
@@ -325,10 +325,15 @@ impl CompositeDAG {
                     data,
                 },
             ) => {
-                if data.len() != 8 {
+                let expected = match kind {
+                    NumericKind::Float32 => 4,
+                    NumericKind::Int64 | NumericKind::Float64 => 8,
+                };
+                if data.len() != expected {
                     return Err(Error::MergeError(format!(
-                        "invalid counter increment data for field '{}': expected 8 bytes, got {}",
+                        "invalid counter increment data for field '{}': expected {} bytes, got {}",
                         field_name,
+                        expected,
                         data.len()
                     )));
                 }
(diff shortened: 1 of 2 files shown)


vertexclique (changes requested):
Correct diagnosis, in code that production does not run. The two length gates in `crates/crdt/src/composite.rs` do hardcode 8, a 4-byte Float32 delta is rejected there, and an 8-byte payload for a Float32 field is truncated by the `data[..4]` slice. But `CompositeDAG` has no user outside the crdt crate: the db crate merges composites through its own handler, so "the path real writes take" in the description is not the case and no production write is affected by the bug or the fix.

What needs to change before this merges:

- The width literals. The fix repeats the `4`/`8` match in two places, on top of the `[..4]`, `[..8]` and `[u8; 4]` literals already in this file and in `counter.rs`. The width belongs in one place. Either a method on `NumericKind` built from `size_of` of the value type that both gates and the slices use, or drop the two pre-checks and let `CounterDelta::decode_int64`/`decode_float32`/`decode_float64`, which already validate the length through the array conversion, be the single check.
- The oversized case CodeRabbit asked for: a 9-byte payload on an Int64 field must be rejected, since that is the boundary the truncating slice depends on.
- CI has not run on this branch (workflows are held pending approval), so the verification in the description is local only.

PR #1705, opened 2026-09-05
fix(crdt): reject trailing bytes in decode_priority

Ayush7614:
**Bug:** `decode_varint` returns `(value, bytes_consumed)` but `decode_priority` (`crates/crdt/src/priority.rs`) discarded the count, so prefix + garbage (e.g. `[0x01, 0xFF, 0x00]`) decoded as `Ok(1)`. Both LWW (`lww.rs:272`) and CompositeDAG (`composite.rs:238`) feed raw stored priority bytes into it for conflict resolution, so a corrupted priority key with appended garbage was silently accepted as its prefix value instead of surfacing as an error. Existing tests covered overflow and truncation but not trailing garbage.

**Fix:** require `consumed == data.len()`, returning `invalid priority encoding` otherwise. Safe: writers always store canonical minimal varints via `encode_priority`, and both callers pass exact stored bytes from `reader.get()`.

**Tests:** added `test_trailing_bytes_rejected` and `test_empty_rejected`.

**Verified:**
- `cargo test -p crdt` — all pass
- `cargo clippy -p crdt --all-targets -- -D warnings` — clean

### DIFF
--- a/crates/crdt/src/priority.rs
+++ b/crates/crdt/src/priority.rs
@@ -48,11 +48,12 @@ pub fn encode_priority(priority: u64) -> EncodedPriority {
 ///
 /// # Returns
 /// * `Ok(priority)` if successful
-/// * `Err(...)` if data is invalid or incomplete
+/// * `Err(...)` if data is invalid, incomplete, or has trailing bytes
 pub fn decode_priority(data: &[u8]) -> Result<u64> {
-    decode_varint(data)
-        .map(|(value, _)| value)
-        .ok_or_else(|| Error::MergeError("invalid priority encoding".into()))
+    match decode_varint(data) {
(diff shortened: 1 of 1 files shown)


vertexclique (approved):
Correct and safe. The varint decoder returns the consumed count and the caller dropped it, so a stored priority with bytes appended decoded to its prefix. Every writer of priority bytes in the repo goes through `encode_priority` (the two CRDT sites) and the db crate writes none with a fixed width, so the stricter decode cannot reject a value we produced.

One nit: the description says writers store "canonical minimal varints", but the decoder still accepts a non-minimal encoding (`0x80 0x00` for zero), so what this change guarantees is "fully consumed", not "canonical". Worth saying that precisely in the doc comment.

CI has not run on this branch yet (workflows held pending approval); fine to merge once it is green.

PR #1706, opened 2026-09-05
feat(cursor): add fallible try_from_doc_id constructor

Ayush7614:
**Gap:** `Cursor::from_doc_id` (`crates/cursor/src/lib.rs`) accepts any string including empty, but `Cursor::decode` rejects an empty doc ID with `EmptyDocId`. So `from_doc_id("").encode()` succeeds yet the resulting token fails `decode` a page later with an opaque error instead of failing at construction.

**Fix (non-breaking):** add `try_from_doc_id` returning `Err(CursorError::EmptyDocId)` for empty input; keep `from_doc_id` unchanged and document the asymmetry on it so existing callers (planner fallback, tests) are unaffected.

**Tests:** added `try_from_doc_id_rejects_empty` and `try_from_doc_id_round_trips` to `crates/cursor/tests/codec.rs`.

**Verified:**
- `cargo test -p cursor` — all pass (9 codec + 2 go_fixtures)
- `cargo clippy -p cursor --all-targets -- -D warnings` — clean

### DIFF
--- a/crates/cursor/src/lib.rs
+++ b/crates/cursor/src/lib.rs
@@ -31,12 +31,31 @@ impl Cursor {
     /// Construct a cursor from a document ID with no key values.
     /// Used when no index is available — the planner falls back to
     /// docID-based iteration.
+    ///
+    /// The ID is not validated: an empty ID encodes fine but `decode`
+    /// rejects it with `EmptyDocId`. Prefer `try_from_doc_id` when the
+    /// ID comes from an untrusted source.
     pub fn from_doc_id(doc_id: impl Into<String>) -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


vertexclique (commented):
Recommend closing. `Cursor::from_doc_id` has no production caller: every cursor the planner builds is a struct literal (`crates/query/src/plan/cursor.rs`, `crates/query/src/planner/builder/cursor.rs`), and its only uses are two unit tests. The new `try_from_doc_id` has no caller either, so this adds public API that nothing exercises to close a gap nothing can reach. If the asymmetry between `from_doc_id` and `decode` needs closing, the change is to validate in the one existing path, not a second constructor.

PR #1707, opened 2026-09-05
fix(cli): harden GraphQL input building against control chars and bad keys

Ayush7614:
**Bugs** (`crates/cli/src/commands/client/validation.rs`, used by `document add`):

1. `escape_graphql_string` passed control characters without a short escape (backspace, form feed, NUL, etc.) through raw. The REST layer (`crates/query/src/rest/gql.rs`, tested in `rest/tests.rs`) encodes them as `\\b`, `\\f`, `\\uXXXX` per Go `valueToGQL` (`encoding/json.Marshal`) parity — the CLI copy diverged.
2. `json_to_graphql_input` interpolated object keys unquoted and infallibly, so a key like `na)me` emitted a malformed mutation that failed server-side instead of being rejected client-side (the REST layer refuses non-`Name` keys).

**Fix:** escape `\\b`/`\\f`/`\\uXXXX` like the REST layer; make `json_to_graphql_input` fallible (`Result`) validating keys via the existing `validate_identifier`, and propagate with `?` at the `document add` call site.

**Tests:** `test_escape_graphql_string_control_chars`, `test_json_to_graphql_input_scalars`, `test_json_to_graphql_input_rejects_invalid_keys`, `test_json_to_graphql_input_valid_object`.

**Verified:**
- `cargo test -p cli --lib` — 76 passed, 1 ignored (rustc 1.95 + protoc; default 1.92 toolchain cannot resolve wasmtime 46 — pre-existing env limitation)
- `cargo clippy -p cli --all-targets -- -D warnings` — clean

### DIFF
--- a/crates/cli/src/commands/client/document.rs
+++ b/crates/cli/src/commands/client/document.rs
@@ -135,7 +135,7 @@ impl DocumentAddArgs {
         let data = get_data_from_args(&self.document, &self.file)?;
         let parsed: JsonValue = serde_json::from_str(&data)?;
 
-        let input_str = json_to_graphql_input(&parsed);
+        let input_str = json_to_graphql_input(&parsed)?;
         let query = format!(
             r#"mutation {{ add_{collection}(input: {input}) {{ _docID }} }}"#,
             collection = collection,
--- a/crates/cli/src/commands/client/validation.rs
+++ b/crates/cli/src/commands/client/validation.rs
@@ -35,6 +35,8 @@ pub fn validate_identifier(name: &str) -> Result<()> {
 /// Escape special characters in a string for use in GraphQL string literals.
 ///
 /// This prevents injection attacks when interpolating user input into GraphQL queries.
+/// Control characters without a short escape are emitted as `\uXXXX`,
+/// matching the REST layer (`query::rest`) and Go's `valueToGQL`.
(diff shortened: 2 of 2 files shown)


vertexclique (changes requested):
Both bugs are real. A raw control character is not a legal GraphQL source character, so a document string carrying one made the CLI's mutation unparsable, and object keys were interpolated unquoted with no validation. The shape of the fix is the problem.

- Duplication. `crates/query/src/rest/gql.rs` already has a `json_to_graphql_input` that validates keys as GraphQL names and encodes string leaves with `serde_json::to_string`, and the CLI already depends on the query crate. This PR rewrites the same converter and the same name check a second time. Expose the REST layer's converter and delete the CLI copy; that also removes the hand-written escaper, since serde_json's string encoding is the escaping the REST layer relies on.
- The parity claim is inexact. serde_json emits the unicode escape with lowercase hex digits and does not escape DEL; this PR emits uppercase and escapes `0x7F`. Both are valid GraphQL, but the two layers still disagree byte for byte, which is the divergence the description says it removes.
- `0x20` and `0x7F` are bare literals in the escaper. Sharing the serde path removes them; failing that, `char::is_control` says what they mean.
- CI has not run on this branch (workflows held pending approval).

PR #1708, opened 2026-09-05
fix(http): reject peer addresses with whitespace at validation

Ayush7614:
**Bug:** `validate_multiaddr` (`crates/http/src/validation.rs`) accepted any non-empty string, so garbage like `'!!! not an addr !!!'` passed HTTP 400-validation and failed deep in the libp2p/iroh transport with a confusing error. Fail-fast at the edge is this module's stated purpose.

**Fix:** reject addresses containing whitespace or control characters. Conservative by design: none of the accepted shapes (libp2p multiaddr, `iroh://` URI, bare endpoint ID — all pinned by the existing `test_validate_multiaddr_valid`) may contain such characters, and full validation still happens in the P2P transport layer.

**Tests:** added `test_validate_multiaddr_rejects_inner_whitespace`.

**Verified:**
- `cargo test -p defra-http --lib` — 143 passed (rustc 1.95 + protoc)
- `cargo clippy -p defra-http --all-targets -- -D warnings` — clean

### DIFF
--- a/crates/http/src/validation.rs
+++ b/crates/http/src/validation.rs
@@ -54,6 +54,17 @@ pub fn validate_multiaddr(address: &str) -> Result<(), HttpError> {
         return Err(HttpError::BadRequest("address cannot be empty".to_string()));
     }
 
+    // None of the accepted shapes (multiaddr, iroh URI, bare endpoint
+    // ID) may contain whitespace or control characters. Rejecting them
+    // here fails fast with 400 instead of a confusing transport-layer
+    // error deeper in the stack.
+    if address.chars().any(|c| c.is_whitespace() || c.is_control()) {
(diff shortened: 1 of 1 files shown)


vertexclique (commented):
Harmless, and less than the description says. It rejects whitespace and control characters before the transport does, with no bare constants, but it is a character filter rather than a validator: `!!!not-an-addr!!!` passes exactly as before, which undercuts the stated fail-fast purpose. A real parse (multiaddr for `/`-prefixed, hex for bare endpoint ids, the `iroh://` URI) would need a new dependency in the http crate, and that is the decision to make rather than adding a partial filter.

If it merges as a partial check: the error message echoes the raw address, control characters included, into the 400 body; describe the rule without echoing the input. CI has not run on this branch (workflows held pending approval).

PR #1709, opened 2026-09-06
feat(storage): support configurable replicator retry schedules

iverc:
## Context
Replicator retry delays are hardcoded in storage, so the configured intervals cannot influence the durable retry clock. This is the storage foundation for wiring the existing flag end to end.

## Changes
- Add a validated retry schedule with the existing default ladder and bounded jitter.
- Apply it to marker registration, legacy migration, and peer rescheduling.
- Keep stored retry records unchanged and preserve existing deadlines when reopening with a different schedule.
- Saturate retry counters and deadlines instead of overflowing.

## Testing
- [x] Full storage test suite.
- [x] Five regressions covering invalid intervals, rung progression, defaults, overflow, and persisted markers across reopen.
- [x] Strict storage clippy for all targets.

Refs #1422

### DIFF
--- a/crates/storage/src/stores/mod.rs
+++ b/crates/storage/src/stores/mod.rs
@@ -26,7 +26,8 @@ pub use headstore::Headstore;
 pub use multistore::Multistore;
 pub use peerstore::Peerstore;
 pub use retry_info::{
-    rewrite_legacy_push_retry_doc_id, PushRetryMarker, PushRetryMarkerStats, RetryInfo, RetryScope,
+    rewrite_legacy_push_retry_doc_id, PushRetryMarker, PushRetryMarkerStats, RetryInfo,
+    RetrySchedule, RetryScope,
 };
 pub use rootstore::RootStore;
 pub use systemstore::Systemstore;
--- a/crates/storage/src/stores/peerstore.rs
+++ b/crates/storage/src/stores/peerstore.rs
@@ -128,6 +128,7 @@ async fn wait_before_marker_retry(_attempt: usize) {
 /// Peerstore provides storage for peer and replication metadata
 pub struct Peerstore<S: Store> {
     store: NamespacedStore<S>,
+    retry_schedule: super::RetrySchedule,
 }
 
 impl<S: Store> Peerstore<S> {
@@ -135,7 +136,14 @@ impl<S: Store> Peerstore<S> {
     pub fn new(store: Arc<S>) -> Self {
         Self {
             store: NamespacedStore::new(store, Namespace::Peerstore),
+            retry_schedule: super::RetrySchedule::default(),
         }
+    }
+
+    /// Set the runtime retry policy without rewriting persisted deadlines.
+    pub fn with_retry_schedule(mut self, schedule: super::RetrySchedule) -> Self {
+        self.retry_schedule = schedule;
+        self
     }
 
     /// Store a replicator's configuration.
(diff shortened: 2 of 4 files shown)


vertexclique (commented):
**Verdict: needs discussion.** The mechanics are correct, but the seam cannot carry the stated goal.

### Major: the schedule cannot reach the sites that bump retries

`crates/storage/src/stores/peerstore.rs:143-147`

`with_retry_schedule` is a builder on a `Peerstore` value, but `Peerstore` is not a long lived object: it is constructed ad hoc at the point of use in 15 or more places, and every site that actually bumps a retry does so on a freshly defaulted instance. `crates/db/src/merge/push_docs_common.rs:207,239,272`, `crates/db/src/merge/push_docs.rs:228,747`, `crates/p2p-adapter/src/retry.rs:39,117,166`, `crates/defra-node/src/p2p_runtime.rs:446`, `crates/embedded/src/node_p2p.rs:116`.

Scenario for the stated goal: an operator sets `replicator.retryintervals = [5, 10]`. The marker registered through `push_docs_common.rs:207`'s `Peerstore::new(db.store().clone())` still uses the 30s to 1920s default ladder, because nothing hands that site the schedule.

Completing the wiring means threading the schedule through all 15 construction sites, which is 15 places that must agree, or moving it onto the store or `DB`. The latter is the seam that satisfies "when two places must agree, they call one function".

### Minor

- `crates/storage/src/stores/retry_info.rs:30-32`: `cap()` does `self.0[(attempt as usize).min(self.0.len() - 1)]`, which underflows and panics on an empty slice. Unreachable today since the field is private, `new` rejects empty and `Default` borrows the non empty const, but `RETRY_INTERVALS_SECS` is `pub`, so the only thing keeping a panic out of the replicator path is that nobody edits that const to `&[]`.
- `retry_info.rs:22`: `new` returns `Result<Self, String>` where the rest of storage uses typed errors, so the eventual CLI wiring must re-wrap or string match. It also takes `Vec<u64>` while the existing knob is `Vec<u32>` (`crates/cli/src/config/mod.rs:47`).
- `retry_info.rs:1-8`: the new `use std::borrow::Cow;` is inserted above the file's leading `///` block, so the doc that describes `RetryInfo` now documents `use std::time::Duration;`. The misplacement is pre existing; the new line makes it more visibly wrong.

### Bounded retries

The delay is bounded, since the last rung is the cap, and `saturating_add` is the right guard now that intervals become configurable: without it a large configured interval would wrap `next_retry_unix` into the past and produce a hot retry loop. `num_retries.saturating_add(1)` also removes a debug build overflow panic. There is no maximum attempt count, so a dead peer retries forever at the cap. That matches Go and the saturating counter is harmless since it only indexes the ladder, but worth confirming it is the intended contract.

### Tests

Sound and not wall clock flaky: every bound is re-anchored (`before + floor <= next <= now() + cap` with `now()` re-sampled after), so they hold under arbitrary descheduling. `default_schedule_preserves_existing_deadlines` compares two bumps with identical key and `num_retries`, hence identical delay, and bounds the difference by elapsed time. `configured_schedule_survives_scope_updates_and_reopen` is a real persistence test that reopens with a different schedule and asserts the stored deadline is unchanged. No sleeps.

Gap: nothing exercises the schedule reaching a real retry through `p2p-adapter` or `push_docs`, which is precisely why the first finding is invisible to the suite.

### Claims

The title, "support configurable replicator retry schedules", over claims: no user visible configuration changes. `replicator_retry_intervals` is still only echoed into the Go parity options dump (`crates/cli/src/commands/start/server_http/node_options.rs:66-73`) and consumed by nothing, so it remains a dead knob. The body is honest about this, calling it a storage foundation. "Apply it to marker registration, legacy migration, and peer rescheduling" is verified at `peerstore.rs:340`, `:432`, `:627`, and no other `bump` site remains in that file.

PR #1710, opened 2026-09-06
fix(p2p): honor configured replicator retry intervals

iverc:
## Context
`--replicator-retry-intervals` was parsed and reported but never used by replication. Failure recording, replay, and the retry sweep each constructed a default schedule.

## Changes
- Wire the configured schedule through live failures, initial replay, and retry dispatch for libp2p and Iroh.
- Reject empty or zero-valued schedules before startup; preserve defaults when a config file omits the setting.
- Keep embedded and reusable node runtimes on the existing defaults.
- Exercise actual retry deadlines through both live transports rather than checking only reported options.

## Testing
- [x] Live libp2p and Iroh retry-schedule tests against the rebuilt node.
- [x] Existing filtered-replication recovery test.
- [x] CLI configuration and database replay regressions.
- [x] CLI, adapter, embedded, and node-runtime suites, apart from the baseline failure noted below.
- [x] Strict clippy across all affected crates and targets with both transports.

The broader CLI suite fails `extract_peer_id_rejects_address_without_peer_id` with Iroh enabled. The same test fails on untouched main (`7751a95f5`); it was excluded only from the remaining-suite run.

Refs #1422

### DIFF
--- a/crates/cli/src/commands/start/mod.rs
+++ b/crates/cli/src/commands/start/mod.rs
@@ -676,6 +676,7 @@ impl StartArgs {
             config.embedding.api_key_env = api_key_env.clone();
         }
         config.api.validate()?;
+        config.retry_schedule()?;
         Ok(())
     }
 }
--- a/crates/cli/src/commands/start/server_p2p/iroh.rs
+++ b/crates/cli/src/commands/start/server_p2p/iroh.rs
@@ -19,6 +19,7 @@ impl Node {
         node_identity: Option<Arc<identity::RawIdentity>>,
         se_key: Option<[u8; 32]>,
     ) -> Result<P2PSetup> {
+        let retry_schedule = config.retry_schedule()?;
         info!("Initializing P2P network (iroh)");
 
         let sync_blockstore = Arc::new(blockstore::DefraBlockstore::new(store.clone(), true));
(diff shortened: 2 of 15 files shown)


vertexclique (commented):
**Verdict: has issues (minor).** The wiring is correct and complete on the three paths it names. One retry path still ignores the configured schedule entirely.

I verified the claim by grepping every schedule writing call site (`record_push_failure`, `observe_push_head`, `reschedule_retry_peer`): all five production sites now receive a `Peerstore` carrying the configured schedule on the CLI paths. The many other `Peerstore::new` sites only clear markers or read replicator info, so they are correctly unaffected.

### Major: the at capacity backoff ignores the configured ladder

`crates/p2p-adapter/src/retry.rs:9-15, 242-247`

The "at capacity" backoff hard codes `PERSISTED_RETRY_SWEEP_INTERVAL` (2s) through `reschedule_retry_peer(peer, Some(delay))` to `RetryInfo::defer_for`, which neither consults the configured `RetrySchedule` nor bumps `num_retries`.

Scenario: a receiver whose pending DAG admission cap is full answers every push with `at capacity: receiver is saturated, back off`. The sender re-dispatches that peer every 2 seconds indefinitely regardless of `--replicator-retry-intervals=1800,3600`, and because the rung never advances the peer never climbs the ladder. This is a fixed interval unbounded retry inside the function this PR rewrites, and the body does not mention it.

### Minor

- `crates/cli/src/config/mod.rs:50-52` versus `crates/storage/src/stores/retry_info.rs:14`: the default ladder is now written twice. `default_replicator_retry_intervals()` returns the literal `vec![30, 60, 120, 240, 480, 960, 1920]` while `RETRY_INTERVALS_SECS` and `RetrySchedule::default()` hold the same list and are what `embedded` and `defra-node` still use. Editing one silently diverges the CLI's reported and applied ladder from the runtime default. This PR extracted the literal into a function and could have called the storage constant instead.
- `crates/cli/src/config/mod.rs:107-119`: `Config::validate()` now hard fails on an explicitly empty `replicator_retry_intervals: []`, which previously loaded fine because the value was unused. The serde default only covers an omitted field. Uncertain whether such a config exists in the wild, but it turns a previously silent no op into a startup abort.
- `crates/db/src/merge/push_docs.rs:95-119`: the public `push_existing_docs` wrapper still hard codes `ReplayPushConfig::default()`. The production path moved to `push_existing_docs_with_config`, so the wrapper is now a divergence trap for the next caller.

`crates/defra-node/src/p2p_runtime.rs:247,342` and `crates/embedded/src/node_p2p.rs` keep the default schedule, the body says so explicitly, and neither crate exposes a retry intervals knob, so nothing is claimed but not shipped.

### Tests

`tools/integration-test/tests/replicator_retry_common.rs` is a deadline bounded poll, 30s budget and 100ms tick, not a fixed sleep, and it asserts a real deadline (`>= before + 1800`) rather than the reported options. The default rung 0 yields 15 to 30s, so the test genuinely fails if the schedule is ignored. Sweep interval is 2s, comfortably inside the budget. `bump_with_schedule` has no underflow or panic for any accepted schedule, since `floor = max(cap/2, 1) <= cap` for all `cap >= 1`, and zeros and empties are rejected.

Issue #1693, opened 2026-09-04
Make TransactionGuard cancellation-safe for embedded transactions


jackzampolin:
An embedded caller that is cancelled after `begin_txn(false)` can leave its uncommitted transaction registered indefinitely unless application code explicitly arranges rollback. `TransactionHandle` is an opaque, cloneable ID; dropping it does not finalize the registry entry. The existing `TransactionGuard` also only logs on drop. This occurs with ordinary Rust future cancellation/task abort, where the caller cannot reach its explicit rollback branch.

Observed while implementing generation-fenced request execution in https://github.com/gents-ai/gents/pull/1371. Gents pins `03e1035f364dc0914016e9cf8bde4e4f1a000f06`; the current default-branch `crates/query/src/txn/guard.rs` still documents and implements the same drop behavior.

Minimal reproduction:

1. Begin an explicit embedded transaction and create a document inside it.
2. Suspend the task before commit/discard, retaining the transaction handle.
3. Abort the task.
4. Query through a saved copy of the handle: the uncommitted transaction remains accessible instead of returning transaction-not-found.

Please make cancellation-safe transaction ownership available in DefraDB itself, preferably by extending the existing guard or providing its owned-executor counterpart. Merely adding rollback-on-drop to every cloneable/serialized transaction ID would be unsafe; distinguish the owning guard from borrowed IDs and HTTP transaction handles.

Acceptance criteria:

- Dropping/aborting an owned embedded transaction removes its registry entry without committing writes.
- Cancellation during begin, execute, commit, and rollback has defined cleanup semantics; no detached handle is silently orphaned.
- Successful commit/rollback disarms cleanup; cleanup racing finalization cannot commit twice or roll back a different transaction.
- Define runtime-shutdown behavior and align the guard's references to an idle cleanup policy with what is actually implemented.
- Include abort/drop regression tests that assert registry removal and absence of uncommitted data, alongside normal commit/signing tests.

Gents is temporarily arranging a best-effort call to DefraDB's own `rollback_txn` from its existing transaction wrapper. It should consume the upstream ownership API and delete that application-side cleanup once available. This report establishes registry retention; it does not claim the leaked explicit transaction alone caused the separately observed cancellation/write stall.

Downstream adoption/removal tracking: https://github.com/gents-ai/gents/issues/1372.

PR #1712, opened 2026-09-06
fix(query): abandon owned transactions on cancellation

iverc:
## Context
Dropping `TransactionGuard` only logged an error, leaving uncommitted transactions registered. Commit and rollback also relinquished the handle before their futures completed.

## Changes
- Remove abandoned transactions synchronously, without relying on a runtime or cleanup task.
- Keep guard ownership until commit or rollback succeeds.
- Leave cloned and serialized handles non-owning; make repeated cleanup harmless.
- Document cancellation before registration, during finalization, and after a durable commit.

`QueryExecutor` and `TransactionRegistry` implementations must now implement or delegate the synchronous abandonment hook.

## Testing
Three real-database regressions reproduce the original leak and pass with the fix.

- [x] All 70 transaction tests, including abort, runtime shutdown, blocked finalization, and store release.
- [x] Complete query suite.
- [x] Node, embedded, and HTTP regression suites.
- [x] Strict native clippy across affected crates and targets.
- [x] CI-equivalent WASM clippy with and without default features.

Refs #1693

### DIFF
--- a/crates/db/src/txn/registry/cleanup.rs
+++ b/crates/db/src/txn/registry/cleanup.rs
@@ -7,9 +7,8 @@ impl<S: Store + 'static> DbTransactionRegistry<S> {
     ///
     /// This method finds all transactions whose last observed request was more
     /// than `max_idle_age` ago and rolls them back, freeing resources. This
-    /// should be called periodically by a background task to prevent resource
-    /// leaks from dropped `TransactionGuard`s or orphaned HTTP transaction
-    /// handles.
+    /// can be called periodically for orphaned, non-owning transaction handles.
+    /// Owning `TransactionGuard`s clean up on drop without waiting for a sweep.
     ///
     /// Returns a `CleanupResult` containing both successfully cleaned transactions
     /// and any failures. Check `result.is_complete()` to verify all cleanups succeeded.
@@ -76,7 +75,7 @@ impl<S: Store + 'static> DbTransactionRegistry<S> {
                 warn!(
                     txn_id = %txn_id,
                     idle_secs,
-                    "Cleaning up idle transaction (orphaned HTTP handle or leaked TransactionGuard?)"
+                    "Cleaning up idle transaction"
                 );
 
                 // Try to take and discard the transaction
--- a/crates/db/src/txn/registry/lifecycle.rs
+++ b/crates/db/src/txn/registry/lifecycle.rs
@@ -254,6 +254,8 @@ impl<S: Store + 'static> DbTransactionRegistry<S> {
             self.broadcaster.clone(),
         ));
 
+        // Registration and returning the handle must be one poll: cancellation
+        // before the caller can construct its guard must not orphan an entry.
         self.transactions
             .write()
             .map_err(|_| TransactionError::lock_poisoned("failed to acquire write lock for begin"))?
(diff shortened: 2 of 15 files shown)


vertexclique (commented):
**Verdict: looks sound.** The mechanism is correct; three things to raise.

The drop chain does roll back and does release locks: `TransactionGuard::drop` to `abandon_txn` to `DbTransactionRegistry::abandon`, which removes the `Arc<DbTransactionContext>` under the write lock, releases the lock, then drops it, then `LensedDocFetcher`, then `Arc<Mutex<Option<DbTxn>>>`, then `DbTxn`, then `RegolithTxn::drop` (`crates/storage/src/backends/regolith/transaction.rs:77-83`), which releases the active txn slot and lets regolith roll back. `doc_guards` and `collection_guards` are plain fields dropped with the `DbTxn`, and `crates/db/src/write/queue.rs` contains no `Drop`, `unwrap` or `expect`, so unwinding cannot panic through this path. The registry lock is never held across `drop(ctx)`, so there is no re-entrant deadlock.

Cancel safety checks out at each boundary. Begin: `begin_with_options` inserts and returns `Ok(handle)` with no await between (`registry/lifecycle.rs:257-263`), so no yield point can orphan an entry, and the added comment is accurate. Commit and rollback: `DbTransactionRegistry::commit` removes the entry before its first await, so a cancelled commit drops the ctx and the taken `DbTxn` and the guard's later `abandon` is a correct no op. Commit failure: keeping the handle until `commit_txn` returns `Ok` means a commit that fails before removal is still cleaned up by `Drop`. Right ordering.

### Minor, latent: abandon drops rather than discards

`crates/db/src/txn/registry/lifecycle.rs:271-282`

`RegolithTxn::discard` runs `callbacks.run_discard()` (`transaction.rs:243-247`); `Drop` runs only `release()`. So any `on_discard` or `on_discard_async` registered on an interactive txn silently will not run on abandonment while it does on `rollback`. Nothing registers one today (`DbTxn::on_discard`, `crates/db/src/txn/mod.rs:330`, has no callers), so it is latent, but it is a live divergence between two finalization paths and the async `take_txn()` makes the mechanical fix awkward. At minimum it deserves a note at the registration site.

### Minor, discussion: a poisoned lock policy carve out

`crates/db/src/txn/registry/lifecycle.rs:274`

`abandon` uses `write().unwrap_or_else(|e| e.into_inner())`, deliberately mutating a poisoned map, while every other method fails closed on the stated grounds (`registry/mod.rs:89-95`) that "a poisoned lock indicates a panic and potential data corruption, continuing operation would be unsafe". The reasoning for the carve out is defensible, but it is a policy change on the one lock the codebase declared fail closed, landed as a doc line.

### Minor

- `crates/query/src/executor.rs:371-377` and `crates/query/src/txn/registry.rs:58-62`: the new methods have no default impl, so every external implementor breaks. All 13 in tree implementors are updated (wasm and ffi implement neither), so this compiles, but it is an unflagged public API break in `query`.
- `crates/db/tests/txn/guard.rs:32-45`: `assert_discarded` proves release via `tokio::time::timeout(2s, store.close())`. The fixed 2s bound races `await_quiescence`'s own `close_timeout` backoff loop on a loaded box, and `active_txns` is decremented in `RegolithTxn::release()` while the handle is an `Arc<Handle>`, so a leaked `NamespaceView` or iterator clone keeps regolith's `OwnedTransaction` alive after the counter hits zero. The assertion proves the `DbTxn` wrapper was dropped, which is weaker than "regolith rolled back".
- `guard.rs:76-88`: `guard_cleanup_does_not_require_a_running_runtime` builds the fixture on one runtime, drops it, drops the guard, then runs `assert_discarded` on a second runtime. Fragile if the DB or store ever spawns a background task, which the registry's own `start_stale_transaction_cleanup` does. Passes today. Marking uncertain.
- `guard.rs:257-287`: the "cannot discard another transaction" half only proves abandoning handle N does not remove N+1, which follows trivially from `id_counter.fetch_add` never reusing ids. The durable commit survives half is real and valuable.

One nuance on the claims: "Keep guard ownership until commit or rollback succeeds" is true at the guard, but in the real registry ownership transfers at the synchronous `remove()` inside `commit()`, before the durable commit, so the retained handle is a backstop for registries that fail before removing rather than a rollback window. The doc's "A cancelled commit therefore has an unknown outcome" is an honest statement of the residual gap.

PR #1713, opened 2026-09-06
feat(node): expose an owning embedded transaction guard

### LINKED ISSUE #1693: Make TransactionGuard cancellation-safe for embedded transactions

iverc:
## Context
Embedded callers need cancellation-safe transaction ownership without using the raw executor, which bypasses the node's signing and default identity context.

## Changes
- Add `EmbeddedNode::begin_transaction_guard` and a non-cloneable `EmbeddedTransaction`.
- Route queries through the existing signed node execution path.
- Consume the guard on commit or rollback and abandon unfinished transactions on drop.
- Keep the existing manually finalized handle API unchanged.

## Testing
- [x] Drop and task-abort regressions with saved handle copies.
- [x] Explicit commit and rollback coverage.
- [x] Cryptographic signature verification before and after guarded commit.
- [x] Existing transaction signing test and complete node, embedded, and HTTP suites.
- [x] Strict native clippy, public example compilation, formatting, and diff checks.

Closes #1693

### DIFF
--- a/crates/defra-node/src/lib.rs
+++ b/crates/defra-node/src/lib.rs
@@ -35,6 +35,7 @@ mod node_acp;
 mod p2p_runtime;
 pub mod search_chunks;
 mod signed_query_runtime;
+mod transaction;
 pub mod version;
 
 use std::path::PathBuf;
@@ -68,6 +69,7 @@ pub use schema::CollectionVersion;
 pub use telemetry::{ConflictMetricsSnapshot, RetryLayerSnapshot};
 #[cfg(feature = "otel")]
 pub use telemetry::{TelemetryConfig, TelemetryHandle};
+pub use transaction::EmbeddedTransaction;
 
 #[cfg(not(target_arch = "wasm32"))]
 use signed_query_runtime::{
@@ -468,7 +470,8 @@ impl EmbeddedNode {
         self.event_bus.subscribe(event_names)
     }
 
-    /// Begin a transaction owned by this embedded node.
+    /// Begin a manually finalized transaction. Dropping its handle does not
+    /// roll it back; use [`Self::begin_transaction_guard`] for scope ownership.
     pub async fn begin_transaction(
         &self,
         readonly: bool,
--- a/crates/defra-node/src/transaction.rs
+++ b/crates/defra-node/src/transaction.rs
@@ -1,0 +1,65 @@
+use crate::{EmbeddedNode, QueryExecutor, QueryRequest, QueryResponse, TransactionHandle};
+use query::{TransactionError, TransactionGuard};
+
+/// An owning transaction that uses the node's signing and identity context.
+///
+/// Dropping it abandons uncommitted writes, including during task cancellation
+/// or runtime shutdown. Cloned handles are non-owning. Cancellation after a
+/// durable commit cannot undo it; see [`TransactionGuard`].
+#[must_use = "dropping the transaction abandons its uncommitted writes"]
+pub struct EmbeddedTransaction<'a> {
+    node: &'a EmbeddedNode,
+    guard: TransactionGuard<'a, dyn QueryExecutor>,
+}
+
+impl EmbeddedNode {
+    /// Begin a transaction that is abandoned automatically unless finalized.
+    ///
+    /// ```no_run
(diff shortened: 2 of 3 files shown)


vertexclique (commented):
**Verdict: looks sound.**

`EmbeddedTransaction` has no `Drop` of its own; the inner `TransactionGuard`'s `Drop` runs on the field, which is both correct and what keeps `commit(self)` and `rollback(self)` able to move the field out. That drop cannot panic while unwinding: `abandon` handles lock poisoning with `into_inner()`, and the entire `DbTransactionContext` to `DbTxn` to `RegolithTxn` chain has no `unwrap` or `expect` (the only `Drop` impl across db, datastore and storage is `RegolithTxn`'s atomic decrement).

All five description bullets check out. `EmbeddedTransaction` is genuinely non cloneable and `#[must_use]`; the manual handle API is untouched apart from a doc comment that now correctly says dropping its handle does not roll back. The signature test verifies `verified_block_signer_did_in_txn` before commit and `verified_block_signer_did` after, which is real coverage rather than a smoke test.

### Minor

- `crates/defra-node/src/transaction.rs:41`: `self.guard.handle().expect("active transaction")` is an `expect` on a public API of `defra-node`. Unreachable today, since the `Option` is cleared only inside `TransactionGuard::commit` and `rollback`, both of which consume the guard, but it violates the no panic on a public path rule and rests on an invariant in another crate that nothing enforces. Storing the `TransactionHandle` in `EmbeddedTransaction` at construction, since it is `Clone`, removes it.
- `crates/defra-node/src/transaction.rs:33`: `begin`, `commit` and `rollback` go through `self.runner()`, whose own doc (`crates/defra-node/src/lib.rs:496-499`) warns it "bypasses the node's signing context and default ACP identity", while `execute` correctly routes through `execute_request_in_txn`. This mirrors the existing `begin_transaction` and `commit_transaction` and the signing test shows it is fine, but a one line comment saying the split is deliberate would stop the next reader reconciling it against that warning.
- `crates/defra-node/tests/transaction_guard.rs:96`: `owned_transaction_keeps_node_signing_context` writes into `defra_core::signing::store_identity`, a process global registry, from a binary whose four tests run in parallel. Safe today; a future test asserting on the absence of a signer would race.

Note for merge order: this is stacked on #1712's branch, so it cannot land first.

Issue #1557, opened 2026-08-21
p2p: a block larger than CAR_MAX_BYTES is permanently unservable and drops its sibling blocks


mike7c2:
Both CAR collectors enforce `CAR_MAX_BYTES` (16 MiB, `crates/p2p/src/sync/car.rs:21`) with a
`break` rather than a `continue` — `collect_dag_blocks` at `car.rs:163` and
`collect_exact_blocks` at `car.rs:211`:

```rust
if total_bytes + data.len() > CAR_MAX_BYTES {
    outcome.truncated_by_bytes = true;
    break;
}
```

`total_bytes` starts at zero, so a block larger than `CAR_MAX_BYTES` fails this check on the
first iteration it is reached. It can never be served — not by this peer, not by another,
not on retry. And because it is `break`, every remaining block in that request is dropped
with it, including small ones the requester could have used.

## Consequences

- The write succeeds locally and logs nothing. The failure is visible only on the receiver.
- The server reports blocks it demonstrably holds as misses: `blockstore_hit_count=0`
  alongside `requested_presence=Some(true)`.
- The requester retries three times (`MAX_FETCH_ATTEMPTS`,
  `crates/p2p/src/sync/coordinator/dag_retry.rs:16`) and terminates with
  `"DAG fetch failed after exhausting retries and providers; document will not converge
  until the root is re-announced"`, roughly four minutes after the write.
- Siblings are collateral: in the run below the small `name` field block was dropped
  alongside the oversized one, though nothing was wrong with it.

## Repro

Two `defra-node` nodes (`--features lark,redb,p2p`, default redb backend), one-way
replicator, one document with a single field value crossing the limit. The cliff is
byte-exact:

| block bytes | result |
| --- | --- |
| 16,777,215 | replicates |
| 16,777,216 | replicates |
| 16,777,217 | fails |

Note the binding constant is `CAR_MAX_BYTES`, not `MAX_MESSAGE_SIZE` (16 MiB,
`crates/p2p/src/iroh/protocols.rs:78`) or `MAX_CAR_SIZE` (64 MiB, `protocols.rs:81`) —
neither of those is what fires.

## Ask

Serve what fits: `continue` past an oversized block instead of `break`, so siblings still
reach the requester. Separately, a block that can never fit in any CAR response needs an
explicit signal back to the requester rather than three silent retries — as it stands the
document never converges and nothing on the writing side ever learns that.

[...]

PR #1715, opened 2026-09-06
fix(p2p): serve smaller blocks after CAR size limits

iverc:
## Context
A block larger than the 16 MiB CAR payload limit stopped collection entirely, so smaller blocks later in the request or DAG were omitted too. Size-limited blocks were also counted as missing in serving diagnostics.

## Changes
- Skip blocks that do not fit and continue serving smaller roots and siblings.
- Bound each collection to 10,000 distinct CID lookups, including skipped and missing blocks.
- Distinguish missing, oversized, and served blocks in collection results and server diagnostics.

The response format and 16 MiB payload limit are unchanged. Requester-side size-limit notices follow in the next PR in this stack.

## Testing
Two regressions fail on unchanged main and pass with the fix.

- [x] Byte-limit boundaries, aggregate limits, sibling traversal, duplicate CIDs, and bounded lookups.
- [x] Serving authorization and presence-counter regressions.
- [x] P2P unit tests with default features and both transports enabled.
- [x] Strict clippy across all P2P features and targets.

Refs #1557

### DIFF
--- a/crates/p2p/src/sync/car.rs
+++ b/crates/p2p/src/sync/car.rs
@@ -11,6 +11,10 @@ use cid::Cid;
 
 use crate::error::{Error, Result};
 
+#[cfg(test)]
+#[path = "../../tests/unit/car_collection.rs"]
+mod collection_tests;
+
 /// Maximum number of blocks allowed in a single CAR response.
 ///
 /// Prevents a malicious or faulty peer from causing the server to collect and
@@ -24,6 +28,9 @@ pub const CAR_MAX_BYTES: usize = 16 * 1024 * 1024;
 #[derive(Debug, Clone, Default)]
 pub struct CarCollectOutcome {
     pub blocks: Vec<(Cid, Bytes)>,
+    pub blockstore_hits: usize,
+    pub blockstore_misses: usize,
+    pub oversized_blocks: Vec<(Cid, usize)>,
     pub truncated_by_blocks: bool,
     pub truncated_by_bytes: bool,
 }
@@ -123,10 +130,10 @@ pub fn decode_car(data: &[u8]) -> Result<CarContents> {
 
 /// Traverse DAGs from a missing frontier, collecting reachable blocks.
 ///
-/// Collection is capped at [`CAR_MAX_BLOCKS`] blocks and [`CAR_MAX_BYTES`] total
-/// bytes.  If either limit is reached the function returns the blocks collected
-/// so far without error; the caller can detect truncation by checking whether
-/// the returned slice represents a complete DAG.
+/// Collection examines at most [`CAR_MAX_BLOCKS`] distinct CIDs and retains at
+/// most [`CAR_MAX_BYTES`] block bytes. Blocks that do not fit are skipped without
+/// walking their links, so later roots and siblings can still be served.
+/// The outcome distinguishes absent blocks, oversized blocks, and truncation.
 /// All roots share the same visited set and response limits, so overlapping
 /// branches are sent once and a large frontier cannot multiply the bounded
 /// CAR response size.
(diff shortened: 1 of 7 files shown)


vertexclique (commented):
**Verdict: has issues.** The skip and continue logic is right. The byte limit still bounds a proxy.

### Major: bytes are counted only after they are already in RAM

`crates/p2p/src/sync/car.rs:170-178` and `226-234`

`blockstore.get(&cid)` materializes the whole block, and only then is `data.len()` compared to the budget. `Blockstore::get_size(cid)` exists for exactly this ("Get the size of a block without reading the entire content", `crates/blockstore/src/traits.rs:79-87`, backed by `txn.get_size` in `crates/blockstore/src/lib.rs:329-334`) and is not used.

Scenario: one 1 GiB block in a document DAG. Every CAR request that reaches that CID allocates 1 GiB on the server, discards it, and sets `truncated_by_bytes`. Because the requester keeps re-requesting on its per root clock (#1716 only stops retries within one fetch), that allocation recurs on every retry. `CAR_MAX_BYTES` bounds what is sent, not what is held.

### Minor

- `car.rs:174-179` and `230-235`: replacing `break` with `continue` on byte exhaustion changes the walk from "stop at the limit" to "read everything, keep 16 MiB". A recursive request over a 10,000 block, 40 MiB DAG previously read about 16 MiB off disk; it now reads all 40 MiB to return the same 16 MiB. Peak RSS is unchanged, one block at a time, so this is IO and CPU amplification, roughly 2.5x here and worse with large blocks.
- `car.rs:174` and `230`: a block that merely does not fit the remaining budget sets `truncated_by_bytes` but is never added to `oversized_blocks` (only `data.len() > CAR_MAX_BYTES` is). That is the right call since it is servable next round, but it means `truncated_by_bytes` and the "CAR response truncated by server side limits" WARN now conflate "come back later" with "never servable", with nothing in the code saying so.
- `event_handler/car.rs:97-102`: `record_car_serve_counts(request.wanted_cids.len(), blockstore_hit_count, ...)` mixes scales on the recursive path. `car_requested_cids` counts request CIDs while `car_present_cids` now counts every block found while walking the DAG, so a 1 CID recursive request over a 500 block DAG records requested=1, present=500. Any consumer computing `missing = requested - present` gets nonsense. The old code had the same mismatch, so not a regression, but the body claims this change makes serving diagnostics accurate.
- `event_handler/car.rs:81-90`: the oversized WARN fires once per request for a permanently unservable condition. On this PR's base there is no requester side stop, so a single oversized block produces a WARN on every retry tick, forever.

### Scope

Commit `0ccc7994c` (`proofs/tests/behavioral/mutation.rs`) is an unrelated indexed LWW flaky test fix bundled into a CAR PR. The helper itself is correct: it retries only when every GraphQL error is `Transaction Conflict. Please retry`, bails on a mixed set, requires non null `data`, and is unit tested for all three. Nits: `Command::output()` blocks the tokio worker inside an `async fn`, and the pause is a fixed 50ms rather than bounded backoff.

### Claims

"Bound each collection to 10,000 distinct CID lookups, including skipped and missing blocks" checks out and actually tightens the old miss path, which was unbounded. "Response format and 16 MiB limit unchanged" checks out. "Distinguish missing, oversized, and served blocks in server diagnostics" is only partly true: `blockstore_misses` is now real, but `oversized` reaches a log line and no counter, plus the scale mismatch above.

Tests are good: `car_collection.rs` covers the boundary exactly at `CAR_MAX_BYTES - 1`, `== CAR_MAX_BYTES` and `+1`, the aggregate case with an explicit re-request of the deferred block, sibling and descendant retention, and duplicates versus the work limit. Note `missing_blocks_count_toward_work_limit_but_duplicates_do_not` does roughly 30,000 blockstore lookups, each opening its own txn, so expect it to be slow.

PR #1716, opened 2026-09-06
fix(p2p): report unservable CAR blocks to fetch owners

### LINKED ISSUE #1557: p2p: a block larger than CAR_MAX_BYTES is permanently unservable and drops its sibling blocks

iverc:
## Context
The collector fix preserves useful sibling blocks, but receivers still need to distinguish an oversized block from a temporary provider miss.

## Changes
- Add optional oversized-block notices to Rust CAR responses, using the same authorization checks as block data.
- Store useful blocks before reporting a size limit, and accept notices only for the receiver's missing DAG frontier.
- Stop futile retries within a fetch while retaining alternate-provider recovery and libp2p's independent Bitswap fallback.
- Correlate two-stream completion with the responding peer and prevent generic failures from overwriting size-limit results.

The 16 MiB payload limit remains unchanged. Pending documents remain recoverable through the existing retry clock; older Rust peers ignore the optional notices.

## Testing
- [x] Full P2P package suite with both transports enabled, including 459 unit tests.
- [x] Default-feature unit tests and strict clippy for all-feature and Iroh-only configurations.
- [x] Malformed notices, authorization, missing-frontier validation, and peer correlation.
- [x] Retry termination and recovery through an alternate provider.
- [x] Real loopback Iroh exchange with a size-limited provider and a delayed healthy alternate.
- [x] Live two-node Iroh document sync and libp2p receiver-pull convergence.

Closes #1557

### DIFF
--- a/crates/p2p/src/iroh/endpoint_rpc.rs
+++ b/crates/p2p/src/iroh/endpoint_rpc.rs
@@ -47,6 +47,7 @@ struct CarFetchAttempt {
 #[derive(Debug, Clone)]
 enum CarFetchOutcome {
     Success,
+    SizeLimited,
     InvalidPeerId(String),
     ConnectFailed(String),
     OpenBiFailed(String),
@@ -65,6 +66,7 @@ impl CarFetchOutcome {
     fn label(&self) -> &'static str {
         match self {
             Self::Success => "success",
+            Self::SizeLimited => "size_limited",
             Self::InvalidPeerId(_) => "invalid_peer_id",
             Self::ConnectFailed(_) => "connect_failed",
             Self::OpenBiFailed(_) => "open_bi_failed",
@@ -84,6 +86,7 @@ impl CarFetchOutcome {
             | Self::WriteFailed(detail)
             | Self::ReadFailed(detail) => Some(detail),
             Self::Success
+            | Self::SizeLimited
             | Self::EmptyResponse
             | Self::HeaderOnlyCar
             | Self::EventChannelClosed => None,
@@ -774,6 +777,8 @@ async fn try_fetch_from_provider(
     // "none returned usable blocks" WARN is suppressed when every provider
     // replies empty (issue #858 review round 3).
     let has_blocks = crate::sync::car::car_has_any_block(&car_data);
+    let has_notices =
+        crate::sync::car::decode_car_oversized(&car_data).is_ok_and(|notices| !notices.is_empty());
 
     debug!(
         provider = %provider,
@@ -803,7 +808,9 @@ async fn try_fetch_from_provider(
     }
     CarFetchAttempt {
         provider: provider.clone(),
-        outcome: if has_blocks {
+        outcome: if has_notices {
+            CarFetchOutcome::SizeLimited
+        } else if has_blocks {
             CarFetchOutcome::Success
         } else {
             CarFetchOutcome::HeaderOnlyCar
@@ -897,6 +904,7 @@ pub(super) async fn handle_block_sync(
     }
 
     let mut any_success = false;
+    let mut size_limited = false;
     let mut failures = Vec::new();
     let provider_count = providers.len();
     let kind = if missing.is_empty() {
@@ -912,7 +920,10 @@ pub(super) async fn handle_block_sync(
                 tasks.abort_all();
                 break;
             }
-            Ok(attempt) => failures.push(attempt),
+            Ok(attempt) => {
+                size_limited |= matches!(attempt.outcome, CarFetchOutcome::SizeLimited);
+                failures.push(attempt);
+            }
             Err(e) => {
                 debug!("Block sync task panicked: {}", e);
                 failures.push(CarFetchAttempt {
@@ -945,7 +956,10 @@ pub(super) async fn handle_block_sync(
     // are durable. An independent success event can otherwise overtake the
     // concurrently dispatched CAR response. Failures have no useful response
     // payload to order behind and retain the aggregate completion event.
+    // Size notices also complete at ingestion; do not overwrite that result
+    // with a generic failure or cancel another provider that can still help.
     if !any_success
+        && !size_limited
         && event_tx
             .send(TransportEvent::BitswapComplete {
                 query_id,
@@ -958,6 +972,10 @@ pub(super) async fn handle_block_sync(
         warn!("Event channel closed, cannot emit BitswapComplete");
     }
 }
+
+#[cfg(test)]
+#[path = "../../tests/unit/iroh_car_size.rs"]
+mod car_size_tests;
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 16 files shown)


vertexclique (commented):
**Verdict: has issues.** The notice protocol, authorization and frontier validation are sound. The cost model and the terminal failure accounting are not.

### Major: one unservable batch abandons the whole fetch

`crates/p2p/src/sync/coordinator/dag_fetcher.rs:460-470, 806-814`

A single `FetchBatchOutcome::Unservable` returns out of the attempt and out of `poll_fetch_dag` entirely, abandoning the remaining batches of the same frontier and all remaining `MAX_FETCH_ATTEMPTS`. `SELECTIVE_FETCH_BATCH_SIZE` is 2048, and `Unservable` only requires that the batch made zero progress with every provider unservable or skipped.

Scenario: single source peer, 5000 missing CIDs. Batch 1 contains the one oversized CID plus 2047 CIDs the provider does not yet have, so `remaining == missing.len()` and `unservable == 1 == providers.len()`. Batches 2 and 3, which the provider does have, are never requested. Skipping the offending batch and continuing the walk would preserve the "store useful blocks" property.

### Major: a response that transferred zero blocks can report success

`crates/p2p/src/sync/coordinator/event_handler/mod.rs:650-665` with `crates/p2p/src/iroh/endpoint_rpc.rs:935-950`

A header only CAR carrying a notice for a CID not in the receiver's frontier yields `CarIngestDisposition::Completed`, and because `has_notices` is true the handler calls `handle_bitswap_complete(query_id, success = true, None)` while `size_limited` suppresses the aggregate `BitswapComplete{success:false}`. On the manager owned path this reaches `clear_pending_dag_fetch_failures(&root_cid)` (`event_handler/bitswap.rs:163`), resetting the root's `fetch_failures` to zero for a response that delivered nothing. A peer answering header only plus irrelevant notice can hold that counter at zero indefinitely.

I could not confirm how often the manager owned path is taken in production. The false success classification itself is definite.

### Major: authorizing a notice re-reads the entire oversized block

`crates/p2p/src/sync/coordinator/event_handler/car.rs:249-256`

To authorize a notice for a block not covered by a rooted grant, the server calls `blockstore().get(&cid)` and feeds the result to `classifier.classify(&cid, &data)`. This is the second full read of a block the collector already read and dropped, pushing an arbitrarily large payload through a DAG-CBOR classifier to decide whether to emit roughly 45 bytes. A 1 GiB block costs about 2 GiB of transient allocation and two full disk reads per request, repeated on every requester retry tick. The size is already known from the collector.

The `_ => continue` arm also swallows a blockstore error with no log, silently dropping the notice so the requester sees a plain truncation instead of a size limit.

### Major: the terminal outcome has no counter

`crates/p2p/src/sync/coordinator/dag_fetcher.rs:186-190`

`Unservable` logs an ERROR and returns with no diagnostics counter, unlike every sibling terminal path (`record_pending_dag_fetch_exhausted`, `record_pending_dag_fetch_deferred_unavailable`, `record_pending_dag_fetch_deferred_contention`). No `SyncDiagnostics` field was added. This is the state that most needs a metric, since the DAG then stays pending on the retry clock forever with no way to alert on it.

### Minor

- `event_handler/car.rs:250`: `oversized.expect("oversized candidate")` is an `expect()` on the CAR serving path. The invariant holds by construction 20 lines above, but a refactor turns it into a panic inside the coordinator event loop.
- `event_handler/car.rs:418`: `decode_car_oversized(&car_data)?` propagates, so a malformed `oversized` header rejects the whole response after `decode_car` succeeded, discarding blocks that verified fine. `decode_car_header` deliberately ignores unknown keys for forward compatibility; `is_ok_and(...)` would match.
- `dag_retry.rs:36, 65-73`: `unservable` stores exactly one CID per peer, overwritten on each `record_size_limit`. A DAG with two oversized blocks under one provider re-asks for the first after the second is recorded.
- `event_handler/car.rs:492-520`: `car_size_limit_disposition` runs a full `find_all_missing_links` walk while the `_storage_owners` guard is held, pushing concurrent arrivals for overlapping CIDs into `Busy`.
- `decode_car_oversized` hard rejects any notice with `size <= CAR_MAX_BYTES`, so two Rust peers built with different `CAR_MAX_BYTES` fail the entire response rather than ignore the notice.

Test quality here is the best of the set I reviewed: `start_paused` for deterministic retry timing, a real loopback exchange with a ready channel ordering the providers, malformed and out of bounds notice coverage. No fixed sleeps gate correctness.

PR #1718, opened 2026-09-07
feat(sync): a push can carry the grants that make what it registers readable

mike7c2:
**TL;DR** — Creating a document from an authoring client takes two steps: create
it, then add the ACL. The node announces the document on the first, so peers are
pushed a document they are not yet allowed to read. Nothing retries, and an empty
pull is indistinguishable from a document that does not exist.

Measured downstream, one writer and one browser holding a synced node: a document
written with a policy **never arrived** — checked at 10 s intervals to 100 s, with
no warning. The same document with no policy on the collection: under 3 s. It
stopped a design.

## The change

`BrowserSyncDocument` gains an optional `relationships` list, applied between
`register_doc_if_needed` and `apply_validated_document` — so registered, granted
and announced are one step. Absent on the wire means empty, so an older client
pushes exactly what it pushed before.

## Authority is unchanged

Grants apply as the authenticated caller of the push, not as the owner the push
registers. `add_actor_relationship` still decides, so a push can do what that
caller could have done with a second call to the relationship endpoint and no
more — a relay forwarding somebody else's DAG still gets `NotOwner`.

## Not covered

- A manager who is not the owner cannot grant this way (`managing_relations` is
  empty), and the relation name is not checked against the policy, so an
  unmatched relation can only ever grant less. Both need the policy store the
  adapter does not hold, and both remain the relationship endpoint's job.
- The browser client sends an empty list: it exports no `sign_*`, so it cannot
  sign a genesis, own a document, or have anything to grant. This is the
  server-side writer's path.

## Testing

Four new adapter tests: grants on the push make a protected document readable at
once; grants cannot be attached to a foreign-signed document; grants require an
authenticated caller; the owner relation and an unbounded list are refused.

`cargo test -p cli --lib` 76 passed, `defra-core` 95 passed, clippy and fmt clean.

Downstream re-measured with the instrument that found the fault: a push carrying
its grant lands in **0.06 s**, where push-then-grant never lands.

---

### DIFF
--- a/crates/cli/src/browser_sync_adapter.rs
+++ b/crates/cli/src/browser_sync_adapter.rs
@@ -6,7 +6,7 @@ use async_trait::async_trait;
 use defra_core::browser_sync::{
     BrowserSyncDocument, BrowserSyncPull, BrowserSyncRequest, BrowserSyncResponse,
     DEFAULT_SYNC_PAGE_SIZE, MAX_SYNC_BODY_BYTES, MAX_SYNC_ID_BYTES, MAX_SYNC_PAGE_SIZE,
-    MAX_SYNC_PAYLOAD_BYTES,
+    MAX_SYNC_PAYLOAD_BYTES, MAX_SYNC_RELATIONSHIPS_PER_DOCUMENT,
 };
 use defra_http::router::{BrowserSyncError, BrowserSyncOperations, BrowserSyncResult};
 use storage::corekv::Store;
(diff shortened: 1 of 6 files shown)


mike7c2:
Folded into #1720, which carries these commits unchanged. The three parts are one story — making the WASM client an authoring client that can use ACP — and splitting them cost more in review overhead than it bought in separation.

PR #1719, opened 2026-09-07
perf(wasm): answer a burst of updates in one sync exchange

mike7c2:
Stacked on #1718 — review that first; this branch contains it.

**TL;DR** — A burst of writes cost one HTTP round trip per document. A burst is
now one exchange, up to the server's limits.

Measured downstream, browser node to a reader watching the server node: a single
write is 3.0 ms median, already good enough; a burst of 20 is 24.7 ms median and
**51.5 ms** at worst. The tail is arithmetic, not congestion — twenty documents
was twenty round trips, because `run_local` called `sync_document` once per ID
and `run_remote` did the same per announced document. Nothing on the wire
required it: a request already carries a list of documents and a pull already
takes a list of IDs.

## The change

`sync_documents` packs a set of IDs into as few exchanges as the server allows —
`MAX_SYNC_DOCUMENTS_PER_REQUEST` documents, `MAX_SYNC_PULL_DOC_IDS` pull IDs, a
body under `MAX_SYNC_BODY_BYTES`.

- Packing lives in `SyncBatch`, apart from the loading, so the rules are testable
  without a node or a network.
- A document too large to push is still named by the pull, as before.
- A pull naming many documents can be cut short by the page limit where a pull
  naming one never could, so a batch follows its cursor to the end.
- The remote side batches from the buffer the SSE decoder already fills: a burst
  arrives as one chunk, and `take_decoded_document_ids` takes the rest of it into
  the same exchange.

## Testing

Five unit tests on the packing: the byte estimate matches a real serialization,
with and without a document behind a pull ID; a batch ends at the document limit,
at the pull-ID limit, and at the body limit. One adapter test on the server side:
a pull naming several documents resumes across its cursor within the named set,
and never serves a document it did not name — the adapter did this already,
nothing pinned it, and the batched client now depends on it.

`just test-wasm` 59 passed in headless Firefox, `cargo test -p cli --lib` 77
passed, `just lint-wasm` and fmt clean.

**Worth flagging:** the unit tests in this module ran nowhere — plain `#[test]`
in a crate that only compiles for wasm32, which `wasm-bindgen-test-runner` does
not collect. They are `#[wasm_bindgen_test]` now, the pre-existing size test
included.

## Not measured here

The round-trip count follows from the request shape and the packing is pinned by
tests, but the wall clock wants the downstream harness that produced the table
above. Worth re-running there before this is called a measured win.

---

### DIFF
--- a/crates/cli/src/browser_sync_adapter_tests.rs
+++ b/crates/cli/src/browser_sync_adapter_tests.rs
@@ -538,6 +538,69 @@ async fn concurrent_changes_converge_through_push_pull_exchange() {
     }
 }
 
+/// A pull naming several documents can be cut short by the page limit, and its
+/// cursor has to resume *within the named set* — resuming across the whole
+/// store would lose whatever fell off the first page.
+#[tokio::test]
+async fn a_pull_naming_several_documents_resumes_across_its_cursor() {
+    let database = Arc::new(db::DB::new(RegolithStore::in_memory().unwrap()).unwrap());
+    database
+        .create_collection(users_schema(false))
+        .await
+        .unwrap();
+
+    let mut named = Vec::new();
+    for name in ["Alice", "Bob", "Carol"] {
+        named.push(create_document(&database, name).await.doc_id);
+    }
+    // In the store, not in the pull: a cursor over the whole store serves it.
+    let unnamed = create_document(&database, "Mallory").await.doc_id;
+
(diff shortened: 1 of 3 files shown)


mike7c2:
Folded into #1720, which carries these commits unchanged. The three parts are one story — making the WASM client an authoring client that can use ACP — and splitting them cost more in review overhead than it bought in separation.

PR #1721, opened 2026-09-07
fix(perf): publish the dashboard, and stop dropping every emitted family

vertexclique:
The first `main` run after #1692 merged did not publish. `Publish` failed at the
render check, which refused to put a dashboard on `gh-pages` that was missing
data. The guard was right; two bugs behind it were mine.

## The collector was pointed at nothing

`collect` reads its family records from `$DEFRA_BENCH_OUT`. The `collect` job
carefully assembled every runner's records into `families.jsonl` and then never
set that variable, so the collector fell back to `./bench-out/`, which does not
exist in that job, and read nothing.

Every family a bench emitted was dropped on every platform: `allocs`, `memory`,
`startup`, `disk_footprint`, `document_ops`, `browser_storage`. The Linux
document came back with 29 families, all 29 harvested from criterion and not one
emitted. That is what hid it: criterion filled the page on Linux and macOS, and
only the browser, which produces no criterion output at all, came back empty
enough to trip the guard.

Reproduced from the failed run's own artifacts. Without the variable the browser
platform holds `criterion: absent` and nothing else, which is exactly the
document CI produced; with it, `document_ops` and `browser_storage` arrive with
their rows. The collector had been printing `no family records found in
$DEFRA_BENCH_OUT (unset)` in the log the whole time.

A new step now fails the collect job when a family reaches it and not the
document, naming the families that went missing, so this class of loss cannot
be silent again. Verified both ways: it passes on a good document and fails on
the broken one with `2 family record(s) reached the collector and not the
document: browser_storage, document_ops`.

## Nothing ever weighed the artifacts

`tools/perf-site/artifact_sizes.py` shipped in #1692 and no workflow called it,
so the `artifact_size` family could never exist. It gets a job: it builds the
release CLI and the wasm bundle and weighs them, raw and compressed.

## The check demanded a trend from a platform with no data

The trend plots scalars, so a platform that recorded none correctly draws
nothing, and the check called that a failure. It is now required only from a
platform that drew at least one family.

## The threshold band is inclusive

A change within plus or minus the threshold is acceptable. At the default that
means exactly -5% and exactly +5% are both fine, and only a move past either is
reported. The pull-request comment says so in its own text rather than leaving
the reader to infer it.

Boundary tested: -5.0% and +5.0% both report no regression, -6.0% reports a
regression.

## Verification

The exact failing scenario, rebuilt from the failed run's Linux, macOS and
browser artifacts, now publishes: `3 of 3 platform(s) verified, 133 cards
across them`, exit 0.

### DIFF
--- a/.github/workflows/perf-pr.yml
+++ b/.github/workflows/perf-pr.yml
@@ -2,7 +2,7 @@ name: Performance (pull request)
 
 # Its own workflow, and that is the whole reason it exists separately.
 #
-# The full matrix is forty jobs across three platforms. Keeping it in the
+# The full sweep is forty jobs across three platforms. Keeping it in the
 # workflow a pull request triggers meant every PR carried forty skipped checks,
 # and GitHub renders a skipped matrix job without expanding its matrix: the
 # check list filled up with literal `${{ matrix.family }}` rows. A pull request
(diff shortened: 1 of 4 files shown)


mike7c2 (approved):
LGTM!

PR #1722, opened 2026-09-07
feat(iroh): multiplex every protocol over one QUIC connection

mike7c2:
Every Defra protocol now shares one QUIC connection per peer.

This is a breaking protocol change!

I discovered whilst trying to make a light client we currently use ALPN for protocol discrimination. This forces a client to use a connection for each protocol. This is a mild perf concern in most contexts but a complete blocker on embedded platforms such as stm32h7 and esp32 where only a few MB of RAM is available.

The current pattern looks somewhat odd as one of the core purposes of QUIC (which Iroh is a layer on top of) is having multiple streams coexist in a single connection.

This also helps to clear ground for future work to make defra-wasm talk iroh directly.

---

The protocol discriminator lived in the QUIC ALPN, which is negotiated once per
TLS handshake and so is connection-scoped. Seventeen protocols therefore meant
up to seventeen connections to the same peer, each with its own congestion
controller, path MTU probe, hole-punch and keepalive timer. An embedded client
touching identity, replication, doc sync, branchable sync and CAR fetch paid
five of each.

## What changes

The discriminator moves in-band. One ALPN, `/defra-iroh/mux/0.1`, carries every
protocol; a stream names itself in its first frame with a length-prefixed tag,
and `dispatch_stream` routes on that tag instead of on `connection.alpn()`. The
tag is read inside the per-stream task rather than in the accept loop, so a peer
that opens a stream and stalls before writing one cannot hold up the other
protocols.

The outbound connection cache is keyed by `EndpointId` alone, and an explicit
dial adopts its connection into that cache rather than leaving the first send to
open a second one. Because one connection carries everything, eviction has to be
narrower: `evict_if_closed` drops a cached entry only once QUIC reports the
connection closed, so a stream timeout no longer tears down the transport
replication is still using. It compares `stable_id` before removing, so it
cannot evict a connection a concurrent dial has already replaced.

iroh-gossip keeps its own ALPN — it owns that handshake and takes whole
connections — so a gossiping node holds two connections to a peer.

## This is a breaking wire change

There is one implementation of this wire, so this is a clean break rather than a
migration: pre-mux peers cannot complete a handshake. Deployed iroh nodes need a
coordinated upgrade. Comments that described the old ALPN split as a
rolling-upgrade bridge are corrected accordingly.

## Tests

`mux_tests` covers the three properties that matter, against a bare iroh
endpoint standing in for a peer so the assertions are about what the transport
actually puts on the wire: six protocols share one connection, a repeated
protocol reuses it rather than redialling, and a stream-level failure does not
evict the connection the other protocols are using.

Locally green: `cargo fmt`, `cargo clippy -p p2p --features iroh-transport
--all-targets`, the 383 `p2p` unit tests, and both iroh gates CI runs —
`--test p2p_iroh connection::` (27 passed) and the two `peer::create` sync
ownership fences.

## Open for review

`read_stream_tag` has no deadline. A peer can open a bidirectional stream, never
write a tag, and pin the spawned task until the connection closes. QUIC's
concurrent-stream limit bounds it, so it is not unbounded, but it is new surface
and a timeout may be worth adding before this leaves draft.

### DIFF
--- a/crates/p2p/src/iroh/README.md
+++ b/crates/p2p/src/iroh/README.md
@@ -3,20 +3,36 @@
 DefraDB's Iroh transport uses the same `P2PTransport` surface as libp2p, but the
 wire mechanics are intentionally different where Iroh has different primitives.
 
-## Direct streams instead of `pubsub_rpc`
+## One connection per peer
 
-DocSync and BranchableSync use Iroh QUIC bidirectional streams rather than the
-`pubsub_rpc` layer introduced for libp2p in #828. The `pubsub_rpc` topic model is
-tied to libp2p peer IDs and gossipsub topic meshes. Iroh peers are addressed by
-`EndpointId`, so reusing the libp2p response-topic convention would require a
-peer-id translation layer that does not exist on the Go-compatible wire path.
+Every Defra protocol shares a single QUIC connection, negotiated on the mux
+ALPN `/defra-iroh/mux/0.1`. A stream names its protocol in-band: the first
+frame it writes is a one-byte length followed by a tag such as
+`/defra-iroh/docsync/0.1`, and `endpoint_streams::dispatch_stream` routes on
+that tag. `protocols.rs` holds the ALPN, the tags, and the frame helpers.
 
-Iroh therefore maps request/response traffic onto dedicated ALPNs:
+The discriminator has to be in-band because ALPN is negotiated once per TLS
+handshake. Naming each protocol with its own ALPN — as this transport did until
+the tags were introduced — costs a separate connection, congestion controller,
+path MTU probe, hole-punch, and keepalive timer per protocol against the same
+peer. An embedded client touching identity, replication, doc sync, branchable
+sync, and CAR fetch paid five of each.
 
-- `/defra-iroh/docsync/0.1` and `/defra-iroh/docsync/0.1/resp`
-- `/defra-iroh/branchable/0.1` and `/defra-iroh/branchable/0.1/resp`
-- `/defra-iroh/se/0.1`
-- `/defra-iroh/se-query/0.1/req` and `/defra-iroh/se-query/0.1/resp`
+Two consequences worth knowing:
+
+- The outbound connection cache in `endpoint_rpc.rs` is keyed by `EndpointId`
+  alone, and an explicit dial adopts its connection into that cache. A stream
+  failure therefore must not evict it — `evict_if_closed` drops a cached entry
+  only once QUIC itself reports the connection closed, since every other
+  protocol is still using it.
+- `iroh-gossip` keeps its own ALPN. It owns that handshake and takes whole
+  connections, so a gossiping node holds two connections to a peer, not one.
+
+DocSync and BranchableSync ride these streams rather than the `pubsub_rpc` layer
+introduced for libp2p in #828. The `pubsub_rpc` topic model is tied to libp2p
+peer IDs and gossipsub topic meshes. Iroh peers are addressed by `EndpointId`,
+so reusing the libp2p response-topic convention would require a peer-id
+translation layer that does not exist on the Go-compatible wire path.
 
 The CBOR message structs stay shared with libp2p; only the transport envelope is
 Iroh-specific. Iroh SE push and SE query messages are signed with the sending
--- a/crates/p2p/src/iroh/endpoint.rs
+++ b/crates/p2p/src/iroh/endpoint.rs
@@ -128,8 +128,12 @@ pub async fn spawn_endpoint(
     Arc<ReplicatorRegistry>,
     JoinHandle<()>,
 )> {
-    let mut alpns: Vec<Vec<u8>> = protocols::ALL_ALPNS.iter().map(|a| a.to_vec()).collect();
-    alpns.push(iroh_gossip::net::GOSSIP_ALPN.to_vec());
+    // Every Defra protocol is multiplexed over one ALPN. Gossip keeps its own:
+    // iroh-gossip owns that handshake and takes whole connections.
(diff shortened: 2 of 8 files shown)


vertexclique (approved):

vertexclique (commented):
**Verdict: needs discussion.** The mechanism is well built and the wire break is honestly disclosed. Three things are unaddressed.

First, the good news I went looking for: **Go interop is not affected**. Iroh is Rust only. Every `tools/integration-test/tests/p2p_iroh/` test builds `rust_nodes` only, no `go_nodes` appears anywhere under `p2p_iroh`, `README.md:21` names libp2p as the Go compatible transport, and no iroh ALPN constant is referenced outside `crates/p2p/src/iroh/`. The break is also clean at the connection level and versioned in the ALPN itself, so a future `/mux/0.2` fails loudly too.

### Major: a fire and forget send to a non understanding peer now returns Ok

`crates/p2p/src/iroh/endpoint_rpc.rs:506`

`send_one_way_message` discards the result of the peer's stream close (`let _ = tokio::time::timeout(..., recv.read_to_end(16)).await;` then unconditional `Ok(())`). The line is pre existing, but this PR changes what reaches it. Before, an unsupported protocol failed at `endpoint.connect()` with `Error::Dial`. Now the connect succeeds and the receiver drops the `SendStream` in the unknown tag arm (`endpoint_streams.rs:547`), which quinn turns into RESET_STREAM, which `read_to_end` returns as an error, which is discarded.

Scenario: a peer built without a protocol, or a future build that renames a tag, receives `send_manage_request` / `send_se_artifacts` / `send_car_response`. The caller gets `Ok(())`, the message is dropped, and nothing is logged above `debug` on either side.

### Major: a half upgraded fleet does not partition cleanly

`crates/p2p/src/iroh/endpoint.rs:134-135`

Gossip is the one ALPN that still interoperates across the break, so gossip peers keep discovering each other while every resulting mux dial fails at `debug` level only (`connect_with_direct_addr_fallback` logs `debug!`; `p2p-adapter/src/retry.rs:150` logs `debug!` for "replicator retry redial failed").

Scenario: an operator upgrades half a fleet, sees peers in the gossip mesh, sees no warning or error at default log level, and replication silently never happens. For an intentional flag day this deserves one `warn` saying the peer runs an incompatible protocol version, rather than a raw ALPN negotiation string at `debug`.

### Major: 17 protocols now share one 100 stream budget

`crates/p2p/src/iroh/endpoint_config.rs` (unchanged by this PR)

`max_concurrent_bidi_streams` is left at the default of 100, so all 17 protocols share one pool where each previously had its own. Combined with `endpoint_rpc.rs:506` holding each one way stream open for up to 5s after `finish()`, and `handle_command` spawning a task per command with no semaphore behind a 256 deep channel, a burst of more than 100 concurrent sends to one peer parks the rest in `open_bi` until `OPEN_STREAM_TIMEOUT` (10s). `apply_multipath_config` is the natural place to raise this.

### Major: the saturation escape hatch was removed

`crates/p2p/src/iroh/endpoint_rpc.rs:218`

The old `evict_connection` dropped the cache entry on any stream failure, so the next request redialled and got a fresh connection with fresh stream credit. `evict_if_closed` now requires `close_reason().is_some()`, so a saturated but healthy connection is never replaced. Once stream credit to a peer is exhausted, every protocol to that peer times out at `open_bi` with no path to a new connection short of an explicit `disconnect`. The narrower eviction is right; it needs a companion condition or a raised stream limit.

### Minor

- `endpoint_rpc.rs:293`: the dial guard is keyed by `EndpointId` alone, so a slow first dial serializes every protocol's first send to that peer for up to about 20s. Intended trade, worth naming in the README's consequences list.
- `endpoint_commands.rs:789`: `handle_dial` dials unconditionally and `remember_connection` overwrites a live cached entry without closing it. Connections created by `connect_with_cache` are never registered in `peer_map`, so `is_transport_connected` does not see them and a later explicit dial opens a second connection to a peer that already has one. Transient rather than a leak, but it undercuts the one connection per peer goal.
- `endpoint_streams.rs:547`: the unknown tag arm returns `Ok(())` after a `debug!`. Resetting the stream with a distinguishable QUIC error code would let the opener fail fast.
- `send_window` is 10 MiB connection wide and now shared while `MAX_CAR_SIZE` is 64 MiB. `send_fairness: true` is the default so I read this as throughput degradation rather than starvation, but I have not measured it.

### Tests

`mux_tests.rs` tests real behaviour; the third test is not vacuous, the bare peer's zero byte `finish()` makes the CAR read fail in milliseconds so `evict_if_closed` genuinely runs and genuinely declines to evict a healthy connection.

Two gaps for a change the description calls a breaking wire change:
- No mixed version test. A bare endpoint advertising only the old ALPNs, dialled by the new transport, asserting a recognizable `Error::Dial`, would be about 20 lines with the existing `localhost_endpoint` helper.
- No many streams test. The central resource question of the change has no coverage.
- `protocols.rs:65`: `ALL_STREAM_TAGS` is now `#[cfg(test)]` and `dispatch_stream` matches tags individually, so the list and the dispatch table can drift silently.

### Description vs diff

- The "read_stream_tag has no deadline is new surface" item overstates: `protocols::read_message` has no deadline either and every pre PR dispatch arm blocked on it identically. Real concern, but pre existing.
- The README's "a gossiping node holds two connections to a peer, not one" undercounts. Replies to the `*_RESP` tags are still reverse dialled on the responder's own outbound connection, so a bidirectional pair holds our mux dial, their mux dial, and gossip.
- `chunked-field-values.md:312` describes a planned `/defra-iroh/value/0.1` ALPN, which assumes the model this PR removes. It should become a stream tag, and its long lived streams are another claimant on the 100 stream pool.

PR #1723, opened 2026-09-07
ci: stop showing a literal ${{ matrix.platform }} on every pull request

vertexclique:
Every pull request carried a check named, literally:

```
CI / Release Artifacts (${{ matrix.platform }})    Skipped
```

GitHub renders a skipped matrix job without expanding its matrix, so a job with
a templated name that never runs on a pull request shows its template rather
than its values. `artifacts` was `if: github.event_name == 'push'` inside a
workflow that pull requests also trigger, so it was skipped on every one of
them and the raw expression is what the check list showed.

Renaming does not help: the name is unexpanded precisely because the matrix
never expanded. The fix is for the check not to be created at all, so the job
moves to a workflow only a push triggers. It is no longer skipped, and its name
always resolves.

The job itself is unchanged: same two-platform matrix, same eight steps, same
artifact names and retention.

## One behavioural change

It no longer waits on `build-binaries`. That was a gate rather than a
dependency: `artifacts` builds its own release binaries with
`cargo build --release -p cli` and consumed nothing from that job, and nothing
in the workflow depends on `artifacts` in turn. On a `main` that does not
compile, the cost is one failing job here instead of a skipped one, which is
the more visible of the two.

## Two others, deliberately left alone

`unit-tests` and `integration` have the same shape, gated on
`needs.ci-scope.outputs.full == 'true'`. They differ in that they are supposed
to run on pull requests, and only skip on a mid-stack PR where `ci-scope`
decides a full run is redundant.

Fixing those means dropping the job-level condition and gating every step
instead, so the job starts, exits early and reports success. That costs a runner
slot for each of the seven matrix entries on every mid-stack PR, and turns a
`skipped` check into a `success` one, which matters if any of them are required
status checks. Both are judgement calls about the stacked-PR workflow rather
than clear improvements, so they are left as they are.


mike7c2 (approved):
LGTM

PR #1740, opened 2026-09-09
Ignore unrelated Chrome APT metadata in CI

jackzampolin:
## Summary

Exclude the hosted runner's Google Chrome repository from DefraDB's
request-local APT source set, alongside the already excluded NodeSource
repository.

Chrome is preinstalled on GitHub-hosted runners and is not a dependency of the
Linux build. Its repository served mismatched `Release` and `Packages.gz`
metadata on 2026-09-09, causing every Linux job to fail before Cargo ran across
multiple unrelated PRs and retries.

This keeps the runner's global APT configuration unchanged and only filters the
source set used by DefraDB's dependency-install step.

## Validation

- `bash -n .github/scripts/apt-update-without-nodesource.sh`
- `shellcheck .github/scripts/apt-update-without-nodesource.sh`
- `git diff --check`

### DIFF
--- a/.github/scripts/apt-update-without-nodesource.sh
+++ b/.github/scripts/apt-update-without-nodesource.sh
@@ -1,17 +1,18 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# Self-hosted runners may have NodeSource configured for operator-managed Node
-# upgrades. That third-party repository is not needed by DefraDB builds, and a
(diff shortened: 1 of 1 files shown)


vertexclique:
hahaha you are amazing! @jackzampolin thanks for taking care of this.

vertexclique (approved):

PR #1744, opened 2026-09-10
feat: share one peer key between libp2p and iroh

mike7c2:
## Summary

Every node now has one Ed25519 peer key, and libp2p and iroh both use it. A node's libp2p peer ID and iroh node ID come from the same public key.

**CLI (`defra start`).** The iroh secret was derived from the keyring peer key with `derive_secret(b"iroh-transport")`. Both transports are Ed25519 and a node only runs one of them, so the derived key wasn't keeping anything apart. The iroh endpoint now uses the peer key's seed directly.

**Embedded / FFI / mobile.** A node used to keep two unrelated keys:
- a libp2p keypair stored as a fake replicator entry (`__local_p2p_identity__`)
- an iroh key in a `<db>.iroh.key` file

Now there is one seed in the peerstore at `/p2p/local-peer-key`, loaded by `embedded::node_peer_key::load_or_create` for whichever transport starts.

## Migration

Existing embedded nodes keep their identity. On the first start after this change, the key the configured transport already had is imported once:

- **iroh:** the key file (explicit path or `<db>.iroh.key`), falling back to the legacy libp2p entry
- **libp2p:** the legacy libp2p entry

After the import, the legacy libp2p entry is removed. This also ends the `failed to decode replicator info` warning that replicator restore logged for that entry on every start.

## Behaviour changes

- **CLI iroh node IDs change once.** A CLI node on the iroh transport gets a new node ID on its first start after this change. libp2p peer IDs are unchanged.
- **Corrupt keys now fail the start.** A corrupt or unreadable key fails the start instead of silently creating a new identity.
- **The iroh key path options are import-only.** `IrohConfig::secret_key_path`, `NodeInitOptions.iroh_key_path` and `p2p.iroh.keyPath` keep their shape, so the C ABI is untouched, but nothing writes the file any more. As a result, a node with an in-memory store and an explicit key path that doesn't exist yet now gets a temporary identity. Nothing in the repo uses that combination.
- **Databases that have run both transports.** If a database has both a legacy libp2p entry and an iroh key file, the key of the transport it starts with wins, and the other identity is dropped.

`defra-node` is iroh-only, so it has no second key and is unchanged.

## Test plan

- [x] `cli`: `iroh_secret_key_is_the_peer_key`
- [x] `embedded` unit tests (`node_peer_key`): key created then reused; iroh file imported once and left untouched; missing file not created; legacy libp2p entry imported and removed; iroh file wins over the legacy entry when starting on iroh; corrupt stored key, file or legacy entry fails the start
- [x] `embedded` end-to-end (`tests/peer_key.rs`): the same database started under libp2p and then iroh gives matching public keys
- [x] `embedded` `iroh_smoke`
- [x] clippy `-D warnings`: `cli` (iroh), `embedded` (libp2p+iroh, iroh-only), `ffi`, `storage`, `p2p`
- [ ] `just integration-suite p2p_iroh`: not run locally because the disk was full; relying on CI

### DIFF
--- a/crates/cli/src/commands/start/server_p2p/iroh.rs
+++ b/crates/cli/src/commands/start/server_p2p/iroh.rs
@@ -479,15 +479,21 @@ impl Node {
             manage_requester: Some(manage_requester),
         })
     }
+    /// Both transports are Ed25519, so the iroh endpoint reuses the peer key's
+    /// seed and the node keeps one identity whichever transport it runs.
     fn iroh_secret_key(peer_keypair: Option<&p2p::Keypair>) -> Result<iroh_net::SecretKey> {
-        if let Some(kp) = peer_keypair {
-            let seed = kp.derive_secret(b"iroh-transport").ok_or_else(|| {
-                Error::InvalidConfig("iroh transport requires Ed25519 key".into())
+        let Some(kp) = peer_keypair else {
+            return Ok(iroh_net::SecretKey::generate());
+        };
+        let ed25519 = kp
+            .clone()
+            .try_into_ed25519()
+            .map_err(|_| Error::InvalidConfig("iroh transport requires Ed25519 key".into()))?;
+        let seed: [u8; 32] =
+            ed25519.secret().as_ref().try_into().map_err(|_| {
+                Error::InvalidConfig("Ed25519 peer key seed must be 32 bytes".into())
             })?;
-            Ok(iroh_net::SecretKey::from_bytes(&seed))
-        } else {
-            Ok(iroh_net::SecretKey::generate())
-        }
+        Ok(iroh_net::SecretKey::from_bytes(&seed))
     }
 
     fn iroh_relay_mode(config: &Config) -> Result<p2p::iroh::IrohRelayModeConfig> {
@@ -546,5 +552,20 @@ impl Node {
             (false, None, None) => Ok(p2p::iroh::IrohDiscoveryConfig::Disabled),
             (true, None, None) => Ok(p2p::iroh::IrohDiscoveryConfig::N0),
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn iroh_secret_key_is_the_peer_key() {
+        let keypair = p2p::Keypair::generate_ed25519();
+        let libp2p_public = keypair.public().try_into_ed25519().unwrap().to_bytes();
+
+        let iroh_key = Node::iroh_secret_key(Some(&keypair)).unwrap();
+
+        assert_eq!(iroh_key.public().as_bytes(), &libp2p_public);
     }
 }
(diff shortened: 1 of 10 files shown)


iverc (commented):
reads clean overall — like that corrupt keys fail instead of silently regenerating, and the cross-transport identity test is nice.

one thing: when the iroh key file wins, you still remove the legacy libp2p entry, so the old libp2p identity is deleted without ever being imported. a node that ran both transports at different times loses that identity for good. any reason not to only clear the legacy entry when it was the one imported?

also: the shared seed sits plaintext at /p2p/local-peer-key (same as the old fake-replicator trick, to be fair) — fine for embedded/mobile, or should it go through the keyring when one is configured?

mike7c2:
Thanks @iverc — both fair, and both addressed in `0713f3b0`. Also rebased onto current `main`.

**The dropped libp2p identity.** You were right, and there was no good reason for it. The removal ran unconditionally after any import, so a database that had run both transports lost its libp2p identity the first time it started on iroh. `load_or_create` now tracks which source it imported from and removes the legacy entry only when that entry was the one imported.

What made the unconditional removal tempting is worth saying, because it needed a second fix: deleting the entry was also what silenced the `failed to decode replicator info` warning that replicator restore logged for it on every start. Keeping the entry brings that back — so `node_recovery` now passes over that slot by name. It is not a replicator, and recognising it beats deleting it to stop it complaining.

`iroh_key_file_wins_over_legacy_libp2p_key` is now `iroh_key_file_wins_without_discarding_the_libp2p_identity`, and asserts the losing key is still byte-for-byte recoverable.

**The plaintext seed.** Splitting your question in two:

On the CLI there is no issue — the peer key already lives in the keyring (`keyring::PEER_KEY`), and this PR makes iroh reuse it instead of deriving a second key and writing a key file, so that path gets strictly better.

Embedded is the real case, and the answer is that there is no keyring to route through: `crates/embedded` has no keyring dependency. It is not a regression — the fake replicator entry and the `.iroh.key` file this replaces were both unencrypted, so two plaintext secrets became one — but "not a regression" is not "fine", and on mobile the identity is readable by anything that can read the database.

Giving embedded nodes a keyring is a bigger change than this PR: it reaches the node identity and DEK material too, and on mobile it wants the platform keystore through the FFI rather than a new crate dependency. So it is #1832, and `node_peer_key`'s module docs point at it, so the next reader finds the reasoning rather than just the plaintext.

**Rebase note.** The branch carried your `sync_status().await` fix; that dropped out automatically, since your `66ae8b46` landed on `main` with #1795.

Verified: `cargo test -p embedded` green, `cargo clippy -p embedded -p cli -p ffi --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean. Still a draft — the `p2p_iroh` suite is the one I want CI to run before I take it out of draft.

PR #1752, opened 2026-09-14
fix(sync): let a /sync write join its branchable collection's DAG

mike7c2:
A document pushed to `/api/v0/sync` was stored, queryable and replicated, and never got a collection-level commit. `BranchableSync` is the only question a peer can ask without already knowing a document id, and it answers with the collection's heads — so a document no collection commit reaches is invisible to such a peer forever.

The cost is not only discovery. The workaround available to a constrained device today is to have the node author the command instead, which means the node signs it with its own key, while the device verifies against a controller DID pinned in its firmware. Those agree only if the node and the controller are the same identity. Give them separate keys, as any real deployment must, and no path remains by which an external signer reaches the device at all.

## Cause

Neither the collection DAG nor replication is implicated. A local mutation makes a head, and heads replicate node-to-node; both were re-measured against a current node before this change.

A local write writes the collection commit in the same transaction as the document. `/sync` goes through the merge handler, which stores blocks, registers ACP, replicates and serves queries — and never wrote one. `announce_merged_document` told peers so explicitly, with `collection_block: None`.

## Fix

Routing `/sync` through the local write path does not fit: a push arrives as finished blocks, not as a document to write. So the merge writes the commit, the way a local write does — in the transaction that lands the document.

- **The request.** `BlockMetadata::authoring_collection_commit(&CollectionCommitSlot)` marks a block whose merging node is the document's first holder — an ingress such as `/sync`, where no peer will ever send the commit.
- **The write.** The standalone composite merge path calls `write_collection_block` (in `merge_handler/collection_commit.rs`) inside the merge's transaction, before it commits — for the root the push named, not for any unmerged history the merge walks on the way to it. Either the document and its commit are both durable or neither is, so a push that fails part-way is a real merge again on re-send rather than a no-op. The conflict retry around `handle_block` covers the head append too.
- **The return.** The merge fills the caller's slot after its transaction commits. `/sync` owns one slot per root, so nothing is shared across pushes and nothing outlives an early return.
- **The announcement.** Each merged root's commit travels with it to the broadcaster, so it replicates like a local write's.
- **What doesn't author.** Replication passes no slot — there the commit arrives with the document, and authoring one per receiving peer would fork the collection DAG. `MergeBlock`, which the batch path takes, has no field for one. A non-branchable collection gets none. A re-sent payload merges nothing and so commits nothing.

The local mutation path is untouched; it is the reference behaviour here.

## What it costs

- **Signing.** The commit is signed with the thread's signing config, as local writes are, and the `/sync` handler sets none — so on a server these commits are unsigned. No merge path enforces signatures (`verify_block_signature` accepts an unsigned block), and a test pins that a peer takes the unsigned commit, head and document both. The document's own blocks keep the pusher's signature, which is what a device verifies, so the node never needs the controller's key.
- **Announcement after a partial failure.** A push whose later root is rejected returns an error before announcing, so an earlier root that did merge is not announced by that push. This is how document blocks already behaved on this path. For a branchable collection the next successful commit is built on the earlier one, so a peer fetching it pulls both.

## Tests

`crates/db/tests/merge/browser_sync_collection_commit.rs`, on the `/sync` path where the bug lives, reading heads through the same provider `BranchableSync` serves:

- `a_sync_write_into_an_empty_branchable_collection_makes_a_head`
- `a_sync_write_extends_a_collection_that_already_has_heads` — onto a head from a local write, then an update to the pushed document
- `a_resent_push_commits_nothing` — no new head and no announcement
- `the_collection_commit_is_announced_with_the_document`
- `a_replicated_merge_authors_no_collection_commit`
- `a_push_into_a_non_branchable_collection_writes_no_commit`

`crates/db/tests/merge/browser_sync_collection_commit_recovery.rs`:

- `a_push_whose_later_root_is_rejected_keeps_the_earlier_commit_and_recovers`
- `concurrent_pushes_leave_every_document_in_the_collection_dag` — sixteen pushes on eight threads, walking the whole collection DAG rather than counting heads, since concurrent appends may fork; 30 of 30 runs pass
- `a_peer_merges_the_unsigned_commit_it_was_announced`
- `a_push_carrying_unmerged_history_commits_once_for_its_root` — create-then-update synced as one payload: the merge walks the unmerged genesis, and only the root the push named gets a commit

The first four fail before this change. The `db`, `defra-core` and `p2p` unit suites pass (1,683 tests), as does the `client_authored` integration binary (9 tests, client-signed pushes through `/sync` end to end). `clippy --all-targets -D warnings` is clean on the changed crates and on `defra-wasm` for `wasm32-unknown-unknown`.

`issue1211_repro` could not run locally: the harness CLI rejects a retired store backend named in the local `~/.defradb/config.yaml` before the test starts. Unrelated to this change.

## Related, not included

`fix/iroh-branchable-sync` carries two commits on the other side of the same peer question — an error reply for a failed collection-head lookup, and naming a non-branchable collection as such instead of answering with an empty head set. They are complementary to this one and touch no file this PR touches, so they want their own PR.


crates/db/src/merge/browser_sync.rs
```diff
@@ -507,7 +508,16 @@ impl<S: Store + 'static> BrowserSyncEngine<S> {
             .await
             .map_err(|error| BrowserSyncError::Storage(error.to_string()))?;
 
-        self.announce_merged_document(&document).await;
+        let collection_commits = collection_commit::write_collection_commits(
```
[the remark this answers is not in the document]
mike7c2: Valid, and the error variant is the sharper half — the PR text had flagged only the crash. Fixed in 308f89ff by making the two atomic rather than adding reconciliation state.

The merge now writes the commit. `BlockMetadata::authoring_collection_commit` marks a block whose merging node is the document's first holder — an ingress such as `/sync`, where no peer will ever send the commit — and the standalone composite path writes the collection commit into the transaction that lands the document, before it commits. Either both are durable or neither is, so a resend after a failure is a merge again rather than a no-op at the empty-roots guard. The conflict retry already wrapping `handle_block` now covers the head append too.

Replication leaves the flag false: there the commit travels with the document, and authoring one per receiving peer would fork the collection DAG. `a_replicated_merge_authors_no_collection_commit` pins that, and `a_push_into_a_non_branchable_collection_writes_no_commit` pins the collection that keeps no DAG of its own. The separate-transaction helper is gone.

crates/db/src/merge/browser_sync.rs
```diff
@@ -471,18 +471,24 @@ impl<S: Store + 'static> BrowserSyncEngine<S> {
...
+                    )
+                    .authoring_collection_commit(),
                 )
                 .await
                 .map_err(|error| BrowserSyncError::Merge(error.to_string()))?
             {
-                MergeOutcome::Merged => merged = true,
+                MergeOutcome::Merged => merged_roots.push(*root),
                 MergeOutcome::Skipped { terminal: true, .. } => {}
                 MergeOutcome::Skipped { reason, .. } | MergeOutcome::Rejected { reason } => {
```
vertexclique, line 494: If an earlier root merged (and `record_authored_collection_commit` stored its commit) and a later root lands here, we return before `take_authored_collection_commit` runs. The entry stays in `authored_collection_commits` forever, and that earlier root's commit is never announced. The same happens if the storage call between the loop and the take fails. Repeated partial failures grow the map without bound. Suggest taking the commit right after each `Merged` outcome inside the loop, or returning it from the merge.

crates/db/src/merge/merge_handler/mod.rs
```diff
@@ -111,6 +112,14 @@ pub struct DbMergeHandler<S: Store, B: blockstore::Blockstore> {
...
     /// fetches.
     prefetched_dek_cids: Arc<std::sync::Mutex<HashSet<Cid>>>,
+    /// Branchable collection commits this handler authored, by the composite
+    /// root each one covers, for the ingress that asked for them.
+    ///
+    /// An ingress needs the commit to announce it to peers, and it is written
+    /// inside the merge transaction, so it cannot be handed back through the
+    /// return value of a merge that has not committed yet. Keyed by root, so
+    /// concurrent pushes cannot read each other's.
+    authored_collection_commits: std::sync::Mutex<HashMap<Cid, (Cid, bytes::Bytes)>>,
```
vertexclique: This side channel exists only because the commit can't come back through the merge's return value. Could it ride on the outcome instead (e.g. `MergeOutcome::Merged` carrying an optional collection commit, or a return from the standalone composite path)? That removes the shared blocking lock on the handler, and it removes the leak class noted in `browser_sync.rs` by construction.

crates/db/src/merge/merge_handler/mod.rs
```diff
@@ -129,6 +138,24 @@ impl<S: Store, B: blockstore::Blockstore> DbMergeHandler<S, B> {
...
 
+    /// Take the collection commit authored for `root`, if this handler wrote
+    /// one.
+    pub(crate) fn take_authored_collection_commit(
+        &self,
+        root: &Cid,
+    ) -> Option<(Cid, bytes::Bytes)> {
+        self.authored_collection_commits
+            .lock()
+            .ok()
```
vertexclique: A poisoned lock is swallowed here (and in `record_authored_collection_commit`): the document is still announced, just without its collection commit, and nothing is logged. At minimum, log it; ideally this goes away with the return-value approach.

crates/db/src/merge/merge_handler/collection_commit.rs
```diff
@@ -0,0 +1,44 @@
...
+            return Ok(None);
+        };
+
+        let commit = write_collection_block(
+            blockstore,
+            headstore,
+            collection.resolved_root_id(),
+            collection.version_id(),
+            *context.cid,
+            defra_core::signing::get_signing_config().as_ref(),
```
vertexclique, line 44: On a server node the thread-local signing config is unset, so this commit is unsigned. That matches local writes, but have we confirmed that peers enforcing signatures accept an unsigned `collection_block` from an announcement rather than rejecting the document? A test for that would pin it.

crates/defra-core/src/merge.rs
```diff
@@ -200,6 +200,15 @@ pub struct BlockMetadata<'a> {
     /// Schema blocks are governed by node-level access control (NAC) rather than
     /// document-level ACP. Doc-level permission checks must be skipped for schema blocks.
     pub is_schema_block: bool,
+    /// Whether the merging node authors the branchable collection's commit for
+    /// this block, in the merge's own transaction.
+    ///
+    /// Set only where this node is the first to hold the document, so no peer
+    /// will ever send that commit: an ingress such as `/sync`. A block that
+    /// arrived by replication leaves it false, because the commit travels with
+    /// the document. Honoured on the standalone composite merge path, which is
```
vertexclique: The flag is silently ignored on every other merge path. If `/sync` is ever routed through the batch or queued merge, the bug returns with no signal. Worth a `debug_assert`, or a test pinning that `/sync` takes the standalone composite path.

vertexclique (commented):
Overall sound: writing the collection commit inside the merge transaction is the right shape. A few items inline, plus:

**PR description is stale.** It predates 308f89ff9:
- *Durability* still says the merge and the collection commit are separate transactions and a crash between them leaves the document out of the DAG. That is no longer true; they commit together now.
- It names `browser_sync/collection_commit.rs` calling `write_branchable_collection_block`; the file is `merge_handler/collection_commit.rs` and it calls `write_collection_block` directly.

**`write/autocommit/helpers.rs`** (not in the diff): the comment in `write_branchable_collection_block` says every branchable mutation funnels through there. With this PR that stops being true, so it is worth updating.

**Tests.** Good coverage of the main cases. Missing:
- a no-op re-send commits nothing (claimed in the description);
- a multi-root push where a later root is rejected (see the leak comment);
- concurrent `/sync` pushes into the same collection (head fork/merge).

crates/db/src/merge/browser_sync.rs
mike7c2: Right, and it was worse than a leak: the stranded commit was never announced either. Fixed in f4258d9d by removing the map rather than moving the take. `/sync` now owns one `CollectionCommitSlot` per root and pairs each merged root with what its merge authored, inside the loop, so an early return drops the slot and whatever is in it.

`a_push_whose_later_root_is_rejected_keeps_the_earlier_commit_and_recovers` covers the case you describe: a root with an unverifiable signature behind one that merges. The earlier commit is durable, nothing is announced, and the honest payload then lands on top of it.

crates/db/src/merge/merge_handler/mod.rs
mike7c2: Took the spirit of this, with a narrower mechanism. Putting it on `MergeOutcome::Merged` would touch 47 match sites across crates for one caller, and changing `handle_block`'s return type is a trait change.

`BlockMetadata` already carries a lifetime, so f4258d9d replaces the flag with `authoring_collection_commit(&CollectionCommitSlot)`: a `OnceLock` the caller owns, filled after the merge's transaction commits. The request and the return channel are one thing now. Nothing is shared across pushes, so there's no lock on the handler and no leak class, by construction. The handler map and both accessors are deleted.

crates/db/src/merge/merge_handler/mod.rs
mike7c2: Gone with the map in f4258d9d. The slot is a `OnceLock`, which can't poison. The one remaining invariant, that a merge fills a slot at most once, is a `debug_assert`. It holds because a retry only follows a conflicted attempt, and a conflicted attempt never commits.

crates/db/src/merge/merge_handler/collection_commit.rs
mike7c2: Checked: nothing enforces signatures on merge. `verify_block_signature` returns `Ok(None)` for any unsigned block and the merge proceeds, on every path. A bad signature is rejected, a missing one isn't.

That's worth pinning, so f4258d9d adds `a_peer_merges_the_unsigned_commit_it_was_announced`. It takes the commit from the announcement, asserts it carries no signature, merges it on a third node through `handle_block` with replication metadata, and checks the peer ends up with the head and the document. If signature enforcement is ever added, that test is where it will show.

crates/defra-core/src/merge.rs
mike7c2: With the slot this is closed structurally rather than by assertion. `MergeBlock`, which the batch and queued paths take, has no field that can carry a slot, so those paths can't be asked to author a commit and then ignore the request. Only `handle_block` accepts one, and it reaches the standalone composite path. The field's doc comment now says so.

The pin you asked for exists too: if `/sync` were rerouted through a path that doesn't author, the slot stays empty and `a_sync_write_into_an_empty_branchable_collection_makes_a_head` fails on an empty head set.

mike7c2:
@vertexclique thanks. The inline items are answered on their threads. The review-level ones, all in f4258d9d:

- **Stale description.** Rewritten for the current shape: the commit lands in the merge's transaction, it lives in `merge_handler/collection_commit.rs` calling `write_collection_block`, and it comes back through a caller-owned `CollectionCommitSlot`. The durability caveat is gone because it no longer holds.
- **`write_branchable_collection_block` comment.** It now says every *local* branchable mutation funnels through it, and names the merge in `composite.rs` as the one other append, which reclaims its superseded heads itself.
- **Tests.**
  - *No-op re-send:* it was already the last step of the extend test. It's now its own test, `a_resent_push_commits_nothing`, and also asserts the re-send announces nothing.
  - *Later root rejected:* `a_push_whose_later_root_is_rejected_keeps_the_earlier_commit_and_recovers`.
  - *Concurrent pushes:* `concurrent_pushes_leave_every_document_in_the_collection_dag`, sixteen pushes on eight threads. It walks the whole collection DAG rather than counting heads, because concurrent appends may legitimately fork. 30/30 runs pass.
  - *Unsigned commit:* `a_peer_merges_the_unsigned_commit_it_was_announced`, from your signing comment.

The new cases live in a sibling `browser_sync_collection_commit_recovery.rs`, so neither test file passes 400 lines.

crates/db/src/merge/merge_handler/composite.rs
```diff
@@ -535,8 +535,50 @@ impl<S: Store, B: blockstore::Blockstore> DbMergeHandler<S, B> {
...
+                // together: a failure here discards the merge, and the resend
+                // that follows is a merge again rather than a no-op.
+                let commit_views = txn
+                    .blockstore()
+                    .and_then(|blockstore| Ok((blockstore, txn.headstore()?)));
+                let collection_commit = match commit_views {
+                    Ok((blockstore, headstore)) => {
+                        match self
+                            .author_collection_commit(&blockstore, &headstore, &context, &state)
+                            .await
```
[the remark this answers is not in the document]
mike7c2: Confirmed and fixed in the commit after f4258d9d. You're right about the mechanism: each unmerged ancestor's `Exit` frame reaches `process_composite_delta_body` with the same metadata, so the first ancestor filled the slot. I reproduced it before changing anything. In a debug build the `debug_assert` on the slot panics; in a release build the announcement would have carried the genesis commit instead of the root's.

It's also the common case, not an edge: a browser that creates a document and then updates it before syncing pushes one payload whose only root is the update. The existing tests pushed one revision at a time, or merged the parent first, so none of them walked an ancestor.

`process_composite_delta_body` now takes `is_root`, and `author_collection_commit` returns early for anything else. The ancestor needs no commit of its own, since the root's parents reach it. `a_push_carrying_unmerged_history_commits_once_for_its_root` pushes create-then-update as one payload into an empty collection. It asserts exactly one commit in the collection DAG, linking the root with no parents, and that this commit is the one announced. It failed with that panic before the fix.

vertexclique (approved):

Issue #208, opened 2026-01-27
feat(query): Implement CID-based queries (time travel)


jackzampolin:
## Summary

CID-based queries allow retrieving a document at a specific historical version by its content-addressed identifier.

## Current State

Queries with `cid` argument fail with: `CID-based queries are not yet implemented`

## Relationship to _version

This feature and #207 (_version field) are closely related - both require:
- DAG traversal infrastructure
- Block/commit resolution from CIDs
- Historical state reconstruction

Consider implementing together.

## What CID Queries Do

```graphql
query {
  Users(cid: "bafyrei...") {
    Name
    Age
  }
}
```

Returns the document state at that specific CID (historical version).

## Implementation Requirements

1. **Remove Validation Error**: In `runner/plan.rs` (lines 17-21)
2. **VersionedFetcher**: Go uses a specialized fetcher that:
   - Traverses CRDT DAG backwards from target CID to genesis
   - Reconstructs document state by merging deltas
   - Uses transient in-memory stores during reconstruction
3. **Block Resolution**: Map CID → block → document state

## Key Go Files

- `/internal/db/fetcher/versioned.go` - VersionedFetcher implementation
- `/internal/planner/select.go` (lines 289-344) - CID detection and fetcher selection
- `/internal/planner/scan.go` (lines 163-173) - Fetcher initialization

## Test Coverage

~15 tests in `tests/integration/query/simple/with_cid*.go`

## Complexity

HIGH - Requires implementing the versioned fetcher system with DAG traversal and state reconstruction.

jackzampolin:
Closed: VersionedFetcher implemented, CID queries working. query/commits 94/94 = 100%. See #382.

PR #1754, opened 2026-09-14
bump regolith to 0.1.7, store at RepeatableRead

vertexclique:
## What

- regolith 0.1.7 (sourcenetwork/regolith#208), and the store runs at its new `IsolationLevel::RepeatableRead`: every point read validated at commit as at `Serializable`, scans recorded per stretch and never per key. Adya's PL-2.99 over snapshot isolation.
- P2P merges hold the collection read guard for the lifetime of the transaction they write (composite, lww, counter, collection and batch paths), acquired before the merge queue and the transaction, the same order local writes use, so a truncate, delete or patch can no longer interleave with a merge. `DB::collection_write_guards` is public so a test can stand in for an in-flight truncate. One helper, `DbMergeHandler::block_collection`, resolves a block's collection for the guard and for the merge; a lookup error is an error the merge returns (and is retried on), never an absent collection that skips the guard.
- The resolver understands both ids a block can carry (review catch). A block's `schema_version_id` names a version, and metadata reconstructed from a block at crash recovery carries that same version id as its collection id, while the guard and the cache are keyed by the collection id; the two are equal only for a collection's first version. The earlier lookup compared against the collection id alone, so a recovered block of a patched collection, or a batch entry naming a version, resolved to nothing and merged without the guard. `block_collection` now tries the version id and the collection id in the cache, then the stored versions for a version no longer active, and every merge site (composite, lww, counter, batch) goes through it. The counter merge also skips an already-merged replay before it resolves the collection, as the composite does, so a replay after the collection is gone stays a terminal skip.
- HeadSet.tla gains a `ScanReads` knob with `MC_HeadSet_Red_PerKeyScan.cfg` (red: a per-key validated head scan lets the sweep abort an append); DESIGN and Lean notes updated.
- `document_create_branchable` and `index_backfill` bench groups.
- One `DB::create_index`, called by the ffi entry and the cli adapter (which the http server uses). Before this, the ffi path backfilled unguarded in one transaction and the cli/http path wrote the definition only: an index created over HTTP or the CLI did not index the documents that already existed. `next_index_id` also clears every id the schema already carries, since the two paths kept the id counter in different stores.
- Index creation is shaped like Go's: the definition, the `BACKFILL_INDEX` action and the backfill's fence (the first short id no existing document has) commit under the collection write guard, the cache reloads, the guard is released. The documents below the fence are then indexed in transactions of `BACKFILL_BATCH_DOCS` (256), holding no guard, detached from the caller's future (a client that gives up waiting does not stop the build), each document observed as well as scanned so a batch that indexed a value the document no longer has aborts and runs again (bounded by `max_txn_retries`), the last short id indexed committed with each batch under a new `/a/p` progress key. A document stored at an older schema version with a transform on its path is migrated and persisted with every index before it is indexed, per document, which replaces the whole-collection single-transaction reindex the ffi path ran afterwards. `DB::open` resumes every backfill a restart interrupted from its last durable batch, before the database takes a write, and clears an action whose index is gone. The planner does not use the index until the action completes; a failed backfill leaves the definition in place and the failure on the action.
- Writers resolve the collection definition after taking its read guard, not before (`AutoCommitMutator::guarded_collection`, the counter merge), so every write that lands after the definition commit maintains the new index itself, which is what lets the backfill hold no guard and scan only below the fence. A transaction-scoped write reads its collection's definition key again under the guard, so one whose snapshot predates a definition change fails at commit instead of writing under a definition that is gone.

## Why

regolith 0.1.6 (#206) made `Serializable` validate every key a scan yields. The branchable collection head set is derived by a scan over keys the reclamation sweep deletes, so under 0.1.6 the sweep aborted appends whose derived answer had not changed (`heads::reclamation_does_not_abort_a_concurrent_append`, the CI failure on this PR), and every append paid one commit check per head and marker walked. The head-set design needs an observed-set read for that scan and validated point reads everywhere else, which is what `RepeatableRead` is.

The first cut of `DB::create_index` on this branch held the collection write guard for the whole backfill in one transaction: every writer on the collection waited for the backfill, memory and the commit record grew with the collection, and a crash left an empty index with an in-progress action forever. Go batches with a persisted watermark, blocks no writer, and recovers at startup; this does the same.

## Numbers

Branchable append (criterion, in-memory store, same machine), Serializable vs RepeatableRead: 1 prior append 79.4 us vs 76.1 us; 14 prior appends (the most the prune interval lets accumulate) 108.0 us vs 90.1 us.

Index backfill (`index_backfill` group, criterion, in-memory store, one ordered index over a four-field collection, wall time of `DB::create_index` including the definition), the guarded single-transaction backfill of the previous commit on this branch vs the batched one:

| documents | before | after | docs/s before | docs/s after |
|---|---|---|---|---|
| 1,000 | 17.4 ms | 10.2 ms | 57.6 k | 98.0 k |
| 10,000 | 222.3 ms | 112.5 ms | 45.0 k | 88.9 k |

Write path with the guard-then-resolve order (`document_create` and `document_create_branchable`, same setup, the previous commit vs this one, paired legs back to back): 4 fields 57.1 us vs 57.3 us; 16 fields 196.2 us vs 196.6 us; 64 fields 713 us vs 717 us; branchable append with 1 prior 66.3 us vs 68.1 us, with 14 prior 76.6 us vs 78.7 us. Two legs of the unchanged code taken an hour apart differed by 2 to 4 percent on the same rows, so the branchable rows are inside the run-to-run noise; the added work per write is one cache-map read and one short id clone.

## Verification

Against the regolith branch through a local path patch (not committed): fmt, workspace clippy `-D warnings`, the `storage` suite, the full `db` suite (index 58, merge 130, vector 197, write 30, txn 70, read 44, and the rest, all green), `ffi` (169), HeadSet TLC (green plus three reds). The concurrent-writer test below ran 20 times in a row without a failure.

`index/create.rs`: three pre-existing documents indexed and the action cleared; the definition waits on a held collection write guard, writes nothing meanwhile, lands once released; a backfill over more documents than a batch holds indexes every one and clears the progress key; a multi-thread run with creates, updates of documents the batches have and have not reached, and deletes racing the whole backfill ends with exactly one entry per live document for its current value and no other; a batch whose document was updated underneath fails its commit with a transaction conflict; a store holding the definition, the in-progress action and one committed batch, reopened, finishes the backfill from the batch after the durable one and clears the action. Rust-native integration on this tree: `basic` (35 passed; the 19 `go_*` rows need the Go binary CI provides), `query -- index_management` (the http path, now backfilling), `query -- lens` (6 passed, including the three `lens_reindex_secondary_index_979` rows; 3 `go_*` rows need Go), `issue1194_repro` (4), `issue1211_repro` (1), and earlier on this branch the `p2p` and `p2p_iroh` suites with the `go_*` cross-runtime tests skipped.

`merge/collection_guard_race.rs` runs one schedule three ways, a single block with normal metadata, a block whose metadata names the schema version as recovery reconstructs it, and a batch entry naming the version; each proves the merge reached the lock rather than that it was slow: `DB::collection_read_guard` emits a trace event before and after its wait, a process-wide recorder keyed by collection id captures them, the test yields until the merge is observably waiting (deadline, no sleeps), holds the write guard through a further bounded window in which the acquire event must not appear and no key may land, then drops the guard and asserts the acquire event and the merged outcome. With the version-id lookups removed from the resolver, the recovered and batch variants fail at the first assertion (the merge finishes without ever waiting), which is the gap the review named. The recorder is process-wide on purpose: tracing computes a callsite's interest once per process on the first thread that hits it, and with a single registered dispatcher it uses that thread's default, so a thread-local subscriber silently misses any callsite another test in the binary reaches first (the first version of this test did, and passed anyway, which is exactly the false pass the review named).

## Review points

- Public `collection_write_guards`: kept public on purpose; it is the lock a truncate, delete or patch holds, and the tests stand in for one of those operations with it.
- Merge reaching the lock: proven by the trace events, three merge paths covered (composite, recovered metadata, batch). The lww and counter paths take the guard through the same resolver and the same `collection_read_guard`; a standalone field block needs an owning composite already merged, so they are covered by the merge suite rather than by the race harness.
- Lookup errors: propagated everywhere, no `.ok().flatten()` left on a collection lookup in the merge handler.
- RepeatableRead as the store default: the scan-site audit below.
- CI: red until this refresh; every failed job stopped at the unresolved `regolith = "^0.1.7"`. `Cargo.lock` now records the published 0.1.7.

## Scan-site audit: why RepeatableRead is safe as the store default

Under `RepeatableRead` a scanned key is validated when the same transaction writes it (through the write set, anchored at the begin snapshot by `scan_range::cover`) and not otherwise. `Serializable` since regolith 0.1.6 would additionally abort on any concurrent write to a scanned key. Main runs regolith 0.1.4, where `Serializable` validated no scanned key at all, so every row below has run under exactly the semantics `RepeatableRead` names, through the whole unit and integration suite, until this PR. Every transactional scan reachable on a write path in `crates/db`, classified by what the transaction does with what it scanned:

| class | sites | under RepeatableRead |
|---|---|---|
| A. deletes or overwrites the keys it scanned | `block/heads.rs` prune (head and marker scans); `index/vector/kv_store.rs` `clear`; `block/builder/mod.rs` `DocHeadsSnapshot::load` (old field heads deleted in `builder/write.rs`); `merge_handler/lww.rs` `current_field_priority` (field heads replaced in `composite_heads.rs`); `collection/ops/mod.rs` `delete_prefix`; `truncate_filtered.rs` (SE keys, document heads); `delete.rs` `truncate_chunk`, `truncate_collection_metadata`; `definition/migration/shortid.rs` moves (six scans) | the scanned keys are written, so each is validated at commit: identical to Serializable |
| B. writes only new keys of its own | `block/heads.rs` `live_collection_heads` (append and merge paths: own head key, own markers); `block/priority.rs` `backfill_commit_priority_index` (`/p/` keys) | nothing to validate at any level; a concurrent insert into the range (phantom) is undetected at every level including Serializable; for the head set, not validating the scanned keys is the design this PR exists for |
| C. read-only transaction | `merge/head_provider.rs` (3); `merge/push_docs.rs` (2); `merge/push_docs_common.rs` (2); `block/builder/mod.rs` `get_all_field_heads`; `merge/se/storage.rs` `fetch_doc_ids`; `merge/browser_sync.rs`; `collection/provider.rs` `list_collections`; `definition/loader.rs` `get_collection_version_ids`; vector `iterate_nodes` and `iterate_aux` on the search path | no commit validation at any level |
| D. writes other keys derived from the scan, under a lock that excludes concurrent writers | `collection/ops/version.rs` `set_active_collection_version` (collection write guard; the per-version keys are point-read, so validated); `definition/migration/reindex.rs` and `shortid.rs` document scans (patch or migration write guard) | the writer Serializable would catch is excluded by the lock |
| E. writes other keys derived from the scan, and point-reads every scanned key it uses | index creation backfill (`index/backfill.rs`, this PR): `BackfillSource` observes each document it yields, and the deletion marker and version key are point-read as they always were | each document a batch indexed is validated at commit: identical to Serializable for the keys that matter, without validating the rest of the range |
| F. writes other keys derived from the scan, no excluding lock, no observed reads | vector index build (`iterate_nodes`, then centroid, list and adjacency aux keys) | the one class where Serializable would abort on a concurrent overwrite of a scanned key and RepeatableRead commits. An approximate structure: a stale sample lowers recall, not correctness, and the aux keys are rebuilt on the next build. |

Not established by the audit: callers outside `crates/db` of the public `get_all_field_heads` and `fetch_doc_ids` (only read-only callers exist in tree); whether a vector search ever executes inside a writable fetcher transaction in practice (it can hold one; the search path writes nothing either way).


jackzampolin (changes requested):
The upstream isolation implementation is sound, and I locally verified this PR's new merge-guard regression against the Regolith #208 head. I found several integration blockers before DefraDB should make this its default:

1. `crates/db/src/collection/locks.rs` changes `collection_write_guards` from `pub(crate)` to `pub` solely so an integration test can hold it. That exposes an internal synchronization primitive which arbitrary callers can retain indefinitely, and makes the lock ordering part of the public API. Keep it crate-private and place the regression where it can exercise internals, or coordinate a real truncate/delete/patch operation through a deterministic test hook.

2. The new test does not establish that its spawned merge reached the collection-lock acquisition. Sixty-four `yield_now` calls followed by `!task.is_finished()` can pass simply because the task was not scheduled. Please add a barrier/instrumented hook that proves the merge reached the blocking point before asserting that it cannot write, and cover each guarded merge path (batch, collection, composite, counter, and LWW), not only composite dispatch.

3. `merge_handler/composite.rs` converts both `find_collection_by_id` errors to `None`; the fallback lookups in counter and LWW do the same. A poisoned collection registry therefore silently disables the very guard this PR establishes as a correctness invariant. Propagate lookup errors and distinguish "not found" from "lookup failed."

4. The motivating exception is a stale-tolerant head-set scan, but `RegolithStoreOptions::new()` changes every DefraDB transaction from Serializable to RepeatableRead. The HeadSet proof does not cover other scan-then-write transactions. Please either scope RepeatableRead to the operations that require it, or include a concrete audit and regression coverage showing that every transaction which derives writes from scan results has the necessary key writes/application locks to preserve its invariant.

Finally, the dependency cannot resolve until Regolith 0.1.7 is published; refresh `Cargo.lock` and rerun the final CI tree after that release. The current red matrix is the expected unresolved-version failure, not evidence about these code paths.

iverc (commented):
mostly with jack's list — two additions from reading the current diff:

- composite and lww swallow the lookup error and then run the merge with no guard at all (`None => None`). a flaky registry read silently turns off the invariant this PR exists to establish. batch and counter propagate; pick one behavior, and I'd propagate.
- batch now aborts the whole batch on a lookup error where these used to be per-block outcomes. worth checking the p2p batch callers don't lose an entire push because one collection lookup failed.

and +1 on the RepeatableRead scoping question — that's the one I'd want written down before it becomes the default for every transaction.

PR #1765, opened 2026-09-14
fix(p2p): let a successful block-sync completion wait for its blocks to land

### LINKED ISSUE #1759: A successful block-sync completion is misread as a stall, sleeping 2s while holding a fetch slot
## Problem

Under sustained replication load a Rust receiver drains far below its own apply rate. The cause is
not latency: `poll_fetch_blocks` (`crates/p2p/src/sync/coordinator/dag_fetcher.rs:895-908`) treats a
`BitswapComplete{success:true}` whose frontier blocks have not yet landed as `Stalled`. Both event
kinds are `DispatchClass::Completion` and run concurrently, so the completion routinely overtakes
the ~25 ms puts.

The false stall logs "Timeout fetching selective block batch (30s)" after ~13 ms, then
`poll_fetch_dag` sleeps `retry_backoff(2)` = 2 s **while holding one of only 4 global fetch
permits**. Measured on a 20 ops/s soak: 914 such stalls at saturation, 857 of them under 50 ms, and
91% of those roots complete via the manager path within 0.5 s anyway. The sleep accounts for **68% of
all slot-seconds** at saturation. Residency then climbs past `PENDING_DAG_TTL` (300 s) and the
receiver silently discards work.

## Plan

- On `Success`, wait up to a bounded grace period for the frontier to land instead of breaking to the
  stall path.
- Keep the existing stall handling for genuine timeouts; only the success-with-missing-frontier case
  changes.
- Pin with unit tests using paused time so the grace period is asserted without wall-clock sleeps.
- Verify on a soak arm against a control: residency at matched occupancy must fall below the 300 s
  TTL, and the spurious stall count must go to ~0.

edjroz:
## What

A successful block-sync completion whose blocks have not yet been written was being read as a stall.
The fetcher then slept 2 s on the retry backoff while holding one of only four global fetch permits,
which consumed 68% of all slot-seconds under load.

This made a receiver drain far below its apply rate, pushed pending-DAG residency past the 300 s TTL,
and caused work to be silently discarded. The five candidate causes originally suspected — the
metadata mutex, storage, permit exhaustion, the Go round trip, DAG depth — were each measured and
falsified.

## What Changed

- `poll_fetch_blocks` waits up to `SUCCESS_LANDING_GRACE` (1 s) for the frontier on a `Success` completion instead of breaking to the stall path.
- Genuine timeouts keep the existing stall and backoff behaviour.
- Two unit tests added, paused-time, asserting the grace window without wall-clock sleeps.

## Verification

- [x] `cargo test -p p2p --lib` — 353 passed
- [x] `cargo clippy` clean
- [x] Soak arm vs control, 2r2g / 20 ops/s / 900 s load / 3600 s settle
- [x] Spurious stall lines 1,434 → 1; backoff sleeps 1,399 → 0
- [x] Residency at occupancy ≥700: 315 s → 180 s (below the 300 s TTL)
- [x] Service rate 142 → 248 DAGs/min; dispatches 230 → 392/min
- [x] `pending_dag_expired` 50 → 0; window drained 5 min after load end (control: 21 min)
- [x] Unique documents missing on a Rust node 748 → 395

## Notes

- The bound moves to the replication loop after this change: fetch-complete → merged latency rises to
  ~1 s at load, from 0.15 s. That is the next thing to look at, not a regression introduced here.
- `pending_dag_resolved` counts only the manager path. Total throughput is
  `pending_dag_retry_dispatched`, which tracks `pending_dag_terminal_merged` within 2%.

---

Closes #1759


iverc (approved):

jackzampolin (changes requested):
The core completion-versus-storage diagnosis is sound, but the watchdog boundary still produces a false stall. In crates/p2p/src/sync/coordinator/dag_fetcher.rs, the outer BLOCK_SYNC_COMPLETION_WATCHDOG condition still wins over landing_deadline: a success just before 35s can set a full one-second landing grace, then the loop exits at 35s anyway. Once success is observed, switch the active deadline to landing_deadline rather than retaining the outer watchdog. Please add paused-time coverage for success immediately before watchdog expiry. I found no other substantive issue; the known red checks are being handled separately.

iverc (commented):
re-read after the landing-grace commit — `late_success_completion_keeps_the_full_landing_grace` pins exactly the race jack described, nothing else from me. looks ready once CI is green.

PR #1766, opened 2026-09-14
fix(p2p): do not shed a push that would have been acked

### LINKED ISSUE #1760: The pending-DAG shed turns an Ok into an at-capacity nack, losing documents at 4+ source peers
## Problem

`can_process_pushlog` (`crates/p2p/src/sync/manager/process/pending_dag.rs:140-150`) gates an early
shed at `pushlog.rs:326-348`. When it fires the receiver returns `Err(PendingDagCapacity)` before
doing any work. The problem is not when it fires — it is what it replaces. Three paths that sit
*after* the gate return `Ok(())` in the normal case: a descendant block, a head covered by the
current scope obligation, and an existing-root upgrade. The shed converts each into a nack.

A Go sender reads an at-capacity nack as success, marks the document delivered and never retries, so
each conversion is silent permanent loss. It is unreachable below 4 source peers: with N peers each
capped at `max_pending_dags/4` = 250, the map cannot exceed `N × 250`, so at 3 peers it tops out at
750 and the global cap of 1000 is never met. Confirmed live — `pending_dag_high_water` is exactly 750
in every 4-node run, and 1000 in a 5-node run where the shed fires thousands of times.

## Plan

- Widen the shed's bypass so only a genuinely new head — one that would actually consume a slot — is
  shed. Leave the `Ok` routes intact.
- Use the predicates already available at the call site at no cost: `announced_block_kind` is computed
  one line above, and `persisted_roots.contains` and the scope decision are in-memory reads.
- Do not change *when* the shed triggers; `can_process_pushlog` keeps testing the global cap only.
- Verify on a wide mesh (5+ nodes) under overload, where the shed actually engages, against a control
  on the same topology.

edjroz:
## What

The early pending-DAG shed rejects a push before doing any work, which is correct for a new head but
wrong for three cases that would otherwise have been acked: a descendant block, a head already
covered by the current scope obligation, and an existing-root upgrade.

Because a Go sender reads an at-capacity nack as success and never retries, each of those
conversions was silent permanent loss. It only bites at four or more source peers, where the global
pending-DAG cap becomes reachable — which no prior overload run had exercised.

## What Changed

- The shed's bypass at `pushlog.rs:327` now also spares descendant blocks, durably owned roots, and heads covered by the current scope obligation.
- `can_process_pushlog` is unchanged: the shed still triggers on the global cap only.
- Three unit tests added at `max_pending_dags = 8`, where the global cap and the per-peer quota diverge.

## Verification

- [x] `cargo test -p p2p --lib` — 354 passed (351 baseline + 3)
- [x] `cargo clippy` clean
- [x] `p2p_admission` and `p2p_admission_fairness` both pass
- [x] Wide-mesh A/B, 3r2g / 20 ops/s, identical topology and seed
- [x] `pending_dag_high_water` reached 1000/1000 — the first run to engage this path at all
- [x] Rust-origin documents missing on a Rust node 88 → 6
- [x] Go-origin unchanged (409 → 410), as expected — that half is the Go nack-as-success defect

## Notes

- Not behaviour-preserving, deliberately: an over-quota peer pushing a covered or descendant head
  previously received `Ok(())` at global saturation and now receives an at-capacity nack in the
  narrow case where the bypass does not apply. The original code already did this; this change makes
  the common cases take the cheap door instead.
- A prior attempt made the shed fire *more* often by adding the per-peer quota to the gate. It was
  driven-tested and refuted: zero measured benefit and 45% more permanent loss. Do not reintroduce it.

---

Closes #1760


jackzampolin (changes requested):
This reopens the saturation path the cap is intended to close. announced_block_kind classifies malformed DAG-CBOR as Descendant, and the new admission condition exempts every Descendant from shedding. A peer can therefore submit unlimited CID-valid malformed or unrelated non-head blocks while the pending-DAG registry is full; they proceed to verification and storage outside the resource bound. Keep the existing-root and covered/superseding-head exemptions, but admit a descendant only when pending.has_waiters(cid), or through another bounded durable receiver obligation. Please restore malformed/allocation-heavy and valid-but-unrelated rejection regressions. The ignored spike binaries are useful evidence but should not add roughly 550 lines of non-CI harness to this focused fix. Known CI work is separate from this review.

iverc (commented):
the unawaited-descendant fix plus the malformed/unrelated-at-cap regressions read right to me — a descendant only gets in when a registered root is waiting on it, and the garbage-CID case is pinned. that admission contract is subtle enough that the next refactor will miss it, worth a couple lines in the module docs spelling out the three acked conversions vs everything else.

PR #1767, opened 2026-09-14
fix(p2p): probe reachability off the replay ladder

### LINKED ISSUE #1761: A partitioned node takes ~38s to notice a peer is back, because the reachability probe sits behind the retry ladder
## Problem

After a network partition heals, a Rust node takes 37.9 s (p50) to resume replicating to the peer
that was cut, where every other outage kind costs seconds. The delay is not discovery or dialling —
it is scheduling. In `crates/p2p-adapter/src/retry.rs` the peer reachability check sits *behind* the
`is_due()` gate, so a peer that has become reachable again is not probed until its retry ladder rung
elapses.

Each failed attempt while partitioned also bumps that ladder, so the longer the partition lasts the
longer the node waits after it ends. The markers are durable and nothing is lost, but convergence
stalls for as long as the rung, which is the tail the partition tests measure.

## Plan

- Check reachability ahead of the `is_due()` gate rather than behind it.
- Defer the probe to `PERSISTED_RETRY_SWEEP_INTERVAL` instead of bumping the replay ladder, so an
  unreachable peer does not push its own recovery further out.
- Hoist `connected_peers()` to one snapshot per sweep rather than one call per marker.
- Verify against the existing partition suite, with the 60 s node-down cut as the headline case.

edjroz:
## What

After a partition heals, a node waited for its retry ladder rung to elapse before even checking
whether the peer was reachable again — so the reconnect notice took 37.9 s at p50 while every other
outage kind recovered in seconds.

The probe now runs ahead of the due-time gate and defers on the 2 s sweep instead of bumping the
ladder, so a peer that comes back is noticed on the next sweep rather than on the next rung.

## What Changed

- Reachability is checked before the `is_due()` gate in `crates/p2p-adapter/src/retry.rs`.
- An unreachable peer defers to `PERSISTED_RETRY_SWEEP_INTERVAL` instead of bumping the replay ladder.
- `connected_peers()` is hoisted to one snapshot per sweep.
- `p2p_partition` gains coverage for the reconnect-notice path.

## Verification

- [x] `rust_rust_node_down_cut_60s` — 37.88 s → 3.74 s p50 (n=5)
- [x] Full `p2p_partition` suite green — 11 passed, 1 ignored

## Notes

- **This is a deliberate divergence from Go, not a parity fix.** Go's `retryReplicators` has no
  connectivity gate, no explicit dial and no reconnect reset
  (`internal/db/p2p/replicator.go:601,646,735,890`), so Go would still measure ~37.9 s here.
- Two gates remain open and should be settled before this leaves the spike branch: every
  disconnected replicator peer with markers is now dialled every 2 s for as long as it stays
  unreachable, and the fleet-scale cost of that is unmeasured (tests use 2 nodes and ≤60 s outages).
  All measurements are libp2p; Iroh is untested.

---

Closes #1761


jackzampolin (changes requested):
The reconnect path is not yet bounded enough to merge. connected_peers failure is converted to an empty set, causing fleet-wide redials; redial_replicator has no timeout and peers are processed serially; and every pass rewrites the configured retry deadline to two seconds. The failing retry-interval conformance is therefore a real semantic regression, separate from incidental CI red. Please use a separate bounded reconnect schedule with per-peer backoff and dial timeout, preserve state when peer observation fails, and keep PeerConnected as the retry activation signal. Add deterministic unit coverage for observation failure, a hung dial, many disconnected peers, and Iroh. The large partition binary is compiled but not exercised by current CI.

iverc (commented):
the follow-ups cover the list for me — 5s dial timeout, 8 dials per pass, backoff capped at 60s, configured retry interval preserved. one small thing: `RECONNECT_BACKOFF_MIN` is tied to `PERSISTED_RETRY_SWEEP_INTERVAL`, so shortening the sweep silently lowers the reconnect floor too. if that coupling is intended, fine — just make it deliberate.

PR #1768, opened 2026-09-14
fix(p2p): continue the retry sweep past a document failure

### LINKED ISSUE #1762: The durable retry sweep stops at the third failure and the peer's rung never comes back down
## Problem

`run_retry_pass` (`crates/p2p-adapter/src/retry.rs`) abandons a peer's whole remaining marker set
after **three** non-capacity failures (`:252-254`) or **one** replay timeout (`:259-260`). A
`rate limited` reply is not recognised as backpressure — `capacity_retry_delay` (`:9-15`) matches
only `at capacity` — so three rate-limited replies end the sweep as if they were delivery failures.

Every such failure also bumps the peer's ladder, and nothing ever lowers it: `num_retries` only goes
to 0 via `new_initial()`, and the clock key is deleted only when that peer's marker set reaches zero.
The retry clock is per peer, not per marker, so one document's failure delays every other document
queued for that peer. The result is a marker series that is a staircase rather than a drain, and a
sender still holding obligations for documents that have already converged.

## Plan

- Continue past a per-document non-capacity failure or replay timeout instead of breaking; record
  that the pass had failures.
- Treat a `rate limited` reply as backpressure like `at capacity` — defer the peer on the paced 2 s
  sweep rather than counting it as a failure.
- Bump the ladder at most once per pass, and return to the first rung when the pass actually landed
  at least one document.
- Verify on a soak arm against the same recipe at a settle the ladder previously could not drain in.

edjroz:
## What

The durable retry sweep gave up on a peer's entire remaining backlog after three failures, and every
failure pushed that peer's ladder a rung higher with no path back down. A rate-limited reply counted
as a failure rather than as backpressure, so a busy receiver could end a sweep outright.

Because the retry clock is per peer rather than per marker, one document's failure delayed every
other document queued for that peer. Markers accumulated as a staircase instead of draining, and
convergence needed roughly double the settle window to complete.

## What Changed

- A per-document non-capacity failure or replay timeout no longer breaks the pass; it marks the pass as failed and continues.
- A `rate limited` reply now defers the peer on the paced 2 s sweep, alongside `at capacity`.
- The ladder bumps at most once per pass instead of once per failed document.
- `restart_retry_peer` returns a peer to the first rung when a pass landed at least one document.

## Verification

- [x] 5 new unit tests, red before the change
- [x] `test-crate defra-p2p-adapter --lib -- retry` — 7 passed
- [x] `test-crate defra-p2p-adapter --lib` — 66 passed; `storage --lib` — 246 passed
- [x] `cargo clippy -p defra-p2p-adapter -p storage --all-targets -- -D warnings` clean
- [x] Soak arm `1789319122-404`, 2r2g / 20 ops/s / 900 s load / **1800 s settle**
- [x] Documents missing on a Rust node: **0** — the same recipe needed a 3600 s settle to reach 0 before
- [x] Marker series peaked at 3,488 and drained to **0**; the plateau falsifier did not fire

## Notes

- 53 lines, and it deliberately mirrors the Go replicator sweep fix — the two runtimes had the same
  defect shape, which is worth knowing if either is revisited.
- **Not covered:** the `!connected` branch (`retry.rs:196-203`) still bumps and returns, so a peer
  that is unreachable rather than failing keeps its one-rung-per-pass behaviour.
- **Known cost:** a peer that is connected but dead now fails every due marker each rung instead of
  stopping at three — roughly 4,900 warn lines and ~10 s of futile writes per 30 min per such peer.
  That is a pre-existing condition becoming visible, not new loss, but it is the reason a
  classifier-level split of peer-level from document-level failures is worth considering next.

---

Closes #1762


jackzampolin (changes requested):
Continuing past one document failure is the right fairness direction, but the current loop creates a larger starvation path. Each marker gets its own 15s timeout and all failures continue, while peers are serial. A black-holed peer with the measured 3,488 markers can occupy one pass for about 14.5 hours and prevent every later peer from being visited. Bound each peer pass by count and/or elapsed time, stop the peer pass on timeout or connection-like failure, advance a durable cursor between bounded passes, and aggregate warnings per peer/pass. Please add a two-peer regression proving one hung peer cannot delay a healthy peer. Also replace the rate-limited substring match with typed or exact classification and state this as a deliberate Rust scheduling improvement unless a specific Go parity commit exists. Known CI work is separate.

iverc (commented):
exact-sentinel matching, the bounded pass, and the hung-peer regression all look right. nothing to add.

PR #1769, opened 2026-09-14
fix(p2p): stop the Bitswap session when its query ends

### LINKED ISSUE #1763: Every Bitswap fetch opens a session that is never stopped, leaking ~100 KB per applied merge
## Problem

`crates/p2p/src/host/command_handler/bitswap.rs:50` opens a session with `client.new_session()` for
each fetch, and nothing in `crates/p2p/src` ever calls `Session::stop` or `Client::stop_session` —
`grep -rn "stop_session\|session.stop" crates/p2p/src` is empty. iroh-bitswap keeps every session in
`SessionManager.sessions`, and only `Session::stop` removes it, so each fetch leaves behind a live
worker task with a periodic search timer, a 128-slot op queue, a want-sender with its own tasks, and
a deadqueue — for the life of the process.

This is the per-applied-write resident-memory growth that a previous investigation left unattributed.
It is a **leak**, not a cache and not a missing bound. On a 30-minute soak a node logged **12,649
sessions created and 0 stopped**, roughly one per applied merge, with resident memory going 53 MB to
1,363 MB. The "growth continues after load stops" symptom is the receiver still draining and still
leaking, not spontaneous growth.

## Plan

- Create the session before spawning the fetch task and keep its abort handle and id where the task
  can reach them.
- Stop the session on both the success and the error path, after dropping the block-receiver guard.
- On cancellation, abort the task and then stop the session from a spawned task — an aborted future
  never reaches its own cleanup.
- Verify by counting `creating session` against `stopping session` on a soak node, and by resident
  memory per applied merge.

edjroz:
## What

Each Bitswap fetch opened an iroh-bitswap session and nothing ever stopped it, so every fetch left a
live worker task, a search timer, a 128-slot op queue and a want-sender behind for the life of the
process. On a soak node that was 12,649 sessions created and zero stopped.

This is the per-applied-write memory growth that had been measured but never attributed. It is a
leak rather than a cache or a missing bound, which matters because the three have different fixes and
the earlier evidence could not tell them apart.

## What Changed

- The session is created before the fetch task is spawned, with its abort handle and id held in a map shared with the task.
- Both the success path and the `get_blocks` error path drop the block-receiver guard and then stop the session.
- Cancellation aborts the task and stops the session from a separate spawned task, since an aborted future never runs its own cleanup.
- A failed stop logs a warning and leaves current behaviour rather than failing the fetch.

## Verification

- [x] 2 new tests in `crates/p2p/tests/bitswap_session_lifecycle.rs`, red before the change
- [x] Both green after (1.02 s); `bitswap_acp_filter_tests` 4/4; `host_tests` 19/19
- [x] `cargo clippy -p p2p --features test-utils --all-targets -- -D warnings` clean
- [x] Soak arm `1789321464-401` against baseline `1788984777-401`
- [x] Sessions created/stopped: **12,649 / 0 → 2,265 / 2,269** — they now track 1:1
- [x] Resident memory per applied merge: **~113 KB → ~51 KB**
- [x] Negative control: `Bitswap fetch complete … success=true` unaffected; stopping after completion does not cancel wants

## Notes

- **The mechanism signal passed decisively; the magnitude target did not.** The pre-registered win
  was resident memory per applied merge falling to the 10-16 KB range. It fell to ~51 KB. Sessions
  are no longer leaking, so the remaining ~51 KB is a second term — likely the store memtable fill
  already attributed at roughly half the old slope. Worth naming before anyone calls the memory
  question closed.
- The verification arm was 300 s rather than the baseline's 1800 s, so absolute resident memory
  (53→151 MB against 53→1,363 MB) is **not** a like-for-like comparison. The per-merge ratio is, and
  it is what the claim rests on. The "end RSS under 400 MB at 18k ops" threshold has not been tested.

---

Closes #1763


jackzampolin (changes requested):
The leak diagnosis is convincing, but cancellation cleanup is still race-prone. cancel_bitswap aborts the fetch and immediately starts stop_session without awaiting the fetch task and its Session clone to drop; Beetle Session::stop requires the expected Arc count, so normal coordinator cancellation can still fail with too many session refs. Completion can also race registration because the task is spawned before the registry insert, leaving a stale completed entry. Finally, awaiting unbounded session.stop before BitswapComplete lets teardown suppress protocol terminalization. Give cancellation and cleanup one joinable owner, register before execution can complete, and separate bounded teardown from the completion signal. Add deterministic cancel-in-flight, last-block race, immediate failure, and idempotent repeated-cancel tests. Please also remove runs-h3 generated workstation artifacts and split/harden the experimental harness separately. Known red CI is not part of this verdict.

iverc (commented):
the register-before-run barrier fixes the stale-entry race cleanly. one question on cancellation: after `abort()` you call `stop_session` immediately without joining the fetch task, so its `Session` clone can still be alive when stop runs. does client-side stop-by-id tolerate that, or is it the same Arc-count requirement jack described? if it's the latter, the race is still there and awaiting the task handle first would close it.

PR #1770, opened 2026-09-14
fix(p2p): hang up only on a proven-dead Bitswap queue

### LINKED ISSUE #1764: After a partition heals, the cut node's Bitswap is dead on every surviving connection and it does nothing about it
## Problem

A partitioned node loses no data but takes **154 to 212 seconds** to catch up, where every other
outage kind costs 5 to 6 seconds. The cause is not provider rotation and not a TTL.

During the cut the socket stays open, so one outbound want sits in a substream handshake until
libp2p's 10 s substream-request timeout fires. The fork's handler drops the response oneshot, the
queue actor reads that as fatal and exits, and the only restart path is a
`Disconnected → Responsive` transition — which a partition never produces, because there is no ping
behaviour, no yamux keepalive, and gossipsub keeps the idle timeout refreshed. The node's Bitswap
`MessageQueue` is then dead on every connection that survived the cut.

Recovery comes from outside: the first peer to open a *fresh* connection ends the silence to the
millisecond and then serves the whole backlog alone. On the measured schedule that was another
node's crash-kill at heal+136 s and a restart at +174 s — which is the 154-212 s figure. The
provider-rotation loop is the symptom of this state, four fetch slots each burning a 10 s window
against a provider that will never answer.

## Plan

- On the receiver: when a bounded fetch exhausts and every provider is still *connected*, disconnect
  them before returning. The durable ladder and the peers' own re-dials rebuild fresh connections,
  which is exactly what ended every slow case in the data.
- Guard it on having received zero blocks from that provider across the whole budget, so it cannot
  fire on a transient stall.
- Verify on the container topology with the seeded partition schedule, against the seven existing
  runs that established the tail.
- Separately, report upstream that the fork's queue actor exits on a failed send, and that
  go-bitswap's `messagequeue.go` shuts down on the same condition.

## Plan note

The root-cause fix is in the `sourcenetwork/beetle` fork — a failed send must not kill the queue
actor. That is 8 lines and is the better fix, but it requires a fork release. The receiver-side
change above is in our code, cures both halves of the observed failure, and is what this change
implements.

edjroz:
## What

When a network partition heals, the node that was cut cannot receive a single Bitswap block from any
peer it was connected to beforehand, and has no mechanism to notice. Its message queue actor died
during the cut, and the only restart path is a connection state change that a partition never
produces.

Recovery currently depends on an unrelated event — the first peer to open a fresh connection — which
is why catch-up takes minutes where every other outage kind takes seconds.

## What Changed

- When a bounded DAG fetch exhausts and every qualified provider is still connected, the receiver disconnects them before returning.
- The disconnect is guarded on having received zero blocks from that provider across the whole fetch budget, so a provider that is merely slow is not dropped.
- The durable retry ladder and the peers' own re-dials rebuild the connections, which restores the queue actor.

## Verification

- [x] 2 tests added, red before the change
- [x] Container soak arm `1789329230-622`, 6 nodes, seeded partition schedule, same recipe as the seven baseline runs
- [x] rust-1 catch-up **155-170 s p50 → 24 s p50, max 53 s**
- [x] Recovery attributable to the change, not to churn: the new WARN fires at heal+41 s, peers reconnect within 1 s and `recv_msg` resumes in the same bucket; the next churn event is at +136 s
- [x] Falsifier dead — reconnects outnumber disconnects on both cut nodes (4/12 and 10/14), followed by 5,955 and 12,780 `recv_msg`
- [x] Negative control passed — the uncut node emitted the new WARN zero times, and its disconnect count matches baseline once the duplicate log line is accounted for
- [x] The guard declines correctly: a provider that had served blocks was not dropped

## Notes

- **This is the Alternate, and it does not close the hole.** Every Rust-to-Rust Bitswap queue still
  dies on the cut and is only repaired when a root happens to exhaust with that dead peer as its sole
  silent provider. A root whose provider set mixes a dead Rust peer with a live Go peer never
  exhausts, so it never triggers the drop — which is why the slower of the two cut nodes took until
  heal+168 s to drop its remaining dead queues.
- **The root cause is in the `sourcenetwork/beetle` fork**: a failed send kills the Bitswap queue
  actor (`message_queue.rs:489-495`). Eight lines, and it removes the condition outright rather than
  repairing it after the fact. Landing that is the better fix; tightening this change with a per-peer
  silent-window clock is more machinery aimed at a symptom.
- **Worth flagging to the Go team:** go-bitswap's `messagequeue.go` shuts down on the same condition,
  so the same partition shape likely strands a Go node too. Untested.
- One unexplained residual: `final_mismatches` 555 and `unresolved` 237 against baseline ranges of
  134-440 and 64-195. Row composition is identical to the baselines and untagged divergences are
  *lower* (156 against 916-961), so this is not new loss, but the elevation is not accounted for.

---

Closes #1764


jackzampolin (changes requested):
The new silent signal is per-root, but the action is peer-global. DocSync and BranchableSync consider every connected peer as an alternate, so a healthy peer that simply does not have this DAG is classified silent and disconnected. The generic path also applies to Iroh even though the motivating failure is a Beetle Bitswap actor problem. This can interrupt unrelated concurrent DAGs and discard libp2p address/Kademlia state. Require positive transport-level liveness evidence and libp2p scoping, or preferably fix the Beetle queue actor. Add a deterministic node-level test proving disconnect then redial then convergence without root reannouncement; the current mock only records disconnect calls and the real partition test is ignored. Land the corrected 1765 first, then restack and retest this PR because false successful stalls feed this silent set. The error saying convergence requires root reannouncement should also be updated now that PeerConnected redrives it.

iverc (commented):
agree with jack here — for a healthy peer that simply doesn't have the root, 'connected but silent' and 'connected but doesn't hold this DAG' look identical, so the peer-global disconnect can drop exactly the peers you want to keep. scoping to real transport-level liveness (or fixing the actor) feels required, not optional. also seconding the deterministic disconnect -> redial -> converge test; a mock that records disconnect calls won't catch the real path regressing.

edjroz:
> preferably fix the Beetle queue actor. 
@jackzampolin https://github.com/sourcenetwork/beetle/pull/7

Issue #1309, opened 2026-08-05
setup_p2p spawns untracked pending-DAG sweep tasks; store lock outlives node shutdown by up to a sweep interval


jackzampolin:
`defra-node/src/p2p_runtime.rs::setup_p2p` spawns two loops that are **not** recorded in `P2PLifecycleInner` and therefore never aborted/joined by `EmbeddedNode::shutdown`:

- `coordinator.run_pending_dag_resync(60s)` (spawned around the `load_persisted_collections` step)
- `coordinator.run_pending_dag_retry_clock(2s)`

Both check `runtime.shutdown.is_shutting_down()` only at the top of each iteration, then `sleep(interval)`. After `shutdown()` returns, each task still holds its `Arc<SyncCoordinator>` — and through the coordinator's blockstore, the storage `Arc` — until its next wake. For the resync loop that is up to **60 seconds**.

Consequence for in-process embedders: reopening the same data path right after `shutdown()` + dropping the node fails with

```
failed to open redb store: backend error: database '<path>/data.redb' is locked by another process
```

for up to a sweep interval. A real process restart never sees this (the OS releases the lock at exit), which is why it surfaces in embedded restart tests rather than production. Hit while writing the boot-brick repro for source-inc/gents#984; the test now retries the reopen with a bounded deadline as a workaround.

Fix: track both `JoinHandle`s in `P2PLifecycleInner` and abort/await them in `shutdown()` like the other five tasks (they're already cancel-safe — pure sleep/sweep loops).

vertexclique:
In-flight cancellation for all these relevant tasks must be separate issue for tracking. This doesn't exist yet. I will defer that to another issue.

The deferred half of this is now tracked in #1356: the two sweeps here are the acute case, but the same uninterruptible-sleep pattern appears at five more production sites, and `resync_persisted_pending_dags()` takes no cancellation parameter so a sweep already inside the walk runs to completion regardless. #1356 does not block this issue.

PR #1771, opened 2026-09-14
refactor(p2p): one iroh peer for every node

mike7c2:
PR 4 of 4, stacked on #1757. With every writer an ordinary iroh peer, there is no reason for more than one way to build one.

## What

`defra start`, embedded/FFI, `defra-node`, and the browser each assembled their own iroh peer. They now all call **`IrohPeer::start`** in `defra-p2p-adapter`:

| Caller | Before | After |
|---|---|---|
| CLI `server_p2p/iroh.rs` | 648 | 260 (incl. unchanged helpers + tests) |
| defra-node `p2p_runtime.rs` | 406 | 97 |
| embedded `node_p2p.rs` + `node_tasks.rs` + `node_recovery.rs` | 1314 | 862 (libp2p path untouched) |
| browser `wasm/src/p2p/runtime.rs` | 231 | 66 |

Net about −590 lines. `IrohPeerConfig` carries what legitimately differs; `wire_document_acp`, `wire_kms`, and `se_query_transport` bind what each node builds after the peer; `IrohPeerShutdown` stops it in one order. libp2p assembly is deliberately left alone.

## Bugs the duplication was hiding

| Behaviour | CLI | embedded | defra-node | browser |
|---|---|---|---|---|
| KMS requester identity (6fcf59a6) | anonymous → **fixed** | anonymous → **fixed** | ✓ | n/a |
| document ACP on the coordinator | ✓ | ✓ | missing → **fixed** | missing → **fixed** |
| topic-peer `JOINED`/`LEFT` events | ✓ | ✓ | missing → **fixed** | missing → **fixed** |
| `merge_complete` / quarantine events | missing → **fixed** | ✓ | missing → **fixed** | missing → **fixed** |
| resync sweeps owned by coordinator (#1309) | ✓ | bare spawn → **fixed** | ✓ | ✓ |

The anonymous KMS resolver meant `defra start` and embedded nodes denied every DEK request for a policy-bound encrypted document.

## Notes

- pubsub_rpc (Go interop) services start only where they exist, beside the libp2p transport, as before.
- Embedded now wires the peer's document ACP before its sourcehub strict flag, so that flag is not overwritten.
- db's replication stack and txn broadcaster take `MaybeSendSync` blockstores so the shared peer builds for wasm32 over any store.
- The log lines backbone's harness matches for iroh nodes (`Peer connected (iroh)`, `Peer disconnected (iroh)`, `Starting replication loop`) keep their exact text, now emitted by the shared peer.
- CLI `P2PTasks` becomes an enum; the iroh variant holds the peer's shutdown and the hosted relay server.
- Removed CLI shims `iroh_p2p_adapter` and `transport_version_syncer`, embedded's iroh event handler and restore, and `run_iroh_event_handler`.

## Testing

| Suite | Result |
|---|---|
| `defra-p2p-adapter` | 62 passed |
| `embedded --features iroh` | 18 passed |
| `defra-node --features p2p --lib` | 58 passed |
| `cli --lib` | 72 passed |
| `p2p --features iroh-transport,libp2p-transport,iroh-relay-server --lib` | 527 passed |
| `defra-wasm` lib / perf / vector (headless Firefox) | 70 / 2 / 5 passed |
| `integration-test --test p2p_iroh` | 233 passed; `sync::version::with_view*` (2) need the Go lens WASM binary, unrelated |
| `tools/browser-p2p-e2e.sh` (browser ↔ node through the hosted relay) | passed |
| `cargo clippy -D warnings` (adapter, cli, embedded, defra-node with iroh/p2p) | clean |

The first `p2p_iroh` run caught two overlay tests timing out: the shared peer had reworded the log lines backbone's harness matches for iroh nodes. They keep their original text now, and both tests pass.


jackzampolin (changes requested):
The shared assembly is a strong direction and removes meaningful divergence, but IrohPeer::start makes the endpoint reachable, starts replication/event/retry tasks, and restores durable topology before callers invoke wire_document_acp. During that window LateBoundServeAcp fail-closes serving, but AcpCompositeMergeHook has no ACP and returns allow; the doc pusher and broadcast path also operate without their ACP hooks. Stable endpoint IDs and restored peers make inbound traffic during startup realistic, especially in defra-node where ACP is constructed after the peer returns. Make document ACP part of start, or hold endpoint/event/replication readiness behind a barrier until all policy hooks are installed. Add a deterministic startup regression proving a protected inbound document cannot merge before policy wiring. Please also add focused regressions for the claimed KMS requester-identity and orderly shutdown/store-release fixes; this PR changes four assemblies but currently adds no direct test files, so broad suites do not pin those repaired contracts.

iverc (commented):
the direction is great, 1.6k lines of divergent assembly gone. the startup window jack describes is the blocker for me too: merge hook with no ACP returns allow while the endpoint is already reachable, so an early inbound push merges with zero policy. if building ACP inside start is too invasive, holding the doc pusher / broadcast readiness behind a barrier until the policy hooks are wired gets you fail-closed startup — either is fine, but not the current gap.

mike7c2:
Thanks. The startup window is closed by construction, and the requested regressions are in (122c8ee0).

**Startup window**
Document ACP is now a required argument: `IrohPeerConfig::new(endpoint, document_acp)`, with the strict flag beside it. Every caller builds ACP before starting the peer; it depends only on the store and config. `IrohPeer::start` binds the serve gate, coordinator, merge handler (ACP and strict flag), doc pusher and broadcast mutator before it:
- reloads subscriptions
- spawns the replication loop, sweeps, event handler and retry loop
- restores replicators

The endpoint still binds first, because the coordinator needs its transport. It only queues inbound events in its bounded channel until the handler that drains them starts, and that handler starts after ACP is bound. `wire_document_acp` and every caller's late-wiring callback are gone, so there's no longer a path that starts a peer without policy.

`document_acp_is_bound_before_start_returns` asserts that the merge handler (ACP and strict) and the coordinator are bound the moment `start` returns. It fails with the install block moved after start. An inbound-traffic test would need to deliver a push before `start` returns, which the API no longer allows, so this pins the contract directly.

**KMS requester identity**
`a_granted_iroh_peer_obtains_the_key` covers an embedded iroh node serving a policy-bound DEK to a granted peer. It fails with `AnonymousResolver` swapped back in. Its twin, `an_ungranted_iroh_peer_receives_the_blocks_but_not_the_key`, asserts the blocks arrive but the key doesn't.

**Orderly shutdown and store release (#1309)**
- `shutdown_releases_every_store_reference`: after shutdown, the store's `Arc` count returns to 1. It fails when a peer task is detached instead of stopped.
- `iroh_shutdown_releases_persistent_store`: reopens the same RegolithStore path after an iroh node stops.
- The libp2p-only failed-setup shutdown test is now gated, so the shutdown tests build iroh-only.

Also on this PR: the unused dependencies the shared peer left behind are removed (62502d12).

PR #1780, opened 2026-09-16
feat: pluggable access control for iroh embedded apps (1/7)

mike7c2:
Lets an iroh embedded app supply its own access control through three hooks, in place of hard-coded ACP. Unclaimed collections behave exactly as before, and a node with no hooks installed takes no new code paths beyond an empty-queue check after each merge.

This is **PR 1 of 2**. PR 2, stacked on this branch, adds the replication-policy trait: outbound per peer, inbound per peer, selective-sync matcher and topics.

### Merge validator (`db::merge::governance`)

```rust
pub trait MergeValidator {
    async fn validate(&self, candidate: &MergeCandidate<'_>, view: &dyn MergeView)
        -> Result<MergeVerdict, String>;
}
pub enum MergeVerdict { Accept, Reject { reason }, Defer { reason, awaiting: Vec<Awaited> } }
pub enum Awaited { Composite(Cid), ImmutableField { collection, field, value } }
pub enum SignatureStatus { Unsigned, Verified(did), Invalid(reason), NotHeld(sig_cid) }
```

- **Where it runs.** `MergeGovernance::new([collection names]).with_validator(..)` is installed on the `DB` before the transport starts. For a claimed collection, every composite frame goes to the validator instead of the ACP `CompositeMergeHook`. That covers the root, ancestors loaded from the blockstore, composites linked from a collection block, batch merges and genesis composites.
- **What the candidate carries.** Its CID, block, payload and derived doc ID. It also carries the collection, `is_genesis`, and a `SignatureStatus`.
- **Collection resolution.** A governed collection is resolved only from the block's own `schema_version_id`, never from the carrier's `metadata.collection_id`, so a sender cannot pick the validator. When the block's version doesn't resolve to the collection the carrier names, the composite defers (retryable skip) and is not judged. Ungoverned collections keep the existing fallback unchanged.
- **Unidentifiable updates.** In a governed collection, an update whose document can't be identified because an ancestor isn't held defers on the first ancestor not held, rather than failing. The ancestor's arrival re-drives it (MD-1). Ungoverned collections keep today's error.
- **Signature status.**
  - It is verified from the block's own signature block on every call, recovery included, and never from `BlockMetadata`.
  - "Unsigned", "invalid" and "signature block not held" are distinct.
  - A storage read failure is a retryable error, not an invalid signature.
- **`MergeView`, the only reads a verdict may use:**
  - `block(cid)`
  - `composite_fields(cid)`: the decoded values of the linked field blocks, with `Encrypted`, `NotHeld` and `Undecodable` distinguished
  - `genesis(cid)`
  - `immutable_fields(collection, doc_id)`: one merged document's `@immutable` scalar LWW fields, from the same snapshot. Mutable fields are never returned. A missing document (not merged yet, or deleted) or a missing field has to defer, because an immutable field can be set for the first time by a later update.
  - `find_documents(collection, field, value)`, over one snapshot shared by the whole verdict. **Only `@immutable` scalar LWW fields** can be looked up (the rule replicator filters already use); any other field is an `Err`. An immutable field is set once, so a match doesn't depend on which update merged first. An empty result still has to defer.
- **Verdict mapping.**
  - `Accept` → merge.
  - `Reject` → `MergeOutcome::Rejected`, which quarantines.
  - `Defer` → `MergeOutcome::Skipped { terminal: false }`, and the composite is indexed under `awaiting`.
  - `Err` → a retryable `MergeError`, never a verdict.
- **Fail closed.** A claimed collection with no validator defers every composite.
- **ACP stays out.** For claimed collections the ACP hook is skipped entirely, including `on_protected_composite`, `check_protected_update`, `on_encrypted_link` and `post_commit_action`. The node-own-key bypass therefore never touches a governed verdict.

### Purity contract (documented on `MergeValidator`)

The verdict is replicated behaviour. It must be a function of the candidate and of `MergeView` reads only:
- no wall clock
- no randomness
- no node identity
- no ACP registration state
- no transport peers
- no state kept beside the database

`Reject` only for content that can never be accepted on replay. `Defer`, naming what is missing, whenever an input is absent; absence is never a reason to reject. Replication policy (PR 2) is node-local and never an input.

The ACP implementation cannot itself be an implementation of this trait: it reads registration state, sender metadata (explicit replay) and the node's own identity, all of which the contract forbids. So ACP stays the default `CompositeMergeHook` for unclaimed collections, byte-for-byte as before. The validator replaces it for claimed collections.

### Re-drive of deferred composites

- **Index.** `DeferredMerges` indexes deferred root composites by the CIDs a frame awaits, keeping the carrier metadata needed to rebuild `BlockMetadata::normal`.
- **What a deferral can await.**
  - `Awaited::Composite(cid)`: released when that composite commits. A document is named by its genesis.
  - `Awaited::ImmutableField { collection, field, value }`: released when any composite commits into `collection` with that `@immutable` scalar LWW field value. It is for an input that doesn't exist yet and so has no CID to name, such as a write arriving before the entry that authorises it.
- **Field restriction.** The field has the same restriction as `find_documents`; awaiting any other field is a validator error. Values are matched by canonical CBOR.
- **Release.** Releases happen on standalone and batch commits. Field values are read from the merged composite's field blocks, and only while some deferred composite awaits a field.
- **Nothing else triggers re-drive.** A composite awaiting a signature or field block's CID is retried by the pending-DAG retry clock instead. This is documented on `MergeVerdict::Defer`.
- **Drain.** After each `handle_block` / `handle_block_batch`, released composites are re-merged through the normal path: signature verified at dispatch, and the validator judges each one again.
- **It never uses `BlockMetadata::recovery()`.**
- **Bounds.**
  - 16,384 indexed composites
  - 64 awaited inputs per deferred composite
  - 1,024 waiters per awaited CID
  - 256 re-drives per drain, with the remainder draining on the next merge
- **Beyond the bounds.** A composite not indexed is still retried by the existing pending-DAG retry clock.

### Client write and read validators (`query::access_hooks`)

- `WriteValidator::validate_write(&WriteRequest)`:
  - **Input.** Identity, collection, mutation kind, resolved target doc IDs, and the create/update input.
  - **When.** It runs in `execute_single_mutation` before ACP and before any block is built. `Err` refuses the mutation.
- `ReadValidator { governs(collection), may_read(&ReadRequest) }` is checked after ACP, so it only narrows. It applies on every read path that has an ACP gate today:
  - planned selects (`PermissionFilterNode`)
  - nested relations and join fast paths
  - top-level aggregates
  - CID time-travel
  - encrypted-index doc-ID results
  - `_commits`

  It also applies to collections without a policy. `Err` hides the document.

### Embedded registration (iroh only)

```rust
NodeBuilder::default()
    .with_iroh(config)
    .with_access_hooks(
        AccessHooks::new(["Note"])
            .with_merge_validator(Arc::new(MyValidator))
            .with_read_validator(Arc::new(MyReader))
            .with_write_validator(Arc::new(MyWriter)),
    )
```

`build_with_store_and_access_hooks` does the same for callers that bring a store. Building with hooks on any other transport returns an error. Everything here is behind the `iroh` feature.

## Tests

- **`crates/db/tests/merge/governance.rs`**
  - A validator accepts, rejects and defers.
  - A deferred note merges once the grant it awaits merges, with no restart; the validator sees `Verified` both times.
  - A re-drive whose signature block was replaced by a forgery is rejected at dispatch before the validator runs.
  - A claimed collection with no validator defers.
  - An update that arrives before its document's genesis defers and merges once the genesis merges, without a restart (MD-1). The same holds through `handle_block_batch`.
  - A composite awaiting a non-composite CID (a signature block) is not re-driven when that block is stored or another composite merges. Merging it again, as the retry clock does, accepts it.
  - A block whose schema version is unknown, carried with a governed collection's ID, is deferred and never reaches the validator.
  - `find_documents` on a mutable field is refused.
  - `immutable_fields` returns a grant's immutable `writer` but not its mutable `label`, even after an update sets `label`, and returns `None` for a document that isn't merged.
  - A note deferred on `(Grants, writer, <signer DID>)` stays deferred when a grant with another writer merges. It is re-driven and merged, without being merged again or restarting, when a grant naming that writer merges.
  - Deferring on a mutable field is refused and indexes nothing.
  - A deferred forged note, after the DB is reopened on disk and re-merged through `BlockMetadata::recovered` (dispatch skips verification), is judged by its real signer and stays unmerged.
- **`crates/embedded/tests/access_hooks.rs` (iroh)**
  - Two nodes. Notes replicated to a governed receiver are accepted (grant held), rejected (no grant) and deferred (grant not held yet); the deferred note appears once the grant is synced.
  - Write and read validators refuse a mutation, hide documents from anonymous user queries and `_commits`, and show them to an identified reader.
  - Hooks without iroh fail to build.
- **Suites.**
  - `cargo test -p db -p p2p -p query --no-fail-fast` plus `cargo test -p embedded --features iroh --no-fail-fast`: 2,245 passed, 0 failed, 34 ignored.
  - `cargo test -p integration-test --test acp`: all 110 Rust tests pass (16 ignored). The 94 `go_*` tests could not run locally because no Go `defradb` binary was on PATH.
  - Clippy is clean for `db`, `query` and `embedded`.

## Behaviour changes for nodes without hooks

- `DbMergeHandler::handle_block` and `handle_block_batch` check the (empty) re-drive queue after merging.
- `prepare_composite_merge` asks whether the collection is governed before calling the ACP hook. It returns `Ungoverned` at once when no governance is installed. The governed-only collection re-resolution and the unidentifiable-update deferral are skipped for ungoverned collections.
- `PermissionFilterNode`'s ACP check is now optional internally. `new`, `from_optional_did` and every ACP decision are unchanged.
- `verify_signature_data` is `pub(crate)`.

There are no wire, CID, document ID or collection ID changes.

## Known limits

- **Re-drive runs inline** on the merging task, up to 256 re-merges per drain.
- **The deferred index is in memory.** After a restart the pending-DAG retry clock re-merges each deferred composite, which re-indexes it.
- **A re-driven merge does not clear its pending-DAG record itself.** The record is cleared on the retry clock's next attempt, which sees "already merged" (backoff is at most 60 s).
- **`find_documents` scans the collection** rather than using an index, and only looks up immutable fields.
- **Governance is claimed by collection name** on each node that runs the app.
- **The write validator covers GraphQL mutations.**
- **For a root composite, dispatch still treats a missing signature block as a terminal signature failure.** Only the validator's `SignatureStatus` distinguishes `NotHeld`.

## Follow-ups against Fefra §18

| | Status |
|---|---|
| U-1 governance field in the definition delta | **Not done.** It changes collection-ID derivation and needs sign-off first. |
| U-2 merge hook with block, values, genesis | **Done** (`MergeValidator`). |
| U-3 owner from the genesis signer | **Enabled.** `MergeView::genesis` plus a signature verified from the genesis block; there is no registration step for governed collections. |
| U-4 reject unsigned in governed collections | **Enabled.** The validator sees `SignatureStatus::Unsigned`; the host does not force the policy. |
| U-5 "could not check" vs "invalid" | **Done for the validator** (`NotHeld`, `Invalid`, retryable storage error). Dispatch's root check is unchanged because changing it alters default behaviour. |
| U-6 node-identity bypass | **Done.** The ACP hook, and its bypass, never runs for governed collections. |
| U-7 synced definitions keep governance | **Not done** (depends on U-1). |
| U-8 content-derived policy IDs | Out of scope. |
| U-9 no collection blocks in governed collections | **Not done.** Composites linked from a collection block are judged per frame, but collection heads are still installed unchecked. |
| U-10 no fallback to sender collection IDs | **Done for governed collections**, which resolve only from the block's schema version and defer otherwise. Ungoverned collections keep the fallback. |
| U-11 prefix-collision emulation | Out of scope. |
| U-12 re-drive on arrival | **Done** (bounded, in memory), on a composite CID or on an immutable field value for inputs that don't exist yet. |
| U-13 lookup returning all matches | **Partly done.** `find_documents` returns every match on an immutable field but scans instead of using an index. |
| U-14 fetch authority entries by `(aid, sn)` | Out of scope. |


mike7c2:

PR #1787, opened 2026-09-17
feat: forward and release re-driven merges (3/7)

mike7c2:
**PR 3 of 3, stacked on #1781** (`feat/pluggable-replication-policy`, which is stacked on #1780). Review those first. This PR's own diff is its last two commits.

A composite an app's `MergeValidator` defers is re-driven once what it awaits arrives (#1780). Two gaps meant a re-driven merge was a dead end. Both show up as soon as a validator defers anything, which for a governed app is the normal case, not a corner.

## 1. A composite merged by re-drive is forwarded to replicators

Only a block whose *own* first attempt returned `MergeOutcome::Merged` reached the replication layer's post-merge path (`crates/p2p/src/sync/replication/handlers.rs:336-365`). A re-driven merge's outcome was only logged, so the node never marked it merged and never pushed it on.

In a chain A → B → C, if composites reach B out of order, B merged them by re-drive and C never received them. A middle node forwarded almost nothing.

- **The seam:** `RedrivenMergeSink` and `RedrivenMerge` (`crates/db/src/merge/governance/redriven.rs`). `redrive_deferred` reports each composite it merges (`crates/db/src/merge/merge_handler/dispatch.rs:145-150`). `db` gains no dependency on `p2p`.
- **The implementation:** `SyncRedrivenSink` (`crates/db/src/merge/redriven_sink.rs`) calls `mark_as_merged` and then `push_to_replicators_with_creator`, the same call the first-attempt path makes, installed in `create_replication_stack_with_max_merge_depth` (`crates/db/src/merge/replication.rs:98`).
- **The policy still applies:** the push goes through `admit_replicator_push`, so a forwarded re-driven block is subject to `may_send` like any other.
- **Bounds unchanged:** `REDRIVE_BUDGET` and the deferred-index caps.
- **Recovery needs no special case:** a recovery-path `handle_block` calls `redrive_deferred` like any other.

**Exactly one push per replicator.** A re-driven composite's first delivery returned a deferred verdict, so the replication layer never pushed it; and a later redelivery merges as `terminal_skip("already merged")`, which is not `Merged`, so it is not pushed again. Asserted two ways: the embedded test counts exactly one `may_send(Push)` for the note, and the `db` tests assert the recording sink saw exactly the one CID.

## 2. A local write releases the waiters it satisfies

Waiters were released only from the merge path (`release_merged_composite`). A local mutation bypasses it, so a composite awaiting a document *this* node then creates itself stayed deferred until a peer replayed it or the retry clock fired.

This is the ordinary case for a node that both receives and authors: a tower defers a device's write awaiting its own anchor, then creates that anchor locally.

- **The seam:** `LocalCommitRelease` (`crates/db/src/merge/governance/local_commit.rs`), told about each composite a local write commits, installed beside the sink.
- **Where it fires:** the committed-write callback (`crates/db/src/event/emission.rs:105-107`), the one seam both local write paths share (`DbDocMutator` and `BatchMutator`).
- **What it does:** computes the same wait keys as the merge path, by reusing `release_merged_composite`, then drains with `redrive_deferred`, so waiters are re-judged by the normal verifying path and, by construction, forwarded through the sink above.
- **After commit:** the callback runs after the durable commit, so the re-judged verdict's snapshot contains the write.
- **Off the writing task:** the callback still holds the document's write guard, so merging inline would deadlock against the per-document write queue whenever the deferred composite belongs to the document just written. The drain is spawned, so the write's own outcome and latency are unchanged, and it is bounded by `REDRIVE_BUDGET`.
- **Ungoverned writes pay nothing:** one uncontended lock (`DeferredMerges::has_waiters`), skipped before any other work.

## Tests

- **`crates/embedded/tests/merge_redrive.rs`** (iroh, three nodes): author → middle → tail, with the middle governing `Note` and deferring a note on the CID of the grant it names. The note is replicated before the grant, so the middle defers it; once the grant arrives the middle merges it by re-drive, and the tail must receive it. Before the fix the middle merged the note and it never left:
  ```
  test a_composite_merged_by_redrive_reaches_this_nodes_replicators ... FAILED
  Error: Note [], expected ["bae-29ba84e1-29d6-5402-a94d-0d5e1a89773d"]
  ```
  After: `ok`, with the exactly-one-push assertion.
- **`crates/db/tests/merge/governance.rs`:** `a_local_write_releases_a_composite_awaiting_an_immutable_field` and `a_local_write_releases_a_composite_awaiting_its_genesis`, both driving the real mutator layer, with no further delivery, restart or retry tick. Both failed before the fix ("Notes never merged …") and pass after. The existing re-drive tests now also assert what was forwarded.

**Suites:** `cargo test -p db -p p2p -p query` and `-p embedded --features iroh` pass. Clippy is clean with and without `iroh`.

## Known limits

- **The local-write release has no end-to-end iroh test.** Embedded nodes run a 2 s pending-DAG retry clock, which re-merges a deferred composite anyway, so such a test passes with the fix disabled and proves nothing. The release is verified deterministically at the `db` level for both `Awaited` variants, and its forwarding through the recording sink; the full wire path for a locally released waiter is unverified.
- **The re-drive fan-out test's pre-fix failure is a timeout-based negative** (30 s). It reproduced identically on every attempt, and tracing confirmed the merge happened locally while the block never left.


mike7c2:

PR #1789, opened 2026-09-17
perf: governance view reads through an index, not a scan (4/7)

mike7c2:
**PR 4 of 4, stacked on #1787** (`feat/redrive-fanout`, on #1781, on #1780). Review those first. This PR's own diff is its last three commits.

The governance view (#1780) answered `find_documents` and `immutable_fields` by loading every document of the collection and filtering in memory. A merge validator makes tens of these calls per verdict, so a verdict cost O(calls × documents held). One governed app measured 2.8 ms per verdict with 11 documents held and 18.3 ms with 26, against about 80 µs for the same verdict in memory.

No new index kind and no new directive: the existing `@index` serves.

## 1. `immutable_fields` reads one document

It resolves the id through the DocID mapping and point-reads the blob on the verdict's snapshot. Unknown, unmerged, deleted and unparseable ids read as `None`, as before. So does a DocID alias: the scan compared against the canonical id, and the point read keeps that.

## 2. `find_documents` reads through the field's index

When the field carries a usable index, the matching short ids come from it and only those documents are point-read; otherwise the scan runs as before (`crates/db/src/merge/governance/view_index.rs`). The refusal of a field that is not an `@immutable` scalar LWW field is unchanged, and now happens before any read.

- **Same snapshot, same answer.** A merge (`persist_merged_document`) and a local write (`create_with_indexes` / `update_with_indexes`) each put a document's index entries in the transaction that stores the document, so on one snapshot the index lists exactly what a scan would match. A standalone LWW field merge never writes the materialised document, so it cannot drift from the index.
- **The index definition comes from the snapshot, not the collection cache.** An index defined, or still backfilling, after the snapshot opened is incomplete on it, so it is not trusted and the scan runs. The choice is remembered per (collection, field) for the verdict.
- **A unique index is not used.** A merge resolves a unique conflict to one entry per value (#1111), while a lookup must see both halves of a fork. `@index(unique: true)` falls back to the scan.
- **Single-field indexes only.** A composite index falls back to the scan.
- Matches are still compared against the document's value after the read, so the result can only ever be what the scan would return.

One existing behaviour worth knowing, not changed here: if a merge cannot build the collection's `IndexManager` from the schema, it logs an error and skips index maintenance (`composite_persist.rs`). Such an index would miss documents a scan finds.

## 3. `@immutable` and `@index` on one field

Already accepted: `FieldDescription::validate` only requires an immutable field to be an LWW scalar. A parser test pins it.

## Tests

`crates/db/tests/merge/governance/lookup.rs`:

- `immutable_fields_reads_one_document_not_the_collection`: another document's blob is corrupted so any collection scan fails; the read still succeeds.
- `find_documents_through_an_index_matches_the_scan`: an indexed field against an identical unindexed twin, over two documents sharing a value (one merged from a peer, one written locally), a value none hold, and `share` against `shared`.
- `find_documents_through_an_index_finds_redriven_and_local_documents`: a document merged by re-drive and one written locally are both found.
- `find_documents_through_an_index_reads_only_the_matches`: with a corrupted bystander, the indexed lookup succeeds and the unindexed twin's scan errors.

`crates/query/src/sdl_parse/parser_tests.rs`: `test_parse_immutable_field_with_index`.

`authored` in `governance.rs` now delegates to a multi-field `authored_fields`.

Ran: `cargo test -p db --test merge` (150 pass), the `query` SDL parser tests, and `cargo clippy -p db -p query --all-targets -- -D warnings`. Not run: the full workspace or the integration suite. Verdict time itself is not measured here.


mike7c2:

PR #1791, opened 2026-09-17
fix: a synced definition never strips what a collection commits to (6/7)

mike7c2:
**Stacked on #1790** (`feat/governed-collection-blocks`, on #1789, #1787, #1781, #1780). Review those first. This PR's own diff is its last commit.

## The gap

A collection definition learned over p2p is rebuilt from its delta in `process_collection_definition_delta`, then added to the runtime collection cache. The cache is keyed by **name** and `add_collection_to_cache` inserts unconditionally, so a synced definition displaces the local record of the same name.

What the delta can carry is the whole of `CollectionDefinitionDeltaPayload` (a name, a view's query) plus, per linked `FieldDefinitionDeltaPayload`, a field name, kind and CRDT type. It carries **no policy, no `@immutable` flags and no branchable flag**. The rebuilt record is built with `CollectionVersion::new`, which defaults all three away.

So a peer sending a definition block that merely names a collection this node already has silently strips it. Measured against the unfixed tree, a local `Grants` with a policy and an `@immutable writer` read back as `(None, false)` — no policy, no immutable field — immediately after an unrelated peer's definition for the name "Grants" merged.

That is live, not cosmetic: `is_immutable_scalar_field` backs the merge validator's `find_documents` and its awaited-field deferrals, and the policy is what ACP enforces on. A reload from the systemstore restores it, so the window is until the next cache reload, but nothing forces one.

Note the two records are not even the same collection: an initial synced definition takes `collection_id = version_id = <block CID>`, which never equals a local collection's ID. They collide only in the name, and only in the cache.

## The change

The synced version is still stored under its own version ID and version index, exactly as before. Only the cache insert is now conditional: it is skipped when a collection of that name already commits to something the delta cannot express. `uncarried_commitments` names the three — a policy, `@immutable` fields, branchable history — and the skip is logged with which ones applied.

Patched versions (`payload.name` is `None`) resolve their collection ID from the previous version, so a genuine patch to a local collection is unaffected. A name this node has never seen is unaffected. A name it has, with none of the three, behaves exactly as before.

## A synced definition cannot carry governance at all today

Worth stating plainly, because it bounds this PR: there is no field in the definition delta for a policy, for `@immutable`, or for a root identity. So "a definition arriving with governance keeps it" is not expressible yet — this PR is only the refusal to overwrite. Putting governance into the identities is the subject of the next PR, which is a design note first.

## Tests

`crates/db/tests/merge/governance/definition.rs`, peer definition blocks authored by hand and fed through `handle_block`:

- `a_synced_definition_does_not_strip_a_governed_collection` — local definition first, then the sync. Failed before the fix with `(None, false)` against the expected policy and immutable flag.
- `a_local_definition_replaces_one_synced_first` — the other order. Passed before and after; recorded so the order stops being an open question.
- `a_synced_definition_for_an_unknown_collection_still_registers` — a name this node does not have still merges and lands in the cache with its fields. Passed before and after.

Ran `cargo test -p db` (all green) and `cargo clippy -p db --all-targets -- -D warnings` (clean).

I tried `cargo test -p integration-test --test p2p_iroh -- schema` as the end-to-end check for this path and **it gave no signal**: all 37 tests fail in that binary, and they fail identically on the base commit (`perf/governance-view-index`) with none of these changes applied. The nodes start — `p2p_listening` is reached — and then `schema_add` fails against the node's HTTP API. That looks like a pre-existing breakage or a missing prerequisite in my worktree rather than anything in this PR, but I could not use it to confirm the change, so **this PR has not been exercised end to end over real p2p.** Worth a reviewer running it in a known-good environment.


mike7c2:

PR #1792, opened 2026-09-17
feat: governance in the collection identities (7/7)

mike7c2:
**Implemented.** This replaces the earlier design note; the decisions it was waiting on are settled and built. Stacked on #1791 (`feat/synced-definition-keeps-governance`, on #1790, #1789, #1787, #1781, #1780). This PR's own diff is its last seven commits.

## The gap

A collection ID is a CID over the type name and its field-definition CIDs; a field-definition CID covered a field's name, CRDT type and kind. The version ID equalled the collection ID. Nothing else reached either, so **all four of these shared one identity**:

```graphql
type Agent { did: String, body: String }
type Agent { did: String @immutable, body: String }
type Agent @policy(id: "p1", resource: "agents") { did: String, body: String }
type Agent @branchable { did: String, body: String }
```

A schema's identity committed to the shape of its data and to nothing about who governs it. A node running the same schema without an app's governance configuration was a replica of the same collection ID that merged whatever it liked, and nothing the two nodes exchanged mentioned the disagreement.

## What now enters the identity

The rule is **commitments, not configuration**: does changing it alter what a writer is promised? A representation or performance choice a node can make and unmake does not qualify.

| Commitment | Where | Carried by |
|---|---|---|
| governance root (`@governed(root:)`) | collection ID | `governance` on the collection delta |
| `@immutable` | collection ID, via the field's own CID | `immutable` on the field delta |
| `@branchable` | collection ID | `branchable` on the collection delta |
| policy reference | **version ID only** | `policy` (a CID) on the collection delta |

Everything else stays out. Indexes especially: `DB::create_index` adds one to a live collection and backfills it *under the same version ID*, so an index cannot belong to an identity it would have to change. Putting indexes in would have turned #1789's plan — adding `@index` to `i`, `pos`, `d`, `target_i`, `signer` on `FefraLogEntry` — into a breaking re-ID instead of an online operation. Also out, and load-bearing: `root_id` is a per-node short ID, so hashing it would give the same schema a different identity on every node.

## The root

`type Agent @governed(root: "...")`, declared in the SDL. The root is a self-addressing identifier, not a bare public key, so one identifier survives key rotation and a later change of signing threshold and the identity commits to the identifier rather than to a key. Per-deployment roots come from substituting the root into an SDL template at build time; distinct roots give distinct collection IDs, which is the point.

Enforcement is structural rather than a handshake: a node without the configuration derives a different collection ID, is not a replica, and never exchanges documents for that collection. Nothing has to check anything, and a content-derived identity cannot be lied about the way a handshake claim can.

## Policy, and why the version only

Attaching or amending a policy mints a new **version** of the same collection rather than a different collection whose existing documents no longer belong to it. The CID covers the policy *reference* — its ID and resource name — not the policy's text, which lives wherever ACP keeps it.

**Known hole:** amending a policy at SourceHub therefore changes the rules without changing this CID.

**Consequence worth a reviewer's eye:** `collection_id` and `version_id` now differ for a collection with a policy, where they were always equal at creation. The stored definition block's CID is the version ID, so for such a collection the collection ID no longer names a stored block. That is deliberate — the collection ID is an identity, not a block address — but it is a real change to an invariant and I want it seen rather than assumed harmless.

Two related fixes fell out: the SDL builder's second pass forced `version_id` back to the collection ID and now carries both forward from the first; and growing the delta pushed `CrdtDelta` past clippy's variant-size threshold, so `EffectiveLinkedDelta::Delta` is boxed.

## What this does to #1791's refusal

The delta now carries the root, `branchable` and per-field `immutable`, so a synced definition **restores** all three — one that dropped them would describe a different collection from the one the block names. That narrows #1791's refusal to displace a local record to two cases:

- **a policy**, which survives only as a CID over its reference: enough to bind the version, not enough to reconstruct the reference, so a record holding one is still never rebuilt from a delta;
- **a differing governance root**, which means a different collection that merely shares a name — their collection IDs differ by construction — and the name-keyed cache would otherwise let it take the local one's place.

## Ungoverned identities are unchanged

Non-negotiable, and guarded. Every new field is `Option` or a flag skipped when false, so a collection using none of them serialises to exactly the bytes it did before and keeps its CID. `ungoverned_collection_identities_are_pinned` is unchanged from the design-note commit and stays green throughout.

## Tests

`crates/query/src/sdl_parse/parser_tests.rs`
- `ungoverned_collection_identities_are_pinned` — unchanged guard rail.
- `commitments_reach_the_identity_and_configuration_does_not` — **replaces** `no_governance_directive_reaches_the_identity_today`, naming exactly which moved and which did not; `@index` and `@policy` still assert no movement of the collection ID.
- `each_commitment_mints_a_distinct_identity` — the three are independent and none collide.
- `a_governance_root_changes_the_collection_id`, `the_same_governed_schema_derives_the_same_identity`.
- `a_policy_moves_the_version_id_and_not_the_collection_id`.
- `test_parse_governed_directive`, `test_ungoverned_collection_has_no_root`, `test_governed_requires_a_non_empty_root`.

`crates/defra-core/src/ipld/tests.rs`
- `an_absent_governance_root_is_not_serialised`, `a_governance_root_round_trips_through_both_codecs`, `a_governance_root_changes_the_delta_bytes`. Both codec paths are covered: the serde one the CID is taken over, and the hand-written IPLD one the p2p decode uses, which would otherwise have dropped the root silently.

`crates/db/tests/merge/governance/definition.rs`
- `a_synced_definition_restores_the_commitments_it_carries`.
- `a_synced_definition_under_another_root_does_not_displace_a_local_one`.
- `a_definition_block_reproduces_its_identity_on_a_fresh_node` — end to end: a governed collection is created through the real SDL path, its definition block is taken from the author's blockstore and fed to a node that has never seen it, and the collection ID, version ID, root and immutable field all come back identical.

## What was run

- `cargo test --workspace --exclude integration-test --exclude ffi-test --exclude conformance` — green.
- `cargo clippy -p query -p schema -p defra-core -p db --all-targets -- -D warnings` — clean.
- `conformance` (the TLA/Lean behavioural proofs) — 31 passed, 0 failed, 34 ignored. The ignored ones are Go-parity tests needing a Go binary that is not present. `cid::cid_determinism_same_content_same_docid` and `replication::two_node_replication_converges` pass, which is the signal that matters most for an identity change.
- Not run: `integration-test`, which fails wholesale in this worktree on the base commit for unrelated reasons.


mike7c2:

PR #1795, opened 2026-09-18
fix: stop the p2p merge path discharging work it never did

mike7c2:
Two defects in the p2p merge path, found while building on it. Both share a shape: **a symptom was fixed by adding a compensating write, and the resulting breakage has been worked around ever since rather than corrected.** 

Both are demonstrated by tests that fail on `main` before the fix.

---

# 1. A synced definition displaces the collection it shares a name with

## What happens

`crates/db/src/collection/ops/lookup.rs`

```rust
cache.insert(name, Collection::new(schema));
```

The collection cache is keyed by **name**. A collection's identity is its **collection ID**. The insert was unconditional, so anything sharing a name took the held entry's place.

The p2p definition path reaches this with a record rebuilt from the delta alone. `CollectionDefinitionDeltaPayload` carries a name and a view's query; each linked `FieldDefinitionDeltaPayload` carries a field name, kind and CRDT type. It carries **no policy and no `@immutable` flag**, and the rebuild uses `CollectionVersion::new`, which defaults both away.

Measured on `main` (`synced_definition_cache.rs`): a local `Agent` with a policy and an `@immutable did` reads back, immediately after an unrelated peer's definition for the name "Agent" merges, as

```
(None, false, "bafyreih4plcjlmm3lfnnwyzwtocfzumigbr3lhztuwga4t6k3boedzagwi")
```

No policy. No immutable flag. And the collection ID replaced by the incoming block's CID — the two records are not even the same collection, they collide only in the name.

This is live, not cosmetic: `is_immutable_scalar_field` reads those flags, and ACP enforces on that policy. A `reload_cache()` restores it, so the window is until the next reload, but nothing forces one.

## The history

`cddc228c` (2026-02-04) introduced `add_collection_to_cache`:

> Previously, synced collections were stored in systemstore but not added to the runtime cache, making them invisible to GetCollections even with IncludeInactive=true.

The real problem was that `get_collections` could not see a record that was already durably stored. The fix was to write it into a second, name-keyed structure as well. That made the test pass and created this defect.

The workarounds since are visible in the code:

- `get_collection_by_version_id_full` exists as an "also check the store" fallback, because the cache cannot answer reliably
- `find_collection_by_id` and `get_collection_by_version_id` both **linear-scan** the name-keyed map looking for an identity — `cache.values().find(|c| c.collection_id() == …)`
- `reload_cache()` repairs the drift afterwards

A name-keyed map cannot hold a local and a synced collection of the same name, so those scans can only ever find whichever wrote last.

## The fix

Refuse the insert when the held entry names a **different** collection. A placeholder is a stand-in for a definition that has not arrived, so it still yields. An entry for the same collection is still replaced, which is what the index-update path (`crates/cli/src/index_adapter.rs`) and recovery rely on.

The guard is inside `add_collection_to_cache` rather than at the p2p call site, because it is an invariant about the cache, not a p2p special case.

`cddc228c`'s intent is preserved: a synced collection under a free name still registers and is still visible. That is the second test.

## One property this changes

A patched version is now built by overlaying the block onto the version it supersedes, so a peer-authored version inherits the **local** predecessor's indexes, embeddings and downsample configuration. That is the intent — those are exactly what the delta cannot carry — but it makes the stored `CollectionVersion` under a content-addressed `version_id` explicitly node-local, where it was previously a function of the delta alone. Nothing reads it for verification (`_commits` goes through blocks, and `is_active` already differed per node), so this needs no change; it is recorded because a future reader will otherwise trip over it.

## What is *not* fixed here

The cache is still keyed on a mutable label while two lookups scan it for identities. Re-keying it on `collection_id` with a name index beside it is the real repair, and it touches every `get_collection(name)` caller. This PR stops the data loss; it does not re-key the cache.

---

# 2. A collection block discharges itself even when its links went unprocessed

## What happens

`crates/db/src/merge/merge_handler/collection.rs`

Merging a collection block walks its links, then writes the collection head. The walk tolerated failure three ways:

```rust
Ok(None) => { tracing::warn!(…, "Linked block not found in blockstore"); continue; }
Err(e)   => { tracing::warn!(…, "Failed to load linked block");          continue; }
…
Err(e)   => { tracing::debug!(…, "Composite merge failed"); }   // swallowed
```

Control then falls through to the head write. `any_merged` is still false, so the outcome is `terminal_skip("no linked composites needed merging")` — and replication treats a terminal skip as done, calling `mark_as_merged`.

So a collection block installs a head naming history this node does not hold, **and permanently discharges the push that carried it.** The document it named is never merged and never retried. A transient storage error on one linked composite is enough.

Measured on `main` (`collection_block_links.rs`): a collection block whose linked composite was never delivered returns `Skipped { reason: "no linked composites needed merging", terminal: true }` and installs exactly one head.

## The history

`851bd97d` (2026-02-11):

> Instead of relying on Bitswap to fetch field blocks (which breaks after connection reset), push the complete DAG (field blocks + composite) directly via PushLog. **Field blocks are sent before the composite so receivers have all blocks when processing the composite.**

The real problem was a Bitswap fetch that broke after a connection reset. The fix was to have the *sender* push everything in a helpful order. That is a convention, not an invariant — and the receiver was left tolerating violations of it rather than enforcing anything.

Neither half enforces the property that actually matters.

**What this PR makes hold, precisely:** *the root block's discharge reflects the root block's own completeness.* Not the stronger "a block merges only when its DAG is complete" — see the deliberate gap below.

## Go does enforce it

`internal/db/merge.go`, `processBlock`:

```go
for _, link := range dagBlock.Links {
    nd, err := mp.blockLS.Load(…)
    if err != nil { return NewErrLoadChildBlock(err, link.Link.String()) }
    childBlock, err := coreblock.GetFromNode(nd)
    if err != nil { return NewErrDecodeChildBlock(…) }
    if err := mp.processBlock(ctx, childBlock, link.Link); err != nil {
        return NewErrProcessChildBlock(…)
    }
}
```

Three error returns where Rust has `continue`. `loadComposites` does the same. Structurally: Go calls `UpdateHeads` *before* that loop, but any child failure propagates to `executeMerge` and the transaction never commits, so the head write is rolled back.

## The fix

Track links that could not be processed **but might be later** — a block not held, or a merge that failed with a `Retryable` disposition — and return a retryable skip instead of writing the head. Both the single-block and the batch path check it.

The gate applies to the **root frame only**, and that is a deliberate, tested gap: an ancestor with an unheld link still writes its head and still drops the document it named, exactly as `main` does. `an_ancestor_missing_a_link_does_not_abort_the_root` pins that as intended. The reason is that a partial DAG is ordinary on the receive path — a CAR is truncated at its block and byte caps and the receiver is handed what fitted — so gating ancestors would stall deep catch-up over a large branchable collection entirely. Replication discharges the root, so the root is what the discharge must answer for. Closing the ancestor gap needs the truncation path to re-request, which is not this PR.

A **terminal** failure still skips the link and still installs the head, because replaying it cannot succeed. That deliberately preserves Go's own tolerance:

```go
// The field may not be known to this node - it may belong to a collection version that does not
// exist locally. In this case, we must ignore it as we cannot merge when we do not know what
// kind of CRDT the field is.
if !shouldProcess { return nil }
```

Go's model is three-way — deliberate skip, hard error, success — and `main` collapsed all three into `continue`. This restores the distinction rather than turning every skip into an error.

## Expected behaviour change, and the risk

A block that today frees its pending-DAG slot immediately will now hold one until the missing link arrives or the peer re-pushes. `DEFAULT_MAX_PENDING_DAGS` is 1000, with per-source-peer quotas.

- A link that arrives late now **resolves correctly** instead of being discharged early against history the node lacks. Better, at the cost of a held slot while waiting.
- A link that never arrives changes from **silent loss** to **slot occupancy**. That is the trade this PR makes deliberately: losing a document silently is worse than holding a slot.

**Where an unresolved link ends up.** Nothing re-drives the merge in-process — the retry machinery in `dag_retry.rs` (`MAX_FETCH_ATTEMPTS = 3`, exponential backoff) is the *fetch* phase, not a re-attempt of a merge that skipped. What bounds it is a reaper: `PENDING_DAG_TTL = 300s` (`crates/p2p/src/sync/manager/pending.rs`), evicted by `evict_expired_pending_dags` with a `record_pending_dag_expired()` counter.

So the never-arriving-link case moves from **silent loss at merge time** to **loss at five minutes, with a metric**. Occupancy is bounded rather than open-ended, and the failure becomes observable. That is the trade this PR makes.

`MissingMetadata` ("Collection not found for schema_version_id") is classified `Retryable` and so becomes a held slot under this change. Go treats the analogous case as a deliberate skip. If that proves to be a common shape in cross-version replication, it may want reclassifying — flagging rather than changing it here.

**Partial progress diverges from the Go model cited above.** When some links merge and one is unprocessed, the batch transaction still commits and `MergeComplete` fires for the merged ones, but the outcome is `Skipped`, so `push_to_replicators` does not run. Go rolls the whole thing back instead. Retries are idempotent so this is benign, but the argument above leans on Go's transactional rollback, and Rust does not follow it here.

---

## Tests

`crates/db/tests/merge/synced_definition_cache.rs`
- `a_synced_definition_does_not_displace_a_collection_of_the_same_name` — **failed before**, with the `(None, false, …)` above
- `a_synced_definition_for_a_free_name_still_registers` — passed before and after; guards `cddc228c`'s intent

`crates/db/tests/merge/collection_block_links.rs`
- `a_collection_block_whose_link_is_not_held_installs_no_head` — **failed before**, with the terminal skip and one installed head
- `a_collection_block_whose_links_are_held_merges_and_installs_its_head` — passed before and after; the normal path is untouched
- `an_ancestor_missing_a_link_does_not_abort_the_root` — a partial DAG at an ancestor must not stall deep catch-up

`crates/db/tests/merge/synced_definition_cache.rs` (continued)
- `a_document_for_a_collection_kept_out_of_the_cache_still_merges` — the composition stall; **failed** once fix 1 was in place
- `a_patched_definition_keeps_what_the_delta_cannot_carry` — **failed before** the carry-forward, with `policy: None`

## Verification

- `cargo test -p db` — green
- `cargo test --workspace` (excluding `integration-test`, `ffi-test`, `conformance`) — green
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `conformance` (the TLA/Lean behavioural proofs) — 31 passed, 0 failed, 34 ignored. The ignored ones need a Go binary that is not present. `replication::two_node_replication_converges`, the `partition::convergence_*` set and `cid::cid_determinism_same_content_same_docid` all pass, which is the signal that matters for a change to merge semantics.
  - Note for anyone re-running: `partition::convergence_delete_update_race_preserves_tombstone` fails when run *in isolation* with a name filter, and does so identically on `origin/main`. It passes in the full suite. That looks like test-isolation state, and it is not from this branch.
- Not run: `integration-test`, which fails wholesale in my worktree on the base commit for unrelated reasons.


mike7c2:
## Review

Both defects are real and the diagnoses hold up. Verified `cargo test -p db --test merge` on the PR head: 140 passed, 0 failed.

Three things I'd want settled before merge.

---

### 1. The two fixes compose into a permanent stall — in exactly the scenario test #1 constructs

`find_collection_by_id` (`crates/db/src/collection/ops/lookup.rs:88`) is a **cache-only** linear scan. `crates/db/src/merge/merge_handler/composite.rs:216` uses it to resolve `collection_lookup`; a miss raises `MissingMetadata` (`composite.rs:407`), which `merge_handler/error.rs:108` classifies `Retryable`.

Chain it together: fix #1 keeps the peer's `Agent` out of the cache → its documents can no longer resolve a collection → `MissingMetadata` → fix #2 turns that into a non-terminal skip that withholds the head *and* the discharge. Two collections named `Agent` now means the other one's documents are permanently unmergeable, where before they merged (against a displaced cache entry, but they merged).

The repair is small and already written elsewhere: `get_collection_by_version_id_full` (`collection/ops/version.rs:167`) does cache-then-store. Give `find_collection_by_id` the same fallback and the collision case resolves instead of stalling.

Related, and it strengthens the PR's case: `get_collections_in_txn` reads the **systemstore prefix scan**, not the cache (`crates/db/src/txn/registry/collections.rs:25-30`). `cddc228c`'s stated symptom — "invisible to GetCollections" — is covered without the cache insert at all. That's a better argument than the free-name test.

### 2. A non-terminal skip in an *ancestor* poisons the whole chain

`crates/db/src/merge/merge_handler/collection.rs:207`:

```rust
if is_root || (!outcome.is_merged() && !outcome.is_terminal_skip()) {
    return Ok(outcome);
}
```

A historical collection block with one unheld link now aborts the **root** merge, not just its own head. Under `main` the walk stepped past it.

That matters because partial DAGs are an expected, documented condition on the receive path: `collect_dag_blocks_from_roots` truncates at `CAR_MAX_BLOCKS`/`CAR_MAX_BYTES` and returns what it got (`crates/p2p/src/sync/car.rs:147`), and truncation is only *logged* — nothing re-requests (`sync/coordinator/event_handler/car.rs:158`). So deep catch-up on a large branchable collection goes from "installs a possibly-wrong head" to "never progresses at all". Worth a test at that shape — ancestor block missing a link, root holding everything — before this lands.

### 3. The open question in the description has an answer

There is a reaper: `PENDING_DAG_TTL = 300s` (`crates/p2p/src/sync/manager/pending.rs:15`), `evict_expired_pending_dags` (`sync/manager/process/pending_dag.rs:67`), with a `record_pending_dag_expired()` counter. Nothing re-drives the merge in-process — the entry just expires.

So the never-arriving-link case goes from *silent loss at merge time* to *loss at five minutes, with a metric*. Still the better trade, and occupancy is bounded rather than open-ended. Worth folding into the description, since it closes the thing flagged as unsettled.

---

### Smaller

**Fix #1 is incomplete for the loss it names.** The guard compares `collection_id` only. On the patched-definition path `collection_id` comes from the previous version (`merge_handler/definition.rs:28-38`), so a peer's schema patch for the *same* collection passes the guard and still overwrites with a `CollectionVersion::new` rebuild — no policy, no `@immutable`, plus `is_active = false`. Same data loss, different door. If the invariant is "the cache never loses what the delta cannot carry", the guard needs to be a merge rather than a gate.

**The refusal is invisible to callers.** `add_collection_to_cache` returns `Ok(())` either way, so `definition.rs:173` goes on to log `"Registered synced collection schema in systemstore and cache"` — now false — and `recovery.rs:166` silently no-ops. Return a bool or a small enum.

**Batch path conflates two failures.** `_ => { unprocessed.get_or_insert(... "not held") }` catches `Ok(None)` and `Err(e)` together, so a transient storage error reports as "not held". The non-batch path distinguishes them; cheap to match. (Not setting `unprocessed` on decode failure is right in both — a CID's bytes won't decode differently later.)

**Partial-progress divergence from the Go model cited.** When some links merge and one is unprocessed, the batch txn still commits (`merge_handler/batch.rs:224`) and `MergeComplete` fires for the merged ones, but the outcome is `Skipped`, so `push_to_replicators` is skipped (`p2p/src/sync/replication/handlers.rs:349`). Go rolls the whole thing back. Retries are idempotent so it's benign — but the description leans on Go's transactional rollback as the model, so it's worth noting where Rust doesn't follow.

**Tests:**
- `a_collection_block_whose_link_is_not_held_installs_no_head` asserts `!matches!(outcome, Skipped { terminal: true, .. })`, which also passes on `Merged`. Assert the exact shape being claimed.
- `collection_heads` filters `key.contains("/c/")` across every namespace in the store. Use the headstore prefix.
- `a_synced_definition_does_not_displace_...` discards the `MergeOutcome`. It is `Merged` for a definition that registered nothing in the cache — assert it, and decide whether that's the intended answer.

mike7c2:
All three majors were real. Fixed in `fff0b10b`; `cargo test --workspace` (less the harness crates), `clippy --all-targets -D warnings` and the conformance proofs are green.

### 1. The composition stall — confirmed, and the cause is one step further in

Reproduced it before fixing: `MissingMetadata("Collection not found for schema_version_id: bafyreih4plc…")`, exactly the chain you traced. New test `a_document_for_a_collection_kept_out_of_the_cache_still_merges`.

The cause turned out to be inside `block_collection` rather than `find_collection_by_id`. Its third step already does the store lookup you were pointing at — and then discards the result:

```rust
if let Some(stored) = self.db.get_collection_by_version_id_full(id).await? {
    return Ok(self.db.find_collection_by_id(stored.collection_id())?);  // cache again → None
}
```

It finds the definition in the store and throws it away by re-entering the cache. That has been latent the whole time *because* the clobber kept the cache populated — fixing the clobber exposed it. It now falls back to `stored`, preferring the cached entry where there is one for its index action state. That leaves `find_collection_by_id` cache-only for its other callers, who may well want that.

You're right that `get_collections_in_txn` reads the systemstore prefix scan, so `cddc228c`'s stated symptom is covered without the cache insert at all. I've put that in the description in place of leaning on the free-name test.

### 2. Ancestor poisoning — confirmed

`an_ancestor_missing_a_link_does_not_abort_the_root` failed with `Skipped { reason: "linked block bafyreibxxs… not held", terminal: false }` where the root held everything it named.

The gate now applies to the root frame only. The reasoning I've put in the comment: replication discharges the root, so the root's completeness is what the discharge must reflect; an ancestor is walked as context, and the walk already tolerates an ancestor's hard errors (`Err(error) if !is_root => continue`), so this is consistent with the existing design rather than a carve-out for it.

### 3. The TTL — thank you, that closes it

`PENDING_DAG_TTL = 300s` with `evict_expired_pending_dags` and `record_pending_dag_expired()`. Folded into the description: the never-arriving case goes from silent loss at merge time to loss at five minutes with a metric, so occupancy is bounded and the failure is observable.

### Smaller

**Patch path** — right that it was a second door, though `@immutable` survives it: `prev_fields` is `prev.fields.clone()`, so existing fields keep their flags. What was genuinely lost is the **policy** and the **branchable flag**. Now carried forward from the superseded version (`carry_forward`), the way Go merges onto the previous. `a_patched_definition_keeps_what_the_delta_cannot_carry` fails without it with `policy: None`.

I left `is_active = false` alone — synced versions being inactive until activated looks like stated intent rather than collateral. Say if you read it the other way.

**Invisible refusal** — `add_collection_to_cache` now returns `bool`, and the definition path logs `cached` rather than claiming it cached something it refused.

**Batch conflation** — split into `Ok(None)` and `Err(e)`, matching the single-block walk. Decode failure still sets nothing on both paths.

**Go divergence on partial progress** — noted in the description rather than changed, since the retries are idempotent.

**Tests** — all three: the retryable outcome is asserted by shape rather than by what it is not; `collection_heads` reads the headstore's own prefix; and the definition merge's outcome is now asserted. On the last one, it *is* `Merged` for a definition that registered nothing in the cache, and I think that is right — the version is durably stored under its own ID and is what `get_collections` reads, so the merge did its job. Asserted with a comment saying so, but happy to be argued out of it.

### One thing I did not split

The fixes landed as a single commit rather than one per concern — the changes interleave within `definition.rs` and the two test files, and splitting them cleanly wasn't worth a rewrite. The message covers all five.

mike7c2:
## Second pass — `fff0b10b`

`cargo test -p db --test merge` on the head: 143 passed, 0 failed.

A correction to the first review before anything else: those line references came from a worktree predating `05907a37`, so `block_collection` didn't exist where I looked and I cited `merge_handler/composite.rs:216` directly. The conclusion held, but the citation pointed at an older shape — this pass is against the PR branch.

**All four accepted findings are genuinely fixed.** The `block_collection` fallback is the right repair; the `is_root` gating is correct in both walks (`is_root: false` is set only on parent frames, and the root's `batch_merged_collections` insert stays behind the gate); and the three new tests pin real behaviour.

Two new things.

---

### 1. `carry_forward` covers 2 of ~9 attributes, and indexes are the one that bites

Adding `@index` to the SDL in `a_patched_definition_keeps_what_the_delta_cannot_carry` and asserting `!patched.indexes.is_empty()`:

```
panicked at crates/db/tests/merge/synced_definition_cache.rs:324:
a patch carries no indexes either: []
```

A peer's schema patch wipes `indexes`, `encrypted_indexes`, `fulltext_indexes`, `vector_embeddings`, `downsample_*`, `collection_set` and `is_embedded_only` from the version this node writes — and because the collection IDs match, the guard lets that version into the cache under the live name. That is precisely the entry `merge_handler/mod.rs:151` now says to prefer "for its index action state".

By code reading, not verified: the same shape flips a view. `carry_forward` runs at `merge_handler/definition.rs:116`, and `definition.rs:142` then unconditionally sets `is_materialized = true` and leaves `query` as `None` whenever the patch carries no `query_select`.

The additive form has to be extended every time `CollectionVersion` grows a field, and that has now gone wrong twice. Inverting it is much harder to get wrong:

```rust
let mut schema = previous.clone();   // or CollectionVersion::new(..) when there is none
schema.name = collection_name;
schema.version_id = version_id;
schema.fields = fields;
// then the delta's own overlay: is_active, query/is_materialized, previous_version
```

Same result for policy and `is_branchable`, and `indexes` stop being a special case anyone has to remember.

### 2. The immutable loop in `carry_forward` is dead code

`fields` starts as `prev.fields.clone()` (`definition.rs:87`), so existing fields already carry `immutable`, and only names *absent* from that set are appended. The loop cannot flip a flag that is not already set.

Deleting it and rerunning: all four `synced_definition_cache` tests still pass — including the one whose assertion names it, "the superseded version's immutable fields carry through". It passes for a different reason than it claims.

---

### Smaller

**`add_collection_to_cache` returns `bool` and two of three callers drop it.** `merge_handler/recovery.rs:166` and `cli/src/index_adapter.rs:104,244` still do `...?;`, and `bool` is not `must_use`, so nothing warns. Recovery is the one that matters: it restores a previous version, silently no-ops when the name holds another collection, and returns `Ok(Some(prev))` as though it had. `#[must_use]` on the method would surface both.

**The description's stated invariant no longer matches the code.** "A block merges only when its DAG is complete" was the original framing, and the `is_root` narrowing is the right call — but an ancestor with an unheld link now writes its head and drops its document exactly as `main` did. The new test pins that as intended, so the body should say what actually holds: the root's discharge reflects the root's completeness, and the ancestor case is a known, deliberate gap. Better to name it than to leave a reader to find it.

mike7c2:
Both findings were right, and the inversion was the correct call — my additive list had indeed gone wrong twice. Fixed in `e2f30997`.

### 1. Inverted, rather than extended

`carry_forward` is gone. The version is now built by overlaying the block onto the version it supersedes:

```rust
let mut schema = match &previous {
    Some(previous) => {
        let mut schema = previous.clone();
        schema.name.clone_from(&collection_name);
        schema.version_id.clone_from(&version_id);
        schema.collection_id.clone_from(&collection_id);
        schema.fields = fields;
        schema
    }
    None => CollectionVersion::new(&collection_name, &version_id, &collection_id, fields),
};
schema.is_active = false;
schema.is_placeholder = false;
```

Nothing has to be remembered, and `indexes` stops being a special case. `a_patched_definition_keeps_what_the_delta_cannot_carry` now carries `@index` in the SDL and asserts `!patched.indexes.is_empty()`, with the comment noting this is the very entry the merge path prefers for its index action state.

**Your unverified view claim was right.** `a_patched_definition_does_not_unmake_a_view` fails without the fix on `and it must still be a view`. The `else` branch is now `else if schema.query.is_none()`, so a patch that carries no `query_select` leaves an inherited view alone. A delta cannot express "no longer a view", so inheriting is the same principle as the rest of the overlay.

`is_placeholder = false` is set explicitly — inheriting `true` from a stand-in would have been a new way to get this wrong.

### 2. Dead loop — removed

You're right, and I had actually noticed the `prev_fields` propagation in my last reply and then kept the loop anyway as "safety". That was wrong: it was unreachable, and the assertion naming it passed for a different reason than it claimed. The loop is gone and the assertion now says what it actually tests — that a superseded version's fields carry their own flags.

### Smaller

**`#[must_use]`** added — but worth flagging that it does **not** fire through `?`: `expr?;` yields a bare `bool`, which is not itself `must_use`, so neither call site warned. I handled them explicitly instead. Recovery logs that the version stayed out of the cache; the index adapter now returns an error, since a refusal there would mean an index change that is durable but invisible. (Also only one cli site on this branch, at `index_adapter.rs:201` — your `104,244` were from the older worktree.)

**Description** — rewritten to state what actually holds: *the root block's discharge reflects the root block's own completeness*, not the stronger DAG-completeness claim. The ancestor case is now named as a deliberate, tested gap, with the reason (CAR truncation makes partial DAGs ordinary) and what closing it would need (the truncation path re-requesting), rather than left for a reader to find.

### One thing to be aware of when re-running

`partition::convergence_delete_update_race_preserves_tombstone` fails when run **in isolation** with a name filter — 3/3 for me — and passes in the full suite. It fails identically on `origin/main`, so it is not from this branch; it looks like test-isolation state. Noted in the description.

I also want to correct myself: my earlier "31 passed, 0 failed" claims were from full-suite runs, and I did not notice until now that the isolated form behaves differently. The full suite is green on this branch (31/0/34).

mike7c2:
## Third pass — `e2f30997`

`cargo test -p db --test merge` on the head: 144 passed, 0 failed.

Both findings from the last pass are properly fixed, and the overlay is the right shape — it's the version I'd have written. `else if schema.query.is_none()` is the correct refinement: a materialized view (`query: Some` with `is_materialized: true`) still survives a patch that carries no `query_select`. The two new tests pin exactly the right things.

One thing I flagged last time is void: `cli/src/index_adapter.rs` has only one call site on this branch — `d0bc261d` had already moved index creation under `DB::create_index`.

Three small things, one a real defect in the fix itself.

---

### 1. The `#[must_use]` does not fire

Reverting `merge_handler/recovery.rs` to the dropping form:

```rust
self.db
    .add_collection_to_cache(prev.clone())
    .map_err(MergeError::Database)?;
```

and running `cargo clippy -p db --lib` is clean. No `unused_must_use`.

`#[must_use]` on a function only requires its *return value* be used, and `.map_err(...)` uses it. The `bool` that falls out of `?` is then an ordinary expression statement, which is not linted. So the attribute is decorative, and the commit message's claim that it guards the call sites does not hold. Same reason `crates/db/tests/collection/cache.rs:105`'s `.expect("cache active");` draws nothing.

It works if the *value* carries the attribute rather than the function:

```rust
#[must_use]
pub enum Cached { Taken, NameHeldByAnother }
```

`expr?;` on a `#[must_use]` type does get linted, through `map_err` and all. That also reads better at the call sites than a bare `bool`. Otherwise the attribute should go, since all three callers now handle the answer explicitly and it is only promising something it cannot deliver.

### 2. The attribute is wedged inside the doc comment

`collection/ops/lookup.rs:24-33`:

```rust
/// The collection can be inactive (synced collections start inactive...).
/// Cache `schema` under its name, returning whether it was taken.
#[must_use = "..."]
///
/// `false` means an entry naming a different collection already holds the name
```

Legal — both are attributes — but unintended: the new summary line landed mid-paragraph of the old description, and the attribute splits the block in two. rustdoc renders it as one run-on. Attribute below all the `///` lines, and fold the two summaries into one.

### 3. Worth a sentence in the description

The overlay means a peer-authored version now inherits the *local* predecessor's indexes, embeddings and downsample configuration. That is the intent and the right default — but it makes the stored `CollectionVersion` under a content-addressed `version_id` explicitly node-local, where before it was a function of the delta alone. Nothing reads it for verification (`_commits` goes through blocks, and `is_active` already differed per node), so I do not think it needs changing. It is a property a future reader will trip over, and one line in the body would save them.

mike7c2:
All three taken, in `1f852a38` — and you were right that flagging the `#[must_use]` problem while leaving the attribute in place was the worst of both.

### 1. `#[must_use]` moved onto the value

Your reading is exactly right, and I reproduced your experiment: reverting recovery to the dropping form drew nothing.

The answer is now a type that carries the attribute:

```rust
#[must_use]
pub enum Cached {
    Taken,
    NameHeldByAnother,
}
```

With the same revert, it fires:

```
warning: unused `collection::ops::lookup::Cached` that must be used
   --> crates/db/src/merge/merge_handler/recovery.rs:167:25
    |
167 | /                         self.db
168 | |                             .add_collection_to_cache(prev.clone())
169 | |                             .map_err(MergeError::Database)?;
```

Through `map_err` and `?`, as you said it would. It also caught the two sites in `crates/db/tests/collection/cache.rs` that were dropping the answer — they now assert `Cached::Taken`, which is what those tests actually mean. And it does read better than a bare `false` at the call sites.

### 2. Doc comment unwedged

Rewritten into one block with the attribute below it. That was my `sed` dropping the new summary into the middle of the old description and splitting the comment with the attribute — thanks for catching it rather than leaving it to rustdoc.

### 3. Node-local `CollectionVersion` — added

Added to the description as its own short section: a peer-authored version now inherits the local predecessor's indexes, embeddings and downsample configuration, which makes the stored record under a content-addressed `version_id` explicitly node-local where it used to be a function of the delta alone. Recorded with your reasoning for why it needs no change — nothing reads it for verification, and `is_active` already differed per node.

Noted too that `cli/src/index_adapter.rs` has one call site, not three; `d0bc261d` had already moved index creation under `DB::create_index`.

---

`cargo test --workspace` (less the harness crates), `clippy --all-targets -D warnings` and conformance (31/0/34) all green on the head.

crates/db/src/merge/merge_handler/collection.rs
```diff
@@ -362,6 +380,29 @@ impl<S: Store, B: blockstore::Blockstore> DbMergeHandler<S, B> {
...
+        // stall deep catch-up entirely. The walk already tolerates an
+        // ancestor's hard errors for the same reason.
+        if let Some(reason) = unprocessed {
+            if is_root {
+                return Ok(MergeOutcome::retryable_skip(reason));
+            }
+            tracing::debug!(
+                %cid,
+                reason,
+                "Ancestor collection block has an unprocessed link; continuing the walk"
```
jackzampolin, line 387: **[bug]** When `unprocessed` is set and `is_root` is false, this branch only logs and then falls through into the head write and the `merged_collections` insert. The frame loop treats that `Merged` or terminal skip as finished, so the root can still be discharged — but the ancestor CID is now in `merged_collections`. Both Enter checks then return `terminal_skip("collection already merged")` before the body runs. A later attempt, including the retryable re-drive of this same root once the missing link is stored, never walks those links. The composite is dropped, and a later push of the block that named it is discharged. The batch copy at line 802 does the same via `batch_merged_collections`, which is committed into `merged_collections` even when the root outcome is a non-terminal skip. `an_ancestor_missing_a_link_does_not_abort_the_root` only checks the first attempt.

**Suggestion:** For an ancestor, continue the walk without writing its head and without inserting the CID into `merged_collections` or `batch_merged_collections`. The root can still commit. A later delivery or retry then sees the ancestor as not merged, and the root gate can apply the link. Extend the ancestor test so a second merge, after the missing composite is stored, actually merges that document.

crates/db/src/merge/merge_handler/collection.rs
```diff
@@ -731,6 +792,17 @@ impl<S: Store, B: blockstore::Blockstore> DbMergeHandler<S, B> {
...
         }
 
+        if let Some(reason) = unprocessed {
+            if is_root {
+                return Ok(MergeOutcome::retryable_skip(reason));
+            }
+            tracing::debug!(
+                %cid,
+                reason,
+                "Ancestor collection block has an unprocessed link; continuing the walk"
```
jackzampolin, line 785: **[bug]** Same fallthrough as the single-block walk: an ancestor with `unprocessed` set still inserts into `batch_merged_collections` and writes its head. That set is copied into `merged_collections` when the batch commits, including when the root outcome is a non-terminal skip. The root's retry then `has_merged_collection`-skips the ancestor, so a link that arrives between the two attempts is never merged.

**Suggestion:** Continue the ancestor walk without the head write and without `batch_merged_collections.insert`. Leave the root free to commit.

crates/db/src/collection/ops/lookup.rs
```diff
@@ -21,21 +36,44 @@ impl<S: Store> crate::database::DB<S> {
...
+                    collection_name = %name,
+                    held = %existing.collection_id,
+                    offered = %schema.collection_id,
+                    "Refusing to displace a cached collection with a different collection ID"
+                );
+                return Ok(Cached::NameHeldByAnother);
+            }
+        }
+
         cache.insert(name, Collection::new(schema));
```
jackzampolin: **[suggestion]** This insert is always `Collection::new`, which puts every index into both `write_indexes` and `queryable_indexes`. `with_index_actions` is what keeps `ERRORED` indexes out of writes and in-progress indexes out of queries, and only `load_collections` applies it. A patched definition of the same collection passes the new guard and replaces that filtered entry. `block_collection` then prefers the cache because it "carries index action state," so the entry just rebuilt without action status wins until the next reload.

**Suggestion:** When taking a schema whose `collection_id` is already cached, rebuild with `index_action_statuses` the way `load_collections` does, instead of `Collection::new`.

jackzampolin (commented):
## Summary

The two fixes target the right defects. A different collection can no longer displace a live name-keyed cache entry, and a root collection block whose own links are missing or failed retryably now returns a non-terminal skip without installing a head. Overlaying a patched definition onto the version it supersedes is the right way to keep policy, indexes, and view state the delta cannot express. The ancestor arm of the new gate undoes part of that: it logs and then falls through into the success path, so an incomplete ancestor is recorded as merged. A later retry of the same root — the retry the new gate exists to allow — skips that ancestor and never merges the composite, even after the link arrives.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

mike7c2:
Thanks — all three landed, in `a52f174f`, plus the tests you asked for in `9aa8ee89`.

**The ancestor fallthrough (both walks).** Both arms now return before the head write and before the `merged_collections` / `batch_merged_collections` insert, so an unprocessed ancestor records nothing and the walk still continues. `Merged` when a link of its own merged, a terminal skip otherwise.

**Writing the test you asked for turned up a limit worth stating.** I first wrote it the way you described — a second merge of the *same root* after the missing composite is stored — and it fails. The root's own links were all held, so it merges and is discharged; a re-drive hits the Enter guard, and the walk prunes at that merged CID before it ever reaches the ancestor. A later root superseding it fails the same way, for the same reason.

So the gate does not create a retry of the ancestor. What it buys is that the ancestor was never falsely recorded, so it merges whenever that block is delivered again. That is what `an_incomplete_ancestor_merges_once_its_missing_link_arrives` pins, and the limitation is written into its doc comment rather than left for the next reader to rediscover. If you think the root's discharge should also be gated on its ancestors, that is a bigger change than this PR and I'd rather do it separately — say so and I'll open an issue.

`the_batch_walk_leaves_an_incomplete_ancestor_mergeable` covers `merge_collection_block_batch`, which had no test at all. I checked it bites by reverting the 785 arm: it fails, and passes again with it restored.

**The cache insert** (`lookup.rs`) now reads `index_action_statuses` and builds with `with_index_actions`, as `load_collections` does. The read happens outside the cache swap so the swap stays non-blocking.

**CI.** It was red on eight jobs from one cause: `add_collection_to_cache` became async with that change and the CLI's index adapter was left un-awaited, so `map_err` was being applied to the future. Fixed in `9aa8ee89` along with the `cargo fmt` drift.

**CodeRabbit's `recovery.rs` finding** — the reconstructed predecessor carrying only name, IDs and fields — is real, and the overlay in `3824ede1` makes it matter more, since the overlay now starts from that partial version. It predates this PR (`git blame` → `627ca7b9`), so I've filed it as #1831 rather than widen this one.

Local: `cargo test -p db -p cli` green across 36 binaries, `cargo clippy -p db -p cli --all-targets -D warnings` clean.

vertexclique (approved):

Issue #1470, opened 2026-08-14
Vector index param validation gaps: SSG bypasses HNSW limits, IVF-PQ accepts an m that cannot divide the width


jackzampolin:
Two related gaps let an index be created that either does unbounded work or can
never build.

**SSG bypasses the HNSW limits.** `Params::validate()` runs only when
`algorithm == Hnsw` (`crates/db-index/src/vector/index.rs:57`), but `engine()`
hands the same unvalidated `m` / `ef_construction` / `ef_search` to
`Ssg::try_new`, which builds an HNSW staging graph from them. The comment says
"each algorithm validates only its own block, so parameters belonging to another
are inert" — true for FLAT and IVF-PQ, but for SSG the HNSW block is not inert,
it drives the staging graph. Posting
`{Algorithm: "SSG", HNSW: {EfConstruction: 100000000, M: 100000}}` bypasses
`MAX_EF_CONSTRUCTION` / `MAX_M`, and one document insert then does unbounded
work — exactly what the limits at `params.rs:13-19` exist to prevent.

**IVF-PQ accepts an m that can never divide the width.** `resolved_m`
(`engine/ivfpq/params.rs:68`) clamps a caller-supplied `m` to `min(dimensions)`
without requiring it to divide the width, and index creation validates IVF-PQ by
constructing a throwaway `IvfPq::try_new(MemoryNodeStore::new(), ...)`
(`index.rs:87`) that never sees the dimensions. So
`@vectorIndex(dimensions: 768, algorithm: IVF_PQ, IVFPQ: {m: 5})` is accepted at
create time, and every later `build()` fails inside `ProductQuantizer::train`
with "product quantizer needs m to divide the width: 5 does not divide 768".
The index can never leave exact-scan mode, and the error surfaces far from the
call that caused it.

Issue #1469, opened 2026-08-14
IVF-PQ tie-breaks distances opposite to Flat and HNSW, so equidistant hits differ from the oracle


jackzampolin:
`crates/db-index/src/vector/engine/ivfpq/search.rs:117` — IVF-PQ's `Ranked`
breaks distance ties with `other.id.cmp(&self.id)`, the opposite of
`Flat::Ranked` and `Candidate` (`candidate.rs:38`, `flat.rs:141`), which order
ties ascending by id.

With two documents at identical distance, ids 4 and 9, and `k` already full:
Flat and HNSW drop the larger id; this heap treats the smaller id as "greater"
and pops id 4 instead. So the engines return different documents for the same
query, and the differential test against the Flat oracle
(`ivfpq_recall_baseline.rs`) disagrees on any tied pair.

Ties are common with quantized codes, which is precisely where IVF-PQ lives, so
this is not as rare as exact float ties would be.

PR #1798, opened 2026-09-18
fix(index): preserve HNSW connectivity during updates

### LINKED ISSUE #1468: HNSW: re-inserting a document creates a self-link at every layer, permanently wasting an edge slot
`crates/db-index/src/vector/engine/hnsw/insert.rs:62` — re-inserting an existing
id gives the node a self-link at every layer.

`put_node` at line 62 stores the new vector with empty layers, but neighbours
from the node's prior insert still point at `id`, so `search_layer` reaches it
and `self.candidate()` scores it at distance 0. `Heuristic::select` has no
zero-distance guard — unlike `Angular::select`, which checks
`norm_c <= f64::EPSILON` — so `id` is selected first. `layers[layer]` then gets
`id`, and `add_link(id, id, layer, max)` writes a self-edge.

Every document update takes this path. One of `m` edge slots per layer is
permanently spent on a link the visited-set makes a no-op, so recall degrades
for the life of the index in proportion to how often documents are updated.

iverc:
## Context
Updating an existing HNSW node could introduce self-links and erase the routes needed to find its replacement neighbors, disconnecting parts of the graph.

## Changes
- Preserve existing links until replacement neighbor selection completes.
- Exclude the updated node from its own neighbors and avoid duplicate backlinks.
- Retain descent routes when the updated node is alone on an upper layer.

## Testing
- [x] Repeated-update regression reproduces lost connectivity before the fix.
- [x] Two-node regression covers replacement of the only upper-layer node.
- [x] Combined validation with #1469 and #1470: 203 vector tests and 20 embedded-node query tests passed; 12 existing measurement tests remain ignored.
- [x] Strict Clippy, formatting, and diff checks.

Closes #1468.


crates/db/src/index/vector/engine/hnsw/insert.rs
```diff
@@ -130,7 +142,10 @@ impl<S: VectorNodeStore> Hnsw<S> {
         while node.layers.len() <= layer {
             node.layers.push(Vec::new());
         }
-        node.layers[layer].push(to);
+        node.layers[layer].retain(|id| *id != from);
+        if !node.layers[layer].contains(&to) {
+            node.layers[layer].push(to);
+        }
```
vertexclique, line 148: retained must be purged in the next index rebuild in async background. You are not doing that in the code here.

vertexclique (approved):

PR #1802, opened 2026-09-18
fix(tooling): make browser setup optional on macOS

### LINKED ISSUE #1459: just setup fails on macOS without a system Firefox
#1377 adds `setup-browser` to the default `just setup` chain:

```
setup: setup-rust setup-jq setup-cargo-tools setup-protoc setup-go setup-jdk setup-lean setup-tla setup-browser
```

`setup-browser` installs geckodriver on any platform, then looks for a Firefox
to drive. If none is on `PATH` and none is in `.tooling`, it falls through to a
platform case that only has download URLs for Linux:

```
case "$(uname -s)-$(uname -m)" in
    Linux-x86_64|Linux-amd64)  ff_platform="linux-x86_64" ;;
    Linux-aarch64|Linux-arm64) ff_platform="linux-aarch64" ;;
    *) echo "error: no firefox on PATH and no pinned build for $(uname -s)-$(uname -m)" >&2
       echo "       install Firefox, then re-run: just setup-browser" >&2; exit 1 ;;
esac
```

On `Darwin-arm64` with no system Firefox that is `exit 1`, which fails the whole
`just setup`. macOS is this project's primary development platform and CI runs
on self-hosted macOS ARM64, so a fresh Mac without Firefox cannot complete
setup — and the failure lands at the end, after every other tool has installed.

Mozilla does publish macOS builds (`mac/en-US/Firefox <version>.dmg`), but they
ship as a DMG rather than the `tar.xz` the Linux path untars, so this is not
just a matter of adding two case arms.

Either add a Darwin branch that mounts and copies from the DMG, or make a
missing Firefox non-fatal: warn, and let `just test-wasm` be the thing that
errors when it is actually needed. `test-wasm` already has its own clear
"run: just setup-browser" guard, so the second option keeps setup cheap without
losing the signal.

iverc:
## Context
General onboarding depended on browser setup, which failed on macOS when Firefox was absent from PATH, even when installed as an application bundle.

## Changes
- Keep browser setup explicit rather than requiring it for `just setup`.
- Share Firefox discovery across setup and both browser test recipes.
- Recognize PATH, the local tooling directory, and macOS system/user application bundles.
- Document optional browser setup; leave pinned Linux downloads unchanged.

## Testing
- [x] Shell regressions cover discovery precedence, paths with spaces, and non-executable or missing browsers.
- [x] Shell syntax checks passed.
- [x] Parsed setup recipes and dry-ran both browser test recipes with `just`.
- [x] Diff checks.

Closes #1459.


justfile
```diff
@@ -515,10 +512,8 @@ test-wasm:
     #!/usr/bin/env bash
     set -euo pipefail
     command -v geckodriver >/dev/null 2>&1 || { echo "error: no geckodriver; run: just setup-browser" >&2; exit 1; }
-    if ! command -v firefox >/dev/null 2>&1 && [ -x "{{ tooling }}/firefox/firefox" ]; then
-        export PATH="{{ tooling }}/firefox:$PATH"
-    fi
-    command -v firefox >/dev/null 2>&1 || { echo "error: no firefox; run: just setup-browser" >&2; exit 1; }
+    firefox="$(bash tools/find-firefox.sh "{{ tooling }}")" || { echo "error: no firefox; run: just setup-browser" >&2; exit 1; }
+    export PATH="$(dirname "$firefox"):$PATH"
```
jackzampolin: Prepending `dirname` of every resolved Firefox undoes the tool order at justfile:49-58 (`.tooling/bin` and rustup's Cargo bin before the system `PATH`).

`tools/find-firefox.sh` returns `command -v firefox` first. On Linux that is usually `/usr/bin/firefox` or `/snap/bin/firefox`, and `setup-browser` then skips the pinned tarball. This export becomes `PATH="/usr/bin:$PATH"` (or `/snap/bin`). I reproduced it: before the prepend, `cargo` and `geckodriver` resolved to the rustup and `.tooling/bin` copies; after it, both resolved to the binaries sitting next to `firefox`.

The `geckodriver` check on line 514 runs *before* this export, so it can pass on the pinned 0.37.1 driver while `wasm-bindgen-test-runner` walks `PATH` and starts the other one. The runner does not need Firefox's directory on `PATH` when `firefox` is already there. The old recipe only prepended `.tooling/firefox` when `command -v firefox` failed. Prepending `.tooling/firefox` or `Firefox.app/Contents/MacOS` is fine — those directories do not contain `cargo` or `geckodriver` — but that is not this branch.

`tools/test-find-firefox.sh` only checks the printed path, so it stays green.

Prepend only when Firefox is not already on `PATH`:

```bash
firefox="$(bash tools/find-firefox.sh "{{ tooling }}")" || { echo "error: no firefox; run: just setup-browser" >&2; exit 1; }
if ! command -v firefox >/dev/null 2>&1; then
    export PATH="$(dirname "$firefox"):$PATH"
fi
```

Same bug at line 526, with `dirname` inserted ahead of `cargo_bin`. `tools/browser-p2p-e2e.sh:35` and `:76` then run distro `cargo` instead of rustup. Keep `.tooling/bin` first, and insert `dirname` only inside that `if`.

justfile
```diff
@@ -527,10 +522,8 @@ test-wasm:
 test-browser-p2p:
     #!/usr/bin/env bash
     set -euo pipefail
-    if ! command -v firefox >/dev/null 2>&1 && [ -x "{{ tooling }}/firefox/firefox" ]; then
-        export PATH="{{ tooling }}/firefox:$PATH"
-    fi
-    export PATH="{{ tooling_bin }}:$PATH"
+    firefox="$(bash tools/find-firefox.sh "{{ tooling }}")" || { echo "error: no firefox; run: just setup-browser" >&2; exit 1; }
+    export PATH="{{ tooling_bin }}:$(dirname "$firefox"):$PATH"
```
jackzampolin: `{{ tooling_bin }}` stays first, but `$(dirname "$firefox")` is still inserted in front of `cargo_bin`. When discovery returned `/usr/bin/firefox` (the common Linux case; the pinned build is skipped whenever `firefox` is on `PATH`), a distro `cargo` in that directory beats rustup. `tools/browser-p2p-e2e.sh` runs `cargo build` (line 35) and `cargo test` (line 76) with this `PATH`.

Prepend `dirname` only when `command -v firefox` fails (app bundle or `.tooling/firefox/firefox`). Leave `PATH` alone when Firefox was already resolved from `PATH`, and keep `.tooling/bin` first either way. See the note on `test-wasm` line 516.

justfile
```diff
@@ -82,9 +82,9 @@ _sha256 file:
 # Setup
 # ---------------------------------------------------------------------------
 
-# Install every dependency needed to develop, test and verify the database.
+# Install core development tools. Browser tests also need `just setup-browser`.
 [group('setup')]
-setup: setup-rust setup-jq setup-cargo-tools setup-protoc setup-go setup-jdk setup-lean setup-tla setup-browser
+setup: setup-rust setup-jq setup-cargo-tools setup-protoc setup-go setup-jdk setup-lean setup-tla
```
jackzampolin: This drops `setup-browser` on Linux too, where that recipe does not fail: it installs geckodriver and, when `find-firefox.sh` finds nothing, still downloads the pinned tarball (justfile:237-255). A fresh Linux `just setup` used to be enough for `just test-wasm`. It no longer is, and the recipe still finishes by running `doctor`.

`doctor` (justfile:449-452) lists docker/node/npm/python3 as optional and never mentions `firefox` or `geckodriver`, so the skip is invisible in the report `setup` tells people to trust. #1459 asked to warn and continue when macOS has no Firefox, not to stop provisioning the OS that already had a pinned build. The README note only helps if someone re-reads onboarding.

Keep `setup-browser` on this chain and make only the `*)` arm in `setup-browser` a warning. If opt-in on every OS is deliberate, add both tools to doctor's optional list.

jackzampolin (commented):
## Summary

Browser discovery itself is right: `tools/find-firefox.sh` checks `PATH`, then `.tooling/firefox/firefox`, then the macOS user and system app bundles, and `tools/test-find-firefox.sh` passed here (precedence, a home directory with spaces, and a non-executable bundle). Removing the hard failure from `just setup` is what unblocks a Mac whose Firefox is only an application bundle.

The recipe integration regresses the non-macOS path. Both browser test recipes now prepend `dirname` of whatever binary was found, including a Firefox that was already on `PATH`. That undoes this justfile's tool order. Separately, `setup-browser` is dropped from `just setup` on Linux as well, where that recipe still succeeds, and `just doctor` never mentions the tools that disappeared.

## Issues

### Issue 1 -- Severity: bug

justfile:516

`test-wasm` and `test-browser-p2p` always prepend `dirname` of the resolved Firefox onto `PATH`. `find-firefox.sh` returns `command -v firefox` first, so on Linux this is usually `/usr/bin/firefox` or `/snap/bin/firefox` (and `setup-browser` then skips the pinned tarball, because a system binary counts as found). The export becomes `PATH="/usr/bin:$PATH"` or `PATH="/snap/bin:$PATH"`.

That reverses the order set at justfile:49-58, where `.tooling/bin` and the rustup Cargo bin directory are supposed to beat the system path. Reproduced with a throwaway `PATH`: before the prepend, `cargo` and `geckodriver` resolved to the rustup and `.tooling/bin` copies; after `export PATH="$(dirname "$firefox"):$PATH"`, both resolved to the copies sitting next to `firefox`.

`test-wasm` checks `geckodriver` on line 514 before the prepend, so the check can pass on the pinned 0.37.1 driver while `wasm-bindgen-test-runner` later walks `PATH` and starts the other one. The runner does not need Firefox's directory on `PATH` when `firefox` is already there; geckodriver locates it. The previous recipe only prepended `.tooling/firefox` when `command -v firefox` failed.

justfile:526 has the same hole with `tooling_bin` kept first. `dirname` is still inserted in front of `cargo_bin`, and `tools/browser-p2p-e2e.sh` (line 35 and line 76) runs `cargo build` / `cargo test`. A distro `cargo` in `/usr/bin` then wins over rustup. Prepending `.tooling/firefox` or `Firefox.app/Contents/MacOS` is safe — those directories do not contain `cargo` or `geckodriver` — but that is not the branch taken when Firefox is already on `PATH`.

`tools/test-find-firefox.sh` only asserts the printed path. It never runs these exports, so it stays green.

Suggestion: prepend only when Firefox is not already on `PATH`:

```bash
firefox="$(bash tools/find-firefox.sh "{{ tooling }}")" || { echo "error: no firefox; run: just setup-browser" >&2; exit 1; }
if ! command -v firefox >/dev/null 2>&1; then
    export PATH="$(dirname "$firefox"):$PATH"
fi
```

For `test-browser-p2p`, keep `.tooling/bin` first either way, and insert `dirname` only in that same `if`.

Status: open

### Issue 2 -- Severity: suggestion

justfile:87

`setup` no longer runs `setup-browser` on any OS. On Linux that recipe does not fail: it installs geckodriver and, when `find-firefox.sh` finds nothing, still downloads the pinned tarball (justfile:237-255). A fresh Linux `just setup` used to leave the machine able to run `just test-wasm`. It no longer does, and the run still ends in `just doctor`.

`doctor` (justfile:449-452) reports docker, node, npm, and python3 as optional and does not mention `firefox` or `geckodriver` at all, so the skip is invisible in the command the recipe tells people to trust. #1459 asked to warn and continue when macOS has no Firefox, not to stop provisioning the platform that already had a pinned build. The README note helps only if someone re-reads onboarding; `just setup` itself prints nothing.

Suggestion: keep `setup-browser` on the `setup` chain, and make only the `*)` arm in `setup-browser` a warning (non-zero only when a later `test-wasm` / `test-browser-p2p` actually needs a browser). If browser setup is meant to be opt-in everywhere, add `firefox` and `geckodriver` to doctor's optional list so the final report shows them missing.

Status: open

PR #1803, opened 2026-09-18
refactor: key the collection cache by collection ID

mike7c2:
**Stacked on #1795** (`fix/collection-merge-swallows-failures`). Review that first; this PR's own diff is its last commit.

#1795 fixes the data loss. This is the repair underneath it: the collection cache was keyed on the wrong thing, and this keys it on identity.

## What was wrong

The process-wide collection cache (`DB::collections`) was a `RapidHashMap<String, Collection>` keyed by **name**. A collection's identity is its **collection ID**; its name is a label that two collections can share. Three consequences, all visible in the code:

- **Both identity lookups were linear scans** over the name-keyed map:
  ```rust
  // find_collection_by_id
  cache.values().find(|c| c.collection_id() == collection_id)
  // get_collection_by_version_id
  cache.values().find(|c| c.version_id() == version_id)
  ```
- **One entry per name meant one version per collection**, so the cache-only `get_collection_by_version_id` could never answer for a superseded version. That's why `get_collection_by_version_id_full` exists — and `block_collection` got exactly that wrong, which was the third fix in #1795.
- **The guard #1795 added protected one entry point.** It lived in `add_collection_to_cache`, while eleven other writers did `cache.insert(name, Collection::new(schema))` directly and would have displaced anything.

## What this does

`CollectionMap` (`crates/db/src/collection/map.rs`) stores entries by collection ID, with a name index and a version index kept beside them, so all three lookups are direct.

Names and entries stay **one-to-one**. GraphQL resolves a collection by name, so a name must answer with exactly one collection — re-keying doesn't remove the need to arbitrate a name, it moves that decision into one place.

**Every writer now names its authority:**

| Method | For | On a taken name |
|---|---|---|
| `put` | a version this node committed as current | takes it, as the raw inserts did |
| `offer` | a definition that arrived from elsewhere | takes it only if free, the same collection, or a placeholder; otherwise returns `Cached::NameHeldByAnother` |

The rule that guarded one method is now a property of the type, and which rule applies is visible at each call site rather than depending on which function happened to be called. `add_collection_to_cache` becomes a thin `offer` with logging.

## Behaviour

Preserved everywhere but one place, and that one is a fix.

Each of the eleven raw inserts keyed the map by a variable (`name`, `actual_name`, `collection_name`, `coll_name`) rather than by the schema's own name. I traced all of them: every one equals the schema's name, because validation refuses to rename a collection (`"collection name cannot be mutated"` in `update_validators.rs`). So keying by the schema's own name is identical for all of them.

**The exception is a placeholder**, which validation *does* let be renamed. Previously a placeholder defined under a new name left the entry reachable under the placeholder's old name. It now moves to the new name. `a_collection_under_a_new_name_moves_rather_than_answering_to_both` pins it.

One incidental cleanup: `patch/store.rs` carried each candidate's name through a loop only to use it as the cache key. The map derives the key itself, so the candidates are now just schemas.

## What this does *not* remove

Worth being plain about, since "re-key it and drop the workarounds" is the natural expectation:

- **`get_collection_by_version_id_full` stays.** Only each collection's current version is cached, so a superseded version still has to come from the store.
- **`block_collection`'s store fallback stays**, for the same reason, and for anything not yet loaded.
- **`reload_cache` stays.** `load_collections` puts into the existing map rather than rebuilding it, as before.

I considered caching *every* version by keying on `version_id` instead. It would let `_full` short-circuit for any version seen since startup, but it changes what the cache means and what `get_collection` has to choose between, and `_full` would still be needed for versions never loaded. Declined here.

So this makes the lookups direct and the collision rule structural. It doesn't delete the store fallbacks, because they aren't workarounds for the *keying* — they're the only way to reach history the cache was never meant to hold.

## Tests

`crates/db/tests/collection/map.rs` pins the invariants:

- `a_put_collection_resolves_by_name_id_and_version`
- `a_new_version_replaces_the_old_one_in_every_index` — the superseded version leaves the version index
- `offering_a_different_collection_for_a_held_name_leaves_it_alone` — #1795's rule, now on the type
- `offering_takes_a_free_name_the_same_collection_or_a_placeholder`
- `putting_a_different_collection_takes_the_name_and_evicts_the_holder` — the displaced entry leaves every index, rather than lingering nameless
- `removing_a_name_clears_every_index`
- `a_collection_under_a_new_name_moves_rather_than_answering_to_both`
- `iteration_and_snapshots_see_each_collection_once_under_its_name`

Every test from #1795 passes unchanged, which is the regression bar this was stacked for.

## Verification

- `cargo test -p db` and `cargo test --workspace` (excluding `integration-test`, `ffi-test`, `conformance`) — green
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `conformance` — **flaky on both this branch and its parent, at the same rate.** Over eight full-suite runs each: this branch 2 failures, parent (`1f852a38`) 2 failures. The failures are `partition::convergence_delete_update_race_preserves_tombstone` and `index::index_no_stale_no_missing` — both timing-sensitive multi-node tests, both failing on the parent too. At that sample size I can't claim to have *proven* no regression, but I found no evidence of one, and nothing in this diff touches the paths those tests exercise beyond lookups whose results are unchanged.


vertexclique:
cc/ @mike7c2 
This makes collections a RwLock<CollectionMap>, and my open #1788 makes it an Atom<RapidHashMap>. Same field, same call sites, incompatible. Worth noting that CollectionMap takes &mut self, so under the lock-free shape every write becomes a read-copy-update that clones three maps. Whichever lands second redoes the other's work, and the second author should know that before starting.

crates/db/src/collection/map.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn take_name(&mut self, collection: Collection) {
+        let id = collection.collection_id().to_string();
+        let name = collection.name().to_string();
+        if let Some(holder) = self.id_by_name.remove(&name) {
+            self.evict(&holder);
+        }
+        self.evict(&id);
+        self.id_by_name.insert(name, id.clone());
+        self.id_by_version
+            .insert(collection.version_id().to_string(), id.clone());
```
jackzampolin, line 140: `take_name` always seizes `collection.name()` and evicts whatever id holds it, then drops this id's previous name. Callers that still look the collection up by a different name (`store_new_version`'s `actual_name`, `load_collections`' name-index key) lose the entry `get_collection` / `patch_collection` need. If that name belongs to a different collection id, this also removes it from `by_id` and `id_by_version`. `offer` is the path that must refuse that displacement; `put` should not delete another id.

crates/db/src/definition/patch/store.rs
```diff
@@ -511,7 +511,7 @@ impl<S: Store> crate::database::DB<S> {
             if new_schema.is_active {
                 // New version is active - cache it under the actual collection name
                 // (not collection_name, which might be a version_id for branching patches)
-                cache.insert(actual_name.to_string(), Collection::new(new_schema.clone()));
+                cache.put(Collection::new(new_schema.clone()));
```
jackzampolin, line 514: The comment above still says to cache under the actual collection name, and the name-index write earlier in this function still uses `actual_name` (`CollectionNameKey::new(actual_name)`), as do `schema_heads` and `maybe_reindex_on_version_switch`. `put` indexes `new_schema.name` instead. For a placeholder rename those differ: `get_collection(actual_name)` and `patch_collection(actual_name)` miss (cache-only name lookup; a miss is an unknown collection), and if another cached id already holds `new_schema.name` it is evicted. The old `insert(actual_name, ...)` did neither. Same-name patches are unchanged.

crates/db/src/collection/ops/mod.rs
```diff
@@ -216,7 +215,7 @@ impl<S: Store> crate::database::DB<S> {
                     .get(&schema.collection_id)
                     .cloned()
                     .unwrap_or_default();
-                cache.insert(name, Collection::with_index_actions(schema, &actions));
+                cache.put(Collection::with_index_actions(schema, &actions));
```
jackzampolin, line 218: `name` is the `/collection/name/{name}` key this loop just loaded, and the old `insert(name, ...)` cached under it. `put` ignores `name` and keys by `schema.name`. When those differ (placeholder rename leaves the name index on the old name), reload does not restore lookup by the registered name, so `get_collection` / `patch_collection` of that name keep missing after restart. Cache under `name`, or only `put` after the name index has moved to `schema.name`.

jackzampolin (commented):
## Review

Same-name updates, id lookup, and `offer` refusing a different non-placeholder holder all match the old name-keyed map. The break is `put`: it rekeys by `collection.name()` and seizes that string, while two writers still address the collection by the name-index key.

### Issue 1 -- bug

`store_new_version` still persists `/collection/name/{actual_name}` (heads and reindex stay on that name too) but now does `cache.put(new_schema)`, which indexes `new_schema.name`. Those differ for a placeholder rename, the rename `validate_collection_name_not_mutated` allows. `get_collection(actual_name)` and `patch_collection(actual_name)` only read this cache, so the registered name stops resolving; previously `insert(actual_name)` hit. `load_collections` used to `insert` the name-index key it had just read and now ignores it, so reload keeps the split. `take_name` also evicts a different collection id that already holds `schema.name` (inactive synced rows are cached on purpose and skipped by the unique-name check). Same-name updates are unchanged.

`put` should not evict another collection id. Either keep caching under the name-index key, or move that key (and heads) to `schema.name` in the same write.

mike7c2:
Addressed in `ee8f58f7` (the branch was rebased onto the lock-free `Atom` shape at the same time, so the diff moved).

**The name key.** You were right that `put` keying on `collection.name()` broke the two writers that address the collection by the name index key. There is now a second writer, `put_named(name, collection)`, which caches under the registered lookup name, and the three sites you flagged use it:

- `store_new_version` → `put_named(actual_name, ..)`, the same name `/collection/name/{name}`, `schema_heads` and `maybe_reindex_on_version_switch` were written under.
- `load_collections` → `put_named(name, ..)`, the key it just read out of the name index, so a reload restores lookup by the registered name rather than re-splitting it.
- `create_cross_collection_one_to_one_indexes` carries each candidate's name through the loop again, for the same reason: it reads candidates out of the cache by name, so it has to write them back under the name it read. I'd removed that as an incidental cleanup; it wasn't one.

`put` stays for the writers whose schema name *is* the registered name — creation, and a rename this node completed. Entries are still keyed by collection ID, so the two spellings can't fork the cache; they only decide which name reaches the entry.

**The eviction.** `take_name` no longer deletes another collection ID. A displaced holder loses the name and keeps its entry, reachable by ID and by version — a name collision must not cost a collection its data, and `offer` remains the path that decides whether to take a held name at all. `putting_a_different_collection_takes_the_name_not_the_entry` pins that.

Moving the registered name (and the heads) onto `schema.name` as part of the rename write is the other resolution you named. I agree it's the better end state, but it's a behaviour change in the store rather than a re-keying of the cache, so I've kept it out of this one.

New tests: `putting_named_caches_under_the_registered_name`, `putting_a_different_collection_takes_the_name_not_the_entry`, `putting_under_a_new_name_drops_the_collections_previous_name`.

`9ad930e1` is unrelated CI repair: the rebase made `add_collection_to_cache` async and the CLI index adapter's call was left un-awaited, which is what red-lit Lint, surfaces and the WASM build.

@vertexclique — #1788 landed first, so this branch is rebased onto it. `CollectionMap` writes go through `rcu` now, as you described.

mike7c2:
Rebased onto the updated #1795, so the SHAs in my previous comment have moved: the refactor is now `2b0bf40b`.

The separate CI-fix commit is gone — the un-awaited `add_collection_to_cache` originates in #1795 and is fixed there, so all that was left here was formatting drift in this PR's own files, which is folded into the refactor commit.

For the record on the last run: the one failure was `partition::convergence_delete_update_race_preserves_tombstone`, the conformance flake this PR's description already reports as failing at the same rate on the parent branch.

vertexclique (approved):

PR #1809, opened 2026-09-19
fix(query): score and route nested similarity per parent

### LINKED ISSUE #1475: Nested similarity queries never reach the vector index
A `_similarity` query nested under a relation never reaches the vector index. It full-scans the child collection, so the index buys nothing on exactly the shape a hybrid retrieval path uses most.

Top-level similarity was fixed in #1474; this is the remaining half.

## Root cause

`ScanNode::with_vector_route` is attached at exactly one site, the top-level scan in `crates/query-plan/src/planner/builder/mod.rs:348`.

Every scan a join builds is constructed without it:

- `planner/joins/child_plan.rs:255` (the one-to-many child plan, which is the nested-similarity case)
- `planner/joins/secondary_id.rs:118`
- `planner/joins/aggregate_joins.rs:318`
- `planner/joins/multi_level.rs:98` and `:222`
- `planner/joins/filter_relation.rs:81`
- `planner/joins/filter_only.rs:133`
- `planner/joins/join_one.rs:215` and `:299`

The routing itself is not the problem: `vector_routing::route` is pure and works on any `Select`. The child plans simply never call it.

## Evidence

One `Session` with 30 `Note` children, `Note.embedding` carrying `@vectorIndex(dimensions: 4, metric: "DOT")`, the same query shape at both levels with `limit: 5`, read from an execute-explain:

```
top-level: vectorIndex=true   docFetches=5
nested:    vectorIndex=false  docFetches=31
```

Top level bounds the read to the 5 documents asked for. Nested reads all 30 children. The gap grows with the child collection, so it is a scan, not a constant overhead.

## Why it was not caught

Results are correct either way. A full scan scores every child and orders them exactly as the index would, so nothing about the returned rows distinguishes the two paths. Only `vectorIndex` in the explain output, added in #1474, separates them.

## Fix

Attach the route wherever a join builds a `ScanNode` for a collection whose child select carries a routable similarity, so a nested scan narrows the same way a top-level one does. Worth checking whether the attachment belongs in a shared helper rather than repeated at nine call sites, since that repetition is what let the top-level fix miss them.

A regression test belongs next to the existing ones in `crates/defra-node/tests/vector_index_query_path.rs`, asserting `vectorIndex` is present and `docFetches` is bounded for the nested shape.

iverc:
## Context
Nested similarity selections did not compute their score fields or use the vector index. Simply attaching a global route would also let one parent's candidates crowd out another's page.

## Changes
- Add similarity score slots and evaluation to child plans.
- Scope eligible one-to-many leaf routes to the current parent before counting accepted candidates.
- Preserve exhaustive execution for deferred filters, grouping, protected collections, and other shapes where early narrowing could lose rows.
- Cover multiple parents, scalar filters, offsets, deferred score filters, and explain output.

## Testing
- [x] The nested-ordering regression fails before the fix and passes afterward.
- [x] All 21 vector query-path tests and three post-scan filter tests passed.
- [x] Full query test suite passed.
- [x] Strict native Clippy, formatting, and diff checks.

Full WASM validation was blocked locally by clang lacking wasm32 support while compiling `ring`.

Closes #1475.


crates/query/src/planner/joins/child_plan.rs
```diff
@@ -222,17 +222,64 @@ impl Planner {
...
+                    field,
+                    Requestable::Select(_)
+                        | Requestable::Aggregate(_)
+                        | Requestable::FullTextSearch(_)
+                )
+            })
+            && !nested_select
+                .filter
+                .as_ref()
+                .is_some_and(|filter| filter.has_relation_filters() || filter.has_alias_filter())
```
jackzampolin, line 256: `has_alias_filter` is only a top-level `conditions().contains_key("_alias")`, and `has_relation_filters` skips `_alias` while recursing into `_and` / `_or` / `_not`. So `{_and: [{title: {_neq: "a7"}}, {_alias: {sim: {_gt: 103}}}]}` and `{_not: {_alias: {sim: {_gt: 103}}}}` still take this branch.

Line 312 then puts that raw filter on the scan. `ScanNode::next` evaluates it before `SimilarityNode` writes the score. The `sim` slot is null, and `_gt`/`_lt` against null is false.

- Under `_and`, every child fails the scan. Widening never fills `k`, and the fallback scan applies the same pre-score filter, so the parent is returned with no notes.
- Under `_not`, the null score makes the predicate true, the scan stops after the nearest `k`, and the post-score `SelectNode` then rejects those rows. Matches past `k` are never read.
- Under `_or` with a scalar, only the scalar side can fill `k`; rows that pass on score alone are dropped.

A top-level `_alias` key correctly skips routing, because `SelectNode` sits above `SimilarityNode`. Treat any `_alias` at any depth the same way: do not route, and do not install the unsplit filter on the scan. A scalar-only `_and` can still route.

jackzampolin (commented):
## Summary

Nested one-to-many similarity now gets a score slot, a SimilarityNode, and a vector route that stays dormant until TypeJoinMany binds the current parent. Widening counts only rows that match that parent (and any scalar residual). A top-level `_alias` predicate correctly stays on the exhaustive path.

The new routing gate misses score predicates nested under `_and` / `_or` / `_not`. Those are applied on the scan before the score exists, so a parent that has hits can come back empty or with the wrong page.

## Issues

### Issue 1 -- Severity: bug

`crates/query/src/planner/joins/child_plan.rs:256`

`has_alias_filter` only checks the top-level key, and `has_relation_filters` skips `_alias` while walking logical operators. The raw filter is then installed on the scan (line 312) and runs before SimilarityNode writes `sim`. Comparisons against that null score drop or truncate the parent's real matches. A scalar-only filter is fine; any `_alias` at any depth needs to stay on the post-score SelectNode.

Status: open

Issue #1476, opened 2026-08-14
Collection index lists are action-aware only on some load paths


vertexclique:
Whether a collection's index list excludes indexes with pending or errored actions depends on which code path loaded it. The planner and the fetchers can therefore disagree about which indexes are queryable, and a half-built index can be both routed to and searched.

Raised on #1474 against the vector search path; it is not specific to vector indexes.

## Two ways a Collection is built

- `Collection::with_index_actions` filters: `write_indexes` drops errored indexes, `queryable_indexes` drops any index with an action still pending.
- `Collection::new` filters nothing: every index is both writable and queryable.

## Where each is used

Action-aware:

- `collection_loader.rs:106` (the lazy loader the transaction-scoped fetchers use)
- `collection_provider.rs:120` and `:142`

Not action-aware, and these populate the process-wide cache:

- `collection_ops/create.rs:247`, `:315`, `:452`
- `collection_ops/version.rs:129`
- `patch/mod.rs:273`
- `migration/set_migration.rs:237`

## Consequence

`DbCollectionProvider::get_collection` reads that cache (`db.get_collection()` then `schema_for_queries()`), so the planner's view of which indexes are queryable is whatever the last writer put there. If the entry came from a `Collection::new` path, a pending index is offered to index selection and to vector routing.

The two fetchers also disagree with each other:

- `LensedDocFetcher` loads through `collection_loader.rs:106`, so it sees the filtered list.
- `LensedAutoCommitFetcher` loads through `db.get_collection`, so it sees whatever is cached.

For a vector index this means the auto-commit path can search a half-built index and return incomplete results, while the transaction path refuses the same index with "no queryable vector index with id N". Same query, two answers, decided by which fetcher served it.

Note this is not a divergence introduced by the vector read path: index selection for scalar indexes reads the same cache and has the same exposure.

## Fix

Make the collection cache action-aware at the point of insertion, so every consumer sees one answer, rather than filtering at each read site. Fixing only a fetcher cannot help: the planner has already chosen the index from the same source by the time a fetcher is asked for it.

Worth a test that a pending index is invisible to both fetchers and to index selection.

PR #1814, opened 2026-09-20
fix(db): preserve index action state across cache refreshes

iverc:
## Context
Some collection-cache writers rebuilt collections without their pending or errored index state. A cache refresh could therefore expose an unfinished index to the planner and autocommit fetcher.

## Changes
- Load index action state when rebuilding collections after creation, activation, patches, migrations, and replication.
- Preserve restrictions in transaction commit callbacks and version-based query lookups.
- Await action-aware cache refreshes from CLI index updates.
- Cover pending and errored indexes across recaching, activation, transaction commits, and schema patches.

## Testing
- [x] Collection, definition, read, transaction, and write suites: 305 tests passed.
- [x] DB and CLI all-target Clippy, formatting, and diff checks.

Based on #1803 to build on the existing collection-cache refactor rather than duplicate it.

Closes #1476.


crates/db/src/txn/registry/collections.rs
```diff
@@ -196,13 +197,13 @@ impl<S: Store + 'static> DbTransactionRegistry<S> {
...
 
         let db = self.db.clone();
-        let committed = target.clone();
+        let committed = Collection::load_index_actions(target.clone(), &txn.systemstore()?).await?;
         txn.on_success(Box::new(move || {
             if let Ok(mut cache) = db.collections.write() {
-                if committed.is_active {
-                    cache.put(Collection::new(committed));
+                if committed.schema().is_active {
+                    cache.put(committed);
```
jackzampolin, line 204: `committed` was loaded at line 200 from this explicit transaction. `begin` anchors that regolith transaction at `register_snapshot_at_horizon()`, and serializable validation does not reject a phantom insert into a scanned prefix, so a `BACKFILL_INDEX` row committed after `begin` is invisible here. A `define_index` that also wrote a schema key this transaction read will conflict and skip the callback. A settlement that happens after validation will not: storage commit returns, async `on_success` hooks (including the ACP hook `begin` registers) await, and only then does this `cache.put` run. `complete_action` / `fail_action` plus `reload_cache` can land in that gap and then be overwritten.

Restoring `IN_PROGRESS` over a finished backfill keeps the index out of `schema_for_queries`. Restoring it over `ERRORED` puts the index back into `write_indexes`, so uniqueness checks use a partial index. The in-txn `cache_collection` above should stay on this snapshot; the process cache should be built from a new read at put time, under the same lock `reload_cache` holds across its own read and put. The new test begins only after the action is committed and never settles it before this callback, so it misses both orders.

crates/db/src/txn/registry/schemas.rs
```diff
@@ -72,7 +76,7 @@ impl<S: Store + 'static> DbTransactionRegistry<S> {
                     if let Ok(mut cache) = db.collections.write() {
                         if let Some(cached) = cache.get(&updated_destination.name) {
                             if cached.schema().version_id == destination_version_id {
-                                cache.put(Collection::new(updated_destination.clone()));
+                                cache.put(updated_collection);
```
jackzampolin, line 79: Same publish of a caller-snapshot `Collection`. `updated_collection` was loaded at lines 46–48, then this async callback awaits `lens_store.add_with_id` before `cache.put`. A backfill can finish and `reload_cache` in that await; the version-id check still replaces the entry when the schema version matches. `maybe_reindex_after_migration` then rebuilds whatever `write_indexes` the stale entry has.

Load action status with a fresh read immediately before this put, serialized with `reload_cache`. The `add_schema_in_txn` callback is a different case: those collections are created in this transaction, so their action rows are its own writes.

crates/db/src/collection/ops/lookup.rs
```diff
@@ -31,17 +40,18 @@ impl<S: Store> crate::database::DB<S> {
     /// [`Cached::NameHeldByAnother`] means an entry naming a different
     /// collection already holds the name and was left alone, so the cache is
     /// unchanged.
-    pub fn add_collection_to_cache(&self, schema: CollectionVersion) -> Result<Cached> {
+    pub async fn add_collection_to_cache(&self, schema: CollectionVersion) -> Result<Cached> {
         let name = schema.name.clone();
+        let offered = schema.collection_id.clone();
+        let collection = self.collection_with_index_actions(schema).await?;
```
jackzampolin, line 46: `collection_with_index_actions` opens a discarded read, so unlike the registry callbacks it does see actions committed before the call. Nothing ties that read to this `offer` (or to the `cache.put`s in `set_active_collection_version`, both patch updates, `set_migration`, and `create_cross_collection_one_to_one_indexes`). `backfill_index` does not take the collection write guard, so `complete_action` / `fail_action` and `reload_cache` can finish during the awaits inside the helper. The later put then restores the pre-settlement filters. Holding the collection guard around the caller does not close it.

The status read and the cache write need to share a lock with `reload_cache`. A late re-read that is not under that lock still loses.

jackzampolin (commented):
The direct cache writers now keep a pending or errored `BACKFILL_INDEX` out of `queryable_indexes`, which is the #1476 fix for create, activate, patch, migration, and replication when the action row is already durable before the refresh. The new test covers that order.

Explicit-transaction commit callbacks still `cache.put` a `Collection` built from the caller's snapshot, and `collection_with_index_actions` is published only after its read returns. `backfill_index` settles the action and `reload_cache`s without the collection write guard, so either publish can put an older filter back onto the process cache. The planner and the autocommit fetcher both read that cache.

Details on the three sites below.

mike7c2:
@vertexclique @iverc — I rebased this onto the current head of #1803 (`2b0bf40b`). Your previous head was `2d575341`, recoverable from the reflog or from this PR's timeline if you need it.

Why: #1803 was rebased onto the updated #1795, which rewrote `ee8f58f7` — the commit this branch was based on — so the base here was orphaned and the diff would have read wrong.

The replay was clean, no conflicts. Your two commits are now `1a5ac1d4` and `bfaa099f`, unchanged in content but for one thing: `crates/cli/src/index_adapter.rs` has dropped out of the diff. The `.await` on `add_collection_to_cache` that this branch carried is now in #1795, where the async change originates, so it arrives from the base instead of from here.

Verified on the rebased head: `cargo test -p db -p cli` green across 36 binaries, `cargo clippy -p db -p cli --all-targets -- -D warnings` clean (no `-A` needed now — the pending-DAG warning your description mentions is gone from the base), `cargo fmt --all --check` clean.

Nothing else touched. Sorry for moving your branch under you; shout if you'd rather have re-done it yourself.

jackzampolin (approved):

PR #1815, opened 2026-09-20
feat(events): identify replicator replay completions

### LINKED ISSUE #1505: ReplicatorCompleted event has no payload: concurrent installs are indistinguishable and replay-vs-skip is invisible
`events::Message::replicator_completed()` publishes `MessageData::ReplicatorCompleted` as a unit variant (`crates/events/src/event.rs`) — no peer id, no collections, no doc/error counts. It is emitted both when initial replay finishes and when an install short-circuits ("Replicator already exists with same collections, skipping initial replay", `crates/p2p-adapter/src/iroh.rs`).

Consequences for an embedding application:
- With two concurrent `add_replicator` installs (e.g. a hub pairing two clients, or one client re-pairing while another installs), the completion events are indistinguishable — a subscriber cannot tell which peer's backfill finished.
- There is no way to distinguish "replayed N documents" from "skipped replay entirely", although the two have very different meanings for a sync-progress UI ("backfill complete" vs "nothing was sent").
- `push_existing_docs` already knows the peer, the collection list, and per-doc outcomes (`crates/db-merge/src/push_docs.rs`); the information is dropped at the event boundary.

Ask: give `ReplicatorCompleted` a payload — at minimum `peer_id` and `collections`; ideally also `replayed_doc_count`, `skipped: bool`, and `failed_doc_count` (failures already persist into the retry ladder, so a count is cheap). Additive CBOR/JSON fields keep existing subscribers working.

Context: found while building sync indicators for a mobile client; the current workaround is serializing all replicator installs app-side so the payload-less event stays attributable, which defeats concurrent pairing.

iverc:
## Context
Concurrent replicator installs emitted indistinguishable completion events. Subscribers could not tell which peer finished or whether replay was skipped.

## Changes
- Include the peer, collection names, skipped status, and any replay-level error.
- Publish outcomes from both libp2p and Iroh adapters.
- Expose the payload through HTTP SSE and FFI while preserving legacy payload-less events and their wire format.

Completion describes the replay task outcome, not an acknowledgement of remote delivery. Per-document counts are not included.

## Testing
- [x] Both-transport adapter suite: 78 tests passed.
- [x] Event suite: 21 tests passed.
- [x] HTTP SSE and FFI serialization regressions passed.
- [x] Clippy for both adapters and event consumers, formatting, and diff checks.

Closes #1505.


crates/p2p-adapter/src/libp2p.rs
```diff
@@ -567,34 +567,55 @@ impl<B: Blockstore + 'static> P2POperations for P2PAdapter<B> {
...
+                    if let Err(ref error) = result {
                         tracing::error!(error = %error, "Failed to push existing docs to replicator");
                     }
                     if let Some(bus) = push_event_bus {
                         tracing::debug!("publishing ReplicatorCompleted event");
-                        bus.publish(events::Message::replicator_completed());
+                        bus.publish(events::Message::replicator_completed_with_data(
+                            events::ReplicatorCompletedData {
+                                peer_id: peer_id.to_string(),
+                                collections: collection_names_requiring_replay,
```
jackzampolin, line 587: `collections` here is only the replay subset from `collections_requiring_replay` (new names, or names whose filter/capability changed), not the set passed to `add_replicator`. An install of `[Note, User]` that only replays `User` completes as `(peer_id, ["User"])`, the same identity as a concurrent install of just `User`. Filters are not in the payload, so two filter changes of the same names collide the same way.

Two calls with the same peer and requested set also share this identity: both can pass the existing-replicator read before either `create_replicator_info` lands, and the first `Ok` publishes `skipped: false` while the other `push_existing_docs` is still running.

The already-registered branch makes that worse. It publishes `effective_collections` with `skipped: true` immediately at line 614, after the registry write and with no wait for the in-flight task. A waiter matching the requested set takes that skip and never sees `skipped: false` for it, because this path later completes under the subset. `iroh.rs:735` and the skip at `iroh.rs:761` do the same thing.

Publish the requested collection set on every path, and do not emit a completion for a peer that still has a replay task running (or stamp each install with its own id).

crates/p2p-adapter/src/libp2p.rs
```diff
@@ -567,34 +567,55 @@ impl<B: Blockstore + 'static> P2POperations for P2PAdapter<B> {
...
+                                skipped: false,
```
jackzampolin, line 588: `skipped: false` with `error: None` is not "this replay finished". `push_existing_docs` returns `Ok(())` from inside the document scan at `crates/db/src/merge/push_docs.rs:368`, before `push_handles` are awaited at line 472. Tasks spawned at line 418 are dropped unjoined, so those pushes are still running when this event is published. The `Ok(())` at `push_docs.rs:239` is the same false success when the replicator row is already gone and nothing was pushed.

`iroh.rs:736` maps `Ok` the same way. Those paths should return an error (and the mid-scan return should join or cancel the handles it already spawned) before either adapter reports a clean completion. This is separate from per-doc failures that stay `Ok` after a fully joined replay, which this PR explicitly does not surface.

crates/p2p-adapter/src/iroh.rs
```diff
@@ -716,32 +716,53 @@ impl<B: Blockstore + 'static> P2POperations for IrohP2PAdapter<B> {
...
+                        .await;
+                    if let Err(ref error) = result {
                         tracing::error!(error = %error, "Failed to push existing docs to replicator");
                     }
                     if let Some(bus) = push_event_bus {
-                        bus.publish(events::Message::replicator_completed());
+                        bus.publish(events::Message::replicator_completed_with_data(
+                            events::ReplicatorCompletedData {
+                                peer_id: push_peer.to_string(),
+                                collections: collection_names_requiring_replay,
```
jackzampolin, line 735: Same identity bug as `libp2p.rs:587`. This list is the replay subset, so an install of `[Note, User]` that only replays `User` collides with an install of `User`, and two in-flight adds of the same peer and requested set share one `(peer_id, collections)`. The first `skipped: false` lands while the other push is still running.

The skip at line 761 publishes `effective_collections` with `skipped: true` on this task, immediately after the registry write, so a waiter for the requested set can observe that skip before the in-flight replay completes under the subset.

crates/p2p-adapter/src/iroh.rs
```diff
@@ -716,32 +716,53 @@ impl<B: Blockstore + 'static> P2POperations for IrohP2PAdapter<B> {
...
+                                skipped: false,
```
jackzampolin, line 736: Same false success as `libp2p.rs:588`. `push_existing_docs` returns `Ok(())` at `crates/db/src/merge/push_docs.rs:368` before the `push_handles` spawned at line 418 are awaited (line 472), and at line 239 when the replicator is already gone. This publish then reports `skipped: false` and no error while that replay has not finished. Return an error from those paths, and join or cancel the already-spawned handles on the mid-scan return.

jackzampolin (commented):
## Summary

Existing `ReplicatorCompleted` subscribers still receive these events. Both the unit variant and `ReplicatorCompletedWithData` use `EventName::ReplicatorCompleted`, and `ChannelBus` matches on that name only. HTTP SSE and FFI still emit the event; `Message::replicator_completed()` keeps the old wire shape (`data: {}` / `{"type":"replicator_completed"}`), and `remove_replicator` still publishes that unit variant.

Distinct peers do not share a `peer_id`: libp2p publishes `PeerId`'s display form, and iroh publishes the same string stored on the replicator record. The payload still fails as an install identity. Overlapping installs of one peer publish the same `(peer_id, collections)`, and a `skipped: false` completion can be published before that replay has finished.

## Issues

### Issue 1 -- Severity: bug

`crates/p2p-adapter/src/libp2p.rs:587` (same pattern at `crates/p2p-adapter/src/iroh.rs:735`, skip publishes at `libp2p.rs:614` and `iroh.rs:761`)

`collections` is not unique across in-flight installs, so one completion is observable before the other replay finishes.

`collections_requiring_replay` returns only names that are new or whose filter or explicit replay capability changed. The post-push event stores that subset, not the set passed to `add_replicator`. An install of `[Note, User]` that only replays `User` completes as `(peer_id, ["User"])`, the same identity as a concurrent install of just `User` on that peer. Filters are not in the payload, so two filter changes of the same names collide the same way.

Two calls with the same peer and requested set also share one identity. Both can pass the existing-replicator read before either `create_replicator_info` lands, both spawn `push_existing_docs`, and the first `Ok` publishes `skipped: false` while the other task is still pushing.

The already-registered branch publishes a different list at a worse time. After the registry write, it publishes `effective_collections` with `skipped: true` on the `add_replicator` task itself, with no wait for a replay spawned by the earlier call. A subscriber matching the requested set accepts that skip and never sees `skipped: false` for it, because the in-flight task later completes under the subset.

The new tests only publish hand-built messages, so they do not catch this.

Suggestion: publish the requested collection set on every path, include the filter when it distinguishes the install, and do not emit a completion for a peer that still has a replay task running. A per-call id is the alternative if overlapping installs must stay distinguishable.

Status: open

### Issue 2 -- Severity: bug

`crates/p2p-adapter/src/libp2p.rs:588` (same mapping at `crates/p2p-adapter/src/iroh.rs:736`)

`skipped: false` and `error: None` are published when `push_existing_docs` returns `Ok(())` before this call's replay has finished.

`crates/db/src/merge/push_docs.rs:368` is inside the document scan and returns `Ok(())` before `push_handles` are awaited at line 472. Tasks spawned at line 418 are dropped unjoined, so those pushes are still running when the adapters publish. The earlier `Ok(())` at `push_docs.rs:239` is the same signal when `acquire_replicator_retry_guard` finds the replicator already gone and nothing was pushed.

Per-document failures that are persisted and still return `Ok` at the end of a fully joined replay are a separate, documented choice (no per-doc counts, not a remote ack). These two returns are not that: the replay task has not finished.

Suggestion: make those paths return an error, and on the mid-scan return join or cancel the handles already spawned, before either adapter treats `Ok` as `skipped: false` with no error.

Status: open

PR #1819, opened 2026-09-20
feat(storage): add read-only storage diagnostics

### LINKED ISSUE #1805: Storage diagnostics: per-store, per-collection and per-field byte accounting
## Summary

There is no supported way to answer "what is using the space in this node?" To attribute bytes in a 1.7 GiB embedded database to stores, collections and fields, we had to write an external scanner against regolith and re-derive DefraDB's key layout and block encoding (context: gents-ai/gents#1543). A small first-party diagnostic would have made that a one-line command and would make storage regressions visible to every embedder.

## What we needed and had to build

- Open read-only without a node (`regolith::Db::open_read_only` works well, and `regolith.levelstats` / `regolith.total-sst-files-size` / `regolith.estimate-num-keys` are useful — but DefraDB does not surface them).
- Key count, key bytes and value bytes per namespace (`d`, `b`, `h`, `s`, `p`, `e`, `a`).
- Datastore bytes per collection (document bodies, versions, secondary indexes), with collection IDs resolved to names.
- Blockstore bytes per collection and field, by decoding `delta.{lww,composite,collection}.collectionVersionID` / `fieldName`, plus counts of signature/composite/collection blocks.
- Versions per document per field (via the `/d/b/` block→doc index) to find rewrite-heavy fields.
- Headstore and systemstore index overhead.

## Proposal

A read-only `storage stats` surface (CLI subcommand and an embedded API so hosts can expose it), reporting the above as JSON: per store, per collection, per field; counts and logical bytes; optionally the top-N fields by versions-per-document. It should run against a closed database directory or a read transaction on a live node, never write, and never emit values.

Nice to have: expose the engine properties (`levelstats`, SST totals) through the same surface, and note which numbers are exact (logical bytes) versus estimated (compressed share).

The scanner we wrote (~300 lines, aggregates only) can be shared as a starting point.

iverc:
## Context
Storage growth was difficult to attribute without reading raw keys or opening a full node.

## Changes
- Add `DB::storage_stats` and an offline `storage stats [--versions]` command.
- Report logical key/value bytes by namespace, collection, field, and block kind without exposing document values.
- Count shared block ownership and optional document versions.
- Keep unknown or ambiguous keys unattributed rather than guessing their collection.
- Open offline storage read-only without creating node configuration or keyring state.

Reported bytes are logical storage sizes, not compressed disk usage. Offline inspection rejects encrypted stores; the live DB API can inspect them.

## Testing
- [x] Storage suite: 460 passed, 9 existing ignored.
- [x] Database suite: 15 passed.
- [x] CLI storage-stats tests: 2 passed.
- [x] Regression checks cover unchanged files after offline reads, rejected writes, shared history, malformed schemas, and index-key ambiguity.
- [x] DB, CLI, and storage all-target Clippy, formatting, and diff checks.

Closes #1805.


crates/db/src/database/storage_stats.rs
```diff
@@ -0,0 +1,310 @@
...
+                        }
+                        owners.close().await?;
+                    }
+                }
+            } else {
+                report.unattributed_blocks.record(&pair.key, &pair.value);
+            }
+        }
+        if report.total.keys % 1024 == 0 {
+            tokio::task::yield_now().await;
```
jackzampolin, line 217: `tokio::task::yield_now()` does not compile on the wasm client. `tokio` is optional in `db` and is linked only by the `native` feature; `crates/wasm` builds `db` with `default-features = false` and does not enable `native`, and this module is not cfg-gated. `cargo build -p defra-wasm` fails resolving `tokio`.

Yield with a one-shot waker future instead (`futures` is already a required dependency) so the scan still yields under both tokio and `wasm-bindgen-futures`. Cfg-gating this call out would compile, but a full-store scan would then block the browser thread.

jackzampolin (commented):
## Summary

Read-only storage diagnostics for a closed data directory and for a live `DB` snapshot. Offline open goes through `regolith::Db::open_read_only` (shared, non-blocking directory lock; WAL is replayed into memory and not rewritten; manifest is not repaired). The live API takes `new_txn(true)` and always `discard`s it, including on error. Aggregates are summed from the root iterator: `total` matches a full scan, namespace buckets partition on the first key byte, and datastore / block bytes are attributed only when the key or block schema version resolves to a loaded collection. Ambiguous vector/ordered index prefixes are left in `unattributed_datastore`. Document values, doc IDs, and signature identities are not serialized. Encrypted stores are rejected on the CLI and decrypted only through the existing live `EncryptedStore` wrapper.

The scan does not take a lock that stalls commits. A read snapshot pins compaction GC the same way any other read transaction does; writers are not blocked. `close` waits on `active_txns`, which is existing transaction shutdown behavior.

## Issues

### Issue 1 -- Severity: bug

`crates/db/src/database/storage_stats.rs:217`

`collect` calls `tokio::task::yield_now()` with no `cfg`. In `db`, `tokio` is optional and is enabled only by the `native` feature (`native = ["dep:tokio", ...]`). `crates/wasm/Cargo.toml` depends on `db` with `default-features = false` and does not turn `native` on, and `pub mod storage_stats` is not cfg-gated. `cargo build -p defra-wasm` and `cargo check -p db --no-default-features` fail resolving the `tokio` crate. The browser client is a supported target.

Status: open

Suggestion: yield with a one-shot waker future (the `futures` crate is already an unconditional dependency). That works under both tokio and `wasm-bindgen-futures`. Cfg-gating the call out would compile, but a full-store scan would then run without yielding on the browser thread.

PR #1820, opened 2026-09-20
feat(p2p): honor negotiated PushLog retry-after hints

### LINKED ISSUE #1305: feat(p2p): PushLog admission retry-after — pusher-side backoff hint (W2b from #1088)
Split from #1088 (closed: W1–W5 shipped via #1089, #1143, #1152).

## Remaining work (W2b)

Capacity-drop nacks are now byte-exact `RATE_LIMITED` replies with per-peer quotas, but the pusher has no backoff hint:

- `MIN_REQUEST_REFILL_RATE = 1.0` floor in `crates/p2p/src/sync/manager/rate_limiter.rs` (~line 101) means a saturated hub refills at a fixed minimum with no signal to the pusher about when to retry.
- No retry-after metadata in the nack reply, so pushers retry on their own schedule, wasting round-trips under sustained fan-in.

## Scope

- Add a retry-after (or refill-rate) hint to the PushLog nack reply.
- Teach the pusher retry path to honor the hint instead of fixed backoff.
- Extend `tools/integration-test/tests/p2p_admission.rs` to assert pushers back off per the hint under saturation.

Optimization only — the no-silent-divergence invariant is already enforced and tested.

iverc:
## Context
PushLog backpressure did not tell senders when to retry, so live sends and replay could keep probing a saturated peer.

## Changes
- Negotiate retry-after support without changing legacy request bytes or error sentinels.
- Sign optional retry hints, validate them, and cap accepted delays at 60 seconds.
- Persist peer retry deadlines and honor them across live delivery, initial replay, reconnects, and restarts.
- Preserve active deadlines when scopes are acknowledged and leave healthy peers unblocked.
- Use the existing retry machinery rather than adding another delivery loop.

## Testing
- [x] P2P library with both transports: 484 tests passed.
- [x] Adapter retry suite: 21 tests passed.
- [x] Storage retry scheduling and DB replay admission tests passed.
- [x] Live P2P admission test passed against the rebuilt node.
- [x] Legacy protocol fallback and wire/signature compatibility regressions passed.
- [x] Affected-package all-target Clippy, formatting, and diff checks.

Closes #1305.


crates/db/src/merge/push_docs_replay.rs
```diff
@@ -57,6 +57,54 @@ pub struct ReplayDocumentFailure {
...
 
+/// Persist the typed hint before crossing the legacy string error boundary.
+pub async fn persist_retry_after<S: storage::corekv::Store>(
+    peerstore: &storage::stores::Peerstore<S>,
+    peer_id: &PeerId,
+    reply: &PushLogReply,
+) -> Result<(), String> {
+    if let Some(delay) = reply.retry_after() {
+        peerstore
+            .reschedule_retry_peer(peer_id.as_str(), Some(delay), 0)
```
jackzampolin, line 68: `persist_retry_after` calls `reschedule_retry_peer`, which returns `Ok(false)` and writes nothing when no `/rep/retry/id/{peer}` row exists. That is the first history replay: the replicator row exists, the schedule row does not, and `Ok(false)` is treated as success.

`persist_replay_failures` then stores `RetryInfo::new_initial()`, so `register_scope_marker_once` installs a 15–30s ladder rung with `not_before_unix = 0`. The in-memory replay pacer dies with this call. A hint longer than that rung (wire cap 60s; tests use 45s) is sent early, and `activate_retry_peer` → `defer_for(0)` makes the next sweep immediate because there is no deadline to preserve. Markers stay peer-scoped, so this does not drop the docs or block another peer.

Insert the schedule (`new_initial` + `defer_for(delay)`) when the replicator exists but the retry-info row does not, and do not return success if the hint was not written. Skip the insert only when the replicator itself is gone.

jackzampolin (commented):
## Summary

Negotiated PushLog retry-after is capped at 60s, signed, and stored as a peer-wide `not_before_unix`. The live failure recorder, in-memory backlog cooldown, and durable sweep honor that deadline once it is actually stored: scope markers are kept, one peer's hint is not applied to another, and `activate_retry_peer` does not clear a future `not_before_unix`.

Initial history replay does not store the deadline. `persist_retry_after` only updates an existing retry-info row, and the first replay of a new replicator has no such row, so the hint is discarded and replaced by an ordinary ladder rung with `not_before_unix = 0`. A later reconnect can then send immediately.

## Issues

### Issue 1 -- Severity: bug

crates/db/src/merge/push_docs_replay.rs:68

Description: `persist_retry_after` is how replay and `retry_doc` record a receiver hint. It calls `Peerstore::reschedule_retry_peer`, which returns `Ok(false)` and writes nothing when `/rep/retry/id/{peer}` is missing (`crates/storage/src/stores/peerstore.rs:643`). That `Ok(false)` is treated as success, so `push_existing_docs` (`crates/db/src/merge/push_docs.rs:433`) does not even warn.

Adding a replicator persists the replicator row and then replays history. Nothing creates a retry-info row before the first nack, so this no-op is the common first-contact path. After the send tasks join, `persist_replay_failures` (`crates/db/src/merge/push_docs_replay.rs:131`) writes each failed scope with `RetryInfo::new_initial()`. `register_scope_marker_once` (`crates/storage/src/stores/peerstore.rs:348`) merges `not_before_unix` only when the requested bytes already carry one. `new_initial` does not, and because the row is still absent the write `bump_with_schedule`s the first ladder rung (15–30s) and leaves `not_before_unix` at 0. The `ReplayPushGate` pacer blocks sibling sends only inside this call; the gate is created at `crates/db/src/merge/push_docs.rs:264` and dropped when replay returns.

Markers are still written, and the key is peer-scoped, so this does not drop the documents or block some other peer. It does fail to delay this peer until the negotiated hint:

- A hint longer than the first rung is not waited out. The wire cap is 60s and the new tests use 45s, but the sweep's `remaining_retry_after` sees no deadline and sends when the 15–30s rung is due.
- `activate_retry_peer` (`crates/storage/src/stores/peerstore.rs:709`) calls `defer_for(0)`. With `not_before_unix == 0`, that sets `next_retry_unix` to now (`crates/storage/src/stores/retry_info.rs:224`). The next sweep sends immediately, and live admission allows it too, because admission only blocks while `not_before_unix` is still in the future. `PeerConnected` runs that activation; a flap after the ladder row exists skips both the rung and the hint.

The live recorder path does not have this hole: it passes `defer_for(delay)` bytes into `record_push_failure`, and `register_scope_marker_once` creates the row from those bytes. `retry_doc` and `retry_collection_commit` are safe only once a row already exists, which is true for the sweep and false for the first replay.

Suggestion: When the replicator row exists but the retry-info row does not, `persist_retry_after` should insert `RetryInfo::new_initial()` after `defer_for(delay)` and must not return success if that write did not happen. If the replicator itself is gone, keep skipping the insert. The later `new_initial` write in `persist_replay_failures` preserves an existing deadline, so it is fine only after that insert. Add a regression that replays into a replicator with no prior retry-info row, receives a 45s hint, and asserts `not_before_unix` survives both `persist_replay_failures` and `activate_retry_peer`.

Status: open

Issue #1727, opened 2026-09-08
Bound rooted CAR authorization memory and verify slow-history continuation


jackzampolin:
## Context

Review of #1726 identified a residual resource/coverage limitation in `crates/p2p/src/sync/car_authorization.rs`. Rooted authorization has a two-second wall-clock budget, yields during traversal, and never returns a partial rooted grant on timeout. Independent replicator/ACP grants remain available. The traversal does not have an explicit frontier-memory bound or resumable progress; repeated requests restart at the root.

This is a hardening and coverage follow-up, not a demonstrated production regression or authorization bypass. Reintroducing the old 10,000-node authorization cutoff would break valid paginated history recovery.

## Acceptance

- Exercise very wide/shared DAGs, including duplicate frontier links, and measure/bound authorization memory and scheduling impact.
- Exercise a valid history whose rooted traversal exceeds one request budget, and define a bounded way for subsequent recovery to make progress without broadening authority.
- Preserve exact root reachability, encryption-link exclusion, independent per-block grants, and fail-closed behavior when authorization cannot be established.
- Include adversarial unreachable requested CIDs and slow blockstore cases; distinguish the CAR response-size limit from authorization-work limits.

PR #1825, opened 2026-09-21
fix(p2p): bound the rooted authorization walk and make it resumable

iverc:
## Context

The rooted CAR authorization walk had three unbounded behaviors (#1727): it queued every link and remembered nodes only at pop time, so a dense DAG with duplicate frontier links grew the frontier without bound; a request for adversarially unreachable CIDs explored the whole DAG, terminating only on wall-clock timing; and it read the full payload of every visited node however large, including oversized ones. A walk that exhausted its budget also stranded its root: retries restarted at the root, so a valid history deeper than one budget never got a rooted grant.

## Changes

- nodes are remembered at push time, so duplicate frontier links cost one queue entry per distinct block
- the walk carries a deterministic node budget alongside the deadline, counted across resumed passes so retries cannot multiply a single root's work
- an oversized node is authorized when requested but never read for links, bounding per-node work however large a linked block is; the blocks behind it fail closed
- a walk that outlives its budget retains its traversal state per root (capped at sixty-four roots, oldest evicted first), and a retried request resumes from the frontier. The node in flight when the budget fired goes back to the front of the frontier; without that, it had been popped but already remembered as seen, so nothing could ever re-enqueue it and the resumed walk drained to an empty grant

Exhaustion still fails closed to independent per-block grants, a CID is still granted only once the walk reaches it, and encryption links still never confer authority.

## Testing

- `cargo test -p p2p --lib`: 390 passed
- a dense ring where every node links both neighbours proves each distinct block is payload-read exactly once
- unreachable requested CIDs end deterministically on the node budget with the retryable error
- an oversized walk node is granted, the block behind it is not, and its payload is never read
- a chain no single pass can finish converges across resumed retries, where a restarting walk never would; the grant contains exactly the reached CID

Fixes #1727


crates/p2p/src/sync/car_authorization.rs
```diff
@@ -13,73 +13,306 @@ use crate::error::{Error, Result};
...
+            loop {
+                if self.remaining.is_empty() {
+                    self.in_flight = None;
+                    return Ok(());
+                }
+                let Some(cid) = self.queue.pop_front() else {
+                    self.in_flight = None;
+                    return Ok(());
+                };
+                if self.expanded >= max_nodes {
```
jackzampolin, line 106: Node-budget exhaustion drops the CID that was just popped. `in_flight` is cleared and the `Err` returns inside the timeout future, so the `Ok(inner)` arm never `push_front`s it. The CID is already in `seen`, and `retain` only persists `state.queue`, so a resume can never visit it. `get_size` / `get` errors at lines 121 and 134 take the same path. Once the frontier is empty, line 104 returns `Ok(())` with `remaining` still non-empty and the caller `forget`s the cursor. That is a truncated allow-list: a chain of length `max_nodes + 1` errors once and then `Ok`s without the leaf. Put the popped CID back on every non-visit exit, and do not return `Ok` after a budget stop.

crates/p2p/src/sync/car_authorization.rs
```diff
@@ -13,73 +13,306 @@ use crate::error::{Error, Result};
...
+        root: Cid,
+        requested: RapidHashSet<Cid>,
+        budget: Duration,
+        max_nodes: usize,
+    ) -> Result<RapidHashSet<Cid>> {
+        let prior = self.take_incomplete(&root);
+        let expanded_before;
+        let mut state = match prior {
+            Some(prior) => {
+                expanded_before = prior.seen.len();
```
jackzampolin, line 222: `expanded` is rebased to `prior.seen.len()`, which includes the unexpanded frontier (`push` inserts into `seen` before a visit). One block under `CAR_MAX_BYTES` (16 MiB) can enqueue on the order of 4e5 links. After that single expansion, a timeout retains the frontier, and the next pass starts already above `AUTHORIZATION_MAX_NODES` (200_000) even though one node was payload-read. Those children are then discarded one CID per retry and never granted. Persist the number of nodes actually expanded instead of `seen.len()`.

crates/p2p/src/sync/car_authorization.rs
```diff
@@ -13,73 +13,306 @@ use crate::error::{Error, Result};
...
+            }
+        }
+        let mut queue = VecDeque::new();
+        for cid in prior.frontier {
+            queue.push_back(cid);
+        }
+        Self {
+            remaining,
+            authorized,
+            seen: prior.seen,
```
jackzampolin, line 77: Reusing `prior.seen` across requests skips descendants that were expanded for a different want-list. Only `prior.authorized ∩ requested` is pre-granted, so an intermediate visited while authorizing some other CID is in `seen`, not on the frontier, and not granted when a later `selective_blocks` / `selective_dag` request asks for it. The walk can then return `Ok` without that CID. A `get_size` miss (`continue` at line 122) sticks the same way if the pass times out before the block is stored. Key the cursor to the unresolved want-list, and do not treat a miss as visited.

crates/p2p/src/sync/car_authorization.rs
```diff
@@ -13,73 +13,306 @@ use crate::error::{Error, Result};
...
+    }
+
+    fn retain(&self, root: Cid, walk: IncompleteWalk) {
+        let Ok(mut incomplete) = self.incomplete.lock() else {
+            return;
+        };
+        let Ok(mut order) = self.eviction_order.lock() else {
+            return;
+        };
+        if !incomplete.contains_key(&root) {
```
jackzampolin, line 267: `take_incomplete` removes the map entry but not the order slot, so every resume `push_back`s another copy of the same root. `forget` does not remove order entries either, so the cap is 64 slots rather than 64 roots. A hot root's retries evict an idle cursor (a quiet root is gone after 63 consecutive retains; a full table drops the oldest on the next retry of an existing root). That walk restarts at the root and a history deeper than one 2s budget does not converge. `run` also awaits between take and retain, and up to 8 recovery tasks share this map: an `Ok` `forget` deletes whatever cursor is stored for the root, including another walk's. If `eviction_order.lock()` fails, `retain` returns without inserting after `take` already dropped the previous state. Keep one order entry per root and a generation so a stale `forget` cannot delete a newer cursor.

crates/p2p/src/sync/car_authorization.rs
```diff
@@ -13,73 +13,306 @@ use crate::error::{Error, Result};
...
+            }
+            None => {
+                expanded_before = 0;
+                WalkState::fresh(root, requested)
+            }
+        };
+        state.expanded = expanded_before;
+        match state.run(blockstore, budget, max_nodes).await {
+            Ok(()) => {
+                self.forget(&root);
```
jackzampolin, line 233: Every `Ok` forgets the root, and `run` returns `Ok` as soon as `remaining` is empty, including on entry (line 98), before the frontier is used. `filter_car_response_blocks` calls `walk` for every rooted request, including a collection that found no blocks (the catch-up race this handler already documents). An empty candidate set, or a later request whose CIDs are already in `prior.authorized`, deletes the frontier saved for a deeper CID. The next request starts at the root and can time out at the same depth on every retry. Do not walk an empty candidate set, and do not forget a non-empty frontier just because this request had nothing left to find.

crates/p2p/src/sync/car_authorization.rs
```diff
@@ -107,29 +341,293 @@ mod tests {
...
+            let (cid, data) = ipld_block(value);
+            store.inner.put(&cid, &data).await.unwrap();
+            ring.push(cid);
+        }
+        let mut root_links = Vec::new();
+        for (index, cid) in ring.iter().enumerate() {
+            let previous = ring[(index + ring.len() - 1) % ring.len()];
+            let next = ring[(index + 1) % ring.len()];
+            let data =
+                DagCborCodec::encode_to_vec(&ipld!({"prev": previous, "next": next, "self": cid}))
```
jackzampolin, line 448: These `prev` / `next` / `self` links point at the placeholder CIDs from `ipld_block`, not at the rewritten ring. The walk returns once `root_links` are authorized, before those duplicate targets are read, so the payload-read assertion does not prove dedup (a pop-time visited set also reads 17 blocks). The oversized test at line 482 has the same gap: `leaf` is not linked from the oversized bytes, and `get` is not counted. Link the rewritten CIDs into a real ring, and build an oversized parent that actually links `leaf`. Also retry the node-budget case on one `RootedAuthorizationProgress` and assert the second call stays `ResponseTimeout`.

crates/p2p/src/sync/car_authorization.rs
```diff
@@ -13,73 +13,306 @@ use crate::error::{Error, Result};
 // budget is a retryable request failure, never a truncated authorization set.
 const AUTHORIZATION_TIMEOUT: Duration = Duration::from_secs(2);
 
-pub(crate) async fn requested_descendants<B: Blockstore>(
-    blockstore: &B,
-    root: Cid,
-    requested: RapidHashSet<Cid>,
-) -> Result<RapidHashSet<Cid>> {
-    requested_descendants_with_budget(blockstore, root, requested, AUTHORIZATION_TIMEOUT).await
+/// Deterministic bound on nodes expanded per rooted walk, so a request for
```
jackzampolin, line 16: This doc comment (and the one on `RootedAuthorizationProgress` at lines 170-174) restates the #1727 design history. A one-line invariant is enough. The `in_flight` note at lines 35-38 should stay; it records why a popped CID has to return to the frontier.

jackzampolin (commented):
## Summary

The timeout arm does put the in-flight CID back on the frontier, duplicate links are deduped at push time, a missing block is not granted, and encryption links still go through `extract_ipld_links`. Those parts hold.

The resumable cursor does not. Several exits drop a popped CID or delete the cursor, then a later call returns `Ok` with a partial set. `filter_car_response_blocks` treats that `Ok` as a finished walk. Blocks the classifier denies (composites and other payloads `DbBlockClassifier` cannot attribute) are served only by the rooted grant, so a CID the walk should still reach is omitted from a successful CAR. No path turns a missing block or a failed link parse into a grant.

## Issues

### Issue 1 -- Severity: bug

crates/p2p/src/sync/car_authorization.rs:106

Description: Only the timeout arm restores the popped CID. The node-budget check runs after `pop_front`, sets `in_flight` to `None`, and returns `Err` from inside the future, so the outer `Ok(inner)` arm does not `push_front`. `get_size` and `get` errors (`?` at lines 121 and 134) leave the same way. The CID is already in `seen` and is not in the queue `retain` saves (lines 239-242). Nothing can enqueue it again. A chain of length `max_nodes + 1` errors once, then the next call sees an empty queue with `remaining` still non-empty, returns `Ok(())` at line 104, and `forget`s the cursor. That `Ok` is a truncated allow-list, which the file header says exhaustion must not produce. `unreachable_requested_cids_end_on_the_node_budget` never retries, so it misses this.

Suggestion: On every exit that is not a finished visit, put the popped CID back before returning. Persist that queue. Return `Ok` only when every requested CID was decided by a visit, not when the frontier was discarded. A node-budget stop should stay `Error::ResponseTimeout` across retries.

Status: open

### Issue 2 -- Severity: bug

crates/p2p/src/sync/car_authorization.rs:222

Description: Resume sets `expanded` from `prior.seen.len()`, and `push` inserts into `seen` before a visit, so the unexpanded frontier counts as already expanded. `IncompleteWalk` does not store the real counter. `CAR_MAX_BYTES` is 16 MiB, so one linked block can enqueue on the order of 4e5 CIDs. After that block is expanded, a timeout restores the in-flight child and retains the frontier, but the next pass starts with `expanded` already above `AUTHORIZATION_MAX_NODES` (200_000) even though one node was read. Combined with issue 1, each retry drops one frontier CID without authorizing it, then a later call returns `Ok` and forgets the cursor. Direct children that were inside the expansion budget are never granted. A narrower frontier is still subtracted in full, so a walk that payload-read far fewer than 200_000 nodes refuses to continue.

Suggestion: Persist the number of nodes actually expanded (incremented after a successful visit, or at least not including CIDs that are only queued). Do not rebase the budget to `seen.len()`.

Status: open

### Issue 3 -- Severity: bug

crates/p2p/src/sync/car_authorization.rs:77

Description: Progress is keyed only by root. `resumed` keeps `prior.seen` and pre-grants only `prior.authorized` intersected with this request. A CID expanded or skipped for a different want-list is in `seen`, not on the frontier, and not authorized, so a later request never visits it and can return `Ok` without it. That happens for `selective_blocks` / `selective_dag` (collection is the want-list or the frontier closure, but the walk still starts at `root_cid` and expands intermediates), and for a request that names an existing CID that is not under the root (the walk explores until the budget and authorizes nothing). A `get_size`/`get` miss `continue`s at line 122 after the CID is already in `seen`; if this pass then times out, a block stored before the retry is never read. The previous visited set did not stick across requests. `DbBlockClassifier` denies composites and other blocks it cannot attribute, and those are exactly the blocks the rooted grant exists to serve.

Suggestion: Do not reuse `seen` across different requested sets. Keep a cursor only for the same unresolved want-list, and do not mark a missing block seen until a later pass has a chance to observe it. A CID already walked must still be granted if a later request asks for it and the walk reached it.

Status: open

### Issue 4 -- Severity: bug

crates/p2p/src/sync/car_authorization.rs:267

Description: `take_incomplete` removes the map entry and leaves `eviction_order` unchanged. `retain` then sees `!contains_key` and `push_back`s the same root again. `forget` does not remove order entries either. The cap is 64 slots, not 64 roots. A hot root's retries fill the deque with duplicates and evict an idle cursor (one quiet root is dropped after 63 consecutive retains of another; a full table evicts the oldest on the next retry of an existing root). That walk restarts at the root, so a history deeper than one 2s budget does not converge while anything else is retained. CAR fetch runs on the recovery pool (`MAX_ACTIVE_RECOVERY` is 8) and `run` awaits between `take` and `retain`/`forget` (line 231). Two walks of one root can both start fresh; the one that returns `Ok` calls `forget` and deletes the other's cursor, because `forget` is unconditional. If `eviction_order.lock()` fails, `retain` returns without inserting after `take` already dropped the previous cursor.

Suggestion: Keep one order entry per root: update it in place on resume, and remove it in `forget` and `take`. Hold a single mutex for the map and the order. Give each checkout a generation so a stale `forget` cannot delete a newer cursor. Serialize walks per root across the await.

Status: open

### Issue 5 -- Severity: bug

crates/p2p/src/sync/car_authorization.rs:233

Description: Any `Ok` calls `forget`. `run` returns `Ok` as soon as `remaining` is empty, including on entry (line 98), before the frontier is touched. `filter_car_response_blocks` always calls `walk` when the peer has rooted authority, including when collection produced no blocks. That empty response is the catch-up race the handler already documents (the peer asked for blocks this node does not have yet). An empty candidate set, or a later request whose CIDs are already in `prior.authorized`, deletes the frontier saved for a deeper CID. The next non-empty request starts at the root and can time out at the same depth forever.

Suggestion: Do not call `walk` for an empty candidate set, and do not `forget` a non-empty frontier just because this request's `remaining` started empty. Forget only when this checkout actually finished the search that owned the cursor.

Status: open

### Issue 6 -- Severity: suggestion

crates/p2p/src/sync/car_authorization.rs:448

Description: `duplicate_frontier_links_expand_each_block_once` points `prev` / `next` / `self` at the placeholder CIDs from `ipld_block`, not at the rewritten ring. The requested set is only `root_links`, and the walk returns when those are authorized, before the duplicate targets are read. The old pop-time visited set would also payload-read 17 blocks, so the assertion does not prove dedup. `oversized_walk_node_is_authorized_but_not_traversed` (line 482) never links `leaf` from the oversized bytes (the payload is zeros; the root links only to `oversized`), and it does not assert that `get` was skipped. The leaf would be ungranted even if the payload were parsed.

Suggestion: Link the rewritten CIDs into a real ring and count payload reads of those nodes. Build an oversized parent whose DAG-CBOR links to `leaf`, and assert the leaf is absent and `get` was not called for the oversized CID. Retry the node-budget case on the same `RootedAuthorizationProgress` and assert the second call is still `ResponseTimeout`, not `Ok`.

Status: open

### Issue 7 -- Severity: suggestion

crates/p2p/src/sync/car_authorization.rs:16

Description: The doc comments on `AUTHORIZATION_MAX_NODES` (lines 16-20) and `RootedAuthorizationProgress` (lines 170-174) restate the #1727 design history (why the count exists, that legitimate histories stop early, that retries must not multiply work). That is rationale, not an invariant the next reader cannot see in the code. The `in_flight` comment (lines 35-38) is the one that should stay: it records why a popped CID has to return to the frontier.

Suggestion: Collapse those two docs to a single line each (the cap, and that the map is the resume cursor). Leave the `in_flight` note.

Status: open

